PR #26, opened 2022-09-01
⭐️ add 'managedBy' field for assets

joelddiaz:
This is related to work to extend the representation of assets on the
database to hold this 'managedBy' field so that different Mondoo clients
that run periodic scan jobs can "claim" an asset (so they can later be
found for garbage collection purposes).


motor/discovery/resolve.go
```diff
@@ -167,6 +167,9 @@ func ResolveAsset(ctx context.Context, root *asset.Asset, cfn common.CredentialF
 			}
 			assetObj.Category = root.Category
 
+			// copy over managedBy from root
+			assetObj.ManagedBy = root.GetManagedBy()
```
czunker, line 171: Where does root get its `managedBy` from? Seems like I'm missing some magic here.

joelddiaz: This would be something set by the client performing the scan. So an example inventory (the one I've used for testing) would look like:
```json
{
  "inventory": {
    "apiVersion": "v1",
    "kind": "Inventory",
    "metadata": {
      "name": "mondoo-scan-inventory",
      "labels": {
        "environment": "production"
      }
    },
    "spec": {
      "assets": [
        {
          "connections": [
            {
              "backend": 19,
              "discover": {
                      "targets": ["pods", "jobs", "cronjobs", "deployments", "replicasets", "daemonsets", "statefulsets", "container-images"]
                     },
                "options": {
                        "all-namespaces": "true"
                }
            }
          ],
          "managedBy": "someMondooClient"
        }
      ]
    }
  }
}
```
The idea here was to give some flexbility for how how different environments might want to set this field when doing scans and creating/updating assets. And then some approach to garbage collection would allow making a DeleteAssets() request that had a condition based on the values in the database found on the assets.

PR #27, opened 2022-09-02
🧹 Use /proc/sys to read Linux kernel params

imilchev:
Use `/proc/sys` to determine the kernel parameters instead of running `sysctl -a`. This allows us to determine the parameters also when running a static FS scan.

I also added a dev container config. I needed it to be able to test the implementation on my Mac. I thought it might be useful for others as well, so I added it to the repo.

I still need to fix the tests. The current mock FS doesn't support `lstat` for files/dirs. This is used by `afero.Walk` and that's why the tests fail. I will need to extend the mock FS to cover the functionality needed.

### DIFF
--- a/resources/packs/core/kernel_test.go
+++ b/resources/packs/core/kernel_test.go
@@ -4,21 +4,40 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.mondoo.com/cnquery/motor"
+	"go.mondoo.com/cnquery/motor/providers/local"
+	"go.mondoo.com/cnquery/resources/packs/core/info"
+	"go.mondoo.com/cnquery/resources/packs/core/kernel"
+	"go.mondoo.com/cnquery/resources/packs/testutils"
 )
 
 func TestResource_KernelParameters(t *testing.T) {
 	t.Run("kernel parameters", func(t *testing.T) {
-		res := x.TestQuery(t, "kernel.parameters")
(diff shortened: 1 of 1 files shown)


resources/packs/core/kernel_test.go
```diff
@@ -4,12 +4,39 @@ import (
...
+		require.True(t, ok)
+		mapS := make(map[string]string)
+		for k, v := range params {
+			mapS[k] = v.(string)
+		}
+
+		assert.NotEmpty(t, mapS)
 	})
 
 	// TODO: something is wrong with /proc parser, once fixed we need to activate this test
```
czunker: What about this test?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.245.2/containers/go/.devcontainer/base.Dockerfile
```
czunker, line 1: Was the devcontainer used for debugging?

imilchev: debugging and development. I wouldn't be able to verify that the approach works without it since I work on a Mac

resources/packs/core/kernel_test.go
imilchev: this was already there but I think it can be deleted now

.devcontainer/Dockerfile
czunker: Sorry, should have read your comments first.

PR #29, opened 2022-09-04
Detect Ubiquiti UbiOS operating system

tas50:
This runs on their Dream Machine Pro / SE products.

### DIFF
--- a/motor/platform/detector/detector_test.go
+++ b/motor/platform/detector/detector_test.go
@@ -902,6 +902,19 @@ func TestGoogleCOSDetector(t *testing.T) {
 	assert.Equal(t, []string{"linux", "unix", "os"}, di.Family)
 }
 
+func TestUbiOSDetector(t *testing.T) {
+	detector, err := newDetector("./testdata/detect-ubios.toml")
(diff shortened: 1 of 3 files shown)


motor/platform/detector/detector_test.go
```diff
@@ -902,6 +902,19 @@ func TestGoogleCOSDetector(t *testing.T) {
...
 }
 
+func TestUbiOSDetector(t *testing.T) {
+	detector, err := newDetector("./testdata/detect-ubios.toml")
+	assert.Nil(t, err, "was able to create the transport")
+	di, err := detector.Platform()
+	require.NoError(t, err)
+
+	assert.Equal(t, "ubios", di.Name, "os name should be identified")
+	assert.Equal(t, "UbiOS", di.Title, "os title should be identified")
```
chris-rock: This test fails because it returns the version number which is included in `PRETTY_NAME="UbiOS 1.12.24.4315`. If you do not want the version number, we need to write a custom detector for ubios and use `NAME` instead

tas50: Full title with version is actually ideal

PR #31, opened 2022-09-05
✨Add docker tag when scanning a container registry image.

preslavgerchev:
When scanning an image from a container registry, also append the image name:tag combo as a label:
image

We have this behaviour for running `scan container image`:
image

### DIFF
--- a/motor/discovery/container_registry/registry.go
+++ b/motor/discovery/container_registry/registry.go
@@ -180,6 +180,7 @@ func (a *DockerRegistryImages) toAsset(ref name.Reference, creds []*vault.Creden
 	}
 	imgDigest := desc.Digest.String()
 	repoName := ref.Context().Name()
+	imgTag := ref.Context().Tag(ref.Identifier()).Name()
 	imageUrl := repoName + "@" + imgDigest
 	asset := &asset.Asset{
 		PlatformIds: []string{containerid.MondooContainerImageID(imgDigest)},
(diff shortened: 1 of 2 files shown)


motor/discovery/k8s/list_images.go
```diff
@@ -181,7 +181,7 @@ func newPodImageAsset(i containerImage) (*asset.Asset, error) {
 	if len(i.image) > 0 {
 		tag, err := name.NewTag(i.image, name.WeakValidation)
 		if err == nil {
-			tagName = tag.TagStr()
+			tagName = tag.Name()
```
czunker, line 184: Nice fix.

motor/discovery/container_registry/registry.go
```diff
@@ -199,8 +200,9 @@ func (a *DockerRegistryImages) toAsset(ref name.Reference, creds []*vault.Creden
 		Labels: make(map[string]string),
 	}
 
-	// store digest
+	// store digest and tag
 	asset.Labels["docker.io/digest"] = imgDigest
+	asset.Labels["docker.io/tags"] = imgTag
```
czunker, line 205: With the fix in this PR, this is no longer just the tag. It is `<registry_server>/<user_name|project|repo>/<image_name>:<tag>`.

I think this new value is more useful than only the tag. Perhaps we can change the label name? Amazon ECR calls this an URI: https://docs.aws.amazon.com/emr/latest/EMR-on-EKS-DevelopmentGuide/docker-custom-images-tag.html Others use URL.

We also have cases, where the "docker.io" part looks strange in the UI:
image
The image is actually from the gcr.io registry.

I think, this was also a discussion in a previous PR. @imilchev can you remember? But that's not the only place with "docker.io". So, perhaps we should make this a follow-up issue.

imilchev: This change makes the "tags" content for containers, images and registries consistent. However, I do think that having just the tag in this label would be better. So having v1.8.3 instead of registry/repo:v1.8.3

I think we can add another label to store the registry and repo. 

For the docker.io part, I agree it's strange. I think we should define some generic labels for container properties that are not Docker-specific. I think we should get this PR merged and just refine the way forward and fix it in a follow-up

PR #37, opened 2022-09-07
⭐ Improvments for related assets for K8s

jaym:
- I've introduced an asset type for K8s nodes. It currently only shows up as a related asset, not a scanned asset
- From the K8s nodes, we can try to detect the cloud account. This works for AKS and GKE, as the node info has the subscription/project identifiers. This information is not available in the EKS node metadata. If this is detectable, we can link the cluster to the cloud account.
- The K8s node is linked to the underlying host node. If its a google cloud compute node or azure cloud compute node, we create the platform identifier from the project/subscription identifier + instance identifier. For EKS, this is skipped because we do not have the AWS account id.

### DIFF
--- a/motor/discovery/k8s/list_nodes.go
+++ b/motor/discovery/k8s/list_nodes.go
@@ -1,0 +1,149 @@
+package k8s
+
+import (
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/pkg/errors"
+	"go.mondoo.com/cnquery/motor/asset"
+	"go.mondoo.com/cnquery/motor/motorid/gce"
+	"go.mondoo.com/cnquery/motor/platform"
+	"go.mondoo.com/cnquery/motor/providers"
+	"go.mondoo.com/cnquery/motor/providers/k8s"
+	v1 "k8s.io/api/core/v1"
+)
+
+// ListNodes lits all nodes in the cluster.
+func ListNodes(p k8s.KubernetesProvider, connection *providers.Config, clusterIdentifier string, namespaceFilter []string) ([]*asset.Asset, []nodeRelationshipInfo, error) {
+	nodes, err := p.Nodes()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	assets := []*asset.Asset{}
+	nodeRelationshipInfos := []nodeRelationshipInfo{}
+	for i := range nodes {
+		node := nodes[i]
+		asset, err := createAssetFromObject(&node, p.Runtime(), connection, clusterIdentifier)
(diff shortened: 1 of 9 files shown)


motor/discovery/k8s/list_nodes_test.go
```diff
@@ -0,0 +1,323 @@
...
+
+import (
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/require"
+	"go.mondoo.com/cnquery/motor/providers"
+	"go.mondoo.com/cnquery/motor/providers/k8s"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
joelddiaz: This is imported twice simply under different names...
FWIW, i prefer the "corev1" naming.

motor/discovery/k8s/list_nodes.go
```diff
@@ -0,0 +1,150 @@
...
+	return assets, nodeRelationshipInfos, nil
+}
+
+type nodeRelationshipInfo struct {
+	cloudAccountAsset *asset.Asset
+	hostInstanceAsset *asset.Asset
+}
+
+var (
+	eksProviderIDInstanceIDRegexp = regexp.MustCompile("^aws:///[\\-0-9a-zA-Z]+/(.+)$")
```
czunker: Is the triple slash correct? GCE only has two. But Azure also has three. Just wondering...

motor/discovery/k8s/list_nodes.go
```diff
@@ -0,0 +1,150 @@
...
+	"go.mondoo.com/cnquery/motor/asset"
+	"go.mondoo.com/cnquery/motor/motorid/gce"
+	"go.mondoo.com/cnquery/motor/platform"
+	"go.mondoo.com/cnquery/motor/providers"
+	"go.mondoo.com/cnquery/motor/providers/k8s"
+	v1 "k8s.io/api/core/v1"
+)
+
+// ListNodes lits all nodes in the cluster.
+func ListNodes(p k8s.KubernetesProvider, connection *providers.Config, clusterIdentifier string, namespaceFilter []string, od *k8s.PlatformIdOwnershipDirectory) ([]*asset.Asset, []nodeRelationshipInfo, error) {
```
preslavgerchev: The `PlatformIdOwnershipDirectory` here is unused

motor/discovery/k8s/list_nodes.go
jaym: i think only azure should have 3. good catch. i can just remove this since its not used

PR #38, opened 2022-09-07
🐛 walk through /proc/sys in a mockable way

joelddiaz:
Using os.Open() when running tests locally bypasses the mocked up data in the toml files causing mocked up data to not be available in the test cases.

Access files through the afero.Fs interface to preserve the access to the mocked data.

### DIFF
--- a/resources/packs/core/kernel/manager.go
+++ b/resources/packs/core/kernel/manager.go
@@ -4,7 +4,6 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"path"
 	"strings"
 
 	"github.com/cockroachdb/errors"
@@ -14,6 +13,8 @@ import (
 
 	"go.mondoo.com/cnquery/motor"
 )
+
+const sysctlPath = "/proc/sys/"
 
 type KernelInfo struct {
 	Version   string            `json:"version"`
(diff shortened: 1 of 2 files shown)


resources/packs/core/kernel/manager.go
```diff
@@ -103,17 +104,24 @@ func (s *LinuxKernelManager) Info() (KernelInfo, error) {
 
 func (s *LinuxKernelManager) Parameters() (map[string]string, error) {
 	sysctlPath := "/proc/sys/"
+	cleanedSysctlPath := filepath.Clean(sysctlPath)
```
jaym: i'd be careful with filepath. The docs say its target dependent. So you might get a different result if this was run from a windows computer

imilchev: Why do we need the `filepath.Clean` in the first place when we are providing the path ourselves?

joelddiaz: Removing the trailing '/' from the filepath is the difference from having to mock up a TOML file named '/proc/sys' and '/proc/sys/'. Since all the other file access while performing the Walk() was done without the trailing '/', I removed it from the explicit '/proc/sys/' path for consistency.

joelddiaz: > i'd be careful with filepath. The docs say its target dependent. So you might get a different result if this was run from a windows computer

Ahh...you mean someone literally running this test from a Windows machine? Do we have any lying around to test this out with?

joelddiaz: Actually, it looks like we can just use the "path" package instead as it seems to be dedicated to the one-true-separator (ie '/').

resources/packs/core/kernel/manager.go
```diff
@@ -103,17 +105,24 @@ func (s *LinuxKernelManager) Info() (KernelInfo, error) {
 
 func (s *LinuxKernelManager) Parameters() (map[string]string, error) {
 	sysctlPath := "/proc/sys/"
```
imilchev: can we move that as a `const` on top of the file. I just saw that I had that change locally but never pushed it to my branch.

resources/packs/core/kernel/manager.go
```diff
@@ -103,17 +105,24 @@ func (s *LinuxKernelManager) Info() (KernelInfo, error) {
...
+	cleanedSysctlPath := path.Clean(sysctlPath)
```
imilchev: if we want to remove the `/` from `sysctlPath` then why not just fix the variable above to not have it and then remove the `Clean`? We are now cleaning our own input instead of just fixing it 😆

joelddiaz: I kept it there to make it more clear that paths with trailing slashes had to be handled appropriately. I suppose I can add a comment explaining what's going on...

PR #39, opened 2022-09-08
🧹 Do not use k8s manifest path in ID for CI/CD scans

imilchev:
This makes sure that we don't create 2 assets whenever we scan exactly the same k8s object via our admission controller. This is only needed until we move away from the manifest scanning for the admission controller

### DIFF
--- a/motor/providers/k8s/manifest_provider.go
+++ b/motor/providers/k8s/manifest_provider.go
@@ -122,7 +122,24 @@ func (t *manifestProvider) SupportedResourceTypes() (*resources.ApiResourceIndex
 }
 
 func (t *manifestProvider) ID() (string, error) {
-	_, err := os.Stat(t.manifestFile)
+	// If we are doing an admission control scan, we have 1 resource in the manifest and it has a UID.
+	// Instead of using the file path to generate the ID, use the resource UID. We do this because for
+	// CI/CD scans, the manifest is stored in a random file. This means we can potentially be scanning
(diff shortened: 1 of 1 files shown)


motor/providers/k8s/manifest_provider.go
```diff
@@ -122,7 +122,24 @@ func (t *manifestProvider) SupportedResourceTypes() (*resources.ApiResourceIndex
...
+	// Instead of using the file path to generate the ID, use the resource UID. We do this because for
+	// CI/CD scans, the manifest is stored in a random file. This means we can potentially be scanning
+	// the same resource multiple times but it will result in different assets because of the random
+	// file name.
+	resourceObjects, _, err := t.resourceIndex()
+	if err == nil {
+		if len(resourceObjects) == 1 {
+			o, err := meta.Accessor(resourceObjects[0])
+			if err == nil {
+				if o.GetUID() != "" {
```
czunker, line 135: Should we log something in case this didn't work?

We have the fallback with the manifest name, but for debugging, it might be nice to know, that we used the fallback.

imilchev: I intentionally didn't put any message there. If you scan a manifest with just 1 resource in it that you manually wrote, it won't have a UID. Then it will go through this code but because it won't find the UID, it will go down and use the filepath hash. In that case it is expected to have the hash. If you happen to have the UID, then it will be used but that's also fine (I guess).

PR #40, opened 2022-09-13
⭐️ Allow k8s resource filtering via inventory options

imilchev:
Allow filtering k8s resources such that we scan only specific ones instead of all of them. This is needed for scanning resources on create/update via the Mondoo operator.

An open question is what do we do with the cluster asset. Currently, it is still being discovered and scanned even when the resource filters are present. I can remove it but then we won't be able to link the specific resource to the cluster. I could add the relationship but do not discover the cluster but I think that would mean if the cluster asset is not already present in the space, the scan will fail. I am open for suggestions.

P.S.: I also added a settings file for VSCode, which is the same as the one that we have to the mondoo repo.

### DIFF
--- a/motor/discovery/k8s/resolver.go
+++ b/motor/discovery/k8s/resolver.go
@@ -130,30 +130,35 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 		}
 	}
 
-	clusterAsset := &asset.Asset{
-		PlatformIds: []string{clusterIdentifier},
-		Name:        clusterName,
-		Platform:    pf,
-		Connections: []*providers.Config{tc}, // pass-in the current config
-		State:       asset.State_STATE_RUNNING,
-	}
-	resolved = append(resolved, clusterAsset)
-
-	ownershipDir := k8s.NewEmptyPlatformIdOwnershipDirectory(clusterIdentifier)
-
-	// nodes are only added as related assets because we have no policies to scan them
-	nodes, nodeRelationshipInfos, err := ListNodes(p, tc, clusterIdentifier, namespacesFilter)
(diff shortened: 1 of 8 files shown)


motor/discovery/k8s/resolver.go
```diff
@@ -122,39 +130,51 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
...
+	if err != nil {
+		return nil, err
 	}
-	resolved = append(resolved, clusterAsset)
 
-	ownershipDir := k8s.NewEmptyPlatformIdOwnershipDirectory(clusterIdentifier)
+	// Only discover cluster and nodes if there are no resource filters.
+	var clusterAsset *asset.Asset
+	var ownershipDir *k8s.PlatformIdOwnershipDirectory
+	if len(resourcesFilter) == 0 {
```
chris-rock, line 141: so does that mean going forward that I cannot scan the cluster only itself? once we activate the resource resource as default, we run into the issue that we cannot scan the cluster only.

imilchev: at the moment we cannot, indeed. There is also no use for it, so I intentionally left it out. If we want it, we can easily add it.

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
tas50, line 1: Thanks for adding this. These are important for OSS

imilchev: If that's the case we should add it to the operator as well. I will open a PR.

motor/discovery/k8s/resolver.go
chris-rock: Then lets merge it as is.

PR #42, opened 2022-09-15
⭐ Add kubelet as MQL resource

czunker:


resources/packs/k8s/kubelet.go
```diff
@@ -0,0 +1,131 @@
...
+		return nil, nil, err
+	}
+	(*args)["process"] = p
+
+	kubeletParameters, err := p.Flags()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Check kublet for "--config" flag and set path to config file accordingly
```
imilchev: nit: typo

resources/packs/k8s/kubelet.go
```diff
@@ -0,0 +1,131 @@
...
+		return args, nil, nil
+	}
+
+	p, err := k.getKubeletProcess()
+	if err != nil {
+		return nil, nil, err
+	}
+	(*args)["process"] = p
+
+	kubeletParameters, err := p.Flags()
```
imilchev: can we unify the naming a bit? The function is called `Flags`, the variable is `Parameters` and the MQL property is `options`

resources/packs/k8s/kubelet.go
czunker: fixed

resources/packs/k8s/kubelet.go
czunker: I tried to make it more clear.

We actually have three things:
- flags: kubelet CLI flags
- the kubelet config file settings/parameters
- options: this is a merge of the defaults, the cli flags and the config file

The MQL controls should use the options.
In case you have a better name for the `options` field, I'm open to changing this. That's the best that came to my mind. Naming things ...

imilchev: Isn't `configuration` or `config` a better name?

czunker: Switched to configuration. Thanks for the suggestion.

PR #43, opened 2022-09-15
⭐️ Create relationships when using AWS discovery

jaym:


motor/discovery/aws/resolver.go
```diff
@@ -98,6 +98,7 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 			}
 			log.Debug().Int("instances", len(assetList)).Msg("completed ssm instance search")
 			for i := range assetList {
+				assetList[i].RelatedAssets = append(assetList[i].RelatedAssets, resolved[0])
```
preslavgerchev: Is resolved[0] the account asset? I'd personally suggest having this as a separate variable so it's clear

vjeffrey: nit: a comment on this line explaining that resolved[0] is the account associated with the instances would be useful, it took me a sec

vjeffrey: heh, i hadn't seen preslav had already made a similar suggestion

PR #46, opened 2022-09-16
Lumi embedded resources

preslavgerchev:
This allows us to embed resources into other resources, i.e.
```
docker.container {
 ....
 embed os.any
}
```
thus allowing queries such as 
```
docker.containers[0].hostname
```

the embedded resources can also be aliased to control how they get generated under the hood
```
docker.container {
 embed os.any as myalias
}
```

internally, a new property gets assigned to the object, which (in further PRs) will allow us to open a connection to that resource and switch context to it


resources/lr/cli/cmd/markdown.go
```diff
@@ -228,11 +228,17 @@ func (l *lrSchemaRenderer) renderResourcePage(resource *lr.Resource, schema *res
 		builder.WriteString("\n\n")
 	}
 
+	inits := []*lr.Init{}
+	for _, f := range resource.Body.Fields {
+		if f.Init != nil {
+			inits = append(inits, f.Init)
+		}
+	}
```
arlimus: nit: I think this is used twice, so we could think about extracting it into its own call (to get all the inits from a resource body fields)

resources/lr/lr.go
```diff
@@ -96,11 +96,19 @@ type SimpleType struct {
...
 }
 
-// Init calls
+// ResourceDef carrying the definition of the field
+// nolint: govet
+type Field struct {
+	Comments   []string    `{ @Comment }`
+	Init       *Init       `( @@ `
+	Embeddable *Embeddable `| @@`
+	BasicField *BasicField `| @@ )?`
```
arlimus, line 108: great simplification! :+1:

resources/lr/cli/cmd/markdown.go
preslavgerchev: good catch, extract that to a func

PR #58, opened 2022-09-19
🐛 Write recorded data to file

czunker:


cli/shell/shell.go
```diff
@@ -297,6 +297,11 @@ func (s *Shell) RunOnce(cmd string) (*llx.CodeBundle, map[string]*llx.RawResult,
 		panic(err)
 	}
 
+	// run onClose handler if set
```
chris-rock, line 300: This approach prevents users of Shell to run multiple queries with RunOnce. Alternative is available here https://github.com/mondoohq/cnquery/pull/59

PR #60, opened 2022-09-19
🐛 Check if /proc/sys/ files can be read before retrieving kernel parameters

preslavgerchev:
While going through `/proc/sys/` files to get kernel params on a linux machine, check if the file can be read before trying to fetch it.

### DIFF
--- a/resources/packs/core/kernel/manager.go
+++ b/resources/packs/core/kernel/manager.go
@@ -111,6 +111,18 @@ func (s *LinuxKernelManager) Parameters() (map[string]string, error) {
 	kernelParameters := make(map[string]string)
 	err := fsUtil.Walk(sysctlPath, func(path string, f os.FileInfo, err error) error {
 		if f != nil && !f.IsDir() {
+			stat, err := s.provider.FS().Stat(path)
+			if err != nil {
+				log.Error().Err(err)
+				return nil
+			}
+			details := pOs.FileModeDetails{
+				FileMode: stat.Mode(),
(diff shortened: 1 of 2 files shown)


resources/packs/core/kernel/manager.go
```diff
@@ -119,8 +131,8 @@ func (s *LinuxKernelManager) Parameters() (map[string]string, error) {
 
 			content, err := io.ReadAll(f)
 			if err != nil {
-				log.Error().Err(err)
-				return err
+				log.Error().Err(err).Msg("cannot read content")
+				return nil
```
preslavgerchev, line 135: There are some files that are readable yet they cannot be read, hence why I decided to just log the err and continue, that way we get the other params normally

chris-rock: This is also the same behaviour as sysctl is doing, those kernel parameters do not show up there.

imilchev: The writeable-only files are not supposed to be read in the first place. I think you don't need to log an error here

preslavgerchev: But there are files that re readable but cannot be read, i.e. 
image

Explanation of this particular case: https://unix.stackexchange.com/questions/251401/cannot-read-key-net-ipv6-conf-all-stable-secret-in-sysctl (thanks to @atomic111)

In those cases reading the file content will fail so this is why I chose to return nil here

PR #63, opened 2022-09-19
⭐️ Allow discovering k8s AdmissionReview assets

imilchev:
- [x] Discover `AdmissionReview` assets
- [x] Discover workloads from `AdmissionReview` requests
- [x] Specify `AdmissionReview` via inventory options
- [x] Unit tests


motor/discovery/k8s/list_admission_reviews.go
```diff
@@ -0,0 +1,100 @@
...
+
+		assets = append(assets, asset)
+	}
+
+	return assets, nil
+}
+
+func assetFromAdmissionReview(a admissionv1.AdmissionReview, runtime string, connection *providers.Config, clusterIdentifier string) (*asset.Asset, error) {
+	// Use the meta from the request object.
+	obj, err := resources.ResourcesFromManifest(bytes.NewReader(a.Request.Object.Raw))
```
czunker, line 48: Do we need this because the object does not implement runtime.Object?

motor/discovery/k8s/list_admission_reviews_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	s := json.NewYAMLSerializer(json.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
+
+	var b bytes.Buffer
+	foo := io.Writer(&b)
+	err := s.Encode(&pod, foo)
+	require.NoError(t, err)
+	// Seed AdmissionReviews
+	admissionReviews := []admissionv1.AdmissionReview{
+		{
```
czunker, line 52: Shouldn't we check here for ObjectMeta because of the Workaround?

motor/discovery/k8s/resolver_test.go
```diff
@@ -35,6 +36,29 @@ func TestManifestResolver(t *testing.T) {
...
...AgICAgICAgICAgICAgICAgICAia2V5Ijoibm9kZS5rdWJlcm5ldGVzLmlvL25vdC1yZWFkeSIsCiAgICAgICAgICAgICAgICAgICAgICJvcGVyYXRvciI6IkV4aXN0cyIsCiAgICAgICAgICAgICAgICAgICAgICJlZmZlY3QiOiJOb0V4ZWN1dGUiLAogICAgICAgICAgICAgICAgICAgICAidG9sZXJhdGlvblNlY29uZHMiOjMwMAogICAgICAgICAgICAgICAgICB9LAogICAgICAgICAgICAgICAgICB7CiAgICAgICAgICAgICAgICAgICAgICJrZXkiOiJub2RlLmt1YmVybmV0ZXMuaW8vdW5yZWFjaGFibGUiLAogICAgICAgICAgICAgICAgICAgICAib3BlcmF0b3IiOiJFeGlzdHMiLAogICAgICAgICAgICAgICAgICAgICAiZWZmZWN0IjoiTm9FeGVjdXRlIiwKICAgICAgICAgICAgICAgICAgICAgInRvbGVyYXRpb25TZWNvbmRzIjozMDAKICAgICAgICAgICAgICAgICAgfQogICAgICAgICAgICAgICBdLAogICAgICAgICAgICAgICAicHJpb3JpdHkiOjAsCiAgICAgICAgICAgICAgICJlbmFibGVTZXJ2aWNlTGlua3MiOnRydWUsCiAgICAgICAgICAgICAgICJwcmVlbXB0aW9uUG9saWN5IjoiUHJlZW1wdExvd2VyUHJpb3JpdHkiCiAgICAgICAgICAgIH0sCiAgICAgICAgICAgICJzdGF0dXMiOnsKICAgICAgICAgICAgICAgInBoYXNlIjoiUGVuZGluZyIsCiAgICAgICAgICAgICAgICJxb3NDbGFzcyI6IkJlc3RFZmZvcnQiCiAgICAgICAgICAgIH0KICAgICAgICAgfSwKICAgICAgICAgIm9sZE9iamVjdCI6bnVsbCwKICAgICAgICAgImRyeVJ1biI6ZmFsc2UsCiAgICAgICAgICJvcHRpb25zIjp7CiAgICAgICAgICAgICJraW5kIjoiQ3JlYXRlT3B0aW9ucyIsCiAgICAgICAgICAgICJhcGlWZXJzaW9uIjoibWV0YS5rOHMuaW8vdjEiCiAgICAgICAgIH0KICAgICAgfQogICAKfQo=",
```
czunker: What is this string? A complete object base64 encoded?

motor/discovery/k8s/resolver.go
```diff
@@ -19,15 +19,16 @@ import (
...
+	DiscoveryAll              = "all"
+	DiscoveryPods             = "pods"
+	DiscoveryJobs             = "jobs"
+	DiscoveryCronJobs         = "cronjobs"
+	DiscoveryStatefulSets     = "statefulsets"
+	DiscoveryDeployments      = "deployments"
+	DiscoveryReplicaSets      = "replicasets"
+	DiscoveryDaemonSets       = "daemonsets"
+	DiscoveryContainerImages  = "container-images"
+	DiscoveryAdmissionReviews = "admissionreviews"
```
czunker, line 31: I do not understand why we need this discovery option. On the other hand, the AdmissionReviews have their own provider. Are they actually discoverable?

motor/providers/k8s/admission_provider.go
```diff
@@ -0,0 +1,157 @@
...
+	"go.mondoo.com/cnquery/motor/platform"
+	"go.mondoo.com/cnquery/motor/providers"
+	"go.mondoo.com/cnquery/motor/providers/k8s/resources"
+	os_provider "go.mondoo.com/cnquery/motor/providers/os"
+	"go.mondoo.com/cnquery/motor/providers/os/fsutil"
+	admission "k8s.io/api/admission/v1"
+	"k8s.io/apimachinery/pkg/version"
+)
+
+func newAdmissionProvider(data string, selectedResourceID string) (KubernetesProvider, error) {
```
czunker, line 20: The spec mentioned, the request should be read from a file:
> disk and then starting the scan from the disk

So, should a path be passed here?

motor/discovery/k8s/list_admission_reviews.go
imilchev: It does implement `runtime.Object` but it has no meta information. It has no properties but the request and the response

motor/discovery/k8s/resolver_test.go
imilchev: this is how the admission review scanning should be invoked. The string is a base64-encoded `admissionv1.AdmissionReview` object

motor/discovery/k8s/resolver.go
imilchev: with this implementation they are. This allows us to decide whether we want to see the admissionreview assets in the UI or not. If this flag is not provided a k8s CI/CD scan will create 1 asset and that will be the resource being created/modified. If the flag is enabled, there will be 2 assets created - the admission review and the object itself

motor/providers/k8s/admission_provider.go
imilchev: It's not supposed to use the disk. This is also what the spec says. It says ideally, it shouldn't create any files on the disk and start the scan from there 😆

motor/discovery/k8s/list_admission_reviews_test.go
imilchev: I guess I don't understand... what should we check?

motor/discovery/k8s/list_admission_reviews.go
czunker: Ah, ok got that wrong on discord. This explains it: https://github.com/kubernetes/api/issues/44

motor/discovery/k8s/list_admission_reviews_test.go
czunker: As you had to add this to get meta data:
```
         // Use the meta from the request object.
	obj, err := resources.ResourcesFromManifest(bytes.NewReader(a.Request.Object.Raw))
```
Should we check this part of the code? It's different from what we do for other objects.

motor/discovery/k8s/resolver_test.go
czunker: I find it hard to read and perhaps we will wonder in the future why this is the way it is.

Could you re-use this part of the admission test:
https://github.com/mondoohq/cnquery/pull/63/files#diff-8f59126e36e6934269bceb2193979631be10917e0b987edac4fc157f1269850fR13-R17

Or is the base64 encoded data there for a specific test? Perhaps a comment would make this clear?

imilchev: will do that

motor/discovery/k8s/list_admission_reviews_test.go
imilchev: This part is checked in this test. The properties are verified. If this code fails, the test will fail too

PR #70, opened 2022-09-20
🐛Fix bugs related to k8s cicd assets showing up as unscored

jaym:
They boil down to parsing platform identifiers. CICD backend changes them. The rest of the code assumes they dont change. We should avoid parsing platform identifiers

### DIFF
--- a/motor/discovery/k8s/asset_data_test.go
+++ b/motor/discovery/k8s/asset_data_test.go
@@ -53,7 +53,7 @@ func TestAssetData(t *testing.T) {
 	assert.Equal(t, "batch/v1", asset.Platform.Version)
 	assert.Equal(t, "k8s-cronjob", asset.Platform.Name)
 	assert.ElementsMatch(t, []string{"k8s", "k8s-workload"}, asset.Platform.Family)
-	assert.Equal(t, "test123", asset.Labels["namespace"])
+	assert.Equal(t, "test123", asset.Labels["k8s.mondoo.com/namespace"])
 }
 
 func TestAssetNodeData(t *testing.T) {
--- a/motor/discovery/k8s/list_admission_reviews_test.go
+++ b/motor/discovery/k8s/list_admission_reviews_test.go
@@ -95,5 +95,5 @@ func TestListAdmissionReviews(t *testing.T) {
 	assert.Equal(t, "admission.k8s.io/v1", assets[0].Platform.Version)
 	assert.Equal(t, "k8s-admission", assets[0].Platform.Name)
 	assert.ElementsMatch(t, []string{"k8s", "k8s-admission"}, assets[0].Platform.Family)
-	assert.Equal(t, pod.Namespace, assets[0].Labels["namespace"])
+	assert.Equal(t, pod.Namespace, assets[0].Labels["k8s.mondoo.com/namespace"])
 }
--- a/motor/discovery/k8s/list_cronjobs_test.go
+++ b/motor/discovery/k8s/list_cronjobs_test.go
@@ -158,7 +158,7 @@ func TestListCronJobs(t *testing.T) {
 	assert.Equal(t, "batch/v1", assets[0].Platform.Version)
 	assert.Equal(t, "k8s-cronjob", assets[0].Platform.Name)
 	assert.ElementsMatch(t, []string{"k8s", "k8s-workload"}, assets[0].Platform.Family)
-	assert.Equal(t, nss[0].Name, assets[0].Labels["namespace"])
+	assert.Equal(t, nss[0].Name, assets[0].Labels["k8s.mondoo.com/namespace"])
 }
 
 func TestListCronJobs_Filter(t *testing.T) {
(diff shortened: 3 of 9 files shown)


resources/packs/k8s/common.go
```diff
@@ -92,6 +92,14 @@ func k8sResourceToMql(r *resources.Runtime, kind string, fn resourceConvertFn) (
 	return resp, nil
 }
 
+func getNameAndNamespace(runtime *resources.Runtime) (string, string, error) {
+	asset := runtime.Motor.GetAsset()
+	if asset == nil || asset.Labels == nil {
+		return getPlatformIdentifierElements(runtime.Motor.Provider)
+	}
+	return asset.Labels["k8s.mondoo.com/name"], asset.Labels["k8s.mondoo.com/namespace"], nil
```
czunker, line 100: Nice. I wasn't aware of these labels. That's why I used the selectedResourceID/PlatformIdentifier.

czunker, line 100: `asset.Labels["k8s.mondoo.com/namespace"]` will not be present for non-namespaced objects. I'm not sure how we handle this currently, or if we are doing that at all. Only thing that comes to my mind is the new `AdmissionReview` object. Perhaps @imilchev can help.

resources/packs/k8s/common.go
```diff
@@ -92,6 +92,14 @@ func k8sResourceToMql(r *resources.Runtime, kind string, fn resourceConvertFn) (
 	return resp, nil
 }
 
+func getNameAndNamespace(runtime *resources.Runtime) (string, string, error) {
+	asset := runtime.Motor.GetAsset()
+	if asset == nil || asset.Labels == nil {
```
czunker, line 97: More talking to myself while testing:

The fleet asset has labels, but not necessarily the `k8s.mondoo.com/*` labels. So, this test passes but later the code panics because it tries to access the `k8s.mondoo.com/*` labels.

resources/packs/k8s/common.go
czunker: Ok, now I know why I wasn't aware of these labels. They aren't present for the fleet assets.

motor/discovery/k8s/asset_data.go
```diff
@@ -97,9 +97,30 @@ func createAssetFromObject(object runtime.Object, runtime string, connection *pr
 		name = ns + "/" + objMeta.GetName()
 		platformData.Labels["namespace"] = ns
 		assetLabels["namespace"] = ns
```
imilchev: do we want to keep this label? I think we can remove it and just use the prefixed one for consistency

motor/discovery/k8s/list_admission_reviews.go
```diff
@@ -82,10 +82,24 @@ func assetFromAdmissionReview(a admissionv1.AdmissionReview, runtime string, con
...
 
+	assetLabels["k8s.mondoo.com/name"] = objMeta.GetName()
+	if string(objMeta.GetUID()) != "" {
+		// objects discoverd from manifest do not neccecarily have a UID
+		assetLabels["k8s.mondoo.com/uid"] = string(objMeta.GetUID())
+	}
+	assetLabels["k8s.mondoo.com/kind"] = objectKind
+	assetLabels["k8s.mondoo.com/apiVersion"] = objType.GetAPIVersion()
+	if objMeta.GetResourceVersion() != "" {
+		// objects discoverd from manifest do not neccecarily have a resource version
```
imilchev: nit: typo

motor/discovery/k8s/list_admission_reviews.go
```diff
@@ -82,10 +82,24 @@ func assetFromAdmissionReview(a admissionv1.AdmissionReview, runtime string, con
...
 		platformData.Labels["namespace"] = ns
 		assetLabels["namespace"] = ns
+		assetLabels["k8s.mondoo.com/namespace"] = ns
 	} else {
 		name = objMeta.GetName()
 	}
 
+	assetLabels["k8s.mondoo.com/name"] = objMeta.GetName()
+	if string(objMeta.GetUID()) != "" {
+		// objects discoverd from manifest do not neccecarily have a UID
```
imilchev: here too

motor/discovery/k8s/asset_data.go
```diff
@@ -97,9 +97,30 @@ func createAssetFromObject(object runtime.Object, runtime string, connection *pr
...
+		assetLabels["k8s.mondoo.com/namespace"] = ns
 	} else {
 		name = objMeta.GetName()
 	}
+	assetLabels["k8s.mondoo.com/name"] = objMeta.GetName()
```
imilchev, line 122: this is copy-pasted twice. Perhaps you can define a function for setting the labels that can be re-used between the admission review stuff and the other assets generation

motor/discovery/k8s/asset_data.go
czunker: Removed.

motor/discovery/k8s/list_admission_reviews.go
czunker: Fixed

motor/discovery/k8s/list_admission_reviews.go
czunker: Fixed

motor/discovery/k8s/asset_data.go
czunker: Extracted it to a function.

PR #76, opened 2022-09-21
🧹 replace ioutil.ReadFile with os.ReadFile

chris-rock:
replaces the deprecated ioutil.ReadFile method with os.ReadFile

### DIFF
--- a/cli/shell/shell.go
+++ b/cli/shell/shell.go
@@ -3,7 +3,6 @@ package shell
 import (
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
 	"os/signal"
 	"path"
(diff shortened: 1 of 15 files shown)


resources/runtime.go
```diff
@@ -64,7 +64,6 @@ func (c *Cache) Delete(key string) { c.Map.Delete(key) }
 type UpstreamConfig struct {
 	AssetMrn    string
 	SpaceMrn    string
-	Collector   string
```
imilchev: is this change intended since it doesn't really fit into the context of this PR? Just making sure it doesn't disappear by accident 😆

chris-rock: That is a change for another PR, will remove it from this here

PR #88, opened 2022-09-22
⭐️ Add AdmissionReview MQL resource

imilchev:


resources/packs/k8s/admission.go
```diff
@@ -0,0 +1,116 @@
...
+		"username", userInfo.Username,
+		"uid", userInfo.UID)
+}
+
+func (k *mqlK8sAdmissionreview) id() (string, error) {
+	return k.Id()
+}
+
+func (r *mqlK8sAdmissionreview) init(args *resources.Args) (*resources.Args, K8sAdmissionreview, error) {
+	return args, nil, nil
```
imilchev: this doesn't work for the case when we want to scan a single `k8s.admissionreview` or if we write a query like `k8s.admissionreview(id: 'admission:default:test-dep-5f65697f8d-fxclr'){ id request { * } }`. Retrieving the `AdmissionRequest` and its `UserInfo` depends on the resources being present in the lumi cache at the moment. When we init the resource from args, then those cannot be retrieved because there is no cache. What is the best way to tackle this?

resources/packs/k8s/k8s.lr
```diff
@@ -688,4 +690,36 @@ k8s.customresource {
...
   // Full resource manifest
   manifest dict
+}
+
+// Kubernetes AdmissionReview
+k8s.admissionreview {
+  // Mondoo ID for Kubernetes AdmissionReview
+  id string
+  // The requested admission
+	request() k8s.admissionrequest
```
chris-rock: nit: whitespace

resources/packs/k8s/admission.go
chris-rock: Right now this implementation essentially does nothing. If that is intended, you can just remove it because `init` is optional.

If you want do make it easier to ask for a specific event, then you need to load all and filter. A good example for that is https://github.com/mondoohq/cnquery/blob/main/resources/packs/core/packages.go#L16-L40

resources/packs/k8s/k8s.lr
imilchev: tabs and whitespaces... fixed it now

PR #105, opened 2022-09-26
🐛 Adjust test reporter permissions

imilchev:
Seems like the suggested fix for this error is to modify the permissions. Not sure why this suddenly stopped working though.  For reference https://github.com/dorny/test-reporter/issues/149#issuecomment-1004820190

### DIFF
--- a/.github/workflows/test-report.yml
+++ b/.github/workflows/test-report.yml
@@ -8,7 +8,6 @@ on:
 jobs:
   report:
     permissions:
-      id-token: write
       contents: read
       checks: write
     runs-on: ubuntu-latest


.github/workflows/test-report.yml
```diff
@@ -7,6 +7,10 @@ on:
       - completed
 jobs:
   report:
+    permissions:
+      id-token: write
```
chris-rock: Do we know why this needs a write token?

imilchev: It's writing some markdown to the checks page. That's the tests table. I see that the operator runs into exactly the same problem. I assume github has changed something... we can try this solution and if it doesn't work I can look for other test reporting solutions

chris-rock: checks: write makes sense to me, but with an id token write you can manipulate the repository

imilchev: let's try it without. I saw someone opened this PR https://github.com/dorny/test-reporter/pull/174/files 

I wonder if `contents: write` is needed too. I want to try without it first

imilchev: This is the other solution that I have in mind in case this one doesn't work https://github.com/test-summary/action

PR #115, opened 2022-09-27
⭐️ access to all resources in terraform state file

chris-rock:
before this PR we had to iterate over all terraform modules to get access to the resources:

```graphql
cnquery> terraform.state.rootModule.resources { providerName == "registry.terraform.io/hashicorp/null" }
terraform.state.rootModule.resources: [
  0: {
    providerName == "registry.terraform.io/hashicorp/null": true
  }
```

With this change it is much easier to access resources in a state file:

```graphql
cnquery> terraform.state.resources
terraform.state.resources: [
  0: terraform.state.resource id = null_resource.ls
]
```

### DIFF
--- a/resources/packs/terraform/terraform.lr
+++ b/resources/packs/terraform/terraform.lr
@@ -93,7 +93,7 @@ terraform.state {
   rootModule() terraform.state.module
   // flat list of all modules
   modules() []terraform.state.module
-  // flat list of all resources of all modules
+  // A flat list of all resources across all modules
   resources() []terraform.state.resource
 }
 


resources/packs/terraform/terraform.lr
```diff
@@ -93,6 +93,8 @@ terraform.state {
   rootModule() terraform.state.module
   // flat list of all modules
   modules() []terraform.state.module
+  // flat list of all resources of all modules
```
scottford-io:
```suggestion
  // A flat list of all resources across all modules
```

PR #129, opened 2022-09-28
🐛 assign correct name and type to aws account when relationship is created

jaym:


motor/motorid/awsec2/id_test.go
```diff
@@ -21,5 +22,11 @@ func TestParseInstanceId(t *testing.T) {
 	path = "//platformid.api.mondoo.app/runtime/aws/ec2/v1/accounts/185972265011/regions/us/instances/i-07f67838ada5879af"
 	id, err = ParseMondooInstanceID(path)
 	assert.Error(t, err, "invalid mondoo aws ec2 instance id")
```
scottford-io: Are these error messages only for us on the backend? Do we need to use the word "mondoo" here? 

```suggestion
	assert.Error(t, err, "invalid aws ec2 instance id")
```

PR #134, opened 2022-09-29
DRAFT: 🐛 Fail k8s discovery for non-existing namespaces

imilchev:
Fail if the namespace filter specifies a namespace that doesn't exist. You can try it with: `cnquery shell k8s --option namespace=bogus`

```
cnquery shell k8s --option namespace=bogus                                          
→ loaded configuration from /Users/ivanmilchev/.config/mondoo/mondoo.yml using source default
x could not connect to asset error="namespace bogus does not exist" asset=
FTL could not resolve assets
```


czunker:
This is superseded by https://github.com/mondoohq/cnquery/pull/256

PR #135, opened 2022-09-29
🐛Fix tags filtering for AWS instances.

preslavgerchev:
We already prepend the tags with `tag` here:https://github.com/mondoohq/cnquery/blob/main/motor/discovery/aws/ec2_instances.go#L92

Therefore, until now `dev:dev` would equal to `tag:tag:dev:dev`, this removes the redundant 2nd tag prepending

### DIFF
--- a/motor/discovery/aws/resolver.go
+++ b/motor/discovery/aws/resolver.go
@@ -86,23 +86,21 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 	instancesPlatformIdsMap := map[string]*asset.Asset{}
 	// discover ssm instances
 	if tc.IncludesDiscoveryTarget(DiscoveryAll) || tc.IncludesDiscoveryTarget(DiscoverySSM) {
-		if val := discoverFilter["ssm"]; val == "true" {
-			// create a map to track the platform ids of the ssm instances, to avoid duplication of assets
-			s, err := NewSSMManagedInstancesDiscovery(provider.Config())
-			if err != nil {
-				return nil, errors.Wrap(err, "could not initialize aws ec2 ssm discovery")
-			}
-			s.FilterOptions = AssembleEc2InstancesFilters(discoverFilter)
-			assetList, err := s.List()
-			if err != nil {
-				return nil, errors.Wrap(err, "could not fetch ec2 ssm instances")
-			}
-			log.Debug().Int("instances", len(assetList)).Msg("completed ssm instance search")
-			for i := range assetList {
-				assetList[i].RelatedAssets = append(assetList[i].RelatedAssets, resolvedRoot)
-				log.Debug().Str("name", assetList[i].Name).Msg("resolved ssm instance")
-				instancesPlatformIdsMap[assetList[i].PlatformIds[0]] = assetList[i]
-			}
+		// create a map to track the platform ids of the ssm instances, to avoid duplication of assets
+		s, err := NewSSMManagedInstancesDiscovery(provider.Config())
+		if err != nil {
+			return nil, errors.Wrap(err, "could not initialize aws ec2 ssm discovery")
+		}
+		s.FilterOptions = AssembleEc2InstancesFilters(discoverFilter)
+		assetList, err := s.List()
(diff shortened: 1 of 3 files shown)


motor/discovery/aws/resolver.go
```diff
@@ -158,9 +158,7 @@ func AssembleEc2InstancesFilters(opts map[string]string) Ec2InstancesFilters {
 		for _, tagkv := range tags {
 			tag := strings.Split(tagkv, "=")
 			if len(tag) == 2 {
-				// to use tag filters with aws, we have to specify tag:KEY for the key, and then put the value as the values
-				key := "tag:" + tag[0]
-				ec2InstancesFilters.Tags[key] = tag[1]
+				ec2InstancesFilters.Tags[tag[0]] = tag[1]
```
vjeffrey, line 159: 👍 i think we may also need to adjust line 122 of https://github.com/mondoohq/cnquery/blob/main/motor/discovery/aws/ssm_instances.go ?

preslavgerchev: Good point, I will try this with an SSM instance (once I get one up and running)

motor/discovery/aws/ssm_instances.go
```diff
@@ -190,12 +190,11 @@ func ssmInstanceToAsset(account string, region string, instance types.InstanceIn
 		for j := range tagresp.Tags {
 			tag := tagresp.Tags[j]
 			if tag.Key != nil {
-				key := ImportedFromAWSTagKeyPrefix + *tag.Key
```
preslavgerchev: @vjeffrey do we still need this prefix here, is that used by the UI? I found it difficult to filter tags as I'd need to prepend `aws.tag` to each to have the search work. But then what if we're searching just for a mondoo-platform tag?

motor/discovery/aws/resolver.go
```diff
@@ -86,23 +86,21 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 	instancesPlatformIdsMap := map[string]*asset.Asset{}
 	// discover ssm instances
 	if tc.IncludesDiscoveryTarget(DiscoveryAll) || tc.IncludesDiscoveryTarget(DiscoverySSM) {
-		if val := discoverFilter["ssm"]; val == "true" {
```
preslavgerchev, line 89: This extra check was preventing SSM discovery, seems redundant with the discovery target check above it

motor/discovery/aws/ssm_instances.go
vjeffrey: a user trying to filter by tags should be able to filter without the additional aws.tag thing. and then we should be able to adjust to make it match. we do use it in the ui, theres code that looks for that aws.tag prefix, to differentiate between other tags and tags that came from aws. u can only filter by aws tags at this point, bc we're calling the aws api with that filter applied

motor/discovery/aws/resolver.go
vjeffrey: oh, i just noticed we're not prepending it with "tag:" anymore ? the aws api requires that for searching i think?

motor/discovery/aws/resolver.go
vjeffrey: ya im not sure why that was ever there! 🤷

motor/discovery/aws/resolver.go
preslavgerchev: yeah, that now (only) happens here: https://github.com/mondoohq/cnquery/blob/main/motor/discovery/aws/ec2_instances.go#L92.

The original problem was that we were doing it in 2 places so we were getting the prefix twice instead of once

motor/discovery/aws/ssm_instances.go
preslavgerchev: Fixed here: https://github.com/mondoohq/cnquery/pull/135/files#diff-4c571e1b4077280a59097de1dd35cd6209b23fd5bd42338de4d9eed7f42dbe44R117

I wanted to do any labels modifications outside of the filter functions such that those remain reusable

PR #137, opened 2022-09-29
⭐️ asset context for resources

jaym:
We're adding the ability the change context within a resource. For example, when running a query against `local`, we should be able to get the docker containers. Each of these containers is now a different context, where you could access the same information you do for the `local` target (packages, hostname, etc).

> Examples below require using the MQLAssetContext feature. For example,
>  FEATURES=MQLAssetContext go run apps/cnquery/cnquery.go run local -c 'docker.containers[0].hostname'

Below is an example using docker:
```
% docker ps -a
CONTAINER ID   IMAGE           COMMAND                  CREATED       STATUS                  PORTS                                       NAMES
dac5706945fb   postgres        "docker-entrypoint.s…"   10 days ago   Up 13 minutes           0.0.0.0:5432->5432/tcp, :::5432->5432/tcp   dev-pg
```

We can run a mql query against a local target to query the docker container and get the container's hostname:
```
> docker.containers[0].hostname
```
```
docker.containers[0].os.unix.base.hostname: "dac5706945fb"
```

which is different than
```
> os.hostname
```
```
os.hostname: "mars"
```

Or a query to list the packages for that container:
```
> docker.containers[0].packages.list
```
```
docker.containers[0].os.base.packages.list: [
  0: package id = deb://adduser/3.118/all
  1: package id = deb://apt/2.2.4/amd64
  2: package id = deb://base-files/11.1+deb11u5/amd64
  3: package id = deb://base-passwd/3.5.51/amd64
  4: package id = deb://bash/5.1-2+deb11u1/amd64
  ...
]
```

## Lumi Changes
There are already changes made to Lumi (https://github.com/mondoohq/cnquery/pull/46) that will also be explained here for completeness when it comes to asset context

### Embedded Types
To get the above example working (`docker.containers[0].hostname`) we need to link the `docker.container` lumi resource and the `os.base` resource that provides the `hostname` property. This is where embedded types come in:
```
os.base {
  hostname string
}

docker.container {
   id string
   ...
   embed os.base
}
```
What the embed here means essentially is: "take the os.base resource and embed that as a property".
The generated go code for this looks as the following:
```
type DockerContainer interface {
        .....
	Os() (resources.ResourceType, error)
	Id() (string, error)
        ......
}
```
The return type is now a `ResourceType`  as this can be used to instantiate an `os.base` resource when switching context to the docker container one. This changes allows the MQL compiler to look up `hostname` when doing `docker.containers[0].hostname` as it now can go through the `os` property - the fully resolved resource path in that case would be `docker.containers[0].os.base.hostname`.

### Aliases
Aliases for embedded fields are supported to allow changing the name that gets generated:
```
docker.container {
   id string
   ...
   embed os.base as dockeros
}
```

```
type DockerContainer interface {
        .....
	Dockeros() (resources.ResourceType, error)
	Id() (string, error)
        ......
}
```

### Changing Context
We now have the ability to change the motor (provider and asset) when creating a resource. The `resource.Runtime` has a new method 
```
CreateResourceWithAssetContext(name string, a *asset.Asset, p providers.Instance, args ...interface{}) (ResourceType, error)
```
that an implementation to change the context. This is used when creating `docker.container` resources:
```go
		mqlDockerContainer, err := p.MotorRuntime.CreateResourceWithAssetContext("docker.container",
			dockerAsset, dockerProvider,
			"id", dContainer.ID,
			"image", dContainer.Image,
			"imageid", dContainer.ImageID,
			"command", dContainer.Command,
			"state", dContainer.State,
			"status", dContainer.Status,
			"labels", labels,
			"names", names,
		)
```

Any fields accessed from this resource would use the new asset and provider given to `CreateResourceWithAssetContext`.

### Field Lookup
Since resources can now embed other resources, for example `docker.container` embedding `os.linux`, field lookup on resources has changed slightly. If you ask for `packages` on a `docker.container`, it is allowed to recursively look into embedded resources for that field. `packages` in this case would get looked up from the embedded `os.linux`, and the embedded `os.unix` on the `os.linux` and then the `os.base` on the `os.unix`. 

### Aliasing and Resources
We use resource aliases to avoid having to redefine a bunch of implementations. For example, in the LR,
we have
```
alias os.base.packages = core.packages
```
This creates an implicit resource `os.base.packages`, which when created, actually creates the `packages` resource from core. To support this, we need to understand the context. For example `docker.containers[0].packages` would want to create a `os.base.packages` resource, but it must be attached to the context of the container and not the local target. To support this, a new global function called `createResource` has been added. When we try to look up `packages` on the container, we see that the field is an implicit resource, and use the `createResource` function to create it. This function takes the binding from which to get the asset context, the resource type (which is the return type), and any additional init parameters for the resource construction. Below is the printed compiled code

```
> docker.containers[0].packages
```
```
-> block 1
   entrypoints: [<1,4>]
   1: docker 
   2: containers bind: <1,1> type:[]docker.container
   3: [] bind: <1,2> type:docker.container (0)
   4: createResource bind: <0,0> type:os.base.packages (ref<1,3>)
```


apps/cnquery/cmd/plugin.go
```diff
@@ -119,7 +119,6 @@ func (c *cnqueryPlugin) RunQuery(conf *proto.RunQueryConfig, out shared.OutputHe
 		// when we close the shell, we need to close the backend and store the recording
 		onCloseHandler := func() {
 			storeRecording(m)
-			m.Close()
```
arlimus, line 122: nice catch on these!

resources/lr/lr.go
```diff
@@ -213,6 +220,26 @@ func Parse(input string) (*LR, error) {
 		resource.title, resource.desc = extractComments(resource.Comments)
 		resource.Comments = nil
 
+		// List types have an implicit list field
```
arlimus, line 223: Good move! :+1:

mqlc/mqlc.go
```diff
@@ -121,6 +124,7 @@ func (c *compiler) newBlockCompiler(binding *variable) compiler {
 		blockRef:   ref,
 		props:      c.props,
 		standalone: true,
+		features:   c.features,
```
arlimus, line 127: I'd love to turn this into a compiler configuration in a follow-up, thank you for getting it started!

preslavgerchev:

arlimus (approved):
Great to get this merged. Thank you so much both @jaym and @preslavgerchev !! Also kudos to the many small improvements along the way (unused vars, motor shutdown)

PR #154, opened 2022-10-03
⭐️ Perform k8s auto discovery by default

imilchev:
Auto discovery means discover all workload types as assets (pods, deployments, daemonsets, etc.). It does not include container images

### DIFF
--- a/motor/discovery/k8s/resolver_test.go
+++ b/motor/discovery/k8s/resolver_test.go
@@ -199,7 +199,7 @@ func TestManifestResolverWrongDiscovery(t *testing.T) {
 	require.NoError(t, err)
 	// When this check fails locally, check your kubeconfig.
 	// context has to reference the default namespace
-	assert.Equalf(t, 1, len(assetList), "discovering pods in a cronjob manifest should only result in the manifest")
+	assert.Equalf(t, 0, len(assetList), "discovering pods in a cronjob manifest should result in no assets")
 }
 
 func TestResourceFilter(t *testing.T) {


motor/discovery/k8s/resolver.go
```diff
@@ -20,6 +20,7 @@ var _ common.ContextInitializer = (*Resolver)(nil)
 
 const (
 	DiscoveryAll              = "all"
+	DiscoveryAuto             = "auto"
```
chris-rock, line 23: the cluster needs to become its own discovery target to make it explicit

motor/discovery/k8s/resolver.go
```diff
@@ -197,10 +199,19 @@ func addSeparateAssets(
 	clusterIdentifier string,
```
chris-rock, line 200: In line 130 we change the cluster name based on namespace input which does not make sense anymore when the discovery is becoming the new default

motor/discovery/k8s/resolver.go
```diff
@@ -40,6 +41,7 @@ func (r *Resolver) Name() string {
 func (r *Resolver) AvailableDiscoveryTargets() []string {
 	return []string{
 		DiscoveryAll,
+		DiscoveryAuto,
 		DiscoveryPods,
 		DiscoveryJobs,
 		DiscoveryCronJobs,
```
chris-rock, line 49: The cluster should only be discovered when all, auto or cluster is set as a discovery target

motor/discovery/k8s/resolver.go
imilchev: @czunker can you elaborate a bit on why this was needed in the first place?

czunker: I'm not sure, I completely understand the problem.
From what I understand:
The cluster name and identifier are different things.
The clusterIdentifier looks somethings like this: `//platformid.api.mondoo.app/runtime/k8s/uid/94a8d558-f07d-4e4d-a7d5-fbbe775e8382`
All other discovered assets are prefixed with this one, because the combination of object namespace, type, and name isn't unique across clusters.

Does this help?

motor/discovery/k8s/resolver.go
```diff
@@ -197,10 +200,19 @@ func addSeparateAssets(
 	clusterIdentifier string,
 	od *k8s.PlatformIdOwnershipDirectory,
 ) ([]*asset.Asset, error) {
+	// If there is no discovery or there are no targets set the targets to DiscoveryAuto
+	if tc.Discover == nil {
```
chris-rock, line 204: I think we should not do the handling here but instead do that in the root resolver. That would avoid that we need to implement this special handling on all individual resolvers. We can do that in a follow-up PR.

imilchev: but this discovery setting is k8s specific. I don't see a benefit in having it in the root resolver

PR #155, opened 2022-10-03
🧹 Disable k8s node discovery

imilchev:
Since we don't have policies for them and we cannot merge them with the physical nodes yet, I disabled the node discovery

### DIFF
--- a/motor/discovery/k8s/resolver.go
+++ b/motor/discovery/k8s/resolver.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/rs/zerolog/log"
+	"go.mondoo.com/cnquery"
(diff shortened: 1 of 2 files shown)


motor/discovery/k8s/resolver.go
```diff
@@ -152,15 +152,18 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 		}
 		resolved = append(resolved, clusterAsset)
 
+		// TODO: we disable this for now since we do not have a way of merging the data for k8s nodes and physical nodes
```
chris-rock: Can we put this in a feature flag instead?

imilchev: done

PR #173, opened 2022-10-05
🧹 Improve logging of k8s namespace filter

imilchev:
Instead of logging multiple messages for every namespace in the cluster now we log just 1 warning at the start

### DIFF
--- a/motor/discovery/k8s/resolver.go
+++ b/motor/discovery/k8s/resolver.go
@@ -81,7 +81,7 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 	if allNamespaces != "true" {
 		namespace := tc.Options["namespace"]
 		if len(namespace) > 0 {
-			log.Warn().Msgf("namespace filter has been set to %q", namespace)
+			log.Info().Msgf("namespace filter has been set to %q", namespace)
 			namespacesFilter = append(namespacesFilter, namespace)
 		} else {
 			// try parse the current kubectl namespace


motor/discovery/k8s/resolver.go
```diff
@@ -81,17 +81,19 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
...
 		namespace := tc.Options["namespace"]
 		if len(namespace) > 0 {
+			log.Warn().Msgf("namespace filter has been set to %q", namespace)
 			namespacesFilter = append(namespacesFilter, namespace)
 		} else {
 			// try parse the current kubectl namespace
 			if k8sctlConfig != nil && len(k8sctlConfig.CurrentNamespace()) > 0 {
-				namespacesFilter = append(namespacesFilter, k8sctlConfig.CurrentNamespace())
+				namespace = k8sctlConfig.CurrentNamespace()
+				log.Warn().Msgf("no namespace filter has been provided, using %q from kubeconfig", namespace)
```
imilchev, line 90: this now also makes it very clear that we are using the namespace from the kubeconfig

motor/discovery/k8s/resolver.go
```diff
@@ -81,17 +81,19 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 	if allNamespaces != "true" {
 		namespace := tc.Options["namespace"]
 		if len(namespace) > 0 {
+			log.Warn().Msgf("namespace filter has been set to %q", namespace)
```
czunker: nit: I assume the user has set this intentionally, so I would only use info.

imilchev: I think that makes sense. I will update it

Issue #171, opened 2022-10-05
Improve user experience for multi-asset handling


chris-rock:
We activated the asset discovery as default for k8s. This works great for:

```bash
cnquery explore k8s # nothing is provided, therefore the value of discover is `auto`
cnspec explore k8s # nothing is provided, therefore the value of discover is `auto`
```

We identified a few shortcomings with `shell` and `run` subcommand:

## Shell

```
cnquery shell k8s

→ discovered 10 asset(s)
name:		kube-system/coredns (k8s-object)
platform-id:	  //platformid.api.mondoo.app/runtime/k8s/uid/a184e66f-32ba-4034-ab1e-2ae9fa1724a9/namespace/kube-system/deployments/name/coredns

name:		kube-system/coredns-64897985d-b7t87 (k8s-object)
platform-id:	  //platformid.api.mondoo.app/runtime/k8s/uid/a184e66f-32ba-4034-ab1e-2ae9fa1724a9/namespace/kube-system/pods/name/coredns-64897985d-b7t87
...
```

Since the shell is by design interactive, it should prompt the user for the asset by using a list. This would make it way easier to interact with may assets.

## Run command

```
cnquery shell k8s
```

Instead of asking the user to specify the target asset, we should just run the query on all assets:

```
spacerocket.local:..ce/cnquery ±> cnquery run k8s --query "k8s.pod.name" --all-namespaces --platform-id //platformid.api.mondoo.app/runtime/k8s/uid/a184e66f-32ba-4034-ab1e-2ae9fa1724a9/namespace/kube-system/pods/name/kube-controller-manager-minikube
→ discover related assets for 1 asset(s)
→ use cluster name from kube config cluster-name=minikube
→ resolved assets resolved-assets=10
k8s.pod.name: "kube-controller-manager-minikube"
k8s.pod.name: "storage-provisioner"
```

Things we need to do to achieve that
- [x] introduce `auto` for all discovery resolvers
- [x] move the current implicit mechanism from k8s resolver to the cli definition to switch all resolvers to auto as default
- [x] make the root asset explicit for all resolvers, specifically important to make the root asset explicit eg. aws account = account etc. essentially all the asset types need to have a short key word
- [x] improve the shell to display the list https://github.com/mondoohq/cnquery/pull/178
- [x] improve the run command to run the query on all assets https://github.com/mondoohq/cnquery/pull/179

chris-rock:
We fixed all of those

PR #176, opened 2022-10-05
⭐️ add auto discovery for all resolvers

imilchev:
Part of #171

### DIFF
--- a/motor/discovery/aws/resolver.go
+++ b/motor/discovery/aws/resolver.go
@@ -15,10 +15,11 @@ import (
 )
 
 const (
-	DiscoveryAll       = "all"
 	DiscoveryAccounts  = "accounts"
 	DiscoveryInstances = "instances"
-	DiscoverySSM       = "ssm"
+	// deprecated: use DiscoverySSMInstances instead
+	DiscoverySSM          = "ssm"
+	DiscoverySSMInstances = "ssm-instances"
 )
 
 type Resolver struct{}
@@ -28,7 +29,7 @@ func (r *Resolver) Name() string {
 }
 
 func (r *Resolver) AvailableDiscoveryTargets() []string {
-	return []string{DiscoveryAll, common.DiscoveryAuto, DiscoveryAccounts, DiscoveryInstances, DiscoverySSM}
+	return []string{common.DiscoveryAuto, common.DiscoveryAll, DiscoveryAccounts, DiscoveryInstances, DiscoverySSM, DiscoverySSMInstances}
 }
 
 func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers.Config, cfn common.CredentialFn, sfn common.QuerySecretFn, userIdDetectors ...providers.PlatformIdDetector) ([]*asset.Asset, error) {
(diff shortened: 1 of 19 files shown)


motor/discovery/aws/resolver.go
```diff
@@ -27,7 +27,7 @@ func (r *Resolver) Name() string {
 }
 
 func (r *Resolver) AvailableDiscoveryTargets() []string {
-	return []string{DiscoveryAll, DiscoveryInstances, DiscoverySSM}
+	return []string{DiscoveryAll, common.DiscoveryAuto, DiscoveryInstances, DiscoverySSM}
```
chris-rock: Just adding it to the list is not doing the trick, the code needs to be adapted to handle that value as well

imilchev: I am still confused about this one. What needs to happen? Right now if I look at the code (let's take AWS as an example) there are 2 conditionals for discovery flags. One for SSM and one for instances. If we want the `DiscoveryAuto` to keep the behaviour the same as what is now when we are not specifying any flags, then this change should be enough. If we want it to behave as `DiscoveryAll` then I need to modify the code.

apps/cnquery/cmd/run.go
```diff
@@ -49,7 +50,7 @@ var execCmd = builder.NewProviderCommand(builder.CommandOpts{
 
 		cmd.Flags().String("path", "", "path to a local file or directory that the connection should use")
 		cmd.Flags().StringToString("option", nil, "addition connection options, multiple options can be passed in via --option key=value")
-		cmd.Flags().String("discover", "", "enables the discovery of nested assets. Supported are 'all|instances|host-instances|host-machines|container|container-images|pods|cronjobs|statefulsets|deployments|jobs|replicasets|daemonsets'")
+		cmd.Flags().String("discover", common.DiscoveryAuto, "enables the discovery of nested assets. Supported are 'all|instances|host-instances|host-machines|container|container-images|pods|cronjobs|statefulsets|deployments|jobs|replicasets|daemonsets'")
```
czunker: Although "auto" is the default, shouldn't it be part of the "Supported ..." list?

The same for `shell` and `explore` command.

motor/discovery/common/discovery.go
```diff
@@ -0,0 +1,3 @@
+package common
+
+const DiscoveryAuto = "auto"
```
czunker: I think with this, we can remove/change L24 + L47 in `motor/discovery/k8s/resolver.go`

imilchev: yup, that's right. I fixed that

apps/cnquery/cmd/run.go
chris-rock: This needs to be applied to all commands that use the --discover flag

motor/discovery/aws/resolver.go
```diff
@@ -16,6 +16,7 @@ import (
 
 const (
 	DiscoveryAll       = "all"
+	DiscoveryAccounts  = "accounts"
 	DiscoveryInstances = "instances"
 	DiscoverySSM       = "ssm"
```
chris-rock: @vjeffrey should this not be better ssm-instances?

apps/cnquery/cmd/run.go
imilchev: I noticed for `shell` the discovery list is much shorter than `run` and `exec`. Shouldn't the lists be identical?

PR #177, opened 2022-10-05
🧹 Remove auto discovery by default from k8s resolver

imilchev:
First drop it from the k8s resolver and then I will see how I can set that in the CLI configuration for `run`/`shell` I suppose that there different changes are going to be required in order to make them work with that discovery setting

### DIFF
--- a/apps/cnquery/cmd/explore.go
+++ b/apps/cnquery/cmd/explore.go
@@ -20,6 +20,7 @@ import (
 	"go.mondoo.com/cnquery/cli/theme"
 	"go.mondoo.com/cnquery/explorer"
 	"go.mondoo.com/cnquery/explorer/scan"
+	"go.mondoo.com/cnquery/motor/discovery/common"
 	v1 "go.mondoo.com/cnquery/motor/inventory/v1"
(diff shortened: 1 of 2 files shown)


apps/cnquery/cmd/shell.go
```diff
@@ -46,7 +47,7 @@ var shellCmd = builder.NewProviderCommand(builder.CommandOpts{
 
 		cmd.Flags().String("path", "", "path to a local file or directory that the connection should use")
 		cmd.Flags().StringToString("option", nil, "addition connection options, multiple options can be passed in via --option key=value")
-		cmd.Flags().String("discover", "", "enables the discovery of nested assets. Supported are 'all|instances|host-instances|host-machines|container|container-images'")
+		cmd.Flags().String("discover", common.DiscoveryAuto, "enables the discovery of nested assets. Supported are 'all|instances|host-instances|host-machines|container|container-images'")
```
chris-rock, line 50: This is great, we are missing the change for `run` and `explore`.

imilchev: done

PR #179, opened 2022-10-05
⭐️ support cnquery run for multiple assets

chris-rock:
after #178 

This change brings multi-asset support to `cnquery run` command. Queries are executed for all discovered results.

Screenshot 2022-10-05 at 16 01 30

Screenshot 2022-10-05 at 16 13 36

### DIFF
--- a/apps/cnquery/cmd/plugin.go
+++ b/apps/cnquery/cmd/plugin.go
@@ -2,8 +2,7 @@ package cmd
 
 import (
 	"context"
-
-	"go.mondoo.com/cnquery/cli/components"
+	"encoding/json"
 
 	"github.com/cockroachdb/errors"
 	"github.com/hashicorp/go-plugin"
@@ -11,7 +10,7 @@ import (
 	"github.com/spf13/cobra"
 	"go.mondoo.com/cnquery/cli/printer"
 	"go.mondoo.com/cnquery/cli/shell"
-	"go.mondoo.com/cnquery/cli/theme"
+	"go.mondoo.com/cnquery/llx"
 	"go.mondoo.com/cnquery/logger"
 	"go.mondoo.com/cnquery/motor/asset"
 	"go.mondoo.com/cnquery/motor/discovery"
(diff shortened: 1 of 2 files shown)


apps/cnquery/cmd/plugin.go
```diff
@@ -91,77 +91,117 @@ func (c *cnqueryPlugin) RunQuery(conf *proto.RunQueryConfig, out shared.OutputHe
...
+		shellOptions := []shell.ShellOption{}
+		shellOptions = append(shellOptions, shell.WithOnCloseListener(onCloseHandler))
+		shellOptions = append(shellOptions, shell.WithFeatures(conf.Features))
+		shellOptions = append(shellOptions, shell.WithOutput(out))
+
+		sh, err := shell.New(m, shellOptions...)
+		if err != nil {
+			return errors.Wrap(err, "failed to initialize the shell")
+		}
+		defer sh.Close()
```
imilchev, line 135: This is now a defer in a loop. Only once the loop is fully finished (passed all the iterations), only then the defer will be called for all the assets in the slice. If this is not an intended behaviour, it should be updated

apps/cnquery/cmd/plugin.go
```diff
@@ -92,76 +91,117 @@ func (c *cnqueryPlugin) RunQuery(conf *proto.RunQueryConfig, out shared.OutputHe
...
+		out.WriteString("[")
 	}
 
-	// when we close the shell, we need to close the backend and store the recording
-	onCloseHandler := func() {
-		storeRecording(m)
-		m.Close()
-	}
+	for i := range assetList {
+		connectAsset := assetList[i]
```
joelddiaz, line 114: I believe the intention here is to actually use the `filteredAssets` list created a few lines above here
```suggestion
	for i := range filteredAssets {
		connectAsset := filteredAssets[i]
```

apps/cnquery/cmd/plugin.go
```diff
@@ -92,76 +91,117 @@ func (c *cnqueryPlugin) RunQuery(conf *proto.RunQueryConfig, out shared.OutputHe
...
+		code, results, err := sh.RunOnce(conf.Command)
+		if err != nil {
+			return errors.Wrap(err, "failed to run")
+		}
+
+		if conf.Format != "json" {
+			sh.PrintResults(code, results)
+		} else {
+			renderJson(code, results, out)
+			if len(assetList) != i+1 {
```
joelddiaz, line 146:
```suggestion
			if len(filteredAssets) != i+1 {
```

PR #187, opened 2022-10-06
🧹 Discover all k8s namespaces by default

imilchev:
Signed-off-by: Ivan Milchev <ivan@mondoo.com>

### DIFF
--- a/motor/discovery/k8s/resolver.go
+++ b/motor/discovery/k8s/resolver.go
@@ -84,27 +84,10 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 		}
 	}
 
-	// If no namespace options have been provided, then use all-namespaces by default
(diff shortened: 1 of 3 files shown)


motor/discovery/k8s/resolver.go
```diff
@@ -84,6 +84,13 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 		}
 	}
 
+	// If no namespace options have been provided, then use all-namespaces by default
+	_, allNsSet := tc.Options["all-namespaces"]
```
chris-rock: Why do we need the all-namespaces property then in the first place? Would it not be easier to remove it completely and just not scan all namespaces if a specific namespace was provided?

imilchev: That would work too. I thought that we want to keep it but that would make the code even simpler. I will adjust it

Issue #186, opened 2022-10-06
Slow/non-returning `kernel.parameters` for running docker container


chris-rock:
The `kernel.parameters` resource does not work for running containers. To reproduce, start a new container:

```
docker run -it rockylinux
```

Then connect with the shell to the container and the response takes way too long / never returns:

```
cnquery shell docker 89e22f207cc6
→ loaded configuration from /Users/chris/.config/mondoo/mondoo.yml using source default
→ discover related assets for 1 asset(s)
→ resolved assets resolved-assets=1
 .--. ,-.,-. .---..-..-. .--. .--. .-..-.™
'  ..': ,. :' .; :: :; :' '_.': ..': :; :
`.__.':_;:_;`._. ;`.__.'`.__.':_;  `._. ;
   mondoo™     : :                  .-. :
               :_:                  `._.' interactive shell

cnquery> kernel.parameters

```

PR #188, opened 2022-10-06
🧹 Use /sbin/sysctl if present when reading kernel parameters.

preslavgerchev:

Partial fix for #186

### DIFF
--- a/resources/packs/core/kernel/manager.go
+++ b/resources/packs/core/kernel/manager.go
@@ -108,14 +108,15 @@ func (s *LinuxKernelManager) Info() (KernelInfo, error) {
 func (s *LinuxKernelManager) Parameters() (map[string]string, error) {
 	if s.provider.Capabilities().HasCapability(providers.Capability_RunCommand) {
 		cmd, err := s.provider.RunCommand("/sbin/sysctl -a")
-		// means sysctl is not present, fallback to /proc/sys walking
-		if err == nil {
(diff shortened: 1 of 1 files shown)


resources/packs/core/kernel/manager.go
```diff
@@ -105,6 +106,14 @@ func (s *LinuxKernelManager) Info() (KernelInfo, error) {
 }
 
 func (s *LinuxKernelManager) Parameters() (map[string]string, error) {
+	if s.provider.Capabilities().HasCapability(providers.Capability_RunCommand) {
+		cmd, err := s.provider.RunCommand("/sbin/sysctl -a")
+		// means sysctl is not present, fallback to /proc/sys walking
```
czunker: I think the opposite is true. Or is the comment meant to be below line L115?

chris-rock: The comment is confusing. Should be more like

```
// if err is nil, the command is available, then we use the command output, otherwise we fallback to /proc/sys walking
```

resources/packs/core/kernel/manager.go
```diff
@@ -105,6 +106,14 @@ func (s *LinuxKernelManager) Info() (KernelInfo, error) {
...
+		if err == nil {
```
chris-rock: tested this with a running container. that is not working since it only returning err if something really when wrong. command exec error is not an error from mondoo perspective. therefore we also need to check for exit code. I'll push a change.

PR #192, opened 2022-10-06
🧹 Rename standard to os resolver. 

preslavgerchev:
~~Fixes issue where scanning via FS would result in uncategorized assets.~~
Rename standard resolver to OS resolver and reuse functions in fs provider.

### DIFF
--- a/motor/discovery/os/os.go
+++ b/motor/discovery/os/os.go
@@ -1,4 +1,4 @@
-package standard
+package os
 
 import (
 	"context"
@@ -16,7 +16,7 @@ import (
 type Resolver struct{}
 
 func (r *Resolver) Name() string {
-	return "Standard Resolver"
+	return "OS Resolver"
 }
 
 func (r *Resolver) AvailableDiscoveryTargets() []string {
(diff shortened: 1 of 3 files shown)


motor/providers/fs/provider.go
```diff
@@ -90,7 +83,8 @@ func (p *Provider) Capabilities() providers.Capabilities {
 }
 
 func (p *Provider) Kind() providers.Kind {
-	return p.kind
+	// TODO: this *could* also be bare metal, how can we differentiate from FS provider point of view?
```
preslavgerchev: See comment: I am not sure how we can differentiate between running in a VM or on bare metal?

chris-rock: I would return runtime = "" and kind Kind_KIND_UNKNOWN, also the os resolver never uses that.

motor/discovery/os/os.go
```diff
@@ -53,6 +53,10 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 		assetObj.Platform = p
 	}
 
+	if assetObj.Connections[0].Backend == providers.ProviderType_FS {
```
chris-rock: It is not save to assume only one connection

PR #195, opened 2022-10-06
Sign windows binaries

jaym:


.github/workflows/goreleaser.yml
```diff
@@ -20,6 +20,28 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: 1.19
+      - name: 'Authenticate to Google Cloud'
+        uses: 'google-github-actions/auth@v0'
+        with:
+          workload_identity_provider: 'projects/855680899344/locations/global/workloadIdentityPools/github-cicd/providers/github-cicd'
```
benr: These need to be protected as secrets.

benr: Secrets add:

- GCP_WIP
- GCP_SERVICE_ACCOUNT

PR #206, opened 2022-10-07
🧹 Reusable list k8s workloads

### LINKED ISSUE #175: Re-write listing k8s assets in a re-usable way
All these functions are almost identical:

- https://github.com/mondoohq/cnquery/blob/main/motor/discovery/k8s/list_cronjobs.go#L13-L76
- https://github.com/mondoohq/cnquery/blob/main/motor/discovery/k8s/list_daemonsets.go#L13-L76
- https://github.com/mondoohq/cnquery/blob/main/motor/discovery/k8s/list_deployments.go#L13-L76
- https://github.com/mondoohq/cnquery/blob/main/motor/discovery/k8s/list_jobs.go#L13-L76
- https://github.com/mondoohq/cnquery/blob/main/motor/discovery/k8s/list_pods.go#L12-L75
- https://github.com/mondoohq/cnquery/blob/main/motor/discovery/k8s/list_replicasets.go#L12-L75
- https://github.com/mondoohq/cnquery/blob/main/motor/discovery/k8s/list_statefulsets.go#L13-L76

We should re-write them such that there is one definition of the function and it is being re-used for all those specific types. This was already done for the `init` functions of the k8s MQL resources. It should be possible for this case as well. See [here](https://github.com/mondoohq/cnquery/blob/main/resources/packs/k8s/common.go#L170-L238) for a generic function re-used by all workload types

imilchev:
Closes #175

### DIFF
--- a/motor/discovery/k8s/list_workloads.go
+++ b/motor/discovery/k8s/list_workloads.go
@@ -110,7 +110,7 @@ func ListWorkloads[T runtime.Object](
 	workloads := []T{}
 
 	if len(resFilter) > 0 {
-		// If there is a resources filter we should only retrieve the daemonsets that are in the filter.
+		// If there is a resources filter we should only retrieve the workloads that are in the filter.
 		if len(resFilter[workloadType]) == 0 {
 			return []*asset.Asset{}, nil
 		}


motor/discovery/k8s/list_workloads.go
```diff
@@ -0,0 +1,164 @@
...
+	resFilter map[string][]K8sResourceIdentifier,
+	od *k8s.PlatformIdOwnershipDirectory,
+	workloadType string,
+	getter func(string, string) (T, error),
+	lister func(v1.Namespace) ([]T, error),
+) ([]*asset.Asset, error) {
+	workloads := []T{}
+
+	if len(resFilter) > 0 {
+		// If there is a resources filter we should only retrieve the daemonsets that are in the filter.
```
czunker: "daemonsets" should now be workloads.

PR #246, opened 2022-10-11
🐛 Do not use missing kubelet config file

czunker:
The kubelet might run without a config file (e.g. on AKS). In this case we cannot use the file to populate the configuration, so skip it.

This PR also fixes some other config parameters which didn't get populated correctly.


resources/packs/k8s/kubelet_flags.go
```diff
@@ -109,41 +112,80 @@ func mergeDeprecatedFlagsIntoConfig(kubeletConfig map[string]interface{}, flags
 		kubeletConfig["readOnlyPort"] = flags["read-only-port"]
 	}
 	if _, ok := flags["anonymous-auth"]; ok {
-		auth := kubeletConfig["authentication"].(map[string]interface{})
-		anon := auth["anonymous"].(map[string]interface{})
+		auth := map[string]interface{}{}
```
chris-rock, line 115: I think it would be nice to add some tests for those config files to ensure we do not break expected behavior in the future.

czunker: Good point. Added an AKS record and tests for it.

PR #255, opened 2022-10-12
⭐️ markdown cli docs generator

chris-rock:
allows you to generate the cli docs:

```
go run apps/gen-docs/main.go --docs-path path/docs/cli
```


arlimus (approved):
Great addition, also thanks for moving this to a separate subcommand. We can probably also move the docs call out of root later on

PR #256, opened 2022-10-12
🌈 Show warning when namespace not present

czunker:

### DIFF
--- a/motor/discovery/k8s/resolver.go
+++ b/motor/discovery/k8s/resolver.go
@@ -87,6 +87,23 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 		return nil, err
 	}
 
+	clusterNamespaces, err := p.Namespaces()
+	if err != nil {
+		return nil, err
+	}
+	for _, ns := range namespacesFilter {
+		found := false
+		for _, clusterNs := range clusterNamespaces {
+			if clusterNs.Name == ns {
+				found = true
+				break
+			}
+		}
+		if !found {
+			log.Warn().Msgf("namespace %q not found in cluster", ns)
+		}
+	}
+
 	clusterIdentifier, err := p.Identifier()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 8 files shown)


motor/discovery/k8s/resolver.go
```diff
@@ -87,6 +87,23 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 		return nil, err
 	}
 
+	clusterNamespaces, err := p.Namespaces()
+	if err != nil {
+		return nil, err
+	}
+	for _, ns := range namespacesFilter {
+		found := false
+		for _, clusterNs := range clusterNamespaces {
```
chris-rock: potentially we have more cluster namespaces than filter namespaces, should we make this the outer loop then?

czunker: Oh yes, you are correct.

I had a look at it, and now I think we can reduce it to just one loop.
`namespaceFilter` is an array, but the cli parameter only allows one namespace. I'd change the filter to a string and just use one loop. Any objections?

chris-rock: Perfect, lets do that

czunker: Removed the array for the namespaceFilter and now I only need one loop.

joelddiaz (commented):
Is there some specific reason we simply don't want to error when given a Namespace that doesn't exist?

joelddiaz:
> Is there some specific reason we simply don't want to error when given a Namespace that doesn't exist?

More specifically: what does it mean to get a shell when the `--namespace` provided didn't exist.

For example when I run: `/cnquery shell k8s --namespace blahblahblah`

I end up in a Mondoo shell prompt, and I see the `k8s` resource available to inspect, but it seems there was no consequence at all to the fact that I provided a `--namespace` that doesn't exist except for the fact that only a single asset was discovered (the k8s cluster itself).

joelddiaz:
Ok, we discussed they "why not error". The --namespace is supposed to be used as a filter, and putting --namespace=doesnotexist will correctly return all the resources under the Namespace doesnotexist.

Perhaps the help text on the parameter can be changed to make it clear that it is to be used as a filter https://github.com/mondoohq/cnquery/blob/main/apps/cnquery/cmd/builder/builder.go#L408

joelddiaz (approved):

PR #259, opened 2022-10-12
🐛 do not auto-discover containers for local provider

chris-rock:


arlimus (approved):
Yeah great call, let's not do this too quickly until our error-handling around discovered targets is further expanded.

PR #262, opened 2022-10-13
🛑⭐ restructure JSON output

arlimus:
1. We used to conditionally create a list of entries for every entrypoint in a query. But this only happened when there were more than 1 entry. This conditional handling creates inconsistencies for the output. It also looses all information about the query that produced it (eg users.list in the example) and it prevents us from attaching more meta-information (since it's just a list item and not structured). So, go for structured info. This is a breaking change.

2. Errors tend to eat up all available JSON output. Try to provide as much output as possible and give access to errors inside of it. There are multiple follow-ups here like aggregating errors (see the point above) and giving structured errors instead of strings in the appropriate fields (which allows for way easier checking ie type == object and field == error)

Given

```
cnquery run local -c "users.list{*}" -j | jq .
```

OLD output

image

NEW output

image

### DIFF
--- a/apps/cnquery/cmd/plugin.go
+++ b/apps/cnquery/cmd/plugin.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"context"
-	"encoding/json"
 
 	"github.com/cockroachdb/errors"
 	"github.com/hashicorp/go-plugin"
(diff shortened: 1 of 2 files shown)


chris-rock (approved):
Thank you @arlimus

PR #270, opened 2022-10-14
⭐ explore report --json

arlimus:
Introduces a custom printer to create reports. It doesn't just take the raw data and dump it all to CLI, but instead carefully constructs a JSON representation of the report that is useful for further processing. It contains MRNs for queries, eliminates pesky lookups, and nicely aggregates data.

image

Features 3 fields on the top level: assets (for quick lookup and aggregation), reports (for all the data) and errors (for error aggregation).

Unlike `json.Marshal` it doesn't dump any compiled code, doesn't dump the resolved pack, and doesn't force lookups of all data fields.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -58,7 +58,7 @@
         "explore",
         "local",
         "-f",
-        "./examples/example.mql.yaml"
+        "./examples/example.unix.mql.yaml"
       ],
     }
   ]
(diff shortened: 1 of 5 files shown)


chris-rock (approved):
Nice addition @arlimus As a follow up I'd like to see some tests to ensure this keeps working.

PR #274, opened 2022-10-14
🐛 properly sort suggestions with matching prefix or empty

arlimus:
It was off when we have the same prefix as seen:

image

Here we want to sort alphabetically, since fuzzy isn't needed for this:

image

This has the beautiful side-effect that it also fixes the edge-case where we don't type anything in, since everything is prefixed by empty string ;)

image


chris-rock (approved):
Nice nice nice

chris-rock (commented):
Thank you @arlimus

PR #276, opened 2022-10-14
🧹 speed up cnquery on windows

jaym:
Using `Invoke-RestMethod` on addresses that don't exist is very slow. Setting the timeout to 1 second takes running a simple query locally from 25 seconds to 5 seconds


chris-rock (approved):
Great call @jaym

PR #281, opened 2022-10-15
⭐ cnquery scan ..

arlimus:
Rename the `cnquery explore` command to `cnquery scan`, because it is more intuitive. It still explores the assets given the provided query packs, but semantically it's easier to remember to "scan an asset" like the crew in Star Trek would scan a planet they were looking at. We explore things by looking at the gathered data or walking the planet, but first we scan it.

Note: the package was not renamed, because it contains a scanner inside of it (see ). So we can think of it as the space ship that has an explorer module with a scanner that does the work.

### DIFF
--- a/apps/cnquery/cmd/scan.go
+++ b/apps/cnquery/cmd/scan.go
@@ -32,178 +32,179 @@ import (
 )
 
 func init() {
-	rootCmd.AddCommand(exploreCmd)
+	rootCmd.AddCommand(scanCmd)
 }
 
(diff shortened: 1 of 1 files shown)


arlimus:
@misterpantz for visibility on this change ^^

chris-rock (approved):
Thank you @arlimus

chris-rock (approved):
Thank you @arlimus, I'll add a follow-up PR to update this in our README

PR #282, opened 2022-10-15
🛑⭐ access resource lists without calling `list`

arlimus:

## Easier list access

Instead of `users.list` just run `users`:

image

Access to individual elements can now be written simply as `users[0]` instead of `users.list[0]`

image

Block operations don't require `list` anymore and always apply to child elements:

image

## Soft deprecations

For V6 compatibility (which we maintain throughout V7 lifetime) you can still write things the old way, but should switch to the new way soon:

```
old> users { list }
new> users

old> ports { listening list }
new> ports
new> ports.listening
```

Additionally we show nice deprecation warnings (e.g. for `ports { listening list }`):

image

### DIFF
--- a/llx/code.go
+++ b/llx/code.go
@@ -19,6 +19,15 @@ func (b *Block) TailRef(blockRef uint64) uint64 {
 	return absRef(blockRef, b.ChunkIndex())
 }
 
+func (b *Block) ReplaceEntrypoint(old uint64, nu uint64) {
+	for i := range b.Entrypoints {
+		if b.Entrypoints[i] == old {
+			b.Entrypoints[i] = nu
+			return
+		}
+	}
+}
+
 // LastChunk is the last chunk in the list or nil
 func (b *Block) LastChunk() *Chunk {
 	max := len(b.Chunks)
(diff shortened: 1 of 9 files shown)


arlimus:
> wards compatibility in v7 for a query like this:

Great catch! We'll have to break users there, this old mode is not supported anymore. I think it will help if we start to revisit and look at a few our resources to update their structure and remove some of these confusions.

arlimus:
@jaym Updated the code for backwards compatibility with v6, wherever it isn't ambiguous.

@chris-rock Updated the code to give nice lists by default (e.g. `users`, `ports`)

chris-rock:
> Don't know if we want backwards compatibility in v7 for a query like this:
> 
> ```
> os.rootCertificates { files }
> ```
> 
> This is expanding it putting the list of files into each certificate instead of the old meaning that was equivalent to `os.rootCertificates.files`

I think `os.rootCertificates { files }` read confuses me since I expect that users would expect that this is returning the files from the certificate. Before we had block access and I agree there is a short period of confusion but I like the new style much better.

chris-rock (approved):
Super cool. Love the new simplicity.

PR #283, opened 2022-10-15
🧹 update readme with new cnquery scan command

chris-rock:
follow up to #281


arlimus (approved):
Great catch, thank you!

PR #285, opened 2022-10-16
🐛 be more defensive with assertion creation

arlimus:
we want to use the assertion approach whenever we see an assessment, not just if the last expression was an assertion


chris-rock (approved):
Great improvement @arlimus

PR #298, opened 2022-10-16
⭐️ add asset synchronization method

chris-rock:
This change defines an asset synchronization method that we can use to sync assets and fetch their queries from upstream.


arlimus (approved):
Thank you for porting these calls

PR #299, opened 2022-10-17
Format all the default values in help the same

tas50:
Makes the docs consistent


chris-rock (approved):
Thank you @tas50

PR #302, opened 2022-10-17
✨ easier configuration for the compiler

arlimus:
... and much easier to grow this in the future as configuration flags get added, some of which may change as sub-compilers get created.

Only merge this after https://github.com/mondoohq/cnquery/pull/137 since it establishes the influence of features in the compiler.


chris-rock (approved):
Great improvement of readability and maintainability. Thank you @arlimus

PR #305, opened 2022-10-17
🧹 Show message what we discover in auto mode

czunker:

### DIFF
--- a/motor/discovery/k8s/resolver.go
+++ b/motor/discovery/k8s/resolver.go
@@ -144,15 +144,17 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 		return nil, err
 	}
 
+	if tc.IncludesDiscoveryTarget(common.DiscoveryAuto) {
+		log.Info().Msg("discovery option auto is used. This will detect the assets: cluster, cronjobs, pods, statefulsets, deployments, replicasets, daemonsets")
+	}
+
 	// Only discover cluster and nodes if there are no resource filters. For CI/CD do not
 	// discover the cluster asset at all. In that case that would be the admission review resource
 	// for which we only care if we have explicitly enabled discovery for it.
 	var clusterAsset *asset.Asset
 	ownershipDir := k8s.NewEmptyPlatformIdOwnershipDirectory(clusterIdentifier)
-	if tc.IncludesDiscoveryTarget(common.DiscoveryAll) ||
-		tc.IncludesDiscoveryTarget(common.DiscoveryAuto) ||
-		tc.IncludesDiscoveryTarget(DiscoveryClusters) &&
-			len(resourcesFilter) == 0 && root.Category != asset.AssetCategory_CATEGORY_CICD {
+	if tc.IncludesOneOfDiscoveryTarget(common.DiscoveryAll, common.DiscoveryAuto, DiscoveryClusters) &&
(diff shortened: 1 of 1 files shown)


czunker:
That's the follow-up to https://github.com/mondoohq/cnquery/pull/176#issuecomment-1271474139

motor/discovery/k8s/resolver.go
```diff
@@ -144,15 +144,17 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 		return nil, err
 	}
 
+	if tc.IncludesDiscoveryTarget(common.DiscoveryAuto) {
+		log.Info().Msg("discovery option auto is used. This will detect the assets: cluster, cronjobs, pods, statefulsets, deployments, replicasets, daemonsets")
```
czunker: We do not discover jobs. Was this intended in this PR?  https://github.com/mondoohq/cnquery/pull/176#issuecomment-1271474139

motor/discovery/k8s/resolver.go
```diff
@@ -144,15 +144,17 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
...
 	// discover the cluster asset at all. In that case that would be the admission review resource
 	// for which we only care if we have explicitly enabled discovery for it.
 	var clusterAsset *asset.Asset
 	ownershipDir := k8s.NewEmptyPlatformIdOwnershipDirectory(clusterIdentifier)
-	if tc.IncludesDiscoveryTarget(common.DiscoveryAll) ||
-		tc.IncludesDiscoveryTarget(common.DiscoveryAuto) ||
-		tc.IncludesDiscoveryTarget(DiscoveryClusters) &&
-			len(resourcesFilter) == 0 && root.Category != asset.AssetCategory_CATEGORY_CICD {
+	if tc.IncludesOneOfDiscoveryTarget(common.DiscoveryAll, common.DiscoveryAuto, DiscoveryClusters) &&
+		len(resourcesFilter) == 0 && root.Category != asset.AssetCategory_CATEGORY_CICD {
```
chris-rock, line 157: I am not sure why we need this, wouldn't it be easier for the operator to say what he wants to scan?

chris-rock: If it is the action, then we should adapt https://github.com/mondoohq/actions/blob/main/k8s-manifest/action.yaml#L47-L51 instead

motor/discovery/k8s/resolver.go
chris-rock: That is a question for @imilchev

imilchev: yup! That seems to have slipped through. Please add it. Jobs should be discovered as well

czunker: Added `jobs` to the auto mode discovery.

motor/discovery/k8s/resolver.go
czunker: I do not understand the question. Isn't this also the code path that is used for CLI scans?

czunker: As just discussed, we will check whether we can get ride of the additional conditions in a separate PR.

imilchev: This code is there because for CI/CD scans we don't want to discover the cluster as an asset for the case of the admission controller. I guess we could technically move that logic to the operator itself, indeed

chris-rock (approved):
Thank you @czunker

PR #313, opened 2022-10-17
🐛 auto-expand the list of resources

arlimus:
It worked for auto-generated list-resources, but not for normal arrays of resources. This requires to grab the resource info early during the first check, which it now does.

image


chris-rock (approved):
Thank you @arlimus

PR #314, opened 2022-10-17
⭐️ support communication with upstream

chris-rock:
This change allows cnquery to gather query packs from logged in Mondoo Platform. When authenticated, cnquery automatically uses upstream.

Screenshot 2022-10-17 at 20 57 51


arlimus (approved):
Upstream we flow!! Thank you :)

PR #316, opened 2022-10-18
🐛 Fix type of block call in auto expansion code

jaym:
block calls during auto expansion were not correctly checking if the result of the block call was supposed to be a list of blocks or just a single block


arlimus (approved):
Cheers for fixing the issue Jay!!

PR #320, opened 2022-10-18
🐛 expand child blocks

arlimus:
So far the auto-expansion (and general post-processing of the compiler) only worked on the outer-most block. Introduce it to the remaining blocks as well, so they all benefit from list- and defaults expansion. Make sure the block types behave as expected.

Old:

image

New

image


chris-rock (approved):
Supercool

PR #321, opened 2022-10-18
✨ auto-expand results in assessments

arlimus:
These were still using their old IDs and not as easy to fix. There is a follow-up with @msg assertions and this, which currently double-print inside of blocks. Worth getting this update in and follow up on the bug.

image


chris-rock (approved):

czunker:
I think we close this, now that https://github.com/mondoohq/cnquery/pull/1211 got merged.

PR #322, opened 2022-10-18
🐛 restructure platform specs into detector package

chris-rock:
We introduced a circular dependency that only happened on Windows:

```
package command-line-arguments
	imports go.mondoo.com/cnquery/apps/cnquery/cmd
	imports go.mondoo.com/cnquery/apps/cnquery/cmd/builder
	imports go.mondoo.com/cnquery/motor/asset
	imports go.mondoo.com/cnquery/motor/platform
	imports go.mondoo.com/cnquery/motor/platform/windows
	imports go.mondoo.com/cnquery/motor/providers/local
	imports go.mondoo.com/cnquery/motor/asset: import cycle not allowed
```

We imported the windows provider in the the platform package. To avoid that I moved the platform specs into the detector package (where they are used anyway).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -205,6 +205,14 @@ explorer/generate:
 
 #   🏗 Binary / Build   #
 
+.PHONY: cnquery/build/linux
+cnquery/build/linux:
+	GOOS=linux go build ${LDFLAGSDIST} apps/cnquery/cnquery.go
+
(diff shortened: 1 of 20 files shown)


arlimus (approved):
Thank you for resolving it!!

PR #326, opened 2022-10-18
Adds defaults for AWS resources

scottford-io:
image

This PR adds a first pass at sane defaults for AWS resources. Wherever possible, I defaulted to the `arn` since it has a bunch of information like the account number, region, and the name of the resource. 


scottford-io:

resources/packs/aws/aws.lr
```diff
@@ -3,7 +3,7 @@ import "../core/core.lr"
 option go_package = "go.mondoo.com/cnquery/resources/packs/aws"
 
 // AWS Resource
-aws {
+aws @defaults("aws.account.id") {
```
arlimus: let's simplify this one to `@defaults("account.id")`

preslavgerchev:
I believe the json files needs updating to add the defaults there too (so they can get picked up later when executing the query), `make cnquery/generate` should do the job

arlimus:
Thank you for catching that @preslavgerchev !!

PR #328, opened 2022-10-18
🧹 Check LR generated json files for resource packs.

preslavgerchev:

Seen in https://github.com/mondoohq/cnquery/pull/326, there is no check if we forget to regen the info json files for the resources.


arlimus (approved):
Great improvement to catch accidental resource changes!! Kudos Preslav :)

PR #329, opened 2022-10-18
Fixes a reference to GitHub for GitLab

scottford-io:
This PR fixes a reference to `GITHUB` that should be `GITLAB`. I also updated text from `access token` to reference `personal access token` to match GitHub and GitLab documentation.

### DIFF
--- a/apps/cnquery/cmd/builder/builder.go
+++ b/apps/cnquery/cmd/builder/builder.go
@@ -639,7 +639,7 @@ func githubProviderOrganizationCmd(commonCmdFlags commonFlagsFn, preRun commonPr
 		},
 	}
 	commonCmdFlags(cmd)
-	cmd.Flags().String("token", "", "GitHub access tokens")
+	cmd.Flags().String("token", "", "GitHub personal access tokens")
 	return cmd
 }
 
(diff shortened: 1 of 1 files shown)


apps/cnquery/cmd/builder/builder.go
```diff
@@ -680,7 +680,7 @@ func gitlabProviderCmd(commonCmdFlags commonFlagsFn, preRun commonPreRunFn, runF
 	commonCmdFlags(cmd)
 	cmd.Flags().String("group", "", "a GitLab group to scan")
 	cmd.MarkFlagRequired("group")
-	cmd.Flags().String("token", "", "GitHub access tokens")
+	cmd.Flags().String("token", "", "q personal access tokens")
```
chris-rock: Here is something wrong

chris-rock (changes requested):

apps/cnquery/cmd/builder/builder.go
tas50: Is the q here a typo?

scottford-io: VIM fat finger. Good catch!

chris-rock (approved):
Thank you @scottford-io

PR #341, opened 2022-10-19
🐛 only render progress bar in tty mode

chris-rock:
The fixes an issue where a user that runs

```
cnquery scan local --json | jq .
```

was erroring out with:

```
cnquery scan local --json | jq .                                                                                                                                                                                
→ connecting to asset spacerocket.fritz.box (baremetal)
parse error: Invalid numeric literal at line 2, column 106
```

The reason was the progress bar that was written to stdout. The new change only prints the progress bar when the stdout is attached to a tty.

### DIFF
--- a/explorer/scan/local_scanner.go
+++ b/explorer/scan/local_scanner.go
@@ -2,9 +2,11 @@ package scan
 
 import (
 	"context"
+	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


czunker (approved):
Nice!

PR #342, opened 2022-10-19
🐛 Fix autocomplete for new k8s resources

### LINKED ISSUE #215: Kubernetes admissionreview lacks autocomplete description
**Describe the bug**

**To Reproduce**
1. Run `cnquery shell k8s`
2. Type `k8s.` and tag to get the autocomplete

**Expected behavior**
admissionreview should have a description

**Screenshots**
image

N/A

**Additional context**
N/A

### LINKED ISSUE #336: No autocomplete for k8s.kubelet
I tested this with cnquery v7. 
image

However, if I type the full resource, I do get autocompletion for its properties.
image

Seems to be related to #215

czunker:
Fixes #215
Fixes #336

### DIFF
--- a/mqlc/builtin.go
+++ b/mqlc/builtin.go
@@ -200,9 +200,11 @@ func publicFieldsInfo(c *compiler, resourceInfo *resources.ResourceInfo) map[str
 		if v.IsImplicitResource {
 			name := types.Type(v.Type).ResourceName()
 			child := c.Schema.Resources[name]
-			if !child.HasEmptyInit() {
-				continue
-			}
+			/*
+				if !child.HasEmptyInit() {
+					continue
+				}
+			*/
 
 			// implicit resources don't have their own metadata, so we grab it from
 			// the resource itself


czunker:
@arlimus Could you please have a look? Removing the check would solve the mentioned issues, but I assume the check was there for a reason.

This would fix autocomplete for resources like `k8s.kubelet`, `k8s.pod`, and all the other k8s resources without an "s" at the end.

arlimus:
Closing this because of https://github.com/mondoohq/cnquery/issues/215#issuecomment-1284402612
In summary: it would show a lot of resources that users can't just call, because they require users to initialize them. Due to the way `k8s.pod` and `k8s,kubelet` are structured right now, it causes them to show up in that list too, despite the fact that both actually auto-init everything. However, there are other resources that don't auto-init and those do require users to write things out, which we don't explain to them in the shell, thus they don't show up right now. Users tha know how they work can still use the resource.

PR #343, opened 2022-10-19
🐛 fixes two issues with the command builder

chris-rock:
- we accidentally attached the ebs subcommand twice to the aws root command
- the gitlab command used the wrong provider type


czunker (approved):
LGTM

PR #346, opened 2022-10-20
✨ time + duration

arlimus:
This was missing and I noticed it while writing the effective MQL guide. Also quickly realized why it was missing. We should revisit the types around time and duration as well.

image


chris-rock (approved):
Thank you @arlimus

PR #349, opened 2022-10-20
✨Support IP addresses and instance ids for ic and ssm connectivity.

preslavgerchev:
Supports both:
```
cnquery shell aws ec2 ssm ubuntu@255.255.255.255
cnquery shell aws ec2 instance-connect ubuntu@255.255.255.255
```
and
```
cnquery shell aws ec2 ssm ubuntu@i-03a9f7b6c706f43d5
cnquery shell aws ec2 instance-connect ubuntu@i-03a9f7b6c706f43d5
```


vjeffrey (approved):
working beautifully!

PR #350, opened 2022-10-20
🧹 Improve error message if no identity file or password are provided for ssh auth

preslavgerchev:
Fixes https://github.com/mondoohq/cnquery/issues/291

```
➜ go run apps/cnquery/cnquery.go shell ssh ubuntu@255.255.255.255
! No identity file or password are provided for ssh authentication, use either --identity-file, --ask-pass or --password. Falling back to ssh agent authentication.
x could not connect to asset error="ssh: handshake failed: ssh: unable to authenticate, attempted methods [none], no supported methods remain" asset=255.255.255.255
FTL could not resolve assets
exit status 1
```

Please feel free to improve the error message if you want to!

### DIFF
--- a/apps/cnquery/cmd/builder/parse.go
+++ b/apps/cnquery/cmd/builder/parse.go
@@ -141,6 +141,9 @@ func ParseTargetAsset(cmd *cobra.Command, args []string, providerType providers.
 		if err != nil {
 			log.Error().Err(err).Msg("cannot parse target")
 		}
+		if identityFile == "" && password == "" {
+			log.Warn().Msg("No identity file or password are provided for ssh authentication, use either --identity-file, --ask-pass or --password. Falling back to ssh agent authentication.")
+		}
 		connection.Host = target.Hostname
 		connection.Port = target.Port
 		connection.Path = target.Path


preslavgerchev:
Realised this now breaks agent auth, putting in draft until I fix this

vjeffrey (approved):

apps/cnquery/cmd/builder/parse.go
```diff
@@ -141,6 +141,9 @@ func ParseTargetAsset(cmd *cobra.Command, args []string, providerType providers.
 		if err != nil {
 			log.Error().Err(err).Msg("cannot parse target")
 		}
+		if identityFile == "" && password == "" {
```
chris-rock, line 144: I think this does not catch all valid cases. There is also aws instance connect and ssm

preslavgerchev: You don't have to provide creds for ic/ssm (far as I know), aws generates a temporary keypair there

chris-rock: You are right, this is covered as a separate asset type. never mind.

PR #356, opened 2022-10-20
⭐️ use upstream configuration if credentials are provided

chris-rock:
This change allows authenticated users to use eol and vulnerability data from mql resources. This is the first step. In a second iteration we connect cnquery to the public Mondoo Vulnerability Database (MVD) so that users do not need to be authenticated.


arlimus (approved):
Yes, yay, awesome!

PR #357, opened 2022-10-20
add --namespaces-exclude --namespaces parameters

joelddiaz:
--namespaces-exclude allows cnquery to ignore a comma-separated list of Namespaces whose resources should be skipped when performing discovery.

On the other hand, specifying --namespaces will only allow resources matching the list of Namespaces provided.

Combining the previously-existing --namespace param (singular) with --namespaces (plural) will just be joined into one big allow-list. So --namespaces=default,kube-system is the same as saying --namespace=kube-system --namespaces=default.

### DIFF
--- a/apps/cnquery/cmd/builder/builder.go
+++ b/apps/cnquery/cmd/builder/builder.go
@@ -394,9 +394,11 @@ func kubernetesProviderCmd(commonCmdFlags commonFlagsFn, preRun commonPreRunFn, 
 		Args:    cobra.MaximumNArgs(1),
 		PreRun: func(cmd *cobra.Command, args []string) {
 			preRun(cmd, args)
+			// FIXME: DEPRECATED, remove in v8.0 vv
 			viper.BindPFlag("namespace", cmd.Flags().Lookup("namespace"))
+			// ^^
 			viper.BindPFlag("namespaces-exclude", cmd.Flags().Lookup("namespaces-exclude"))
-			viper.BindPFlag("namespaces-include", cmd.Flags().Lookup("namespaces-include"))
+			viper.BindPFlag("namespaces", cmd.Flags().Lookup("namespaces"))
 			viper.BindPFlag("context", cmd.Flags().Lookup("context"))
 		},
 		Run: func(cmd *cobra.Command, args []string) {
@@ -407,10 +409,14 @@ func kubernetesProviderCmd(commonCmdFlags commonFlagsFn, preRun commonPreRunFn, 
 		},
 	}
 	commonCmdFlags(cmd)
+	// FIXME: DEPRECATED, remove in v8.0
 	cmd.Flags().String("namespace", "", "filter kubernetes objects by namespace")
+	cmd.Flags().MarkHidden("namespace")
+	// ^^
+
 	cmd.Flags().String("context", "", "target a kubernetes context")
 	cmd.Flags().String("namespaces-exclude", "", "filter out kubernetes objects in the matching namespaces")
-	cmd.Flags().String("namespaces-include", "", "only include kubernetes object in the matching namespaces")
+	cmd.Flags().String("namespaces", "", "only include kubernetes object in the matching namespaces")
 	return cmd
 }
 
--- a/apps/cnquery/cmd/builder/builder.go
+++ b/apps/cnquery/cmd/builder/builder.go
@@ -395,7 +395,8 @@ func kubernetesProviderCmd(commonCmdFlags commonFlagsFn, preRun commonPreRunFn, 
 		PreRun: func(cmd *cobra.Command, args []string) {
 			preRun(cmd, args)
 			viper.BindPFlag("namespace", cmd.Flags().Lookup("namespace"))
-			viper.BindPFlag("skip-namespaces", cmd.Flags().Lookup("skip-namespaces"))
+			viper.BindPFlag("namespaces-exclude", cmd.Flags().Lookup("namespaces-exclude"))
+			viper.BindPFlag("namespaces-include", cmd.Flags().Lookup("namespaces-include"))
 			viper.BindPFlag("context", cmd.Flags().Lookup("context"))
 		},
 		Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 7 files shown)


apps/cnquery/cmd/builder/builder.go
```diff
@@ -395,6 +395,7 @@ func kubernetesProviderCmd(commonCmdFlags commonFlagsFn, preRun commonPreRunFn,
 		PreRun: func(cmd *cobra.Command, args []string) {
 			preRun(cmd, args)
 			viper.BindPFlag("namespace", cmd.Flags().Lookup("namespace"))
+			viper.BindPFlag("skip-namespaces", cmd.Flags().Lookup("skip-namespaces"))
```
chris-rock: I think we should align this to the kubernetes config `namespaceExlude`, so it should be `--namespace-exlude`

apps/cnquery/cmd/builder/parse.go
```diff
@@ -231,6 +231,12 @@ func ParseTargetAsset(cmd *cobra.Command, args []string, providerType providers.
 		} else if namespace != "" {
 			connection.Options["namespace"] = namespace
 		}
+
+		if skipNamespaces, err := cmd.Flags().GetString("skip-namespaces"); err != nil {
+			log.Fatal().Err(err).Msg("cannot parse --skip-namespaces values")
+		} else if skipNamespaces != "" {
+			connection.Options["skip-namespaces"] = skipNamespaces
```
chris-rock: That option should match the wording in the operator too

chris-rock (changes requested):

apps/cnquery/cmd/builder/builder.go
joelddiaz: namespaceExclude or namespaceSExclude?

motor/discovery/k8s/resolver.go
```diff
@@ -80,28 +82,45 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
...
+		nsFilter.include = append(nsFilter.include, strings.Split(includeNamespaces, ",")...)
+	}
+
+	// Put a Warn() message if a Namespace that doesn't exist was part of the
+	// list of Namespaces to include
+	clusterNamespaces, err := p.Namespaces()
+	if err != nil {
+		return nil, err
+	}
+	for _, ns := range nsFilter.include {
```
czunker, line 108: I'll just repeat Chris comment on my PR, when I tried the same: https://github.com/mondoohq/cnquery/pull/256#discussion_r993473673

I tried what Chris suggested, but that made the loop more complicated. Perhaps you find an easy solution or we leave it as is for better readablility.

motor/discovery/k8s/list_workloads.go
```diff
@@ -13,94 +13,99 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+type NamespaceFilterOpts struct {
+	include []string
+	ignore  []string
```
czunker: nit: for consistency I'd suggest exlude.

motor/discovery/k8s/resolver.go
joelddiaz: Yeah, I'm not sure how to restructure it to avoid looping through the provided list to filter on...
We could use sets and do some set comparison to figure out what in the filter list doesn't exist in the actual list of namespaces, but I imagine that implementation would also have to do a similar approach.

motor/discovery/k8s/list_workloads_test.go
```diff
@@ -1550,3 +1550,85 @@ func TestListStatefulSets_Filter(t *testing.T) {
...
+		{
+			TypeMeta:   metav1.TypeMeta{Kind: "Pod", APIVersion: "v1"},
+			ObjectMeta: metav1.ObjectMeta{Name: "some-workload2", Namespace: nss[2].Name},
+		},
+		{
+			TypeMeta:   metav1.TypeMeta{Kind: "Pod", APIVersion: "v1"},
+			ObjectMeta: metav1.ObjectMeta{Name: "some-workload3", Namespace: nss[2].Name},
+		},
+	}
+	p.EXPECT().Pods(nss[0]).Return(defaultNamespacePods, nil).AnyTimes()
```
czunker, line 1601: Nice. Learned something new: `.AnyTimes()`

motor/discovery/k8s/list_workloads_test.go
```diff
@@ -1550,3 +1550,85 @@ func TestListStatefulSets_Filter(t *testing.T) {
...
+
+	// List 'kube-system' and 'other-namespace'
+	assets, err = ListPods(p, pCfg, clusterIdentifier, NamespaceFilterOpts{include: []string{nss[1].Name, nss[2].Name}}, make(map[string][]K8sResourceIdentifier), ownershipDir)
+	require.NoError(t, err)
+	assert.Equal(t, 4, len(assets), "expected only 1 Pod to be returned")
+
+	// Exclude kube-system
+	assets, err = ListPods(p, pCfg, clusterIdentifier, NamespaceFilterOpts{exclude: []string{nss[1].Name}}, make(map[string][]K8sResourceIdentifier), ownershipDir)
+	require.NoError(t, err)
+	assert.Equal(t, 5, len(assets), "expected only 1 Pod to be returned")
```
czunker: That should be "expected only 5 Pod to be returned"

motor/discovery/k8s/list_workloads_test.go
```diff
@@ -1550,3 +1550,85 @@ func TestListStatefulSets_Filter(t *testing.T) {
...
+
+	// List only 'kube-system'
+	assets, err = ListPods(p, pCfg, clusterIdentifier, NamespaceFilterOpts{include: []string{nss[1].Name}}, make(map[string][]K8sResourceIdentifier), ownershipDir)
+	require.NoError(t, err)
+	assert.Equal(t, 1, len(assets), "expected only 1 Pod to be returned")
+
+	// List 'kube-system' and 'other-namespace'
+	assets, err = ListPods(p, pCfg, clusterIdentifier, NamespaceFilterOpts{include: []string{nss[1].Name, nss[2].Name}}, make(map[string][]K8sResourceIdentifier), ownershipDir)
+	require.NoError(t, err)
+	assert.Equal(t, 4, len(assets), "expected only 1 Pod to be returned")
```
czunker: That should be "expected only 4 Pod to be returned"

czunker (approved):

chris-rock (approved):
Thank you @joelddiaz

PR #360, opened 2022-10-21
🐛 select querypacks on cli

arlimus:
This behavior was buggy for querypack. Note, it only works in incognito right now


chris-rock (approved):
Tested and works like a charm with local bundles

PR #361, opened 2022-10-21
🧹 Use extra Error type for 'object not found'

czunker:

### DIFF
--- a/resources/packs/k8s/common.go
+++ b/resources/packs/k8s/common.go
@@ -15,9 +15,9 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
-type K8sObjectNotFound struct{}
+type K8sResourceNotFound struct{}
 
-func (e *K8sObjectNotFound) Error() string {
+func (e *K8sResourceNotFound) Error() string {
 	return "could not find k8s resource"
 }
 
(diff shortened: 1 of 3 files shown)


resources/packs/k8s/common.go
```diff
@@ -15,6 +15,12 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+type K8sObjectNotFound struct{}
+
+func (e *K8sObjectNotFound) Error() string {
```
imilchev: nit: Since we call it "k8s resource" in the log messages, shouldn't the error be called `K8sResourceNotFound`?

resources/packs/k8s/common.go
```diff
@@ -241,7 +247,8 @@ func initNamespacedResource[T K8sNamespacedObject](
 		}
 	}
 
-	return args, *new(T), fmt.Errorf("not found")
+	// the error K8sObjectNotFound is checked by cnspec
+	return args, *new(T), &K8sObjectNotFound{}
```
imilchev: Where will the error be checked? It seems like in `runtime.go` you are not returning that error type anymore but just `error`. Is it intentional?

resources/packs/k8s/common.go
czunker: Good point. Thanks.

resources/packs/k8s/common.go
czunker: It is check in cnspec: https://github.com/mondoohq/cnspec/pull/131/files#diff-0a279e72affdf3ff0b6e366176e86400378905934fa1b2203a50a9cae6ae74cbR323

imilchev (approved):

PR #363, opened 2022-10-21
🧹 remove .list from services in readme

chris-rock:
- new ascii art header
- simplify `services.list` to `services`


arlimus (approved):
nice catch :D

PR #364, opened 2022-10-21
🧹 remove .list from query examples

chris-rock:
We recently removed the requirements for list resources to use `.list`. This adapts the used examples. In addition it adds defaults fields for `auditpol` and `mount`.


arlimus (approved):
Kudos!!

PR #368, opened 2022-10-21
🐛 Make sure k8s-node assets are not created for asset relationships

imilchev:
Fixes a bug that still discovers `k8s-node` assets via asset relationships


chris-rock (approved):
Thank you @imilchev

PR #369, opened 2022-10-21
🧹 refactor mvd client

chris-rock:
This change refactors the MVD client out of the resource packs because they should not have been there in the first place. I moved them over to the upstream directory.


arlimus (approved):
Time to get those vulnerabilities back

PR #370, opened 2022-10-22
🧹 add missing auto-generated file

chris-rock:
A follow-up to #369 I am not 100% sure how the go tests ever passed that PR.


arlimus (approved):
good catch!

PR #371, opened 2022-10-22
🧹 rename transport to provider

chris-rock:
This change covers a few more renames that we forgot.


arlimus (approved):
Nice catch! This really got lost

PR #372, opened 2022-10-22
🐛 fix ssh authentication

### LINKED ISSUE #291: `mondoo scan ssh FOO` authentication message leaves user hanging
When the user tries to ssh scan to a host that requires a password, but the user has not supplied one or specified ask-pass we give a pretty cryptic message

```
cnspec scan ssh tsmith@123.123.123.123 --incognito
x could not resolve asset error="ssh: handshake failed: ssh: unable to authenticate, attempted methods [none password], no supported methods remain" asset=123.123.123.123
FTL failed to run scan error="failed to resolve multiple assets"
```

We should help the user out here by suggesting how they can properly authenticate to this asset.

chris-rock:
When the cli was refactored, the cli always added a credential with password authentication as soon as a username was set. The ssh provider check to enable ssh config reading and ssh agent fallback only triggered when no credential was provided. This change adapts the checks so that the fallbacks to ssh config and ssh agent start working again.

fixes #291

### DIFF
--- a/apps/cnquery/cmd/builder/parse.go
+++ b/apps/cnquery/cmd/builder/parse.go
@@ -141,9 +141,6 @@ func ParseTargetAsset(cmd *cobra.Command, args []string, providerType providers.
 		if err != nil {
 			log.Error().Err(err).Msg("cannot parse target")
 		}
-		if identityFile == "" && password == "" {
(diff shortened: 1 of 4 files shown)


arlimus (approved):
thank you!

PR #380, opened 2022-10-24
🧹 User general error when resources aren't found

czunker:

### DIFF
--- a/resources/packs/k8s/common.go
+++ b/resources/packs/k8s/common.go
@@ -15,12 +15,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
-type K8sResourceNotFound struct{}
-
-func (e *K8sResourceNotFound) Error() string {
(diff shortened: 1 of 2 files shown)


resources/packs/k8s/common.go
```diff
@@ -308,13 +302,13 @@ func initResource[T K8sObject](
...
+	for i := range k8sResources {
+		entry := k8sResources[i].(T)
 		if matchFn(entry) {
 			return nil, entry, nil
 		}
 	}
 
 	// the error K8sResourceNotFound is checked by cnspec
-	return nil, *new(T), &K8sResourceNotFound{}
+	return nil, *new(T), &resources.ResourceNotFound{}
```
preslavgerchev, line 313: Nit: The comment above refers to the old type

czunker: Good catch. Thanks.

PR #390, opened 2022-10-27
🐛 Remove mondoo.com/instance label for containers.

preslavgerchev:
It seems containers get tagged with the `mondoo.com/instance` which is an AWS-specific label.
image


czunker (approved):
LGTM

Besides your changes, I'm wondering why these two blocks look different. Both are titled "// fetch docker specific metadata" and both are about containers. Shouldn't they be the same?

preslavgerchev:
> LGTM
> 
> Besides your changes, I'm wondering why these two blocks look different. Both are titled "// fetch docker specific metadata" and both are about containers. Shouldn't they be the same?

One is related to `ContainerInfo`, the other is about an `Asset`. The labels from `ContainerInfo` do get end up being passed through though eventually.

PR #407, opened 2022-10-31
Update message for when cnquery is already logged in

tas50:
Make this slightly more clear


chris-rock (approved):
Thank you @tas50

PR #413, opened 2022-10-31
🐛 prevent cnquery from panicking if there are no querybundles

jaym:
When in incognito mode and no querypack-bundle specified, this was panicking during scan.


chris-rock (approved):
Thank you @jaym

PR #418, opened 2022-11-01
🐛 propagate errors about port-process-relationships

arlimus:
Previously when reading out the file descriptors, we would silently ignore errors that happened there. However, it is quite common for someone without sufficient permissions to read this list, e.g. when we ask which port is opened by which process. When this happens, now we propagate the errors and wrap it nicely so that users know it is a permissions issue and can sudo their way out of it.

### DIFF
--- a/resources/packs/core/port.go
+++ b/resources/packs/core/port.go
@@ -8,7 +8,9 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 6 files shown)


chris-rock:
Thank you @arlimus

chris-rock (approved):

PR #423, opened 2022-11-01
Be more clear that no provider == local

tas50:
Right now we just say we're "using defaults" but what is the default?

### DIFF
--- a/apps/cnquery/cmd/builder/builder.go
+++ b/apps/cnquery/cmd/builder/builder.go
@@ -54,7 +54,7 @@ func NewProviderCommand(opts CommandOpts) *cobra.Command {
 				os.Exit(1)
 			}
 
-			log.Info().Str("provider", "local").Msg("no provider specified, using defaults.\n  Use --help for a list of available providers.")
+			log.Info().Str("provider", "local").Msg("no provider specified, defaulting to local.\n  Use --help for a list of available providers.")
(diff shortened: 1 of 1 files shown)


apps/cnquery/cmd/builder/builder.go
```diff
@@ -54,7 +54,7 @@ func NewProviderCommand(opts CommandOpts) *cobra.Command {
 				os.Exit(1)
 			}
 
-			log.Info().Str("provider", "local").Msg("no provider specified, using defaults.\n  Use --help for a list of available providers.")
+			log.Info().Str("provider", "local").Msg("no provider specified, defaulting to local.\n  Use --help for a list of available providers.")
```
chris-rock: If we add this to the text, we need to remove `.Str("provider", "local")` otherwise we add the same information twice

chris-rock (changes requested):

tas50:
@chris-rock it looks fine in my local version:

w/o a provider specified:
```
 ~/d/cnquery   $  /Users/tsmith/go/bin/cnquery shell
  Use --help for a list of available providers. provider=local
```

chris-rock:
I update the PR to remove the property, the second line `Use --help for a list of available providers. provider=local` is confusing because you moved the info about the local provider into the first line.

chris-rock (approved):

PR #427, opened 2022-11-01
🐛 clean up error handling on github webhooks resource

vjeffrey:
fixes https://github.com/mondoohq/cnquery/issues/426

we were logging the error -- if it's a 404 that means nothing was found, so we can just move right along. if there is a non-404 error it will be returned, so really no need to log the error here


chris-rock (approved):
Thank you @vjeffrey

PR #428, opened 2022-11-01
🐛 include tags on rds snapshot resource nested in rds dbClusters

vjeffrey:
fixes https://github.com/mondoohq/cnquery/issues/394


chris-rock (approved):
Thank you @vjeffrey

PR #430, opened 2022-11-02
⭐️ Add certificate/password authentication support for Azure.

preslavgerchev:
Similar to how we authenticate against MS365 currently, this allows us to run
```
cnquery shell azure --client-id <id> --certificate-path /Users/preslavgerchev/certificate.pfx --tenant-id <tenant-id> --certificate-secret supersecret
```
or
```
cnquery shell azure --client-id <id> --tenant-id <tenant-id> --client-secret mysecret
```

It now also supports the argument-less version:
```
cnquery shell azure
```
Furthermore, extended support for multiple subscriptions. If the `subscription` flag is not set, we discover all and let the user pick:
image


motor/providers/azure/authentication.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/cockroachdb/errors"
+	"github.com/rs/zerolog/log"
+	"go.mondoo.com/cnquery/motor/vault"
+)
+
+// GetAuthorizer determines what authorizer to use, based on the passed in configs. Possible options are:
+// - Authorizer via the CLI that uses the `az` executable
+// - Authorizer via password that uses the sdk
+// - Authorizer via certificate that uses the sdk
+func GetAuthorizer(clientId, tenantId, resource string, credential *vault.Credential) (autorest.Authorizer, error) {
```
preslavgerchev, line 15: Moved this here from `provider.go` to let us use it w/o neeeding to actually have a provider (i.e. when discovering intitially).

preslavgerchev:
- [ ] there's one issue with credentials type being `unknown`, going to look into that before this can be merged

motor/discovery/azure/resolver.go
```diff
@@ -29,102 +31,123 @@ func (r *Resolver) AvailableDiscoveryTargets() []string {
...
+	// we either discover all subs (if no filter is provided) or just verify that the provided one exists
+	var subs []subscriptions.Subscription
+	subsClient := NewSubscriptions(azureClient)
+	if subscriptionID != "" {
+		subscription, err := subsClient.GetSubscription(subscriptionID)
+		if err != nil {
+			return nil, err
+		}
+		if tc.Options["tenant-id"] == "" {
+			tc.Options["tenant-id"] = *subscription.TenantID
```
vjeffrey, line 72: how does all this work with related assets/asset relationships? i may have missed it, but i didn't think i saw anywhere you're setting the related assets?

chris-rock:
This is a great step forward. What we need to do though is to go away from autorest and use the new azidentity that is used for ms365. See migration guide here https://github.com/Azure/azure-sdk-for-go/blob/main/documentation/MIGRATION_GUIDE.md#authentication

motor/discovery/azure/resolver.go
preslavgerchev: It's indeed not there yet, we don't have anything about asset relationships in Azure so far. That can be a future PR, I don't know yet what we need to link together

apps/cnquery/cmd/builder/builder.go
```diff
@@ -552,6 +552,11 @@ func azureProviderCmd(commonCmdFlags commonFlagsFn, preRun commonPreRunFn, runFn
 		},
 	}
 	commonCmdFlags(cmd)
+	cmd.Flags().String("tenant-id", "", "directory (tenant) ID of the service principal")
```
chris-rock, line 555: Nice that those are aligned with microsoft 365

motor/discovery/azure/compute.go
```diff
@@ -14,44 +14,29 @@ import (
 	azure_transport "go.mondoo.com/cnquery/motor/providers/azure"
 )
 
-type AzureClient struct {
-	Subscription string
-	Authorizer   autorest.Authorizer
-}
-
-func (c *AzureClient) VirtualMachinesClient() compute.VirtualMachinesClient {
+func (c *Compute) VirtualMachinesClient() compute.VirtualMachinesClient {
```
chris-rock, line 29: nit: can we move the method definitions underneath the definition of `type Compute struct`. This would improve readability.

chris-rock (approved):
Thank you @preslavgerchev Great improvement. I added a nit comment. This can also be changed in follow-ups

PR #431, opened 2022-11-02
🧹 add cnquery version and build to client headers on upstream scan

vjeffrey:
add cnquery version and build to client headers on upstream scan so that we can read in the client version and build info server-side on upstream scan

### DIFF
--- a/apps/cnquery/cmd/root.go
+++ b/apps/cnquery/cmd/root.go
@@ -132,17 +132,20 @@ func sysInfoHeader(sysInfo *sysinfo.SystemInfo, features cnquery.Features) range
 	)
 
 	h := http.Header{}
-	h.Set(HttpHeaderUserAgent, scope.XInfoHeader(map[string]string{
+	info := map[string]string{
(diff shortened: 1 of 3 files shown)


apps/cnquery/cmd/root.go
```diff
@@ -132,17 +132,20 @@ func sysInfoHeader(sysInfo *sysinfo.SystemInfo, features cnquery.Features) range
...
 		"cnquery": cnquery.Version,
 		"build":   cnquery.Build,
-		"PN":      sysInfo.Platform.Name,
-		"PR":      sysInfo.Platform.Version,
-		"PA":      sysInfo.Platform.Arch,
-		"IP":      sysInfo.IP,
-		"HN":      sysInfo.Hostname,
-	}))
+	}
+	if sysInfo != nil {
```
vjeffrey, line 139: the only real change here is this nil checking, which i did to avoid setting a log.Fatal() if we can't fetch the sys info

preslavgerchev (approved):

PR #439, opened 2022-11-04
🐛 Fix loading of base64 config from env

czunker:
When a default config file is present, it previously overrode the env var. Now reading the default config file is skipped when the config is loaded as base64 encoded env var.

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -36,6 +36,8 @@ var (
 	Features             cnquery.Features
 )
 
+const configSourceBase64 = "$MONDOO_CONFIG_BASE64"
+
 // Init initializes and loads the mondoo config
 func Init(rootCmd *cobra.Command) {
 	cobra.OnInitialize(initConfig)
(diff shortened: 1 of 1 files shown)


czunker:
This is how it looked before.
Default config connects to us API:
```
cnquery status 
...
→ API ConnectionConfig:	https://us.api.mondoo.com
...
```

New config connects to eu API:
```
cat /tmp/creds.json | jq
{
...
  "api_endpoint": "https://eu.api.mondoo.com"
}
```
But, `status` was still showing the us API:
```
export MONDOO_CONFIG_BASE64=$(cat /tmp/creds.json | base64)
cnquery status
 → loaded configuration from /etc/opt/mondoo/mondoo.yml using source $MONDOO_CONFIG_BASE64
...
→ API ConnectionConfig:	https://us.api.mondoo.com
...
```


Now it looks like this:
```
cnquery status
 → loaded configuration from environment using source $MONDOO_CONFIG_BASE64
...
→ API ConnectionConfig:	https://eu.api.mondoo.com
...
```

cli/config/config.go
```diff
@@ -219,6 +221,8 @@ func DisplayUsedConfig() {
 		log.Warn().Msg("could not load configuration file " + UserProvidedPath)
 	} else if LoadedConfig {
 		log.Info().Msg("loaded configuration from " + viper.ConfigFileUsed() + " using source " + Source)
+	} else if Source == configSourceBase64 {
```
chris-rock, line 226: should we also cover the `MONDOO_CONFIG_PATH` case here?

czunker: That was covered by `} else if LoadedConfig {`.

But this PR has a bug. I have to fix it to have it covered there again. Will push another commit.

czunker: Fixed, now this case looks as before:
```
export MONDOO_CONFIG_PATH=/tmp/creds.yaml
 ~  cnquery status 
→ loaded configuration from /tmp/creds.yaml using source $MONDOO_CONFIG_PATH
→ Platform:	ubuntu
...
```

chris-rock (approved):
Thank you @czunker

PR #453, opened 2022-11-08
🐛 Alloow one or two args

czunker:
Fixes https://github.com/mondoohq/cnspec/issues/171


czunker:
As discussed in the issue, this will be solved in the action.

PR #454, opened 2022-11-09
⭐️ Update Azure SDKs and resources

preslavgerchev:
Updates the Azure SDK + Azure resources. Some fields are deprecated as the new API does not expose them but mostly it's all the same. The main changes here are:
 - using `azidentity` for authN/authZ towards Azure, replacing `autorest`
 - Move `msgraph` adapter/request code to another module so it can be reused by both `ms365` and `azure` providers
 - Using the MS Graph API for `azuread` resources - `users`, `groups`, `domains`
 - Most of the other changes are a product of switching to `azidentity` and the new azure APIs.

I've tried to split changes by commits so hopefully that's a bit easier to read


resources/packs/azure/azure.lr
```diff
@@ -19,9 +19,11 @@ azuread.user {
   accountEnabled bool
   mailNickname string
   mail string
-  objectType string
+  // deprecated, no longer used
+  objectType() string
```
preslavgerchev, line 23: I did not drop the deprecated fields as that's breaking change but rather made them a function. Semantically (in MQL) it's all the same, but that allows us to throw an error when computing the value.

resources/packs/msgraph/msgraphclient/client.go
```diff
@@ -9,8 +9,11 @@ import (
 	"github.com/microsoftgraph/msgraph-beta-sdk-go/applications"
 	"github.com/microsoftgraph/msgraph-beta-sdk-go/devicemanagement"
 	"github.com/microsoftgraph/msgraph-beta-sdk-go/domains"
+
```
chris-rock: nit: whitespace in imports

resources/packs/msgraph/msgraphclient/adapter.go
```diff
@@ -3,6 +3,7 @@ package msgraphclient
 import (
 	nethttp "net/http"
 
+	"github.com/cockroachdb/errors"
```
chris-rock: I think msgraph package is not in the correct spot here? The directory should only contain resource packs. Should this be moved into the ms365 provider? Seems like it is primary a client assigned with the provider

motor/discovery/azure/resolver.go
```diff
@@ -53,12 +53,12 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 		}
```
chris-rock, line 53: line 45 should be "could not create azure provider"

chris-rock (commented):
Great improvement @preslavgerchev I added a few nits that we should fix before we merge it

resources/packs/msgraph/msgraphclient/adapter.go
preslavgerchev: Done, moved to the provider under its own module

chris-rock (approved):
Thank you @preslavgerchev

PR #459, opened 2022-11-10
✨ array delete: [1,2,3,3,4] - [3,4,5] = [1,2]

arlimus:
Pretty straightforward, after implementing ways to concat arrays, add the ability to quickly remove elements. This is going to be the most commonly used intuitive shorthand for this type of operation. We already have ways to filter content (`where`) and will likely use similar approaches to remove elements via index. That however is a different use-case.

```
> [1,2,3,3,4] - [3,4,5]
[1,2]
```


atomic111:
@arlimus, thanks for great improvement

preslavgerchev (approved):
Nice one! 
Works also with more complex queries like `users.map(uid) - users.map(gid)`

PR #460, opened 2022-11-10
⭐ port.tls to access tls params on a port

arlimus:
This is just a first step to kick this off. Next step will be to switch out the port resources to use socket underneath, then work around the fact that a bunch of ports do not support tls (and we want to reflect that; for now it's just empty), finally expand integration testing for this.

image

and an example for non-tls ports

image


chris-rock:
Nice clean addition. The pass-through via sockets also allows us to test things remote.

chris-rock (approved):
Thank you @arlimus

PR #461, opened 2022-11-10
✨Improve msgraph API error handling.

preslavgerchev:
Before:
image

After:
image

### DIFF
--- a/motor/providers/ms365/msgraphclient/errors.go
+++ b/motor/providers/ms365/msgraphclient/errors.go
@@ -1,0 +1,17 @@
+package msgraphclient
+
+import (
+	"github.com/cockroachdb/errors"
+	od "github.com/microsoftgraph/msgraph-beta-sdk-go/models/odataerrors"
+)
+
+func TransformODataError(err error) error {
+	oDataErr := err.(*od.ODataError)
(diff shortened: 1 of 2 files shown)


resources/packs/azure/azure_ad.go
```diff
@@ -41,7 +41,7 @@ func (a *mqlAzuread) GetUsers() ([]interface{}, error) {
 	}
 	userList, err := usersClient.Users().Get()
 	if err != nil {
-		return nil, err
+		return nil, msgraphclient.TransformODataError(err)
```
chris-rock: Should azure really depend on the msgraph package? Is there a odataerror package in the azure sdk

chris-rock (commented):
Nice addition

resources/packs/azure/azure_ad.go
preslavgerchev: I can move the error transformation out. The msgraph package has the odataerror in it, we use the `msgraphclient` for users,domains, etc.

This probably brings up a good point of doing the same thing twice: via `ms365` and via `azure` providers. Should we consider dropping one of those?

chris-rock (approved):
Thank you @preslavgerchev

PR #464, opened 2022-11-10
🐛Fix passing of client secret for ms365.

preslavgerchev:
We were using the wrong arg here, the client secret was being ignored


chris-rock (approved):
Thank you @preslavgerchev

PR #465, opened 2022-11-10
⭐️ powerpc 64 little-endian build

chris-rock:
As per user request, we are adding power pc architecture to our build pipeline.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -13,6 +13,7 @@ builds:
       - 386
       - arm64
       - arm
+      - ppc64le
     # ARM 6= Raspberry Pi A, A+, B, B+, Zero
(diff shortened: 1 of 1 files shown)


tas50:
Let's make sure to update the docs with this support.

.goreleaser.yml
```diff
@@ -13,6 +13,7 @@ builds:
       - 386
       - arm64
       - arm
+      - ppc64le
```
czunker, line 16: Do we also need a docker image for power?

chris-rock: I think we should, but I wanted to split this in iterations. First lets test binary build and then we can test the docker build.

arlimus (approved):
sweet addition!!

PR #466, opened 2022-11-11
Fix resource comment that breaks docs tables

tas50:
You can't use a | in a resource description.


chris-rock (approved):
Thank you @tas50

PR #470, opened 2022-11-12
🧹 be more specific for gcp flags and use `--project-id` and `organization-id`

chris-rock:
Before it was a bit unclear which arguments should be used. `project` can also be the name of the project but we only supported the project-id. This is a non-breaking change because the old flags stay available but are hidden now. If the inventory is used, the old properties are being supported as well.

```
cnquery shell gcp --organization-id 12345678
```

### DIFF
--- a/motor/providers/gcp/provider.go
+++ b/motor/providers/gcp/provider.go
@@ -28,15 +28,22 @@ func New(pCfg *providers.Config) (*Provider, error) {
 		return nil, providers.ErrProviderTypeDoesNotMatch
 	}
 
-	if pCfg.Options == nil || (pCfg.Options["project"] == "" && pCfg.Options["organization"] == "") {
+	if pCfg.Options == nil || (pCfg.Options["project-id"] == "" && pCfg.Options["project"] == "" && pCfg.Options["organization-id"] == "" && pCfg.Options["organization"] == "") {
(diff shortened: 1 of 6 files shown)


tas50:
Do we have any docs or example code in the UI to update with this change?

chris-rock:
Maybe, but that is non-breaking so the current behavior keeps working

tas50:
It will be breaking in 8.0 though so we should make sure to update any examples now. No need to get new users onto old behaviors if we can avoid it.

arlimus:
Good point, I think if we move users to the new flags, we should at least show a deprecation warning to let them know to move to the new flags. That will allow us to deprecate the old flags in v8.

chris-rock:
added the deprecation warning to the flag:

```
cnquery shell gcp --organization 12434456
Flag --organization has been deprecated, --organization is deprecated in favor of --organization-id
```

motor/providers/gcp/provider.go
```diff
@@ -100,7 +107,7 @@ func (p *Provider) Kind() providers.Kind {
 }
 
 func (p *Provider) Runtime() string {
-	return providers.RUNTIME_AWS
+	return providers.RUNTIME_GCP
```
arlimus, line 110: nice catch!

arlimus (approved):
Thank you for the adjustments, I will follow up with a nit, glad to have this in!

PR #471, opened 2022-11-12
🧹 rename the gcp provider to google provider

chris-rock:
after #470

This is no functional change nor a breaking change. CLI commands and inventory settings stay the same. It is just the reflection that the google api's reflect more than just GCP. Therefore we rename the provider to be named google. The resource pack that provides access to GCP resources stays the same. This allows us to use the same provider for more google services going forward.


arlimus (approved):
Thank you Chris!

Full background: we have extended services that use the same provider i.e. we don't have to create another provider to access auxiliary services that are not strictly considered GCP. Since they are not part of GCP, but use that provider, things get a little confusing. To avoid this, we are renaming the provider to better reflect the interaction with APIs going forward.

If you found this PR after 2022: The provider structure is constantly evolving, please check the current structure for an up to date reference to these providers.

PR #474, opened 2022-11-13
🧹 add timeouts to snapshot creation and search for ebs volume scan

vjeffrey:
adding another completion check to the snapshot creation process and adding some timeout checks to ensure we don't eternally loop 
```
! WARNING: you are using an early access feature feature=AWS_EC2_EBS status=experimental
→ validate state instance={"AccountId":"185972265011","Id":"i-04614e3ab48488e5f","PlatformId":"//platformid.api.mondoo.app/runtime/aws/ec2/v1/accounts/185972265011/regions/us-east-1/instances/i-04614e3ab48488e5f","Region":"us-east-1"}
→ find volume id instance={"AccountId":"185972265011","Id":"i-04614e3ab48488e5f","PlatformId":"//platformid.api.mondoo.app/runtime/aws/ec2/v1/accounts/185972265011/regions/us-east-1/instances/i-04614e3ab48488e5f","Region":"us-east-1"}
→ find recent snapshot
→ create snapshot
→ waiting for snapshot completion; sleeping 10 seconds progress=
→ snapshot complete progress=100%
→ checking snapshot region scanner instance=us-east-1 snapshot=us-east-1
→ create volume
→ waiting for volume creation completion; sleeping 10 seconds state=creating
```

### DIFF
--- a/motor/providers/awsec2ebs/setup.go
+++ b/motor/providers/awsec2ebs/setup.go
@@ -165,7 +165,6 @@ func GetVolumeIdForInstance(instanceinfo *types.Instance) *string {
 }
 
 func (t *Provider) FindRecentSnapshotForVolume(ctx context.Context, v VolumeId) (bool, SnapshotId, error) {
-	log.Info().Msg("find recent snapshot")
 	return FindRecentSnapshotForVolume(ctx, v, t.scannerRegionEc2svc)
 }
 
@@ -187,6 +186,7 @@ func FindRecentSnapshotForVolume(ctx context.Context, v VolumeId, svc *ec2.Clien
 			s := SnapshotId{Account: v.Account, Region: v.Region, Id: *snapshot.SnapshotId}
 			log.Info().Interface("snapshot", s).Msg("found snapshot")
 			snapState := snapshot.State
+			timeout := 0
 			for snapState != types.SnapshotStateCompleted {
 				log.Info().Interface("state", snapState).Msg("waiting for snapshot copy completion; sleeping 10 seconds")
 				time.Sleep(10 * time.Second)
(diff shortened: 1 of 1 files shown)


motor/providers/awsec2ebs/setup.go
```diff
@@ -237,16 +241,33 @@ func CreateSnapshotFromVolume(ctx context.Context, cfg aws.Config, volID string,
...
+			if errors.As(err, &ae) {
+				if ae.ErrorCode() == "InvalidSnapshot.NotFound" {
+					time.Sleep(30 * time.Second) // if it says it doesnt exist, even though we just created it, then it must still be busy creating
+					timeout++                    // still count against the timeout to avoid an eternal loop
+					continue
+				}
+			}
 			return nil, err
 		}
 		snapProgress = *snaps.Snapshots[0].Progress
```
chris-rock, line 270: What happens if snaps does not include a snapshot? While very unlikely, we should check that case and throw an error because we expect exactly one result here.

motor/providers/awsec2ebs/setup.go
```diff
@@ -237,16 +241,36 @@ func CreateSnapshotFromVolume(ctx context.Context, cfg aws.Config, volID string,
...
+	for snapState != types.SnapshotStateCompleted || !strings.Contains(snapProgress, "100") {
 		log.Info().Str("progress", snapProgress).Msg("waiting for snapshot completion; sleeping 10 seconds")
 		time.Sleep(10 * time.Second)
 		snaps, err := ec2svc.DescribeSnapshots(ctx, &ec2.DescribeSnapshotsInput{SnapshotIds: []string{*res.SnapshotId}})
 		if err != nil {
+			var ae smithy.APIError
+			if errors.As(err, &ae) {
+				if ae.ErrorCode() == "InvalidSnapshot.NotFound" {
+					time.Sleep(30 * time.Second) // if it says it doesnt exist, even though we just created it, then it must still be busy creating
+					timeout++                    // still count against the timeout to avoid an eternal loop
```
czunker: In theory, this can still end in an infinite loop because of the continue and the timeout check at the end of the loop. I'm not sure whether this could also happen in reality.

vjeffrey: true, good point, i increase the timeout but dont check it right here -- ill add a quick check just in case

PR #482, opened 2022-11-14
🧹 Remove unused azure functions. Add deprecation TODOs.

preslavgerchev:
Small cleanup + deprecation TODOs.

### DIFF
--- a/motor/providers/azure/client.go
+++ b/motor/providers/azure/client.go
@@ -1,11 +1,6 @@
 package azure
 
 import (
-	"encoding/json"
-	"io"
-	"io/ioutil"
-
-	"github.com/cockroachdb/errors"
 	"go.mondoo.com/cnquery/motor/providers/local"
 )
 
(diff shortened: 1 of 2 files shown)


motor/providers/azure/client.go
```diff
@@ -18,44 +13,3 @@ func IsAzInstalled() bool {
 	command, err := t.RunCommand("az")
 	return command.ExitStatus == 0 && err == nil
 }
-
-// shells out to `az account show --output json` to determine the default account
-// call `az account list` displays all subscriptions
-func GetAccount() (*AzureAccount, error) {
```
chris-rock, line 24: Shouldn't that be used by default if users type `cnquery shell azure`?

preslavgerchev: The azure sdk uses `az account get-access-token -o json` to do so, we only need to check if `az` is present

chris-rock: and the az command also gather the default subscription?

preslavgerchev: We only use the `IsAzInstalled` to make sure that if you run `cnquery shell azure` without any params you can use the AZ cli. In that case it discovers all subs and you can pick:
image

preslavgerchev:
Closing this in favour of #534

PR #490, opened 2022-11-15
Generate unique docusaurus IDs for MQL menus

tas50:
These can't match the child resource IDs or we don't get nice menu titles.


chris-rock (commented):
Thank you @tas50

PR #494, opened 2022-11-18
Add ServiceAccountCredentials to Inventory proto.

preslavgerchev:
Allows us to specify upstream credentials to report data back, need it for server-side scanning.
Reason I am putting this here and not in cnspec is that this will also allow us flexibility if we're only gathering asset data instead of scanning.

See `inventory.proto` for the changes, the rest is protoc version bump.


motor/inventory/v1/inventory.proto
```diff
@@ -129,6 +130,9 @@ message Inventory {
 
   // Most recently observed status of the Inventory.
   InventoryStatus status = 3;
+
+  // optional: the upstream credentials to use for the inventory
+  mondoo.upstream.v1.ServiceAccountCredentials credentials = 4;
```
chris-rock: should this be part of the inventory spec?

preslavgerchev: That's probably a better place indeed

motor/inventory/v1/inventory.proto
```diff
@@ -136,6 +137,8 @@ message InventorySpec {
   map<string, cnquery.motor.vault.v1.Credential> credentials = 2;
   VaultConfiguration vault = 3;
   string credential_query = 4;
+  // optional: the upstream credentials to use for the inventory
+  mondoo.upstream.v1.ServiceAccountCredentials upstream_credentals = 5;
```
chris-rock: I think that should be number greater than 15 https://developers.google.com/protocol-buffers/docs/proto3#assigning_field_numbers

chris-rock (changes requested):
Great addition, LGTM once the number is updated

chris-rock (approved):
Thank you @preslavgerchev

PR #496, opened 2022-11-18
🧹 Check if provided github token is valid

### LINKED ISSUE #495: Not apparent when bad github token is provided
**Describe the bug**
If you scan a GitHub org with a bad token it doesn't fail in a way that shows you the token was bad on the CLI. Instead you have to either do full output or view the asset in the platform to see that each control failed on a 401.

**To Reproduce**
1. cnspec scan github org SOME_ORG --token bogus_token
2. exits 0
3. nothing succeeds

**Expected behavior**
It should be very apparent to the user at the time of the scan that it failed on bad auth.

**Screenshots or CLI Output**
image

image

preslavgerchev:
Fixes #495  by performing a quick call when initializing the client

image

### DIFF
--- a/motor/providers/github/provider.go
+++ b/motor/providers/github/provider.go
@@ -57,7 +57,14 @@ func New(tc *providers.Config) (*Provider, error) {
 	}
 
 	client := github.NewClient(oauthClient)
-
+	// perform a quick call to verify the token's validity.
+	_, resp, err := client.Zen(context.Background())
+	if err != nil {
+		if resp.StatusCode == 401 {
+			return nil, errors.New("invalid github token provided")
+		}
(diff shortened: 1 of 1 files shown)


motor/providers/github/provider.go
```diff
@@ -57,7 +57,14 @@ func New(tc *providers.Config) (*Provider, error) {
 	}
 
 	client := github.NewClient(oauthClient)
-
+	// perform a quick call to verify the token's validity.
+	_, resp, err := client.Zen(context.Background())
```
czunker, line 61: :smile: Nice endpoint!

czunker (approved):

PR #498, opened 2022-11-18
GCP account -> organization

tas50:
Let's make sure we use the right term


chris-rock (approved):
Thank you @tas50

PR #504, opened 2022-11-21
🐛 Only delete volumes we created

### LINKED ISSUE #500: cnspec deletes available EBS volume after failed scan
**Describe the bug**
EBS Volume is deleted after failed scan:
```
./cnspec scan aws ec2 ebs volume vol-04b6ac8dc3bff9fd3
! resolver does not support discovery target 'auto', the following are supported:  resolver="AWS EC2 EBS Resolver"
→ syncing assets
→ connecting to asset vol-04b6ac8dc3bff9fd3
! WARNING: you are using an early access feature feature=AWS_EC2_EBS status=experimental
→ validate exists volume={"AccountId":"***","Id":"vol-04b6ac8dc3bff9fd3","PlatformId":"//platformid.api.mondoo.app/runtime/aws/ec2/v1/accounts/***/regions/us-east-1/volumes/vol-04b6ac8dc3bff9fd3","Region":"us-east-1"}
→ attach volume volume id=vol-04b6ac8dc3bff9fd3
x attach volume err error="operation error EC2: AttachVolume, https response error StatusCode: 400, RequestID: e362c3a3-1856-426f-b293-e2aa6b72d21e, api error InvalidVolume.ZoneMismatch: The volume 'vol-04b6ac8dc3bff9fd3' is not in the same availability zone as instance 'i-0b24443c8d18fdbab'" volume=vol-04b6ac8dc3bff9fd3
x unable to complete setup step error="operation error EC2: AttachVolume, https response error StatusCode: 400, RequestID: e362c3a3-1856-426f-b293-e2aa6b72d21e, api error InvalidVolume.ZoneMismatch: The volume 'vol-04b6ac8dc3bff9fd3' is not in the same availability zone as instance 'i-0b24443c8d18fdbab'"
→ unmount volume
x failed to unmount dir error="no such file or directory"
x failed to unmount dir error="no such file or directory"
x unable to unmount volume error="no such file or directory"
→ detach volume
x unable to detach volume error="operation error EC2: DetachVolume, https response error StatusCode: 400, RequestID: 069a6932-dba4-46fd-9d6b-8375a0b960aa, api error IncorrectState: Volume 'vol-04b6ac8dc3bff9fd3' is in the 'available' state."
→ delete created volume
→ remove created dir
```

**To Reproduce**
1. Create EBS volume
2. Get EBS volume ID
3. Scan EBS volume from EC2 instance with above command
4. See delete message at the end of the output
5. Check EBS volumes

**Expected behavior**
We shouldn't delete the volume.

**Screenshots or CLI Output**
-

- cnspec 7.4.0

**Additional context**
Came across this while trying to reproduce mondoohq/cnspec#198

czunker:
In some cases, e.g., snapshot scans we create volumes which we have to delete afterwards. But when scanning an available volume, we shouldn't delete it afterwards.

Fixes #500

### DIFF
--- a/motor/providers/awsec2ebs/provider.go
+++ b/motor/providers/awsec2ebs/provider.go
@@ -156,6 +156,7 @@ type TargetInfo struct {
 
 type tmpInfo struct {
 	// these fields are referenced during setup/mount and close
+	createdExtraVolume  bool      // true if we created an extra volume to attach to the scanner instance
 	scanVolumeId        *VolumeId // the volume id of the volume we attached to the instance
 	scanDir             string    // the tmp dir we create; serves as the directory we mount the volume to
 	volumeAttachmentLoc string    // where we tell AWS to attach the volume; it doesn't necessarily get attached there, but we have to reference this same location when detaching
@@ -192,9 +193,12 @@ func (p *Provider) Close() {
 	if err != nil {
 		log.Error().Err(err).Msg("unable to detach volume")
 	}
-	err = p.DeleteCreatedVolume(ctx, p.tmpInfo.scanVolumeId)
-	if err != nil {
-		log.Error().Err(err).Msg("unable to delete volume")
+	// only delete the volume if we created it, e.g., if we're scanning a snapshot
+	if p.tmpInfo.createdExtraVolume {
+		err = p.DeleteCreatedVolume(ctx, p.tmpInfo.scanVolumeId)
+		if err != nil {
+			log.Error().Err(err).Msg("unable to delete volume")
+		}
 	}
 	err = p.RemoveCreatedDir()
 	if err != nil {
--- a/motor/providers/awsec2ebs/provider.go
+++ b/motor/providers/awsec2ebs/provider.go
@@ -155,11 +155,9 @@ type TargetInfo struct {
 }
 
 type tmpInfo struct {
-	// these fields are referenced during setup/mount and close
-	createdExtraVolume  bool      // true if we created an extra volume to attach to the scanner instance
-	scanVolumeId        *VolumeId // the volume id of the volume we attached to the instance
-	scanDir             string    // the tmp dir we create; serves as the directory we mount the volume to
-	volumeAttachmentLoc string    // where we tell AWS to attach the volume; it doesn't necessarily get attached there, but we have to reference this same location when detaching
+	scanVolumeInfo      *VolumeInfo // the info of the volume we attached to the instance
+	scanDir             string      // the tmp dir we create; serves as the directory we mount the volume to
+	volumeAttachmentLoc string      // where we tell AWS to attach the volume; it doesn't necessarily get attached there, but we have to reference this same location when detaching
 }
 
 func (p *Provider) RunCommand(command string) (*os.Command, error) {
@@ -180,7 +178,7 @@ func (p *Provider) FS() afero.Fs {
 
 func (p *Provider) Close() {
 	if p.opts != nil {
-		if p.opts[NoSetup] == "true" || p.targetType == EBSTargetSnapshot {
+		if p.opts[NoSetup] == "true" {
 			return
 		}
 	}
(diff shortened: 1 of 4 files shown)


czunker:
This is where it happens:
https://github.com/mondoohq/cnquery/blob/main/motor/providers/awsec2ebs/setup.go#L66

We need this structure to unmount the filesystem later on. We could also add another field instead of the bool, holding the original volume ID and a possible new volume ID.

motor/providers/awsec2ebs/provider.go
```diff
@@ -192,9 +193,12 @@ func (p *Provider) Close() {
 	if err != nil {
 		log.Error().Err(err).Msg("unable to detach volume")
 	}
-	err = p.DeleteCreatedVolume(ctx, p.tmpInfo.scanVolumeId)
-	if err != nil {
-		log.Error().Err(err).Msg("unable to delete volume")
+	// only delete the volume if we created it, e.g., if we're scanning a snapshot
+	if p.tmpInfo.createdExtraVolume {
```
chris-rock: Could we use tags for that? eg. we annotate the volume / snapshot that we created and only delete it if it has our tag. Advantage would be that we can recover from states where the execution was interrupted.

czunker: From what I understand, the API does not support an additional filter for tags on the delete call: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DeleteVolume.html

But perhaps @preslavgerchev or @vjeffrey know how this could work.

preslavgerchev: Pushed up a commit to get tags and delete based on them. A quick test shows it works so far, feel free to adjust!

motor/providers/awsec2ebs/provider.go
```diff
@@ -199,6 +199,8 @@ func (p *Provider) Close() {
 				log.Error().Err(err).Msg("unable to delete volume")
 			}
 		}
+	} else {
```
chris-rock, line 203: Since the default behaviour is now that we do not delete the volume, I would add an info log, once the volume was deleted.

motor/providers/awsec2ebs/types.go
```diff
@@ -53,16 +53,18 @@ type SnapshotId struct {
...
+	Tags        map[string]string
 }
 
 func resourceTags(resourceType types.ResourceType, instanceId string) []types.TagSpecification {
 	return []types.TagSpecification{
-		{ResourceType: resourceType,
+		{
+			ResourceType: resourceType,
 			Tags: []types.Tag{
 				{Key: aws.String("Created By"), Value: aws.String("Mondoo")},
```
chris-rock, line 70: Should we stick to the standard AWS tag "createdBy" https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/aws-tags.html?

chris-rock (changes requested):
I love the new tag bases approach, makes it much more reliable and predictable. I added a few nit picks, once that is in, LGTM

motor/providers/awsec2ebs/types.go
czunker: Changed.

motor/providers/awsec2ebs/provider.go
czunker: Done.

czunker:
Thanks for your help @preslavgerchev

preslavgerchev (approved):

motor/providers/awsec2ebs/types.go
```diff
@@ -53,18 +53,20 @@ type SnapshotId struct {
...
 }
 
 func resourceTags(resourceType types.ResourceType, instanceId string) []types.TagSpecification {
 	return []types.TagSpecification{
-		{ResourceType: resourceType,
+		{
+			ResourceType: resourceType,
 			Tags: []types.Tag{
-				{Key: aws.String("Created By"), Value: aws.String("Mondoo")},
+				{Key: aws.String("createdBy"), Value: aws.String("Mondoo")},
```
vjeffrey, line 69: is there a reason this was changed from "Created By" to "createdBy" ? i believe we have other logic that depends on this tag naming in the lambda

czunker: Was requested here: https://github.com/mondoohq/cnquery/pull/504#discussion_r1031076186

Because of compliance with the standard AWS tags.

Where can we find out whether we need this specific form of the tag?

vjeffrey: https://gitlab.com/mondoolabs/mondoo/-/blob/master/apps/mondoo-lambda/cloudformation/mondoo-lambda-cf.json#L490 we currently depend on the tag being "Created By": "Mondoo"

vjeffrey: can we add both tags for now, and then we'll transition the lambda to the new one? i'll make a ticket for the followup

czunker: Sure. Added it.

vjeffrey (approved):

chris-rock (approved):
Thank you @czunker

PR #505, opened 2022-11-21
Extend aws fingerprint detection

preslavgerchev:
- Extend the `PlatformFingerprint` to also be able to return an asset name, next to the platform ids
 - Update the `aws` detectors to be able to fetch a name for an instance. This can either happen via using the tags local endpoint (if enabled) or using the AWS API if there's a config. If neither's present we fallback to using the instance id as the name.
 - Tests

This will help with properly naming EC2 assets when scanning those via ssh (instance-connect,ssm)

### DIFF
--- a/motor/motorid/awsec2/metadata_cmd.go
+++ b/motor/motorid/awsec2/metadata_cmd.go
@@ -1,11 +1,16 @@
 package awsec2
 
 import (
+	"context"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"strings"
 
+	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	"github.com/cockroachdb/errors"
 	"go.mondoo.com/cnquery/motor/platform"
 	"go.mondoo.com/cnquery/motor/providers/os"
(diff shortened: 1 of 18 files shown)


motor/motorid/awsec2/metadata_cmd.go
```diff
@@ -1,55 +1,91 @@
...
 	// parse into struct
 	doc := imds.InstanceIdentityDocument{}
 	if err := json.NewDecoder(strings.NewReader(instanceDocument)).Decode(&doc); err != nil {
 		return Identity{}, errors.Wrap(err, "failed to decode EC2 instance identity document")
 	}
+
+	name := ""
+	// Note that the tags metadata has to be enabled for this to work. If not, we fallback to trying to get the name
+	// via the aws API (if there's a config provided)
+	taggedName, err := m.instanceTags()
```
czunker: nit: The function name suggests that multiple tags are returned, but it is the name tag only.

motor/motorid/awsec2/metadata_cmd.go
```diff
@@ -1,55 +1,91 @@
...
-func (m *CommandInstanceMetadata) instanceIdentityDocument() (string, error) {
+func curlWindows(url string) string {
+	return fmt.Sprintf("Invoke-RestMethod -TimeoutSec 1 -URI %s -UseBasicParsing | ConvertTo-Json", url)
+}
+
+func (m *CommandInstanceMetadata) curlDocument(url string) (string, error) {
 	switch {
 	case m.platform.IsFamily(platform.FAMILY_UNIX):
-		cmd, err := m.provider.RunCommand("curl " + identityUrl)
+		cmd, err := m.provider.RunCommand("curl " + url)
```
czunker, line 88: Not all Linux distros have curl installed by default.

Perhaps for a follow-up PR: We could replace this with go. That should also work on both platforms.

czunker (approved):
Nice addition and tests.

motor/motorid/awsec2/metadata_cmd.go
chris-rock: This is only the remove option, which is difficult to implement via go. We could only fallback to other curl-like methods.

motor/motorid/awsec2/metadata_local.go
```diff
@@ -20,13 +23,30 @@ type LocalEc2InstanceMetadata struct {
...
 	if err != nil {
 		return Identity{}, err
 	}
+	filters := []ec2types.Filter{
+		{
+			Name:   aws.String("resource-id"),
+			Values: []string{doc.InstanceID},
+		},
+	}
+	tags, err := ec2svc.DescribeTags(ctx, &ec2.DescribeTagsInput{Filters: filters})
```
jaym, line 38: is this information not available from the metadata service? This looks like it talks to aws?

preslavgerchev: The metadata service needs the tag service to be enabled (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#allow-access-to-tags-in-IMDS). That being said, I do notice we never check that and instead rely on `DescribeTags`. I will add this in a follow-up PR that we first check if the metadata tags are present and only then try and fetch them from the aws API

PR #511, opened 2022-11-22
🧹 Query tags metadata service for ec2 instance name if available.

preslavgerchev:
Follow up to https://github.com/mondoohq/cnquery/pull/505#discussion_r1029250391

Always prefer the tags metadata service, if that is available. Fallback to aws API as a last resort

### DIFF
--- a/motor/motorid/awsec2/metadata_local.go
+++ b/motor/motorid/awsec2/metadata_local.go
@@ -2,6 +2,7 @@ package awsec2
 
 import (
 	"context"
+	"io"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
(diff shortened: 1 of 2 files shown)


motor/motorid/awsec2/metadata_local.go
```diff
@@ -29,18 +30,26 @@ func (m *LocalEc2InstanceMetadata) Identify() (Identity, error) {
...
-	filters := []ec2types.Filter{
-		{
-			Name:   aws.String("resource-id"),
-			Values: []string{doc.InstanceID},
-		},
-	}
-	tags, err := ec2svc.DescribeTags(ctx, &ec2.DescribeTagsInput{Filters: filters})
 	name := ""
+	// try and fetch this from the metadata, if the tag metadata service is enabled.
+	nameTag, err := m.getMetadataValue(metadata, "tags/instance/Name")
```
czunker, line 35: In the previous PR, we had `m.instanceNameTag()` in `metadata_cmd.go`. Is it possible to combine them somehow? Now, we have two different approaches to get to the same tag.

preslavgerchev: The other one (the remote) relies on curl as we cannot create a metadata client when executing remotely, that would instantiate the client on our own machine rather than the ec2 one.  I can use curl here too but I prefer this as it's got also better err handling

czunker (approved):

PR #513, opened 2022-11-22
✨Add asset-name flag, allowing overwriting of asset names.

preslavgerchev:
A new `asset-name` flag that allows for overwriting the root asset name. Essentially equal to passing an inventory file with the name of the asset being set.


chris-rock (approved):
Thank you @preslavgerchev

PR #517, opened 2022-11-25
🐛 Fix file stat when run with sudo

### LINKED ISSUE #512: Core Resource: File, --sudo yields no user or groupnames
**Describe the bug**
Using the `--sudo` switch when using `cnquery` leads to unusable output.
`cnquery shell local --sudo`

**To Reproduce**

` cnquery> file($filename).group`

` cnquery> file($filename).name`

See error below:
image

**Expected behavior**
The same as above without the `--sudo` switch:
`cnquery shell local`
image

Tested on:
- Debian 11
- Ubuntu 22.04 
using:
- cnquery 7.5.0 (0d6924f0, 2022-11-22T12:33:11Z)
- cnquery v7.3.0 (7594c926, unknown)

czunker:
Fixes #512

### DIFF
--- a/motor/providers/local/local_unix.go
+++ b/motor/providers/local/local_unix.go
@@ -6,11 +6,19 @@ package local
 import (
 	"os"
 	"syscall"
+
+	pos "go.mondoo.com/cnquery/motor/providers/os"
 )
 
 func (p *Provider) fileowner(stat os.FileInfo) (int64, int64) {
 	uid := int64(-1)
 	gid := int64(-1)
+	if p.Sudo != nil {
+		if stat, ok := stat.Sys().(*pos.FileInfo); ok {
+			uid = stat.Uid
+			gid = stat.Gid
+		}
+	}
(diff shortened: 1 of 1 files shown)


motor/providers/local/local_unix.go
```diff
@@ -6,11 +6,19 @@ package local
...
 	"os"
 	"syscall"
+
+	pos "go.mondoo.com/cnquery/motor/providers/os"
 )
 
 func (p *Provider) fileowner(stat os.FileInfo) (int64, int64) {
 	uid := int64(-1)
 	gid := int64(-1)
+	if p.Sudo != nil {
```
jaym: do we need to check for sudo? Are there other cases where we might have this `pos.FileInfo` type?

jaym (approved):

motor/providers/local/local_unix.go
czunker: You are correct, this also works without the sudo check. Removed it.

PR #530, opened 2022-11-30
✨ Unify msgraph and azure transports.

preslavgerchev:
This allows users to use msgraph.beta resources from the azure shell.

Note that `ms365` resources still require `ms365` transport and this is now separate from `msgraph`.
The next step would be to actually put the providers into a single place, there's already some preparations here (making a `microsoft` module under `providers`).

We could also have a new `microsoft` transport that lets us run azure/msgraph queries. This might be useful for writing policies as it provides a gateway into both the MSGraph and the Azure domains.


resources/packs/ms365/msgraph.go
```diff
@@ -9,17 +9,32 @@ import (
...
-		return nil, errors.New("ms365 resource is not supported on this transport")
+		az, ok := t.(*azure.Provider)
+		if !ok {
+			return nil, errors.New("msgraph resource is not supported on this transport")
+		}
+		cred := az.Credential()
+		if cred == nil {
+			return nil, errors.New("msgraph resource requires credentials")
+		}
+		cfg := &providers.Config{
```
preslavgerchev, line 30: Note: this is a temporary bridge to allow running `msgraph` queries onto the azure provider. Ideally, we don't want to create a new provider for each query here

resources/packs/ms365/msgraph.go
```diff
@@ -9,17 +9,32 @@ import (
...
+	"go.mondoo.com/cnquery/motor/vault"
 	"go.mondoo.com/cnquery/resources"
 	"go.mondoo.com/cnquery/resources/packs/core"
 )
 
-func ms365Provider(t providers.Instance) (*ms365_provider.Provider, error) {
+func msGraphProvider(t providers.Instance) (*ms365_provider.Provider, error) {
 	at, ok := t.(*ms365_provider.Provider)
 	if !ok {
-		return nil, errors.New("ms365 resource is not supported on this transport")
```
chris-rock, line 22: it would read easier if we would use the go type switches here https://go.dev/tour/methods/16

chris-rock (approved):
I really like the iterative approach to tackle the problem in smaller iterations. I added a smaller comment that improves readability. You can also tackle that in another iteration.

PR #531, opened 2022-11-30
🐛 Add socket info to processes

### LINKED ISSUE #524: Core Resource: ports --sudo switch does not work properly
**Describe the bug**
The subresource `process.executable` of  the `ports` resource returns `no data available` if not run from a privileged `root` shell.
The `--sudo` switch should yield the same results as running the command from a privileged shell. 

**To Reproduce**
From a regular shell, using the `--sudo` switch run:

```
cnquery> ports.where(port == 25)
! mql[processes]> could not retrieve process information error="process 79821 does not exist: /proc/79821" pid=79821
! mql[processes]> could not retrieve process information error="process 79822 does not exist: /proc/79822" pid=79822
! mql[processes]> could not retrieve process information error="process 79823 does not exist: /proc/79823" pid=79823
Query encountered errors:
1 error occurred:
    * cannot cast resource to resource type: <nil>
ports.where.list: [
  port port=25 protocol="tcp" address="127.0.0.1" process.executable=no data available
]
```

**Expected behavior**
From a shell with root privileges run:
```
cnquery> ports.where(port == 25)
ports.where.list: [
  port port=25 protocol="tcp" address="127.0.0.1" process.executable="exim4"
]
```

`cnquery 7.5.1 (60bf39b9, 2022-11-24T16:56:00Z)`
- OS: Debian 11
- OS Version:
```
root@bullseye:/home/vagrant# uname -a
Linux bullseye 5.10.0-19-amd64 #1 SMP Debian 5.10.149-2 (2022-10-21) x86_64 GNU/Linux
```

czunker:
Fixes #524


czunker:
This needs more testing:

- [ ] FreeBSD
- [ ] MacOS

As far as I know now, it will not work on MacOS, because `ss` is not supported. On the other side, latest Ubuntu versions no longer ship with `netstat` preinstalled.

czunker:
This PR includes changes for two differrent execution paths:
- `linuxproc_unix.go` is used with `local`
- `unixps.go` is used with `ssh`

resources/packs/core/lsof/lsof.go
```diff
@@ -296,6 +299,7 @@ func parseProcessLines(lines []string) (Process, error) {
 			if err != nil {
 				return p, err
 			}
+		} else if strings.HasPrefix(line, "o") {
```
czunker, line 302: The break was at `f` before. That would skip the line with `d`.

resources/packs/core/port.go
```diff
@@ -175,6 +175,45 @@ func (p *mqlPorts) processesBySocket() (map[int64]Process, error) {
 	if c.Error != nil {
 		err = errors.New("cannot read related processess: " + c.Error.Error())
 	}
+
+	if len(res) == 0 {
```
chris-rock, line 213: When does this fall back kicks in? Should we do an explicit platform check if it is for unix?

czunker: Currently, the function `processesBySocket`  is only called when listing ports on Linux.

The fallback only kicks in, when the cached `_socketsMap` is empty. `_socketsMap` is currently also only used on Linux.

chris-rock: Can we expect lsof on linux? I am only aware of its usage on BSD platforms.

czunker: Ubuntu has it installed by default. Arch and Rocky have it in their repos. I can't remember a system where I had to install it. It's more complicated with `netstat` and `ss`. These vary between distros.

chris-rock: It is not default on linux container, so we should prepare ourselves to handle this in many ways. I like lsof but we should check if lsof is installed before we try to run it.

czunker: We catch err on RunCommand. I assume the error shows that `lsof` isn't present or not in the `PATH`.

For the container case: That's more complicated. Matching pid and socket isn't that easy. Especially in the cases where we explicitly disable procfs and switch to commands. Sounds like an extra PR.

czunker: I changed this to use `find`. This also works with containers:
```
cnquery run --sudo container 9df1d9ea2e4b -c "ports"                                                                                                                                                         
! mql[processes]> could not retrieve process information error="process 4465 does not exist: /proc/4465" pid=4465
! mql[processes]> could not retrieve process information error="process 4471 does not exist: /proc/4471" pid=4471
ports.list: [
  0: port port=80 protocol="tcp" address="0.0.0.0" process.executable="nginx"
  1: port port=80 protocol="tcp" address="::" process.executable="nginx"
```
This is an nginx image running, which is based on debian.

Using `find` would also solve https://github.com/mondoohq/cnquery/issues/549 :
```
cnquery run --sudo ssh -i /home/christian/workspace/mondoo/examples/.vagrant/machines/default/virtualbox/private_key vagrant@127.0.0.1:2222 -c "ports" 
ports.list: [
  0: port port=53 protocol="tcp" address="127.0.0.53" process.executable="/lib/systemd/systemd-resolved"
  1: port port=22 protocol="tcp" address="0.0.0.0" process.executable="sshd:"
  2: port port=22 protocol="tcp" address="10.0.2.15" process.executable="sshd:"
  3: port port=22 protocol="tcp" address="10.0.2.15" process.executable="sshd:"
  4: port port=22 protocol="tcp" address="10.0.2.15" process.executable="sshd:"
  5: port port=53 protocol="udp" address="127.0.0.53" process.executable="/lib/systemd/systemd-resolved"
  6: port port=68 protocol="udp" address="10.0.2.15" process.executable="/lib/systemd/systemd-networkd"
  7: port port=22 protocol="tcp" address="::" process.executable="sshd:"
  8: port port=80 protocol="tcp" address="::" process.executable="/usr/sbin/apache2"
]
```

Manually testing the `find` command on Alpine based images (BusyBox) also works. For cnquery, it would need some more testing after https://github.com/mondoohq/cnquery/issues/570 is fixed.

chris-rock:
fixes https://github.com/mondoohq/cnquery/issues/549

chris-rock (changes requested):
This is a super great addition @czunker If we could add a small test to test the regex it would be perfect.

czunker:
> This is a super great addition @czunker If we could add a small test to test the regex it would be perfect.

Tests are now also present.

chris-rock (approved):
Thank you @czunker Great improvement!

PR #534, opened 2022-12-01
🧹 Combine MS365 and Azure providers into one

preslavgerchev:
The ms365 and azure providers are now gone, there's a new `microsoft` provider - internally it has an `assetType`, letting us determine if the target is ms365 or azure.


resources/packs/azure/azure.go
```diff
@@ -16,8 +16,17 @@ func init() {
...
+func azureTransport(t providers.Instance) (*microsoft_transport.Provider, error) {
+	at, ok := t.(*microsoft_transport.Provider)
+	if !ok || len(at.SubscriptionID()) == 0 {
+		return nil, errors.New("azure resource is not supported on this transport")
+	}
+	return at, nil
+}
+
+// temporary second function to be used only in azuread.* resources. for these, a subscription is not required.
+func msGraphTransport(t providers.Instance) (*microsoft_transport.Provider, error) {
```
preslavgerchev, line 31: Note: This also works with the `azureTransport`, we just get a different error then (sub id cannot be empty). I figured we might want a more explicit error if you do `azurerm.vms` from an ms365 shell

resources/packs/azure/azure_ad.go
```diff
@@ -30,7 +30,7 @@ func (a *mqlAzuread) id() (string, error) {
 }
 
 func (a *mqlAzuread) GetUsers() ([]interface{}, error) {
-	at, err := azuretransport(a.MotorRuntime.Motor.Provider)
+	at, err := msGraphTransport(a.MotorRuntime.Motor.Provider)
```
chris-rock, line 33: That reads much better +1

motor/providers/microsoft/provider.go
```diff
@@ -10,18 +10,26 @@ import (
 	"go.mondoo.com/cnquery/motor/vault"
 )
 
+type microsoftAssetType int32
+
+const (
+	OptionTenantID       = "tenant-id"
```
chris-rock, line 16: I like the new constants. Make it way more reliable.

motor/providers/microsoft/provider.go
```diff
@@ -10,18 +10,26 @@ import (
...
+	azure microsoftAssetType = 1
 )
 
 var (
 	_ providers.Instance           = (*Provider)(nil)
 	_ providers.PlatformIdentifier = (*Provider)(nil)
 )
 
-// New create a new Microsoft 365 provider
+// New create a new Microsoft provider
```
chris-rock: I think we should update the description to include also ways how to call it for azure and ms365. The specific example is also outdated and we should update it

motor/providers/microsoft/roles.go
```diff
@@ -0,0 +1,45 @@
+package microsoft
+
+var DefaultRoles = []string{
```
chris-rock, line 4: I think we w need to be a bit more explicit. Those roles only apply to msgraph

resources/packs/azure/azure.go
chris-rock: Lets clearly mark it as `//TODO:` so that the editor hightlights it

motor/providers/microsoft/platform.go
```diff
@@ -0,0 +1,52 @@
...
+	return p.clientID
+}
+
+func (p *Provider) Credential() *vault.Credential {
+	return p.cred
+}
+
+func (p *Provider) PlatformInfo() (*platform.Platform, error) {
+	switch p.assetType {
+	case azure:
```
chris-rock, line 39: I think this highlights that we want to defined azure ad as a separate asset going forward.

motor/platform/detector/detector.go
```diff
@@ -89,20 +88,8 @@ func (d *Detector) Platform() (*platform.Platform, error) {
...
-			Runtime: providers.RUNTIME_AZ,
-		}, nil
-	case *ms365.Provider:
-		return &platform.Platform{
-			Name:    "microsoft365",
-			Title:   "Microsoft 365",
-			Kind:    providers.Kind_KIND_API,
-			Runtime: providers.RUNTIME_MICROSOFT_GRAPH,
-		}, nil
+	case *microsoft.Provider:
```
chris-rock, line 91: I think we want to move forward and take the same approach for all providers. For os detection, we probably can migrate that into the os provider. This is a follow-up task, nothing to do as part of this PR.

chris-rock (approved):
This is a great improvement and make it much easier for us going foreword. Thank you @preslavgerchev

motor/providers/microsoft/platform.go
preslavgerchev: `azure.ad` is marked as deprecated, as that's the same as `msgraph.beta.* We still need a subscription for any other azure resources though (vms, network interfaces, etc)

motor/providers/microsoft/roles.go
preslavgerchev: I added a comment thats only msgraph stuff. We can rework this in the future, once we start getting the token's permissions as they are, instead of using default roles.

PR #536, opened 2022-12-01
🌈 add index to auto-expanded list items

arlimus:
It's much easier to address individual entries if you have the index as a prefix:

image


chris-rock (approved):
Super cool. This makes it so much better.

PR #541, opened 2022-12-02
🐛 empty arrays are falsy

arlimus:
This is had been burried for a bit: Empty arrays should count as falsy. They are often used in statements where people use the `where` clause instead of `contains` and have a hard time figuring out why the query is not behaving right. This is because "where" semantically in a boolean statement often tends to be thought of as "where xyz contains". Users should still use `contains` in these cases, but this change makes it easier to work around it for a moment. Also linting... tba


chris-rock (approved):
This sounds right to me

jaym (approved):

PR #543, opened 2022-12-02
🐛 aws: use default region if none is specified

vjeffrey:
fixes https://github.com/mondoohq/cnquery/issues/539#issuecomment-1335645776

### DIFF
--- a/motor/providers/aws/provider.go
+++ b/motor/providers/aws/provider.go
@@ -78,6 +78,10 @@ func New(pCfg *providers.Config, opts ...ProviderOption) (*Provider, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "could not load aws configuration")
 	}
+	if cfg.Region == "" {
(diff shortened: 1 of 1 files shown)


motor/providers/aws/provider.go
```diff
@@ -78,6 +78,9 @@ func New(pCfg *providers.Config, opts ...ProviderOption) (*Provider, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "could not load aws configuration")
 	}
+	if cfg.Region == "" {
+		cfg.Region = "us-east-1" // in case the user has no region set, default to us-east-1
```
preslavgerchev, line 83: I would suggest we also log a warning here to show the user we're defaulting to some region. Otherwise it might be confusing for them that suddenly we fetch data from AWS w/o them having set a region either via param or via the aws config

vjeffrey: 👍  good point, ill add that in

preslavgerchev (approved):

PR #544, opened 2022-12-03
Replace Mondoo Client with cnquery

tas50:
We still had a few floating around


czunker (commented):
Nice catches.

resources/packs/core/core.lr
```diff
@@ -1,10 +1,10 @@
 option go_package = "go.mondoo.com/cnquery/resources/packs/core"
 
-// Provide contextual info about Mondoo Client and environment
+// Provide contextual info about cnquery and environment
```
czunker, line 3: This requires a re-generate of the docs files. `make cnquery/generate` will do this for you. Then, also the tests should pass.

tas50: Updated and rebased

czunker (approved):

PR #548, opened 2022-12-05
🐛 prevent crash on dict[string].contains

arlimus:
Originally contains was only created for simple strings. However, we have since started to support the keyword for arrays and maps. The problem is: dicts can be either and the behavior betweenboth is very different.

This PR adds the first level of support for contains on dicts, that works on both arrays and strings. The printing is still shot, but this is the first step to add proper support.

### DIFF
--- a/llx/builtin_map.go
+++ b/llx/builtin_map.go
@@ -117,7 +117,6 @@ func _mapWhereV2(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64, inve
 	err = e.runFunctionBlocks(argsList, fref, func(results []arrayBlockCallResult, errs []error) {
 		resMap := map[string]interface{}{}
 		for i, res := range results {
-
 			if res.isTruthy() == !invert {
 				key := argsList[i][0].Value.(string)
 				resMap[key] = list[key]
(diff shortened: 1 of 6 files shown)


llx/builtin_map.go
```diff
@@ -396,6 +395,60 @@ func dictValuesV2(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64) (*R
...
+	}
+
+	funBlock := e.ctx.code.Block(fref)
+	if len(funBlock.Chunks) == 0 {
+		return BoolFalse, 0, nil
+	}
+
+	// We have to rewrite operations in the block to make them behave correctly
+	// with checking the body of the string, instead of the entire thing.
+	for i := range funBlock.Chunks {
```
jaym: is this transformation not possible to do in the compiler? Doing it during execution might not be safe if this code is called concurrently. It may also be unexpected to callers to have the struct change from underneath them

jaym: or maybe a copy of the struct can be made and changes can be made to that

arlimus: Good point, I think I have to move it. 

Unfortunately it is not safe at compile-time, because the behavior changes based on the contents of the dict - which we don't know until it is executed.

The solution is probably a conditional path or something...

arlimus:
@jaym Huge kudos to your suggestion! I have updated it with a new type that is safe for lookups at runtime, without compromising the compiled code. I think it's worth it to sacrifice a type for string handling here.

If you agree with this approach, I'll follow this PR up with another to implement the other functions for it, making contains more powerful. I'll also likely migrate the string implementation so we can share the code-base. Those are follow-ups if this approach is good.

types/types.go
```diff
@@ -235,19 +247,16 @@ func (typ Type) Label() string {
 		return "EMPTY"
 	}
 
-	if typ[0]&'\xf0' == '\x00' {
```
arlimus, line 238: note: I changed the handling here because it was forcing the "simple types" to remain in the last few remaining slots (low numbers). It didn't really give us a big advantage, so instead we are now looking it up as a simple type first and as a complex one second.  `labelfun` is still needed for the more complex types that need a bit of handling for their label

types/types.go
```diff
@@ -87,6 +88,16 @@ const (
 	ResourceLike = Type(rune(byteResource))
 	// FunctionLike is the underlying type of all functions
 	FunctionLike = Type(rune(byteFunction))
+
+	// StringSlice is used to represent special function for searching strings.
+	// Users are never exposed to this type directly and it is not documented
+	// as a primitive. It serves as a way to array functions on top of strings,
```
jaym, line 94: Not sure this next sentence makes sense "serves as a way to array functions  on top of strings"

jaym (approved):

PR #550, opened 2022-12-05
🐛 Set asset name in ebs scanning if it is provided.

preslavgerchev:
If `asset-name` flag is set, use it in the EBS provider.


vjeffrey (approved):

vjeffrey:
thank you!

PR #557, opened 2022-12-06
Document the possible ec2 instance states

tas50:
This makes it easier to craft a query without reading the aws docs.

From: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html

image

Fixes https://github.com/mondoohq/cnquery/issues/552


misterpantz (approved):
SWEET!

PR #559, opened 2022-12-06
🧹 Capitalize comments that feed the docs

tas50:
Generate better docs

Signed-off-by: Tim Smith <tsmith84@gmail.com>

PR #560, opened 2022-12-06
🧹 use self-hosted runner for code generation test

chris-rock:
after #559 

This change reduced the code-generation time from more than 30 min to 3 min

### DIFF
--- a/.github/workflows/pr-test-generated-files.yaml
+++ b/.github/workflows/pr-test-generated-files.yaml
@@ -15,7 +15,7 @@ env:
 jobs:
   # Check if there is any dirty change for generated files
   generated-files:
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
     steps:
       - name: Install Go
         uses: actions/setup-go@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-test-generated-files.yaml
```diff
@@ -15,20 +15,23 @@ env:
...
+      - name: "Install required tooling"
+        run: |
+          sudo apt install -y gcc
       - name: Check generated files
         # Note we do not use apt install -y protobuf-compiler` since it is too old
         run: |
           PB_REL="https://github.com/protocolbuffers/protobuf/releases"
           curl -LO $PB_REL/download/v${PROTO_VERSION}/protoc-${PROTO_VERSION}-linux-x86_64.zip
-          unzip protoc-${PROTO_VERSION}-linux-x86_64.zip -d $HOME/.local
+          unzip -o protoc-${PROTO_VERSION}-linux-x86_64.zip -d $HOME/.local
```
czunker, line 34: Shouldn't we have a clean environment where we do not need to overwrite something?

chris-rock: Yes that would be preferable. We could move the `.local` in the repo and add `.local` to gitignore. We need to make sure we keep a clean git repo, otherwise diff checks may fail.

czunker (approved):
Nice performance improvement!

.github/workflows/pr-test-generated-files.yaml
czunker: Or perhaps something like this? https://github.com/mondoohq/cnquery/pull/571/files

then we do not have to worry about the repo.

PR #562, opened 2022-12-06
✨ Recursively search directory for k8s manifests

czunker:
When a directory is provided as an argument, we will recursively read it and add the k8s manifest files. Currently, only yaml and yml files will be read. Non-manifest yaml files will be skipped.

### DIFF
--- a/motor/providers/k8s/manifest_provider.go
+++ b/motor/providers/k8s/manifest_provider.go
@@ -4,9 +4,8 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"io"
-	"io/ioutil"
+	"io/fs"
 	"os"
-	"path"
 	"path/filepath"
 
 	"github.com/pkg/errors"
@@ -20,6 +19,8 @@ import (
 	"go.mondoo.com/cnquery/motor/providers/os/fsutil"
 	admissionv1 "k8s.io/api/admission/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
 	"k8s.io/apimachinery/pkg/version"
 )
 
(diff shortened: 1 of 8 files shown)


motor/providers/k8s/manifest_provider.go
```diff
@@ -197,29 +198,50 @@ func loadManifestFile(manifestFile string) ([]byte, error) {
...
-				filename := path.Join(manifestFile, f.Name())
 
 				// only load yaml files for now
-				ext := filepath.Ext(filename)
-				if ext == ".yaml" || ext == ".yml" {
-					log.Debug().Str("file", filename).Msg("add file to manifest loading")
-					filenames = append(filenames, filename)
-				} else {
-					log.Debug().Str("file", filename).Msg("ignore file")
+				if !d.IsDir() {
```
chris-rock, line 208: I think we should add tests for that to make sure it works as expected.

chris-rock (commented):
Great addition @czunker

motor/providers/k8s/manifest_provider.go
```diff
@@ -197,29 +198,50 @@ func loadManifestFile(manifestFile string) ([]byte, error) {
...
+					ext := filepath.Ext(path)
+					if ext != ".yaml" && ext != ".yml" {
+						log.Debug().Str("file", path).Msg("ignore file, no .yaml or .yml ending")
+						return nil
+					}
+					// check whether this is valid k8s yaml
+					f, err := os.Open(path)
```
imilchev: I think you can replace `os.Open` and `io.ReadAll` with `os.ReadFile()`. That should do the same but it's a single line

motor/providers/k8s/manifest_provider.go
```diff
@@ -197,29 +198,50 @@ func loadManifestFile(manifestFile string) ([]byte, error) {
...
+						return nil
+					}
+					defer f.Close()
+
+					content, err := io.ReadAll(f)
+					if err != nil {
+						log.Debug().Str("file", path).Err(err).Msg("ignore file, could not read file")
+						return nil
+					}
+					// At this point, we do not care about specific schemes, just whether the file is a valid k8s yaml
```
imilchev, line 220: Instead of initializing a new serializer for each file, you can initialize it above and re-use it here

motor/providers/k8s/manifest_provider.go
imilchev: I agree. I think a test is needed. Perhaps create a test folder with some manifest, some non-manifest yaml and a readme (or some other random file). Just to make sure we test this bit of code since it is kind of complicated

motor/providers/k8s/manifest_provider.go
czunker: Good catch. Moved it up.

motor/providers/k8s/manifest_provider.go
czunker: Thanks, learned something new.

czunker:
@imilchev / @chris-rock Now, the PR also contains tests for different file types and contents.

imilchev (approved):
LGTM

PR #567, opened 2022-12-06
🧹 Replace msgraph-beta-sdk with msgraph-sdk. Replace msgraph.beta.* resources with msgraph.*

preslavgerchev:
- Replace the beta sdk module with the production one.
 - `msgraph.beta.*` -> `msgraph.*` for all resources
 - Dropped every field that's not part of the production API

Note that the resource pack changes are **breaking**. However, now seems like the good time to get those out before we start using this more often.

TODO:
- [x] Verify if we can use the service clients from the module instead of having our own (for building performance purposes) (another PR)
- [x] Double check dropped fields


resources/packs/ms365/ms365.lr
```diff
@@ -146,46 +144,41 @@ msgraph.beta.rolemanagement.roledefinition {
...
+// Microsoft Graph Device Compliance Policy
+msgraph.devicemanagement.devicecompliancepolicy {
   id string
   createdDateTime time
   description string
   displayName string
   lastModifiedDateTime time
-  roleScopeTagIds []string
   version int
-  properties dict
```
preslavgerchev, line 188: I have removed the `properties` from most of the objects. The MSGraph API does not really have properties-like field as everything is exposed as a separete field there (which makes sense since they use odata).

I think this also is a step in the right direction in general - makes writing policies/queries easier as we have a better idea of what the resource contains just by looking at its definition. Else, one would have to see what the `properties` dict contains and then assert against it.

@chris-rock would love your thoughts on this in general. We have some more objects in the ms365 pack that are pure `dict`s now, we could look into migrating those to more specific resources

resources/packs/ms365/ms365.lr
```diff
@@ -2,28 +2,47 @@ import "../core/core.lr"
...
-  groups() []msgraph.beta.group
-  domains() []msgraph.beta.domain
-  applications() []msgraph.beta.application
-  serviceprincipals() []msgraph.beta.serviceprincipal 
-  settings() []dict
-}
-
-// Microsoft Graph Beta Organization
-msgraph.beta.organization {
+alias msgraph.beta = msgraph
```
chris-rock, line 5: This is a great way to make the resources backwards-compatible

resources/packs/ms365/ms365.lr
chris-rock: Throwing an error is the right thing to do. We do not want to pretend the test worked. For old policies, this gives the users a proper error.

resources/packs/ms365/ms365.lr
```diff
@@ -146,47 +166,46 @@ msgraph.beta.rolemanagement.roledefinition {
...
   createdDateTime time
   description string
   displayName string
   lastModifiedDateTime time
-  roleScopeTagIds []string
   version int
-  properties dict
   assignments []dict
+  // deprecated, kept for backwards compatability.
+  properties dict
```
chris-rock, line 208: we should make all properties a lazy function. This allows us to properly return an error.

chris-rock (approved):
Great improvement @preslavgerchev Glad to see the move to the stable SDK.

PR #571, opened 2022-12-07
🧹 Clean protoc installation for each commit

czunker:

### DIFF
--- a/.github/workflows/pr-test-generated-files.yaml
+++ b/.github/workflows/pr-test-generated-files.yaml
@@ -31,9 +31,10 @@ jobs:
         run: |
           PB_REL="https://github.com/protocolbuffers/protobuf/releases"
           curl -LO $PB_REL/download/v${PROTO_VERSION}/protoc-${PROTO_VERSION}-linux-x86_64.zip
-          unzip -o protoc-${PROTO_VERSION}-linux-x86_64.zip -d $HOME/.local
+          mkdir tools
+          unzip protoc-${PROTO_VERSION}-linux-x86_64.zip -d ./tools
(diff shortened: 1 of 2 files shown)


czunker:
Follow-up to this discussion: https://github.com/mondoohq/cnquery/pull/560/files#r1040727520

.github/workflows/pr-test-generated-files.yaml
```diff
@@ -31,11 +31,16 @@ jobs:
         run: |
           PB_REL="https://github.com/protocolbuffers/protobuf/releases"
           curl -LO $PB_REL/download/v${PROTO_VERSION}/protoc-${PROTO_VERSION}-linux-x86_64.zip
-          unzip -o protoc-${PROTO_VERSION}-linux-x86_64.zip -d $HOME/.local
+          mkdir -p $HOME/$GITHUB_SHA
+          unzip protoc-${PROTO_VERSION}-linux-x86_64.zip -d $HOME/$GITHUB_SHA
           rm protoc-${PROTO_VERSION}-linux-x86_64.zip
-          export PATH="$PATH:$HOME/.local/bin"
+          export PATH="$PATH:$HOME/$GITHUB_SHA/bin"
```
chris-rock: Why do we want to put this into the home directory? Would it not be easier to download it into the current directory eg. `.tools/` and then add this directory to the PATH. That would have the advantage that binaries are cleanup up automatically and we do not need the cleanup task

czunker: Good point. Changed it.

chris-rock (approved):

PR #573, opened 2022-12-08
✨ Recursively search for terraform hcl files

czunker:

### DIFF
--- a/motor/providers/terraform/provider.go
+++ b/motor/providers/terraform/provider.go
@@ -4,9 +4,13 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
-	"github.com/rs/zerolog/log"
+	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
+	"regexp"
+
+	"github.com/rs/zerolog/log"
 
 	"github.com/cockroachdb/errors"
 	"github.com/hashicorp/hcl/v2"
@@ -17,6 +21,8 @@ import (
 var (
 	_ providers.Instance           = (*Provider)(nil)
 	_ providers.PlatformIdentifier = (*Provider)(nil)
+	// e.g. mondoo-operator/.github/terraform/aws/.terraform/modules/vpc/examples/secondary-cidr-blocks/main.tf/1/1
+	MODULE_EXAMPLES = regexp.MustCompile(`^.*/modules/.+/examples/.+`)
 )
 
 type terraformAssetType int32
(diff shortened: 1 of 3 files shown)


czunker:
This works for cnquery:
```
cnquery run terraform ~/workspace/mondoo/github.com/mondoo-operator/.github -c 'terraform.providers' 
terraform.providers: [
  0: terraform.block id = terraform.block//home/christian/workspace/mondoo/github.com/mondoo-operator/.github/terraform/gke/main.tf/31/1
  1: terraform.block id = terraform.block//home/christian/workspace/mondoo/github.com/mondoo-operator/.github/terraform/aws/provider.tf/1/1
  2: terraform.block id = terraform.block//home/christian/workspace/mondoo/github.com/mondoo-operator/.github/terraform/aws/provider.tf/5/1
  3: terraform.block id = terraform.block//home/christian/workspace/mondoo/github.com/mondoo-operator/.github/terraform/aks/versions.tf/35/1
]
```

But it seems to break `cnspec`. Have to do some more testing.

motor/providers/terraform/provider.go
```diff
@@ -79,22 +85,36 @@ func New(tc *providers.Config) (*Provider, error) {
...
-			if err != nil {
-				return nil, errors.Wrap(err, "could not parse tfvars files")
-			}
-
-			modulesManifest, err = ParseTerraformModuleManifest(path)
+			filepath.WalkDir(path, func(path string, d fs.DirEntry, err error) error {
+				if err != nil {
+					return err
+				}
+				foundExamples := MODULE_EXAMPLES.FindString(path)
```
czunker, line 92: I'd like to have opinions on this one. I don't think it makes sense when we scan example file like they are included in modules: `mondoo-operator/.github/terraform/aws/.terraform/modules/vpc/examples/secondary-cidr-blocks/main.tf`

Should we perhaps skip more?

motor/providers/terraform/provider.go
```diff
@@ -79,22 +85,36 @@ func New(tc *providers.Config) (*Provider, error) {
...
+					if err != nil {
+						return errors.Wrap(err, "could not parse hcl file")
+					}
+
+					err = ReadTfVarsFromFile(path, tfVars)
+					if err != nil {
+						return errors.Wrap(err, "could not parse tfvars file")
+					}
+				} else {
+					modulesManifest, err = ParseTerraformModuleManifest(path)
```
czunker, line 109: As I don't know which directory holds the `.terraform` subdirectory, I have to scan each directory for it. But it is no error, when it is not found.

motor/providers/terraform/provider.go
imilchev: So from what I read `examples` is a standard folder in terraform. Would there be any reason why someone would ever want to scan the content of that folder?

motor/providers/terraform/provider.go
```diff
@@ -79,22 +85,36 @@ func New(tc *providers.Config) (*Provider, error) {
...
-			}
-
-			modulesManifest, err = ParseTerraformModuleManifest(path)
+			filepath.WalkDir(path, func(path string, d fs.DirEntry, err error) error {
+				if err != nil {
+					return err
+				}
+				foundExamples := MODULE_EXAMPLES.FindString(path)
+				if foundExamples != "" {
+					log.Debug().Str("path", path).Msg("ignoring terraform module example")
```
imilchev: if we are ignoring the folder I would log a warning message so it is explicitly clear we skipped that folder

motor/providers/terraform/provider.go
czunker: I don't know why someone would scan these. When they use the examples, the code will end up in a place where is relevant and scanned.

motor/providers/terraform/provider.go
czunker: Changed.

imilchev (approved):
The code looks good to me but I don't know too much about TF. I tried it out with a few configs and it seems to work. Can't come up with any edge-cases though

PR #576, opened 2022-12-08
🐛 Parse custom resources from k8s manifests

imilchev:
Make sure we can parse custom resources when scanning k8s manifests. The code will treat every unknown k8s type as a custom resource. I also added a test to check this. To verify it manually you can use the debug target that I added or:

```
cnquery run k8s --path motor/providers/k8s/resources/testdata/cr/tekton.yaml -c "k8s.customresources { name namespace }"
```


scottford-io:
@imilchev Thank you for adding this. I tried a Tekton task manifest and it is now showing up as a `k8s.customresource`, which was not working before, but I am seeing one error...

### Error
```javascript
cnquery> k8s.customresources {*}
Query encountered errors:
1 error occurred:
	* cannot get resource from cache
```

### Tekton Task
```yaml
apiVersion: tekton.dev/v1beta1
kind: Task
metadata:
  name: example-task-name
spec:
  params:
    - name: pathToDockerFile
      type: string
      description: The path to the dockerfile to build
      default: /workspace/workspace/Dockerfile
  resources:
    inputs:
      - name: workspace
        type: git
    outputs:
      - name: builtImage
        type: image
  steps:
    - name: ubuntu-example
      image: ubuntu
      args: ["ubuntu-build-example", "SECRETS-example.md"]
    - image: gcr.io/example-builders/build-example
      command: ["echo"]
      args: ["$(params.pathToDockerFile)"]
    - name: dockerfile-pushexample
      image: gcr.io/example-builders/push-example
      args: ["push", "$(resources.outputs.builtImage.url)"]
      volumeMounts:
        - name: docker-socket-example
          mountPath: /var/run/docker.sock
  volumes:
    - name: example-volume
      emptyDir: {}
```

### cnquery shell full output
```javascript
Query encountered errors:
1 error occurred:
	* cannot get resource from cache
k8s.customresources: [
  0: {
    created: 366 days
    namespace: ""
    uid: ""
    resourceVersion: ""
    id: "task:example-task-name"
    manifest: {
      apiVersion: "tekton.dev/v1beta1"
      kind: "Task"
      metadata: {
        name: "example-task-name"
      }
      spec: {
        params: [
          0: {
            default: "/workspace/workspace/Dockerfile"
            description: "The path to the dockerfile to build"
            name: "pathToDockerFile"
            type: "string"
          }
        ]
        resources: {
          inputs: [
            0: {
              name: "workspace"
              type: "git"
            }
          ]
          outputs: [
            0: {
              name: "builtImage"
              type: "image"
            }
          ]
        }
        steps: [
          0: {
            args: [
              0: "ubuntu-build-example"
              1: "SECRETS-example.md"
            ]
            image: "ubuntu"
            name: "ubuntu-example"
          }
          1: {
            args: [
              0: "$(params.pathToDockerFile)"
            ]
            command: [
              0: "echo"
            ]
            image: "gcr.io/example-builders/build-example"
          }
          2: {
            args: [
              0: "push"
              1: "$(resources.outputs.builtImage.url)"
            ]
            image: "gcr.io/example-builders/push-example"
            name: "dockerfile-pushexample"
            volumeMounts: [
              0: {
                mountPath: "/var/run/docker.sock"
                name: "docker-socket-example"
              }
            ]
          }
        ]
        volumes: [
          0: {
            emptyDir: {}
            name: "example-volume"
          }
        ]
      }
    }
    kind: "Task"
    labels: cannot get resource from cache
    annotations: cannot get resource from cache
    name: "example-task-name"
  }
]
```

motor/providers/k8s/manifest_parser.go
```diff
@@ -437,5 +468,31 @@ func load(manifest []byte) ([]k8sRuntime.Object, error) {
...
+	// Every unstructured object here is an object that we couldn't match to an actual type.
+	// Such objects we treat as custom resources and should end up in the k8s.customresources list.
+	// To do that we need to have a CRD for every kind that couldn't be matched to a type. Here we create
+	// the related CRD for every type that needs it.
+	for _, o := range res {
+		if unstr, ok := o.(*unstructured.Unstructured); ok {
+			gvk := unstr.GetObjectKind().GroupVersionKind()
+			if _, err := resTypes.Lookup(gvk.Kind); err != nil {
+				crdName := strings.ToLower(fmt.Sprintf("%s.%s", gvk.Kind, gvk.Group))
+				res = append(
```
czunker, line 497: I'm not sure here:

You are iterating over `res` and append to `res`. Don't we have the object now twice?

imilchev: No because the code only adds entries to the list if it sees an `Unstructured` object. Since we are not adding `Unstructured` objects, the code won't run twice. However, this made me see that we can add the same CRD twice if, for example, we scan a manifest that contains 2 times the same resource. I fixed that now

czunker (approved):

PR #578, opened 2022-12-09
⭐️ add ecr discovery to the aws resolver

vjeffrey:
find public and private ecr registries and their images, use existing container image connection methods to connect to each asset

Screenshot 2022-12-09 at 09 48 09

### DIFF
--- a/motor/discovery/aws/ecr.go
+++ b/motor/discovery/aws/ecr.go
@@ -8,22 +8,25 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	"github.com/aws/aws-sdk-go-v2/service/ecr"
+	"github.com/aws/aws-sdk-go-v2/service/ecr/types"
+	"github.com/aws/aws-sdk-go-v2/service/ecrpublic"
+	publicecrtypes "github.com/aws/aws-sdk-go-v2/service/ecrpublic/types"
 	"github.com/rs/zerolog/log"
 	"go.mondoo.com/cnquery/motor/asset"
+	"go.mondoo.com/cnquery/motor/motorid/containerid"
 	"go.mondoo.com/cnquery/motor/platform"
 	"go.mondoo.com/cnquery/motor/providers"
 	"go.mondoo.com/cnquery/resources/library/jobpool"
 )
 
-func NewEcrImages(cfg aws.Config) (*EcrImages, error) {
+func NewEcrDiscovery(cfg aws.Config) (*EcrImages, error) {
 	return &EcrImages{config: cfg}, nil
 }
 
 type EcrImages struct {
-	config aws.Config
+	config  aws.Config
+	profile string
 }
-
-var aws_ecr_registry_pattern = "https://%s.dkr.ecr.%s.amazonaws.com"
 
 func (k *EcrImages) Name() string {
 	return "AWS ECR Discover"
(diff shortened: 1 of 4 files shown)


motor/discovery/aws/ecr.go
```diff
@@ -140,6 +127,122 @@ func (a *EcrImages) getRepositories() []*jobpool.Job {
 
 // combine with docker image MondooContainerImageID
 func MondooContainerImageID(id string) string {
-	id = strings.Replace(id, "sha256:", "", -1)
-	return "//platformid.api.mondoo.app/runtime/docker/images/" + id
+	return "//platformid.api.mondoo.app/runtime/ecr/images/" + id
```
chris-rock: I think we should keep the identifier stable. This would help us to uniquely identify the image

vjeffrey:
note: i just noticed the connection only works if creds to the ecr repo are already present; we should amend the code so we can run the login cmd for the user as a followup

motor/discovery/aws/ecr.go
```diff
@@ -77,61 +79,47 @@ func (a *EcrImages) getRepositories() []*jobpool.Job {
...
 			for i := range repoResp.Repositories {
+				assets = append(assets, ecrRepoToAsset(repoResp.Repositories[i], region, a.profile))
 				repoName := repoResp.Repositories[i].RepositoryName
+				repoUrl := *repoResp.Repositories[i].RepositoryUri
 				imageResp, err := svc.DescribeImages(ctx, &ecr.DescribeImagesInput{
 					RepositoryName: repoName,
 				})
 				if err != nil {
-					return imgs, nil
+					return assets, nil
```
czunker: Not sure whether my thought is valid:

When we get multiple repos from AWS and an error occurs on the first repo, we will not try to query the other repos. Perhaps this can happen when limited credentials are provided or something like this. Not sure how much of an edge-case this is. My thoughts also only apply to private repos. Don't whether this could also happen in a different situation to public repos. Perhaps something for v2 of the ECR support?

vjeffrey: yes that's a great point!

motor/discovery/aws/ecr.go
```diff
@@ -140,6 +128,134 @@ func (a *EcrImages) getRepositories() []*jobpool.Job {
 
 // combine with docker image MondooContainerImageID
 func MondooContainerImageID(id string) string {
-	id = strings.Replace(id, "sha256:", "", -1)
```
czunker, line 143: To be concise with the other images, we need this back in.

GCR and docker also include the "sha256" bit. Perhaps you can re-use `containerid.MondooContainerImageID`.

motor/discovery/aws/resolver.go
```diff
@@ -66,7 +67,7 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 	}
 
 	var resolvedRoot *asset.Asset
-	if tc.IncludesOneOfDiscoveryTarget(common.DiscoveryAll, common.DiscoveryAuto, DiscoveryAccounts) {
+	if tc.IncludesOneOfDiscoveryTarget(common.DiscoveryAll, common.DiscoveryAuto, DiscoveryAccounts, DiscoveryECR) {
```
preslavgerchev: Unrelated to this PR but: Should this also have the SSM discovery target? If not, the acc will not be discovered when doing a `--discover ssm` filter

motor/discovery/aws/ecr.go
```diff
@@ -140,6 +128,134 @@ func (a *EcrImages) getRepositories() []*jobpool.Job {
...
+
+type EcrImageInfo struct {
+	digest      string
+	tags        []string
+	repoDigests []string
+	repoUrl     string
+	region      string
+}
+
+func ecrToAsset(a *asset.Asset, i EcrImageInfo, profile string) *asset.Asset {
```
preslavgerchev: Why not combine this and just return an `asset.Asset` by passing in solely an `EcrImageInfo`? Now that asset is being built in two places

motor/discovery/aws/resolver.go
vjeffrey: ya, we should probably always discover the account and just remove this if block?  maybe? idk

czunker (approved):
LGTM. Thanks @vjeffrey

PR #581, opened 2022-12-09
⭐️ gcp big query resources

chris-rock:
This PR adds new resources for GCP Big Query


chris-rock:
will be replaced with a newer version

PR #583, opened 2022-12-09
⭐️ gcp compute resources

chris-rock:
builds on #581 and extends the gcp compute resources


chris-rock:
will be replaced with a newer version

PR #584, opened 2022-12-09
⭐️ k8s ingress as resource (part 1)

joelddiaz:
Just a standard passthrough of Ingress resources as Mondoo assets.

should be able to now run things like:
./cnquery run k8s --discover ingresses -c
'k8s.ingress.spec["ingressClassName"]'

At present, the Ingress resources will not be part of the 'auto' discovery for k8s.


motor/discovery/k8s/resolver.go
```diff
@@ -351,6 +353,17 @@ func addSeparateAssets(
 		resolved = append(resolved, admissionReviews...)
 	}
 
+	// discover ingresses
+	if tc.IncludesOneOfDiscoveryTarget(common.DiscoveryAll, DiscoveryIngresses) {
```
imilchev, line 357: Is it intentional that ingresses are not included in the `auto` discovery? That is the default discovered assets

motor/providers/k8s/manifest_parser.go
```diff
@@ -438,6 +439,47 @@ func (t *manifestParser) DaemonSets(namespace v1.Namespace) ([]*appsv1.DaemonSet
...
 }
 
+func (t *manifestParser) Ingress(namespace, name string) (*networkingv1.Ingress, error) {
+	result, err := t.Resources("ingresses.v1.networking.k8s.io", name, namespace)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(result.Resources) > 1 {
+		return nil, errors.New("multiple ingresses found")
```
imilchev, line 449: I know this is not part of this PR but perhaps we want to improve this error message (also for the other resources). If we show this message to a user, it will be very confusing. Yes, there are multiple ingresses found but that is not the problem. The problem is they have the same name and namespace.,

resources/packs/k8s/k8s.lr
```diff
@@ -540,6 +542,32 @@ k8s.service @defaults("namespace name created") {
...
+  // Kubernetes Object Namesapce
+  namespace string
+  // Kubernetes Object Type
+  kind string
+  // Kubernetes Object Creation Timestamp
+  created time
+  // Full resource manifest
+  manifest dict
+  // Ingress Spec
+  spec dict
```
imilchev: don't we want a specific type for the ingress spec? Or at least take out the ingress props in the ingress object? The main thing we want to check is in the spec and right now it's just a dict. I guess if we want to take out those props we should probably do it already in part 1 since otherwise we will need to re-write any policies that have been written in-between.

motor/discovery/k8s/resolver.go
joelddiaz: Yes. In order to avoid the assets being unscored, it is not part of auto for now. When we have policies written to execute against Ingress resources, then we can make Ingress part of "auto".

tas50: @imilchev My hope is that next week we'd get at least one policy written so we could turn this on by default.

resources/packs/k8s/k8s.lr
joelddiaz: It isn't clear to me right now which fields should be elevated outside of the .spec fields. There is definite upcoming work to create the new 'certificates' field that will be generated based out of the Secrets referenced in .spec.tls[], but that is in a separate PR #607

imilchev: I think we can have strong types for path blocks. Now if I want to query my endpoints I need to use JSON and figure my way through it. Instead we can have an actual type that lets us query the endpoints we want and have autocompletion

motor/providers/k8s/manifest_parser.go
czunker: Can this even happen? I'd assume k8s would prevent two resources of the same Kind with the same name in the same namespace.

imilchev: This is for a manfiest, so technically there's nothing stopping you from having the same resource twice in a single file. To make it more complicated, now that we have recursive directory scanning you can end up in this situation even without having the same resource in a single file

czunker: Ok, makes sense to me. But this could also be valid. I opened an issue for that: https://github.com/mondoohq/cnquery/issues/616

But, I think, this needs some more discussion.

resources/packs/k8s/k8s.lr
```diff
@@ -540,6 +542,85 @@ k8s.service @defaults("namespace name created") {
...
+  // Kubernetes Object Namesapce
+  namespace string
+  // Kubernetes Object Type
+  kind string
+  // Kubernetes Object Creation Timestamp
+  created time
+  // Full resource manifest
+  manifest dict
+  // Ingress rules
+  //rules() []k8s.ingressrule
```
imilchev: nit: i guess this can be removed

imilchev:
Apart from that, I tested it and everything seems to be working. Nice addition 🎉

imilchev (approved):

PR #589, opened 2022-12-09
Update no config file error message

tas50:
This is a super confusing error message that makes it unclear what config file is not being found. Is it K8s? Is it AWS? Is it Mondoo? Who knows.

Current message:

image


vjeffrey (approved):
+1 on the changes, i think we should follow up to remove that msg entirely when accessing the shell - i can't think of a reason you would need a mondoo configuration for shell..

tas50:
That would be great. Let me get a bug for that one

PR #590, opened 2022-12-09
⭐ try using ssm or ec2 instance connect to connect to the discovered instances

vjeffrey:
use ssm and instance-connect to try to connect to ec2 instances discovered on cli

Screenshot 2022-12-09 at 15 09 31


motor/discovery/aws/ec2_instances.go
```diff
@@ -146,33 +158,53 @@ func (ec2i *Ec2Instances) List() ([]*asset.Asset, error) {
...
 			Kind:    providers.Kind_KIND_VIRTUAL_MACHINE,
 			Runtime: providers.RUNTIME_AWS_EC2,
 		},
-		State: mapEc2InstanceStateCode(instance.State),
+		State: mapEc2InstanceStateCode(instanceInfo.instance.State),
 	}
 
 	// if there is a public ip, we assume ssh is an option
-	if instance.PublicIpAddress != nil {
+	if instanceInfo.instance.PublicIpAddress != nil {
```
chris-rock, line 186: Great call. Could we further improve it by checking via the API if the ssm agent is installed and or the permissions are set?

motor/discovery/aws/ec2_instances.go
```diff
@@ -146,33 +158,53 @@ func (ec2i *Ec2Instances) List() ([]*asset.Asset, error) {
 	return instances, nil
 }
 
-func instanceToAsset(account string, region string, instance types.Instance, insecure bool, passInLabels map[string]string) *asset.Asset {
+type instanceInfo struct {
```
chris-rock, line 161: +1

motor/discovery/aws/ec2_instances.go
```diff
@@ -183,11 +215,11 @@ func instanceToAsset(account string, region string, instance types.Instance, ins
 		}
 	}
 	// add passed in labels
-	for k, v := range passInLabels {
+	for k, v := range instanceInfo.passInLabels {
 		asset.Labels[k] = v
 	}
 	// add AWS metadata labels
-	asset.Labels = addAWSMetadataLabels(asset.Labels, ec2InstanceToBasicInstanceInfo(instance, region, account))
+	asset.Labels = addAWSMetadataLabels(asset.Labels, ec2InstanceToBasicInstanceInfo(instanceInfo.instance, instanceInfo.region, instanceInfo.account))
```
chris-rock: should we just accept the in `instanceInfo` as input for `ec2InstanceToBasicInstanceInfo`?

motor/discovery/aws/ssm_instances.go
```diff
@@ -202,9 +231,19 @@ func ssmInstanceToAsset(account string, region string, instance types.InstanceIn
...
-	asset.Labels = addAWSMetadataLabels(asset.Labels, ssmInstanceToBasicInstanceInfo(instance, region, account))
+	asset.Labels = addAWSMetadataLabels(asset.Labels, ssmInstanceToBasicInstanceInfo(instanceInfo.instance, instanceInfo.region, instanceInfo.account))
 	if label, ok := asset.Labels[ImportedFromAWSTagKeyPrefix+AWSNameLabel]; ok {
 		asset.Name = label
 	}
 	return asset
 }
+
+func getProbableUsernameFromSSMPlatformName(name string) string {
+	if strings.HasPrefix(name, "RHEL") {
```
chris-rock: Are we sure this works for all rhel? What happens if I spin up RHEL instead of CentOS?

motor/discovery/aws/ec2_instances.go
vjeffrey: we get the ssm info during the ssm part of the discovery and then merge them

not sure about checking if the permissions are set, i'll have to look into that 🤔

motor/discovery/aws/ssm_instances.go
vjeffrey: it probably does not, this was just a quick starter list. i'll spin a few machines up and see what i get

vjeffrey:
note to self: we need to update 
```
func (c *generator) GenerateCredentials(target string, user string) (*InstanceCredentials, error) {
	ctx := context.Background()
..........
```

to see if we can use the ssm api when it's a purely managed instance (instance id is mi-), because ec2 describe instances won't find those instances

motor/discovery/aws/ssm_instances.go
vjeffrey: reel gives me ec2-user, centos gives centos. i updated to just check for centos or ubuntu, and it it's neither one, just use ec2-user

motor/providers/ssh/session.go
```diff
@@ -206,7 +207,16 @@ func prepareConnection(pCfg *providers.Config) ([]ssh.AuthMethod, []io.Closer, e
...
+			log.Debug().Str("profile", pCfg.Options["profile"]).Str("region", pCfg.Options["region"]).Msg("using aws creds")
+
+			loadOpts := []func(*config.LoadOptions) error{}
+			if pCfg.Options != nil && pCfg.Options["region"] != "" {
+				loadOpts = append(loadOpts, config.WithRegion(pCfg.Options["region"]))
+			}
+			if pCfg.Options != nil && pCfg.Options["profile"] != "" {
+				loadOpts = append(loadOpts, config.WithSharedConfigProfile(pCfg.Options["profile"]))
+			}
+			cfg, err := config.LoadDefaultConfig(context.Background(), loadOpts...)
```
preslavgerchev, line 219: Nice!

motor/discovery/aws/ssm_instances.go
```diff
@@ -202,9 +231,19 @@ func ssmInstanceToAsset(account string, region string, instance types.InstanceIn
...
 	// add AWS metadata labels
-	asset.Labels = addAWSMetadataLabels(asset.Labels, ssmInstanceToBasicInstanceInfo(instance, region, account))
+	asset.Labels = addAWSMetadataLabels(asset.Labels, ssmInstanceToBasicInstanceInfo(instanceInfo))
 	if label, ok := asset.Labels[ImportedFromAWSTagKeyPrefix+AWSNameLabel]; ok {
 		asset.Name = label
 	}
 	return asset
 }
+
+func getProbableUsernameFromSSMPlatformName(name string) string {
```
preslavgerchev, line 241: Nit: Can this be merged with the ec2-specific function? 
I don't think the name would differ if SSM agent is enabled

preslavgerchev (approved):

motor/discovery/aws/ssm_instances.go
vjeffrey: it's more that we're looking at different things

for the ec2 instances we look at the image name
for ssm it gives us a platform name

PR #592, opened 2022-12-09
Update AWS docs comments

tas50:
Make these match the names that Amazon uses better (including capitalization)


misterpantz (approved):
nice!

PR #597, opened 2022-12-10
🧹 rename scope `gcloud` for gcp resources to `gcp`

chris-rock:
This is a non-breaking change since we added aliases for all existing resources.


imilchev (approved):
That's a nice trick 😮

PR #605, opened 2022-12-12
🐛 properly wait until progress bar is closed

chris-rock:
The terminal cursor was not properly recovered because we have not waited until the go routine is properly completed

### DIFF
--- a/cli/progress/progress.go
+++ b/cli/progress/progress.go
@@ -94,7 +94,10 @@ func (p *progressbar) Open() error {
 				time.Sleep(time.Second / progressPipedFps)
 				o.ClearLines(2)
 				o.WriteString(p.View())
-				if p.Data.complete {
+				p.lock.Lock()
+				complete := p.Data.complete
+				p.lock.Unlock()
+				if complete {
 					break
 				}
 			}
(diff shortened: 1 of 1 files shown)


cli/progress/progress.go
```diff
@@ -106,7 +111,10 @@ func (p *progressbar) OnProgress(current int, total int) {
 }
 
 func (p *progressbar) Close() {
+	p.lock.Lock()
 	p.Data.complete = true
```
jaym, line 118: if Close and Update are in different threads, you need to take a lock before writing and reading.

jaym (approved):

cli/progress/progress.go
chris-rock: Anything for me todo or is that as it should be?

jaym: Sorry, got the function name wrong. The uses of `p.Data.complete` in `Open`  need to take a lock before looking a `p.Data.complete`

chris-rock: great catch @jaym

chris-rock:
@jaym I updated the PR, can you double-check that I did this right?

jaym:
lgtm

PR #606, opened 2022-12-12
Implement msgraph.settings.

preslavgerchev:
It's now under `/groupSettings`, rather than just `/settings` for the production API.


chris-rock (approved):
Thank you @preslavgerchev

PR #607, opened 2022-12-12
⭐️ k8s-ingress with certificates

joelddiaz:
k8s-ingress part2

Add new 'certificates' field to Ingress resources. This will be
populated when an Ingress has .spec.TLS entries pointing to Secrets
containing TLS certificates.

A properly filled out Ingress should now allow these kinds of queries:

```
$ ./cnquery run k8s --discover ingresses -c
'k8s.ingress.certificates{ expiresIn }'
→ use cluster name from kube config cluster-name=minikube
k8s.ingress.certificates: [
  0: {
    expiresIn: 12 days 2 hours 12 minutes 14 seconds
  }
]
```


joelddiaz:
This PR builds on top of the work found in #584

resources/packs/k8s/ingress.go
```diff
@@ -63,3 +70,63 @@ func (k *mqlK8sIngress) GetAnnotations() (interface{}, error) {
...
+					tlsSecretFound = true
+
+					certs, err := secret.GetCertificates()
+					if err != nil {
+						return nil, errors.New("error getting certificate data from Secret")
+					}
+					if certs == nil {
+						// no TLS data in Secret referenced
+						// Put a blank certificate in the list of certificates?
+						return nil, fmt.Errorf("no certificates in Secret referenced by Ingress")
```
joelddiaz: It isn't immediately obvious to me what the right behavior should be here in the even an Ingress we are trying to scan points to an improperly-formatted Secret.

resources/packs/k8s/ingress.go
```diff
@@ -63,3 +70,63 @@ func (k *mqlK8sIngress) GetAnnotations() (interface{}, error) {
...
+					if !ok {
+						return nil, fmt.Errorf("expected a list of Certificates")
+					}
+					certificates = append(certificates, certList...)
+				}
+			}
+			if !tlsSecretFound {
+				// Ingress references a Secret that was not found
+				// Put a blank certificate in the list of certificates?
+			}
```
joelddiaz: Same here. What should we do in the event the Ingress references a Secret that simply does not exist?

resources/packs/k8s/ingress.go
```diff
@@ -225,3 +232,63 @@ func (k *mqlK8sIngressservicebackend) id() (string, error) {
...
+	certificates := []interface{}{}
+	if len(ingress.Spec.TLS) > 0 {
+		// This returns ALL Secrets found in the cluster!
+		secretsInterface, err := getSecrets()
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch Secrets referenced in Ingress: %s", err)
+		}
+
+		secrets := []*mqlK8sSecret{}
+		for _, secInterface := range secretsInterface {
```
imilchev, line 251: I think you can optimize this a bit:
- You can only need/use secrets that are in the same namespace as your ingress. This means that you can already discard all secrets that are in a different namespace.
- Instead of adding the `mqlK8sSecret` structs to a slice, you can add them to a map with the key of the map being the name (there won't be duplicate names if you keep secrets from 1 ns as mentioned in the first point)

resources/packs/k8s/ingress.go
```diff
@@ -225,3 +232,63 @@ func (k *mqlK8sIngressservicebackend) id() (string, error) {
...
+			}
+
+			secrets = append(secrets, secret)
+		}
+
+		for _, tls := range ingress.Spec.TLS {
+
+			tlsSecretFound := false
+
+			for _, secret := range secrets {
```
imilchev: If you use a map, then this loop can go away

resources/packs/k8s/ingress.go
imilchev: I guess there will be no cert in that case. Shouldn't be an error though, since k8s doesn't throw an error in such a case either.

resources/packs/k8s/k8s.lr
```diff
@@ -622,6 +622,8 @@ k8s.ingress @defaults("namespace name created") {
   manifest dict
   // Ingress rules
   rules []k8s.ingressrule
+  // Certificates referenced by Ingress
+  certificates []core.certificate
```
imilchev: I think we might need a bit more than just a list of certificates here. If I look at the `Ingress` documentation you specify a `tls` section that can have a list of hosts and for each list of hosts you have certificates.

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: tls-example-ingress
spec:
  tls:
  - hosts:
      - https-example.foo.com
    secretName: testsecret-tls
...
```
I can imagine that if you are writing a control for an ingress you might want to locate a specific certificate for a specific set of hosts, which isn't possible with the current resource

resources/packs/k8s/ingress.go
imilchev: I think we should do nothing. This code will run for manifests scanning. In that case I think in 99% of the time the secret won't be available in the manifest. We shouldn't crash/error

joelddiaz: :+1: A good point about the manifest scanning that I hadn't considered.

resources/packs/k8s/k8s.lr
joelddiaz: Good catch. No clue what I was thinking by just dumping all the certificates found in each referenced Secret into a single list of []certificate. I've added a new structure to more closely capture the way the data is structured in the Ingress resource.

imilchev:
Current changes look good to me. Could you add some tests for the secrets parsing?

imilchev:
The tests look good. Perhaps good to also add 1 ingress which references a secret that doesn't exist in the manifest. In this way we also check that we don't crash/panic in these cases. I think that's important since in some time we might forget about this requirement and might break it unintentionally

joelddiaz:
> 

Yes, this PR adds an example TLS certificate with a locally generated self-signed CA including all the pieces needed for a TLS-type of Secret (including the private key). I tried putting fake data in the private key fields, but it later fails Secret parsing during test case execution.

imilchev (approved):
LGTM!

PR #612, opened 2022-12-12
🧹 make lr compiler more robust and not silently ignore errors

chris-rock:
This errors out when an issue was detected. This is helpful to ensure we do not accidentally miss errors as happened for gcp, see #611 

needs #611 merged first to get it green


arlimus (approved):
noice!

PR #617, opened 2022-12-13
🐛 Fix microsoft resource identifier.

preslavgerchev:
Missed during the renaming


chris-rock (approved):
Thank you @preslavgerchev

PR #621, opened 2022-12-14
Capitalize Azure resource titles for docs

tas50:
These were updated as part of the rename, but we need them to be upper case


resources/packs/azure/azure.lr
```diff
@@ -108,13 +108,13 @@ azuread.serviceprincipal {
   id string
 }
 
-// Azure Resource Manager
+// Azure Resource
```
czunker: This description is now the same as L117, but they are different resources.

Perhaps just add an "s" here to distinguish them?

tas50: Fixed

czunker (approved):

PR #622, opened 2022-12-14
🐛 Re-generate azure manifest, add cmd for it.

preslavgerchev:
The azure yaml definitions were not being updated, there  was a missing cmd for it. We should optimise this further to make sure the definitions get updated properly when building new resources ( I will open an issue)


tas50:
Docs are now properly generating. Thanks @preslavgerchev

vjeffrey (approved):

PR #623, opened 2022-12-14
🧹 Reusable k8s container resolution

imilchev:
I needed to perform some assertions on the assets that are being created from a scan. I did that in https://github.com/mondoohq/mondoo-operator/pull/727 For that reason I had to expose some functions from cnquery such that I don't need to copy-paste some complicated code. These are the changes needed for the mondoo-operator tests.

### DIFF
--- a/motor/discovery/k8s/container_utils.go
+++ b/motor/discovery/k8s/container_utils.go
@@ -3,30 +3,31 @@ package k8s
 import (
 	"strings"
 
+	"go.mondoo.com/cnquery/types"
 	v1 "k8s.io/api/core/v1"
 )
 
 const DockerPullablePrefix = "docker-pullable://"
 
-type containerImage struct {
+type ContainerImage struct {
 	image         string
 	resolvedImage string
 	pullSecrets   []v1.Secret
 }
 
-func ResolveUniqueContainerImages(cs []v1.Container, ps []v1.Secret) map[string]containerImage {
-	imagesSet := make(map[string]containerImage)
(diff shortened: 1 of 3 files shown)


czunker:
This somehow feels strange to me.

From what I understand, we want to test for the images that get scanned and reported upstream by the operator. The image scanning is done by cnspec which uses cnquery. But now, we use cnquery to verify the data, that was generated by cnquery. Shouldn't we have a static list of the images? Is this too complicated to maintain? Do I miss something?

imilchev:
> This somehow feels strange to me.
> 
> From what I understand, we want to test for the images that get scanned and reported upstream by the operator. The image scanning is done by cnspec which uses cnquery. But now, we use cnquery to verify the data, that was generated by cnquery. Shouldn't we have a static list of the images? Is this too complicated to maintain? Do I miss something?

Our integration tests are running on multiple versions of k8s which have different container versions. We cannot just hardcode the containers we expect. The current tests are verifying the assets by their names and the name generation for containers is not so simple. I don't want to copy 200 lines of code to the operator just to ensure the right names exist upstream. If that code changes, then we need to find it in the operator and change it too. With this setup it can be as simple as updating a module. 

I don't think this is that weird. There is a specific way for determining asset names and we want to make sure those names are all sent upstream. I don't see a different way of performing the same check. There is also no way to import cnspec in the operator to re-use that code from there because the code actually originates from cnquery.

P.S.: There is also logic that filters out duplicate containers to make sure we don't send multiple assets of the same container upstream. That is also not the most straight-forward code. If we don't want to re-use that, then we need to copy-paste it. We cannot hardcode the containers since the integration tests at the moment can run on any type of cluster and any type of k8s version.

PR #625, opened 2022-12-14
🧹 do not try to use ssm session for `mi-` managed instances

vjeffrey:
follow up to https://github.com/mondoohq/cnquery/pull/590


preslavgerchev:
lgtm! is the fallback here ssh agent since no creds type is specified?

preslavgerchev (approved):

PR #630, opened 2022-12-15
⭐️ ecs discovery: containers and container instances

vjeffrey:
after https://github.com/mondoohq/cnquery/pull/578

Screenshot 2022-12-14 at 22 08 13


motor/discovery/aws/ecs_containers.go
```diff
@@ -0,0 +1,321 @@
...
+	if c.RuntimeId == nil {
+		return nil
+	}
+
+	asset := &asset.Asset{
+		Name:        *c.Name,
+		PlatformIds: []string{containerid.MondooContainerID(*c.RuntimeId)},
+		Platform: &platform.Platform{
+			Kind:    providers.Kind_KIND_CONTAINER,
+			Runtime: providers.RUNTIME_DOCKER_CONTAINER,
```
czunker: Oh, this one can get tricky.

From what I understood, the runtime (from AWS perspective) can be Fargate or EC2. In the EC2 case, it could be Docker. Not sure whether we can detect this. Perhaps we use an ECS runtime type or a generic CONTAINER runtime type? Or does the AWS API provide the information about the runtime?

motor/discovery/aws/ecs_containers.go
```diff
@@ -0,0 +1,321 @@
...
+		if stringx.Contains(containerAttachmentIds, *a.Id) {
+			for _, a := range task.Attachments {
+				for _, detail := range a.Details {
+					if *detail.Name == "networkInterfaceId" {
+						publicIp = getPublicIpForContainer(ctx, clonedConfig, *detail.Value)
+					}
+
+					// add connections here
+					asset.Connections = append(asset.Connections, &providers.Config{
+						Backend: providers.ProviderType_SSH, // ???
```
czunker: Most containers will not have sshd installed. When Fargate is used, the container runtime is also not accessible to us.

I've no clue how we could access a running container.
Perhaps in the EC" case, a user could ssh into the EC2 instance and connect to the container via the container runtime. Perhaps...

motor/discovery/aws/ecs_containers.go
```diff
@@ -0,0 +1,321 @@
...
+		log.Debug().Msg("using ssm discovery for container instance")
+		ec2i.FilterOptions = Ec2InstancesFilters{InstanceIds: []string{*ci.Ec2InstanceId}, Regions: []string{region}}
+		assets, err := ec2i.List()
+		if err == nil {
+			if len(assets) == 1 {
+				return assets[0]
+			}
+		}
+	}
+	asset := &asset.Asset{
```
czunker: What if ECS is running on Fargate?

motor/discovery/aws/resolver.go
```diff
@@ -148,6 +149,11 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 		}
 	}
 
+	// add all the detected ssm instanced and ec2 instances to the list
```
czunker, line 152: nit: "instances" instead of "instanced"

czunker (commented):
LGTM.

I'm only missing Fargate. Perhaps this is intentionally left out?

motor/discovery/aws/ecs_containers.go
IppX: You could try to use ECS Exec, but it requires additional permissions (https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-exec.html).

One caveat is that not all containers are running. They can be stopped in a running task and you can have scheduled tasks as well.

motor/discovery/aws/ecs_containers.go
```diff
@@ -0,0 +1,321 @@
...
+		}
+	}
+	asset.Labels[common.IPLabel] = publicIp
+	if publicIp != "" {
+		asset.Name = *c.Name + "-" + publicIp
+	}
+	if c.Image != nil {
+		asset.Labels["image"] = *c.Image
+	}
+	for j := range task.Tags {
```
IppX: That's nice and I think think the cluster name as well as maybe the task definition would be helpful to identify the origin of the asset easier.

motor/discovery/aws/ecs_containers.go
```diff
@@ -0,0 +1,321 @@
...
+			resp, err := svc.ListClusters(ctx, &ecsservice.ListClustersInput{})
+			if err != nil {
+				return nil, errors.Wrap(err, "could not gather ecs cluster information")
+			}
+			for _, cluster := range resp.ClusterArns {
+				containerInstances, err := svc.ListContainerInstances(ctx, &ecsservice.ListContainerInstancesInput{Cluster: &cluster, Status: types.ContainerInstanceStatusActive})
+				if err != nil {
+					return nil, errors.Wrap(err, "could not gather ecs container instances")
+				}
+				if len(containerInstances.ContainerInstanceArns) > 0 {
```
IppX: I'm not sure if you should discover those as well.
As part of the AWS integration those ec2 instances will already be discovered and scanned. Whether they're part of an ECS cluster or not is not so interesting and will probably already be known from the extracted aws tags.
It also raises the question of how to reconcile the ec2 asset and this one, since it's effectively the same.

IppX (commented):
Looks good! We went for a similar approach and algorithm in a k8s cronjob:
Get all clusters in configured account/region -> Get all containers images in running tasks -> run `mondoo scan container` and add some tags.

I didn't exactly understand the part where you need access to the running container. Scanning the image seemed enough for us.

motor/discovery/aws/ecs_containers.go
vjeffrey: i'll look into what we get from the api so we can more appropriately label this as fargate vs ecs. i also like your idea of a ECS runtime, ill add that in there as well

motor/discovery/aws/ecs_containers.go
vjeffrey: thank you! 

ya i for sure don't expect us to be able to reach the containers right away, this is more discovery to expose the public ip. 

that ecs-exec thing is interesting, im definitely gonna give it a try/play around and see what we can do!

motor/discovery/aws/ecs_containers.go
vjeffrey: you're right, they will be discovered

if we detect it's an ec2 instance we use the ec2 instance resolver to add it in, so reconciling won't be a problem

my intention here was so that we can build out a nice map of infrastructure that is being used within ecs in aws. i was thinking future world where we have an aws group that has instances and images and also an ecs group inside of it that holds a reference to the container instance and the containers.... 🤷 @scottford-io would love to hear your opinion on this one

motor/discovery/aws/ecs_containers.go
vjeffrey: 👍

motor/discovery/aws/ecs_containers.go
vjeffrey: you mean re: the id?

if things are running in fargate the ci.ec2instanceid should be a reference to some external thing that's running, but that's pretty much all we have i think. i was just using the same mondooinstanceid func until we know more 🤷

czunker: Yes, sorry. I meant the EC2 instance Id. I've no idea what this will look like, when running on fargate or whether AWS will expose this.

vjeffrey: i added the comment from the aws docs at the top of the function, looks like this should in theory always be something that can work with the mondoo instance id func

vjeffrey:
this is ready for re-review, i'm going to try the ecs-exec stuff and follow up in another pr with those changes

motor/discovery/aws/ecs_containers.go
scottford-io: @vjeffrey I agree that it is a powerful story to be able to build out a map of the infrastructure as suggested. I believe anything we can do to get users of cnquery more visibility into their infrastructure automatically is something we should always strive for. 

Something to keep in mind is that the person executing cnquery may not be the person who provisioned the infrastructure. I love the idea of this user being able to point cnquery at an AWS account to quickly get a map of what is deployed.

scottford-io:
@vjeffrey can you provide an example of the commands you are running with this change? I am not clear.

vjeffrey:
@scottford-io you mean the cnquery command for this ecs discovery?

`cnquery shell aws --discover ecs`

motor/discovery/aws/ecs.go
```diff
@@ -0,0 +1,335 @@
...
+
+			resp, err := svc.ListClusters(ctx, &ecsservice.ListClustersInput{})
+			if err != nil {
+				return nil, errors.Wrap(err, "could not gather ecs cluster information")
+			}
+			for _, cluster := range resp.ClusterArns {
+				clusterName := ""
+				clusterDescription, err := svc.DescribeClusters(ctx, &ecsservice.DescribeClustersInput{Clusters: []string{cluster}})
+				if err != nil {
+					return nil, errors.Wrap(err, "could not gather ecs cluster information")
```
czunker: Missed this one before:

When AWS returns multiple clusters, but we get an error on the first one, we would skip all the other clusters.
Perhaps, just log and continue?

motor/discovery/aws/ecs.go
```diff
@@ -0,0 +1,335 @@
...
+				}
+				if len(clusterDescription.Clusters) == 1 {
```
czunker, line 110: Can this be anything other than 1? What should happen in this case?

motor/discovery/aws/ecs.go
```diff
@@ -0,0 +1,335 @@
...
+					clusterName = *clusterDescription.Clusters[0].ClusterName
+				}
+				containerInstances, err := svc.ListContainerInstances(ctx, &ecsservice.ListContainerInstancesInput{Cluster: &cluster, Status: types.ContainerInstanceStatusActive})
+				if err != nil {
+					return nil, errors.Wrap(err, "could not list ecs container instances")
```
czunker, line 118: Same as above: Continue and log to check the other clusters?

motor/discovery/aws/ecs.go
```diff
@@ -0,0 +1,335 @@
...
+				}
+				if len(containerInstances.ContainerInstanceArns) > 0 {
+					containerInstancesDetail, err := svc.DescribeContainerInstances(ctx, &ecsservice.DescribeContainerInstancesInput{Cluster: &cluster, ContainerInstances: containerInstances.ContainerInstanceArns})
+					if err != nil {
+						return nil, errors.Wrap(err, "could not gather ecs container instances")
```
czunker: Same as above: Continue and log to check the other clusters/instances?

motor/discovery/aws/ecs.go
```diff
@@ -0,0 +1,335 @@
...
+					}
+					for _, ci := range containerInstancesDetail.ContainerInstances {
+						// container instance assets
+						containerInstancesAssets = append(containerInstancesAssets, ecsContainerInstanceToAsset(account, region, ci, clonedConfig))
+					}
+				}
+				tasks, err := svc.ListTasks(ctx, &ecsservice.ListTasksInput{Cluster: &cluster})
+				if err != nil {
+					return nil, errors.Wrap(err, "could not gather ecs tasks for cluster")
```
czunker, line 133: Same as above: Continue and log to check the other clusters?

motor/discovery/aws/ecs.go
```diff
@@ -0,0 +1,335 @@
...
+				}
+				if len(tasks.TaskArns) > 0 {
+					taskdescriptions, err := svc.DescribeTasks(ctx, &ecsservice.DescribeTasksInput{Tasks: tasks.TaskArns, Cluster: &cluster})
+					if err != nil {
+						return nil, errors.Wrap(err, "could not describe ecs tasks for cluster")
```
czunker: Same as above: Continue and log to check the other clusters?

motor/discovery/aws/ecs.go
vjeffrey: it shouldn't, but i can log it

czunker (approved):
Nice! One more AWS resource. Thanks @vjeffrey

PR #631, opened 2022-12-15
🐛 Set recorder file enoent property correctly when file does not exist.

preslavgerchev:
makes sure the `enoent` property is correctly set so it can be used:
```
[files]
  [files.blabla]
    path = "blabla"
    enoent = true
    content = ""
    [files.blabla.stat]
      mode = 0
      time = 0001-01-01T00:00:00Z
      isdir = false
      uid = 0
      gid = 0
      size = 0
```


chris-rock (approved):
Thank you @preslavgerchev

PR #644, opened 2022-12-19
🧹 update aws dependencies

chris-rock:


go.mod
```diff
@@ -145,7 +145,7 @@ require (
 	github.com/vmware/goipmi v0.0.0-20181114221114-2333cd82d702
 	github.com/vmware/govmomi v0.29.0
 	github.com/xanzy/go-gitlab v0.74.0
-	github.com/zclconf/go-cty v1.10.0
```
preslavgerchev: this might be dangerous, not sure where/how we use it?
this is the same problem we had with the packer plugin

chris-rock: Great catch, removed this update for hcl

PR #645, opened 2022-12-19
⭐️ Container edge releases

imilchev:
Since we cannot use goreleaser directly to do our edge releases I created a separate workflow for that. The workflow is triggered for every commit to `main`. It will do a snapshot release of the binary and then create a container with that binary. The container is then pushed to our registry. This PR will create a container for every architecture. If we agree to the approach I will create a follow-up PR to do the virtual tag that combines all architectures.

### DIFF
--- a/.github/workflows/goreleaser-edge.yml
+++ b/.github/workflows/goreleaser-edge.yml
@@ -68,5 +68,4 @@ jobs:
           args: release -f .goreleaser-edge.yml --rm-dist --timeout 120m
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          CERT_PASSWORD: ${{ steps.gcp_secrets.outputs.code_sign_cert_challenge }}
           NFPM_DEFAULT_RPM_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
--- a/.github/workflows/goreleaser-edge.yml
+++ b/.github/workflows/goreleaser-edge.yml
@@ -22,7 +22,7 @@ jobs:
         os: [linux]
         arch: ['amd64', '386', 'arm64', 'arm', 'ppc64le']
 
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
     timeout-minutes: 120
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


.github/workflows/goreleaser-edge.yml
```diff
@@ -0,0 +1,111 @@
...
+      contents: 'write'
+      # Add "id-token" for google-github-actions/auth
+      id-token: 'write'
+
+    strategy:
+      matrix:
+        os: [linux]
+        arch: ['amd64', '386', 'arm64', 'arm', 'ppc64le']
+
+    runs-on: ubuntu-latest
```
imilchev: I assume we are using self-hosted to speed up the release. Since this workflow will execute 5 builds in parallel it should be okay to use the Github runners. Also in this way we won't block our self-hosted runner. Edge releases are not urgent anyways so even if they are very slow, that should be fine

chris-rock: I would go with self-hosted runner. Much faster.

.github/workflows/goreleaser-edge.yml
```diff
@@ -0,0 +1,111 @@
...
+            code_sign_cert_b64:mondoo-base-infra/mondoo_code_sign_certificate_pfx_b64
+            code_sign_cert_challenge:mondoo-base-infra/mondoo_code_sign_challenge
+      - name: "Write RPM Signing Cert"
+        run: |
+          gpgkey="$(mktemp -t gpgkey.XXX)"
+          base64 -d <<<"$GPG_KEY" > "$gpgkey"
+          echo "GPG_KEY_PATH=$gpgkey" >> $GITHUB_ENV
+        env:
+          GPG_KEY: '${{ secrets.GPG_KEY}}'
+      - name: "Write Windows Signing Cert"
```
chris-rock: Is the intention to also build edge binaries?

chris-rock:
This feels a bit of overload. What if we take the version we determine here https://github.com/mondoohq/cnquery/blob/main/Makefile#L20 and tag it locally (without pushing) and then run goreleaser without the release tasks?

.github/workflows/goreleaser-edge.yml
imilchev: I think this will be faster with the Github runners. I assume we have 1 hosted runner right now. This workflow will start 5 jobs in parallel for which it will need 5 runners. That would mean that they will actually be executed sequentially. 

This is just a guess for now since I haven't run this. I would give it a shot with the default runners first and if it's slow, we can easily swap it for our own

imilchev:
> This feels a bit of overload. What if we take the version we determine here https://github.com/mondoohq/cnquery/blob/main/Makefile#L20 and tag it locally (without pushing) and then run goreleaser without the release tasks?

I can try that with my test repo to see if it would work. Will describe the results once I have them

.github/workflows/goreleaser-edge.yml
imilchev: Well, they are needed for the containers but they are not pushed anywhere. I have the use-case for the containers now and nothing for the actual binaries. I was also not sure where those binaries should go right now, so I didn't address this topic at all.

My goal now is to have cnquery and cnspec edge releases so we can enable cnspec by default for the operator. We can figure out the binaries after that I guess since I am not aware of anyone actually waiting on that right now.

.github/workflows/goreleaser-edge.yml
benr: Given it's a matrix build, I'd agree with using shared runners for now.  We know Go builds can be slower on shared runners, so it's really a question of whether or not the total latency of each job is greater or less than the hit from running them on a single runner.  (There will be additional runners in the near future.)

.github/workflows/goreleaser-edge.yml
```diff
@@ -0,0 +1,111 @@
...
+          CERT_PASSWORD: ${{ steps.gcp_secrets.outputs.code_sign_cert_challenge }}
+          NFPM_DEFAULT_RPM_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
+      - name: Copy Dockerfile
+        run: |
+          cp Dockerfile dist/linux_linux_${{ matrix.arch }}
+      - name: Extract Docker metadata
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
```
benr: I don't see any place that IMAGE_NAME is defined.  Did you mean to set it to ${{ github.repository }} like seen [here](https://docs.github.com/en/actions/publishing-packages/publishing-docker-images)?

benr: Also note that our Docker Hub org is "mondoo", not "mondoohq".

.github/workflows/goreleaser-edge.yml
```diff
@@ -0,0 +1,111 @@
...
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v3
+        with:
+          distribution: goreleaser
+          version: latest
+          args: release --rm-dist --snapshot --id=linux --timeout 5m
```
benr: --id=linux is only a valid argument when invoked for build, not release.  

You should remove the RPM and Windows certs from the job.

benr (changes requested):

.gitignore
```diff
@@ -8,4 +8,5 @@ id_rsa
 id_rsa.pub
 mondoo-test-secret-key
 gha-creds-*.json
-/tools
\ No newline at end of file
+/tools
+.goreleaser-edge.yml
```
imilchev, line 12: We must ignore the copied file with the altered config such that goreleaser doesn't complain for creating a release from a dirty checkout

.github/workflows/goreleaser-edge.yml
```diff
@@ -0,0 +1,72 @@
...
+            ${{ runner.os }}-go-
+      - name: Log in to the Container registry
+        uses: docker/login-action@v2
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      # Make sure no releases are created in Github
+      # Make sure only linux binaries and docker files are built
+      - name: Prepare .goreleaser-edge.yml
```
imilchev, line 47: This will remove all the builds that we don't need for creating the edge containers. I also removed all the secrets that are no longer needed.

chris-rock (approved):
That looks great to me

benr (approved):
Very interesting solution to use yq.  Looks good!

PR #649, opened 2022-12-20
Sync in CLI updates from cnspec

tas50:
We were missing a few things here and were using incorrect terms


chris-rock (approved):
Thank you @tas50

PR #653, opened 2022-12-21
⭐️ GCP GKE cluster resource

imilchev:
Add GCP cluster resource. Can test it by running:
```
gcp.project.clusters{ * }
```


resources/packs/gcp/gcp.lr
```diff
@@ -789,4 +790,69 @@ private gcp.dns.policy {
   enableInboundForwarding bool
   // Indicates if logging is enabled
   enableLogging bool
+}
+
+// GKE cluster resource
```
tas50:
```suggestion
// GKE Cluster Resource
```

imilchev: done

resources/packs/gcp/gcp.lr
```diff
@@ -789,4 +790,252 @@ private gcp.dns.policy {
   enableInboundForwarding bool
   // Indicates if logging is enabled
   enableLogging bool
+}
+
+// GKE Cluster
+private gcp.cluster @defaults("name") {
```
chris-rock: Cluster should be scoped in `gcp.project.cluster`. All the other resources need to be reßscoped too. I think we should also allow users to test individual clusters directly:

```
gcp.project.cluster(projectID: "myproject", name: "clustername")
```

Therefore this resource should not be private. To achieve that you need to implement the init function as done for services https://github.com/mondoohq/cnquery/blob/main/resources/packs/gcp/services.go#L99-L154

chris-rock (changes requested):
This looks great. The only thing we should do is to move the resources into the `gcp.project` scope.

resources/packs/gcp/gcp.lr
imilchev: the rescoping is done. As agreed the init functions will be added at a later point. We will first add all the necessary GCP resources

chris-rock (approved):
Great addition @imilchev

PR #658, opened 2022-12-21
Fix new typos

tas50:
Fix typos in https://github.com/mondoohq/cnquery/pull/654 and https://github.com/mondoohq/cnquery/pull/655


apps/cnquery/cmd/run.go
```diff
@@ -23,7 +23,7 @@ func init() {
 
 var execCmd = builder.NewProviderCommand(builder.CommandOpts{
 	Use:   "run",
-	Short: "Run an MQL query",
+	Short: "Run an MQL query.",
```
chris-rock, line 26: Should we really have a dot in that short description. It always confuses me.

chris-rock (commented):
Thank you

chris-rock (approved):
@tas50 Thank you for the fix. We can discuss about the final sentence dot in a follow up PR

apps/cnquery/cmd/run.go
tas50: This was the only one without a period and it looked really off in the main pack docs

PR #659, opened 2022-12-21
Updates GitHub provider resource descriptions

scottford-io:
Updates resource descriptions for GitHub


tas50:
https://github.com/mondoohq/cnquery/pull/658 needs to get merged first. There was typos in the previous PR that are causing failures as these make it to the docs site.

resources/packs/github/github.lr
```diff
@@ -381,7 +399,7 @@ private github.commit @defaults("sha") {
 
 // GitHub Repository Pull Request 
 private github.mergeRequest @defaults("id state") {
-  // Pull request ID
+  // Pull request id
```
misterpantz: Hey @scottford-io why the lowercase id here and below?

misterpantz (commented):
Thank you for this!!!

resources/packs/github/github.lr
chris-rock: I also think ID is upper case

scottford-io: agreed!

scottford-io: fixed!

tas50 (approved):

PR #660, opened 2022-12-22
🧹 Use proper tag naming for edge releases

imilchev:
The goal here is to have the following tags for edge releases:
- edge-latest
- edge-latest-rootless
- edge-VERSION
- edge-VERSION-rootless

I spent quite some time trying to make the magic happen with `yq` but it seems too much effort. 80% of the goreleaser config for edge releases is the docker configuration. In this case it has to be quite different from the one for the normal release procedure. This is the reason why I decided to just manually configure the file and use that

### DIFF
--- a/.github/.goreleaser-edge.yml
+++ b/.github/.goreleaser-edge.yml
@@ -1,0 +1,169 @@
+---
+project_name: cnquery
+env:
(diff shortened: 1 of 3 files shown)


chris-rock (approved):
Thank you @imilchev

PR #662, opened 2022-12-22
Spellcheck the LR files

tas50:
Avoid introducing spellin mistakes here that get flagged in the docs site. This should avoid a lot of back and forth PRs and wasted time.


misterpantz (approved):
Thank you!

PR #664, opened 2022-12-23
🧹 aws ec2 & ssm discovery: add running instance state filter by default

vjeffrey:
⚠️ this changes the default behavior of the ec2 discovery to only return running/online instances (we'll need to update the lambda behavior as a follow up)

before:
Screenshot 2022-12-23 at 15 09 45

after:
Screenshot 2022-12-23 at 15 04 13

### DIFF
--- a/motor/discovery/aws/ec2_instances.go
+++ b/motor/discovery/aws/ec2_instances.go
@@ -39,6 +39,20 @@ type Ec2Instances struct {
 	profile       string
 }
 
+type Ec2InstanceState int
+
+const (
+	Running Ec2InstanceState = iota
+	All
+)
+
+type Ec2InstancesFilters struct {
+	InstanceIds   []string
+	Tags          map[string]string
+	Regions       []string
+	InstanceState Ec2InstanceState
+}
+
 func (ec2i *Ec2Instances) Name() string {
 	return "AWS EC2 Discover"
 }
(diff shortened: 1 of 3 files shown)


motor/discovery/aws/ssm_instances.go
```diff
@@ -109,6 +109,10 @@ func (ssmi *SSMManagedInstances) getInstances(account string, ec2InstancesFilter
 				input.Filters = append(input.Filters, types.InstanceInformationStringFilter{Key: aws.String("InstanceIds"), Values: ec2InstancesFilters.InstanceIds})
 				log.Debug().Interface("instance ids", ec2InstancesFilters.InstanceIds).Msgf("filtering")
 			}
+			if ec2InstancesFilters.InstanceState == 0 {
+				input.Filters = append(input.Filters, types.InstanceInformationStringFilter{Key: aws.String("PingStatus"), Values: []string{"Online"}})
```
vjeffrey, line 113: i considered adding a specific filter for the ssm ping, instead of relying on the running filter.. 🤷  opinions?

motor/discovery/aws/ec2_instances.go
```diff
@@ -96,6 +110,10 @@ func (ec2i *Ec2Instances) getInstances(account string, ec2InstancesFilters Ec2In
 					log.Debug().Msgf("filtering by tag %s:%s", k, v)
 				}
 			}
+			if ec2i.FilterOptions.InstanceState == 0 {
+				input.Filters = append(input.Filters, types.Filter{Name: aws.String("instance-state-name"), Values: []string{"running"}})
+				log.Debug().Msg("filtering by running state")
```
preslavgerchev: nit:
```suggestion
				log.Debug().Msg("filtering aws instances by running state")
```

gives more context when reading through logs

preslavgerchev (approved):

PR #675, opened 2022-12-27
🐛 Fix bug where concurrent chain executions would not complete

### LINKED ISSUE #674: Queries that cause concurrent execution can hang
To repro:
```bash
mkdir tmp
echo "tmp/a" > tmp/ref1
echo "asdf" > tmp/a

cat > testquery <<EOF
a = file("tmp/ref1").content.trim
file(a).path == "tmp/a"
file(a).content.trim == "asdf"
EOF

go run apps/cnquery/cnquery.go run  local --log-level trace -c -- run  local --log-level trace -c "$(cat testquery)"
```

In this example, the file resource causes concurrent execution. In this case, we rely on the trigger mechanism to get us to completion, unlike in the non-concurrent case where we can just run each chain to completion directly

jaym:
Setup:
```sh
mkdir tmp
echo "tmp/a" > tmp/ref1
echo "asdf" > tmp/a
```

A sample query that failed:
```
a = file("tmp/ref1").content.trim
file(a).path == "tmp/a"
file(a).content.trim == "asdf"
```

Fixes #674

### DIFF
--- a/llx/llx.go
+++ b/llx/llx.go
@@ -838,9 +838,11 @@ func (e *blockExecutor) runChain(start uint64) {
 			res, nextRef, err = e.runRef(curRef)
 		}
 
-		// stop this chain of execution, if it didn't return anything
+		// stop this chain of execution, if it didn't return anything and
+		// there is nothing else left to process
 		// we need more data ie an event to provide info
-		if res == nil && nextRef == 0 && err == nil {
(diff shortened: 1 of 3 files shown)


preslavgerchev (approved):

arlimus (approved):
Sweet catch @jaym Thank you!! :grin:

PR #677, opened 2022-12-27
✨Add subscriptions and subscriptions-exclude arguments for azure scanning

preslavgerchev:
Lets us include or exclude specific subscriptions when running an azure discovery. Very similar to the k8s namespaces filtering in terms of behaviour.

`~/go/bin/cnquery shell azure  --subscriptions-exclude=984df67f-fc2e-4ebf-80a2-98380182f060,1e829eb0-e6a3-4c7b-8212-5868aaebd318`

### DIFF
--- a/motor/discovery/azure/subscriptions.go
+++ b/motor/discovery/azure/subscriptions.go
@@ -10,6 +10,10 @@ import (
 type Subscriptions struct {
 	AzureClient *AzureClient
 }
+type subscriptionsFilter struct {
+	exclude []string
+	include []string
+}
 
 func NewSubscriptions(client *AzureClient) *Subscriptions {
 	return &Subscriptions{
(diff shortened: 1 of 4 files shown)


motor/discovery/azure/subscriptions.go
```diff
@@ -45,8 +49,36 @@ func (client *Subscriptions) GetSubscriptions() ([]subscriptions.Subscription, e
...
+				return false
+			}
+		}
+		// didn't find it, so it must be skipped
+		return true
+	}
+
+	// if nothing explicitly meant to be included, then check whether
+	// it should be excluded
+	if len(filter.exclude) > 0 {
```
imilchev, line 74: If I read this code correctly it means that if you provide both include and exclude it won't work. I think that's good because it makes the code simpler. I do think that you should add a check in the CLI to make sure that we return an error if someone provides both flags at the same time

imilchev (approved):

PR #690, opened 2022-12-28
🧹 Clarify data returned by docker resource

tas50:
Return what format the data is returned in not just the name of the data. Also fix up some other core resource data.


chris-rock (approved):
Thank you @tas50

PR #692, opened 2022-12-30
⭐️ use the ecr keychain helper for ecr image connect

vjeffrey:
unfortunately i haven't found a way to tell the ecr keychain helper to use a specific profile quite yet (to be able to pass in the profile identifier that is passed to the shell)


motor/providers/container/image/registry.go
```diff
@@ -73,7 +79,12 @@ func LoadImageFromRegistry(ref name.Reference, opts ...Option) (v1.Image, io.Rea
 	}
 
 	if o.auth == nil {
-		auth, err := authn.DefaultKeychain.Resolve(ref.Context())
+		client := api.DefaultClientFactory{}
```
czunker: Do we need this? L57 has no explicit factory.

czunker:
> unfortunately i haven't found a way to tell the ecr keychain helper to use a specific profile quite yet (to be able to pass in the profile identifier that is passed to the shell)

I read this, that the helper implicitly checks the env var: https://github.com/awslabs/amazon-ecr-credential-helper#aws-credentials

motor/providers/container/image/registry.go
vjeffrey: we don't! nice catch

vjeffrey:
> > unfortunately i haven't found a way to tell the ecr keychain helper to use a specific profile quite yet (to be able to pass in the profile identifier that is passed to the shell)
> 
> I read this, that the helper implicitly checks the env var: https://github.com/awslabs/amazon-ecr-credential-helper#aws-credentials

thank you!  maybe we could temporarily set the env var according to the awsprofile passed in, 🤔 or just manually get the credential when an awsprofile is passed in

czunker (approved):
Nice! Thanks @vjeffrey

PR #705, opened 2023-01-02
🌙 Add GCP Logging MQL resources

imilchev:
GCP SQL MQL resources:
 - buckets - `gcp.project.logging.buckets{ * }`


resources/packs/gcp/gcp.lr
```diff
@@ -1040,4 +1042,64 @@ private gcp.project.cluster.nodepool.config.confidentialNodes @defaults("enabled
   id string
   // Whether to use confidential nodes
   enabled bool
+}
+
+// GCP Logging resources
+private gcp.project.loggingservices {
```
chris-rock: Should this be a singular? `gcp.project.loggingservice`

resources/packs/gcp/gcp.lr
```diff
@@ -1040,4 +1042,64 @@ private gcp.project.cluster.nodepool.config.confidentialNodes @defaults("enabled
...
+  // Project ID
+  projectId string
+  // List of logging buckets
+  buckets() []gcp.project.loggingservices.bucket
+}
+
+// GCP Logging bucket
+private gcp.project.loggingservices.bucket @defaults("name") {
+  // Internal ID
+  id string
```
chris-rock: Lets remove all `Internal ID` since we do not need them

resources/packs/gcp/logging.go
```diff
@@ -0,0 +1,116 @@
...
+	"fmt"
+
+	"go.mondoo.com/cnquery/resources"
+	"go.mondoo.com/cnquery/resources/packs/core"
+	"google.golang.org/api/logging/v2"
+	"google.golang.org/api/option"
+)
+
+func (g *mqlGcpProjectLoggingservices) id() (string, error) {
+	return "gcp.project.loggingservices", nil
```
chris-rock: This will not work for multiple instances for multiple projects since they return always the same id.

```
func (g *mqlGcpProjectLoggingservices) id() (string, error) {
  projectId, err := g.ProjectId()
  if err != nil { return "", err}
  
  return "gcp.project.loggingservices/"+ projectID, nil
}
```

resources/packs/gcp/logging.go
```diff
@@ -0,0 +1,116 @@
...
+				"type", cfg.Type,
+			)
+			if err != nil {
+				return nil, err
+			}
+			indexConfigs = append(indexConfigs, mqlIndexConfig)
+		}
+
+		mqlBucket, err := g.MotorRuntime.CreateResource("gcp.project.loggingservices.bucket",
+			"id", bucket.Name,
```
chris-rock: We do not need the id since we have the name as field already. Lets move this logic into the id function.

resources/packs/gcp/logging.go
```diff
@@ -0,0 +1,116 @@
...
+			)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		indexConfigs := make([]interface{}, 0, len(bucket.IndexConfigs))
+		for i, cfg := range bucket.IndexConfigs {
+			mqlIndexConfig, err := g.MotorRuntime.CreateResource("gcp.project.loggingservices.bucket.indexConfigs",
+				"id", fmt.Sprintf("%s/indexConfigs/%d", bucket.Name, i),
```
chris-rock, line 82: lets move this logic into the id function. Instead we should add the bucketName as attribute

resources/packs/gcp/logging.go
```diff
@@ -0,0 +1,116 @@
...
+		return nil, err
+	}
+
+	mqlBuckets := make([]interface{}, 0, len(buckets.Buckets))
+	for _, bucket := range buckets.Buckets {
+
+		var mqlCmekSettings resources.ResourceType
+		if bucket.CmekSettings != nil {
+			mqlCmekSettings, err = g.MotorRuntime.CreateResource("gcp.project.loggingservices.bucket.cmekSettings",
+				"id", fmt.Sprintf("%s/cmekSettings", bucket.Name),
```
chris-rock: lets move that logic into the id function

chris-rock (commented):
Great addition @imilchev I left a few comments for improvement.

resources/packs/gcp/logging.go
imilchev: I don't have the index in the MQL resource. I either need to add the index, the bucket name and the project ID to this reosource and then move the construction of the ID to the id function or I should leave it like this. Not sure what's better but this looks like less code

resources/packs/gcp/gcp.lr
```diff
@@ -1040,4 +1042,66 @@ private gcp.project.cluster.nodepool.config.confidentialNodes @defaults("enabled
...
+  // List of logging buckets
+  buckets() []gcp.project.loggingservice.bucket
+}
+
+// GCP Logging bucket
+private gcp.project.loggingservice.bucket @defaults("name") {
+  // Project ID
+  projectId string
+  // CMEK settings of the log bucket
+  cmekSettings gcp.project.loggingservice.bucket.cmekSettings
```
chris-rock: I think we can use a dict here

imilchev: done

chris-rock (approved):
Thank you @imilchev

PR #707, opened 2023-01-03
✨ Add azure.cloudDefender resources

preslavgerchev:
Adds `azure.cloudDefender` resource:
```
// Azure Microsoft Cloud Defender
azure.cloudDefender {
  // indicates if the monitoring agent will be automatically provisioned on new VMs
  monitoringAgentAutoProvision() bool
  // the list of configured security contacts
  securityContacts() []azure.cloudDefender.securityContact
}

// Azure Microsoft Cloud Defender Security Contact 
private azure.cloudDefender.securityContact {
  // the identifier of the security contact
  id string
  // the name of the security contact
  name string
  // the emails that will receive security alerts
  emails []string
  // the alerts notification settings
  alertNotifications dict
  // the notifications by role settings
  notificationsByRole dict
}
```

Note I placed this under `azure.cloudDefender` and not `microsoft.cloudDefender` since Microsoft Cloud Defender supports multiple clouds and this is it's azure-specific settings. Happy to discuss that and change if needed though


chris-rock:
This is only available with the azure resource pack and requires the azure api right? In that case I would stick with the `azure` category

resources/packs/azure/azure_cloud_defender.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	security "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/security/armsecurity"
+	"go.mondoo.com/cnquery/resources/packs/core"
+)
+
+func (a *mqlAzureCloudDefender) id() (string, error) {
+	return "azure.cloudDefender", nil
```
chris-rock, line 22: going forward we need to scope this to a subscription

chris-rock (approved):
Thank you @preslavgerchev

resources/packs/azure/azure_cloud_defender.go
preslavgerchev: Good point. I think currently we do not support multiple subscriptions on the microsoft provider, so I reckon for now it's okay

chris-rock: That will need to change within the next weeks since we need full asset hierarchy for Azure as well.

PR #709, opened 2023-01-03
Make letha the codeowner on the LR files

tas50:
This way we get reviews on potential docs changes before they make it into the docs. This reduces followup PRs and wasted effort.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,6 @@
 ### mondoohq/cnquery CODEOWNERS
 
-# All resource updates require editorial review by Docs Editor Letha with backup of Tim
-*.lr @misterpantz @tas50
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,4 @@
+### mondoohq/docs CODEOWNERS
+
+# All resource updates require editorial review by Docs Editor Letha
+*.lr letha@mondoo.com
```
chris-rock: We need to define a deputy in case Letha is not available

tas50: I can do that if she's not around

benr:
Updated to use a Github Team after discussion with @chris-rock.  This will be non-enforced and thus only a recommendation for review, not a requirement.

chris-rock (approved):
Thank you @benr

PR #718, opened 2023-01-04
🌙 Add GCP Essential contact MQL resources

imilchev:
GCP Essential contact resource:
- `gcp.project.essentialContacts`

Note the resource itself is called `gcp.essentialContact` because I saw the same thing is used for folders and organisations. If we add support for those in the future, the same resource will be used.


chris-rock (approved):
Thank you @imilchev

PR #720, opened 2023-01-04
🧹 detect the aws ec2 instance name

vjeffrey:
before & after:

Screenshot 2023-01-04 at 15 07 05

notice the "connecting to asset" line. with the installed cnquery, it's the instance id. with the local copy of cnquery, it's the name of the instance from the name tag


vjeffrey:
closing, it seems most plain ec2 instances also dont have these perms

PR #722, opened 2023-01-05
⭐ dict.contains supports maps with keys and values

arlimus:
Couple of things:
1. add the `key` and `value` keywords when used inside of `where` clauses in dicts, e.g. when you parse JSON
2. make these keywords now also available in arrays. you can still use `_` to select the current element, but to be consistent and since the JSON parsing type is not known at compile-time we have to support `key` and `value` for anything that can be called on top of it (like `where`, `contains` and friends). For consistency, arrays now also support accessing elements via `key` and `value`
3. support the `_` in maps. Now that we treat dict maps in a way where keys and values are supported, we found a shortcoming in the current implementation of regular maps, where `_` was not pointing to the actual map entries. This was not done in the past since we could have selected either. With the update to dict we by default always select the value, so that has been ported back to maps.

## How is this useful?

Let's say you were looking for a value inside of a map, but didn't know which keys it is under. E.g. what if we were looking for everything that was `true`, or as in this example set to `1` (or not set to `1` as in the second example):

image

The same thing works with keys, e.g. let's say we wanted all entries with an "e" in their key:

image

Here is an example with the same functionality on top of arrays from a parsed JSON (i.e. dict with arrays). Here you can use `key` and `value` too now:

image

As mentioned, we have backported the support for `_` to regular maps:

image

... but we still highly encourage the use of `value` instead, since `_` is hard to guess in the case of maps (with their keys and values):

image


scottford-io:
I just tested with Terraform modules and it is now working as expected:

```hcl
module "uniform_bucket_level_access_module_fail" {
  source  = "terraform-google-modules/cloud-storage/google"

  prefix       = "uniform-bucket-level-access-fail-"
  names = [ "module" ]
  project_id = var.project_id
  location   = var.location

  bucket_policy_only = { 
    module = false 
  }

}
```

## MQL Query:

```javascript
cnquery> terraform.modules.where( source == "registry.terraform.io/terraform-google-modules/cloud-storage/google" ).all( block.attributes['bucket_policy_only']['value'].none(value == false) )
[failed] [].all()
  actual:   [
    0: terraform.module id = terraform.module/key/uniform_bucket_level_access_module_fail/version/3.4.1
  ]
```

scottford-io (approved):

jaym (approved):

chris-rock:
I add the examples here for copy paste:

```json
{
  "1": 1,
  "1.0": 1,
  "date": "2016-01-28T23:02:24Z",
  "hello": "hello",
  "true": true,
  "dict": {
    "ee": 3,
    "ej": 4,
    "ek": 5
  },
  "f": [ { "ff": 3}]
}
```


```javascript
cnquery> parse.json('dummy.json').params
parse.json.params: {
  1: 1.000000
  1.0: 1.000000
  date: "2016-01-28T23:02:24Z"
  dict: {
    ee: 3.000000
    ej: 4.000000
    ek: 5.000000
  }
  f: [
    0: {
      ff: 3.000000
    }
  ]
  hello: "hello"
  true: true
}
cnquery> parse.json('dummy.json').params.where( value == 1)
parse.json.params.where: {
  1: 1.000000
  1.0: 1.000000
}
cnquery> parse.json('dummy.json').params.where( value != 1)
parse.json.params.where: {
  date: "2016-01-28T23:02:24Z"
  dict: {
    ee: 3.000000
    ej: 4.000000
    ek: 5.000000
  }
  f: [
    0: {
      ff: 3.000000
    }
  ]
  hello: "hello"
  true: true
}
cnquery> parse.json('dummy.json').params.where( key != /e/)
parse.json.params.where: {
  1: 1.000000
  1.0: 1.000000
  dict: {
    ee: 3.000000
    ej: 4.000000
    ek: 5.000000
  }
  f: [
    0: {
      ff: 3.000000
    }
  ]
}
cnquery> parse.json('dummy.json').params.where( key == /e/)
parse.json.params.where: {
  date: "2016-01-28T23:02:24Z"
  hello: "hello"
  true: true
}
```


```javascript
cnquery> sshd.config.params
sshd.config.params: {
  AcceptEnv: "LANG LC_*"
  AllowAgentForwarding: "yes"
  AuthorizedKeysFile: ".ssh/authorized_keys"
  Subsystem: "sftp /usr/libexec/sftp-server"
  UsePAM: "yes"
}
```

```javascript
cnquery> sshd.config.params.where (value == "yes") 
sshd.config.params.where: {
  AllowAgentForwarding: "yes"
  UsePAM: "yes"
}
```

```javascript
cnquery> sshd.config.params.where (_ == "yes") 
sshd.config.params.where: {
  AllowAgentForwarding: "yes"
  UsePAM: "yes"
}
```

PR #724, opened 2023-01-05
🌟 --props key=value

arlimus:
Allow users to inject properties into the execution via `--props` by specifying keys and values.                                                               
                                                                                                                                                               
For most common cases we look at UIDs. If you define a property like this:                                                                                     
                                                                                                                                                               
```yaml
props:
- uid: homeFolder
  mql: |
    "~"   
```
                                                                                                                                                               
and use it in queries like this:                                                                                                                               
                                                                                                                                                               
```yaml
- title: Home zsh
  uid: homeZsh 
  props:          
  - uid: homeFolder
  mql: |                     
    file(props.homeFolder) { * }
```                                                                                                                                                            
                                                                                                                                                               
Then you can now set it via                                                                                                                                    
                                                                                                                                                               
```bash
cnquery scan -f my.mql.yaml --props "homeFolder='myHome'"
```                                                                                                                                                            
                                                                                                                                                               
Note: the quotation marks are important! We need to differentiate e.g. strings from resources, since the value has to be valid MQL. 

![image](https://user-images.githubusercontent.com/1307529/210767016-196488cb-7bb8-4648-a961-f2c9cff372c3.png)
![image](https://user-images.githubusercontent.com/1307529/210767056-9cfbf750-bd87-457e-860d-3dd637e300b9.png)


chris-rock (approved):
Great addition @arlimus

PR #725, opened 2023-01-05
🧹 Remove name from gcp.essentialContact

imilchev:
Since the name is just a number, we don't show it


chris-rock (approved):
Thank you @imilchev

PR #729, opened 2023-01-05
Fix spelling of MySQL

tas50:
Add a pattern match on this and also sync over style changes to the spellcheck job to make it easier to diff


misterpantz (approved):
Thanks!

PR #730, opened 2023-01-05
Add AWS product names to spellcheck

tas50:
Make sure we get these right as they flow into docs


misterpantz (approved):
🥳

PR #732, opened 2023-01-06
Use the official Azure names for database offerings

tas50:
Per https://learn.microsoft.com/en-us/azure/?product=databases


misterpantz (approved):
beauty!!!!!

PR #737, opened 2023-01-06
✨ GCP compute network subnetworks retrieval

imilchev:


resources/packs/gcp/compute.go
```diff
@@ -939,8 +939,61 @@ func (g *mqlGcpComputeNetwork) id() (string, error) {
...
+	wg.Add(len(ids))
+	mux := &sync.Mutex{}
+	for _, id := range ids {
+		go func(id resourceId) {
+			defer wg.Done()
+			subnet, err := computeSvc.Subnetworks.Get(id.Project, id.Region, id.Name).Do()
+			if err != nil {
+				log.Error().Err(err).Send()
+			}
+			mqlSubnet, err := newMqlSubnetwork(id.Project, nil, g.MotorRuntime, subnet)
```
imilchev: Not sure how I can set the region here

resources/packs/gcp/compute.go
```diff
@@ -1016,14 +1051,76 @@ func (g *mqlGcpComputeSubnetwork) id() (string, error) {
...
+	}
+
+	ctx := context.Background()
+
+	computeSvc, err := compute.NewService(ctx, option.WithHTTPClient(client))
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := computeSvc.Regions.Get(projectId, regionName).Do()
```
imilchev: @chris-rock this is what we chatted about. Here I would like to use `gcp.compute.regions` if possible, so I don't need to query GCP multiple times for the same region

preslavgerchev (approved):

PR #739, opened 2023-01-06
🧹 update aws debug tests and logging

chris-rock:
improve the logging messages and updates the debug test to be able to test aws discovery and provider quicker

### DIFF
--- a/motor/providers/aws/provider.go
+++ b/motor/providers/aws/provider.go
@@ -52,10 +52,12 @@ func TransportOptions(opts map[string]string) []ProviderOption {
 	}
 
 	if awsRegion, ok := opts["region"]; ok {
+		log.Debug().Str("region", awsRegion).Msg("using region")
 		transportOpts = append(transportOpts, WithRegion(awsRegion))
 	}
 
 	if awsProfile, ok := opts["profile"]; ok {
+		log.Debug().Str("profile", awsProfile).Msg("using aws profile")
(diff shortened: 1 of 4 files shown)


motor/providers/aws/provider.go
```diff
@@ -79,7 +81,7 @@ func New(pCfg *providers.Config, opts ...ProviderOption) (*Provider, error) {
 		return nil, errors.Wrap(err, "could not load aws configuration")
 	}
 	if cfg.Region == "" {
-		log.Warn().Msg("no AWS region found, using us-east-1")
+		log.Info().Msg("no AWS region found, using us-east-1")
```
tas50, line 84: Found where?

chris-rock: either `AWS_REGION` or `.aws/credentials`

tas50: Can we make that part of the message? Otherwise it's not actionable by the user.

chris-rock: I agree but would like not to do this as part of this PR. We need another round of changes for AWS to change the way we use credentials for AWS internally

tas50: I think we need to get that PR as soon as possible. It's important for us to provide an actionable error message and right now this one isn't actionable. We already saw in Slack this week where users were confused around how these regions were set.

vjeffrey (approved):

PR #741, opened 2023-01-08
🧹 socketstats resource was returning an empty error 

chris-rock:
when used on a povider that does not allow to run commands, it returned an empty error:

```
cnspec> socketstats.openPorts
Query encountered errors:

socketstats.openPorts: []
```

Not it returns:

```
cnspec> socketstats.openPorts
Query encountered errors:
socketStats not supported on this provider
socketstats.openPorts: []
```


imilchev (approved):
LGTM

PR #742, opened 2023-01-09
✨ Add GCP compute network mode property

imilchev:
The gcloud CLI returns network mode which can be auto, legacy or custom. This is a computed field according to the gcloud CLI source. I copied the logic behind that field from the CLI implementation (see [here](https://github.com/google-cloud-sdk-unofficial/google-cloud-sdk/blob/ecbea23d5753163a877909040d71d49b19170a15/lib/googlecloudsdk/api_lib/compute/networks_utils.py#L24-L29)).

I guess it's useful to have that embedded in the MQL resource since it is part of the CIS policy for GCP and would make the queries simpler


czunker (approved):
LGTM

PR #747, opened 2023-01-09
🧹 Move gcp.project.clusters to gcp.project.gke.clusters

imilchev:
This makes the resource consistent with the rest of the GCP resources that we have


resources/packs/gcp/gcp.lr
```diff
@@ -46,7 +46,8 @@ gcp.project @defaults("name") {
   recommendations() []gcp.recommendation
   // GCP Compute Resources for the Project
   // compute() gcp.compute
-  clusters() []gcp.project.cluster
+  // GCP GKE resources
+  gke() gcp.project.gkeService
```
chris-rock, line 51: I think it is a great call. Do we need an alias for it?

imilchev: I think not since this was released last week and we are not using it just yet.

chris-rock: Sounds good to me, just coordinate with Tim, that we update the release notes.

PR #749, opened 2023-01-09
🧹 gcp.compute -> gcp.project.compute

imilchev:


resources/packs/gcp/gcp.lr
```diff
@@ -3,9 +3,6 @@ option go_package = "go.mondoo.com/cnquery/resources/packs/gcp"
 alias gcloud.organization = gcp.organization
 alias gcloud.project = gcp.project
 alias gcloud.resourcemanager.binding = gcp.resourcemanager.binding
-alias gcloud.compute = gcp.compute
```
imilchev, line 6: @chris-rock for some reason if those aliases stay in I get the following error when cnquery launches:
```
./cnquery run gcp -c "gcp.project.compute.networks{*}"😂                                                                                                                            
panic: failed to add resource info: could not merge resources because IDs don't match: , gcp.project

goroutine 1 [running]:
go.mondoo.com/cnquery/resources/packs/gcp/info.init.0()
        /Users/ivanmilchev/Projects/cnquery/resources/packs/gcp/info/info.go:27 +0xb8
```

If the aliases are deleted, then everything works correctly. I tried to figure out what the root cause of the error is but I am not entirely sure. Any suggestions what to do with this?

tas50:
Is the end goal to move all the services under project?

chris-rock:
> 

No, only the services that are part of a project and require a project. There are org level services. The way it is designed is to allow a user to query multiple projects. It is the preparation to get the asset hierarchy in.

imilchev:
This is now ready to be merged

resources/packs/gcp/gcp.lr
```diff
@@ -117,35 +119,35 @@ gcp.resourcemanager.binding {
 }
 
 // GCP Compute Engine
-gcp.compute {
```
chris-rock, line 120: Can we add an alias or a temporary deprecated implementation that returns the new object:
```
gcp {
  compute() gcp.project.computeService
}
```

chris-rock (commented):
Thank you @imilchev Please make sure all policies are updated accordingly

resources/packs/gcp/gcp.lr
imilchev: done

PR #750, opened 2023-01-09
⭐️ Expose memory vault.

preslavgerchev:
Allows the usage of memory vault if instantiating from an inventory. Allow storing of all secrets, regardless of encoding.


chris-rock (approved):
Great addition @preslavgerchev

PR #751, opened 2023-01-09
✨ Extend Berglas vault provider with opts.

preslavgerchev:
Add `WithBucket` and `WithKmsKeyId` opts to set the bucket and kms key to be used when using berglas for storing secrets. The vault now also returns the berglas key by using the passed in value and returning it in a formatted manner. This lets us inject the bucket to which we want data to be stored in an easier way, rather than relying on a passed in string.


chris-rock (approved):
Great addition @preslavgerchev

PR #752, opened 2023-01-09
🌙 Rename lr Compute func to MqlCompute

imilchev:
The name `Compute` causes a collision with a resource we want to add for GCP


chris-rock (approved):
Thank you @imilchev

PR #753, opened 2023-01-10
Update GCP service names in help / docs

tas50:
Make these match the google branded names on their site


scottford-io (approved):
Thanks @tas50

PR #770, opened 2023-01-12
🐛 aws: try govcloud region if default region does not work

vjeffrey:
```
[11/01/23 18:41:42] ❯ go run apps/cnquery/cnquery.go shell aws --profile gov
→ loaded configuration from /Users/vj/.config/mondoo/mondoo.yml using source default

cnquery> aws.ec2.instances { * }
aws.ec2.instances: [
  0: {
    launchTime: 2023-01-12 01:23:47 +0000 UTC
```

### DIFF
--- a/motor/providers/aws/provider.go
+++ b/motor/providers/aws/provider.go
@@ -88,8 +88,13 @@ func New(pCfg *providers.Config, opts ...ProviderOption) (*Provider, error) {
 	// gather information about the aws account
 	identity, err := CheckIam(t.config)
 	if err != nil {
-		log.Warn().Err(err).Msg("could not gather details of AWS account")
-		// do not error since this break with localstack
+		// try with govcloud region
+		t.config.Region = "us-gov-west-1"
+		identity, err = CheckIam(t.config)
+		if err != nil {
+			log.Warn().Err(err).Msg("could not gather details of AWS account")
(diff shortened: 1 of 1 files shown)


motor/providers/aws/provider.go
```diff
@@ -88,8 +88,13 @@ func New(pCfg *providers.Config, opts ...ProviderOption) (*Provider, error) {
 	// gather information about the aws account
 	identity, err := CheckIam(t.config)
```
chris-rock, line 89: This is a great addition. I wonder if we could simplify the implementation for more clarity. This can be done in a followup PR. Something like:

```go
regions := []string{}

if cfg.Region != "" {
   regions = []string{cfg.Region}
} else {
   regions  = []string{"us-east-1", "us-gov-west-1"} // all regions we try before giving up
}
 
for _, region range regions {
   identity, err := CheckIam(t.config)
   if err != nil { continue}
   
   // all passed
   found = true
   break
}

if found != true {
  // do what is required if we cannot connect
}

// proceed
```


```

chris-rock (approved):
Great addition @vjeffrey

motor/providers/aws/provider.go
vjeffrey: oh yes, that's nice!

PR #792, opened 2023-01-17
🐛 fix var labels in comparable assertions in blocks

arlimus:
One more special case I had overlooked, kudos @chris-rock for the reminder.

Finalizes https://github.com/mondoohq/cnquery/issues/574

**Problem**

I have a query where I assign a value to a variable and then compare the variable e.g. `["a", "b"] { a=_ a=="a" }` The output is sub-optimal since I do not see anymore in the results which value it was compared to.

```javascript
cnquery> ["a", "b"] { a=_ a=="a" }
[
  0: {
     == "a": true
  }
  1: {
     == "a": false
  }
]
```

**Solution**

The new output includes the proper label of the variable:

```javascript
cnquery> ["a", "b"] { a=_ a=="a" }
[
  0: {
    a == "a": true
  }
  1: {
    a == "a": false
  }
]
```


chris-rock (approved):
This is a great addition @arlimus Makes it much easier to understand the results

PR #801, opened 2023-01-18
⭐️ Allow providing GCP service account via inventory vault

imilchev:
Add support for providing GCP service account via an inventory vault. For testing use `mondoo-dev` project and the following inventory:
```yaml
apiVersion: v1
kind: Inventory
metadata:
  name: inventory
spec:
  assets:
    - name: cool-stuff
      connections:
        - backend: 13
          credentials:
            - secret_id: storage/random-bucket2/foo
              type: 1
              secret_encoding: 3
          options:
          discover:
            targets:
              - auto
  vault:
    name: gcp-berglas
    type: gcp-berglas
    options:
      project_id: mondoo-dev-262313
```

Then the following command can be used: 
```bash
./cnquery scan --inventory-file inv.yaml
```


motor/providers/google/provider.go
```diff
@@ -25,13 +26,18 @@ const (
...
+	var cred *vault.Credential
 	if pCfg.Backend == providers.ProviderType_GCP {
 		// FIXME: DEPRECATED, update in v8.0 vv
 		// The options "project" and "organization" have been deprecated in favor of project-id and organization-id
 		if pCfg.Options == nil || (pCfg.Options["project-id"] == "" && pCfg.Options["project"] == "" && pCfg.Options["organization-id"] == "" && pCfg.Options["organization"] == "") {
 			// ^^
 			return nil, errors.New("google provider requires a gcp organization id, gcp project id or google workspace customer id. please set option `project-id` or `organization-id` or `customer-id`")
 		}
+
+		if len(pCfg.Credentials) != 0 {
```
preslavgerchev, line 38: should we also be checking the credential type? i'd say for gcp you want to pass in a `private_key` credential, passing in a `bearer` doesn't much make sense here.

preslavgerchev (approved):

PR #804, opened 2023-01-19
⭐️ aws resources as assets

vjeffrey:
problem:
it's hard to see the results of aws resources when they're all checks in one policy

in this pr, we:
- gather resources using the mql discovery that already exists
- the resource is converted to an asset
- each resource init function knows how to look at its name and platform ids to understand what it is

then in the policy:
- we change the queries to go from `aws.vpcs.all(test goes here)` -> `aws.vpc { test goes here}`
- and use asset filters to filter on the platform like `aws.s3_bucket`

--------
~note: this is currently using a special temporary discovery target, "mql-objects". this will switch to auto discovery once the policies are ready, and will probably switch names even before then, i was just using something quick to test.~

i split the pr into a couple commits to more clearly show what needs to be done to add a couple more resources as assets, since we should be able to reuse a lot of this logic for the other clouds

there's also a commit to show how to pass the platform thru so that the asset has the correct platform at asset filter evaluation time (it's the platform override stuff)

`go run apps/cnquery/cnquery.go shell aws --discover resources,accounts`

to test it out with cnspec and the console ui:
- `go run apps/cnspec/cnspec.go scan aws --config ../mondoo/demo.agent.credentials.json --discover resources,accounts`
- https://gitlab.com/mondoolabs/mondoo/-/merge_requests/3602

and you'll need an updated policy, like:
[newpolicy.zip](https://github.com/mondoohq/cnquery/files/10476779/newpolicy.zip)


motor/discovery/aws/mql_asset_objects.go
```diff
@@ -0,0 +1,283 @@
...
+	"github.com/cockroachdb/errors"
+	"go.mondoo.com/cnquery/motor/asset"
+	"go.mondoo.com/cnquery/motor/providers"
+)
+
+func getTitleFamily(awsObject awsObject) (awsObjectPlatformInfo, error) {
+	switch awsObject.service {
+	case "s3":
+		if awsObject.objectType == "bucket" {
+			return awsObjectPlatformInfo{title: "AWS S3 Bucket", platform: "aws.s3_bucket"}, nil
```
imilchev: We should align on the platform format. In k8s we have defined those as `k8s-pod`, `k8s-deployment`, etc. If we want to be consistent I guess we should used `aws-s3-bucket`?

motor/discovery/aws/mql_assets.go
```diff
@@ -0,0 +1,180 @@
...
+		},
+		State:       asset.State_STATE_ONLINE,
+		Labels:      addInformationalLabels(mqlObject.labels, mqlObject),
+		Connections: []*providers.Config{t},
+	}
+}
+
+func validate(m mqlObject) error {
+	if m.name == "" {
+		return errors.New("name required for mql object to asset translation")
```
imilchev:
```suggestion
		return errors.New("name required for mql aws object to asset translation")
```

motor/discovery/aws/mql_assets.go
```diff
@@ -0,0 +1,180 @@
...
+	d, _ := mapstructure.NewDecoder(&mapstructure.DecoderConfig{
+		Result: &a,
+	})
+	d.Decode(value.Value)
+	return a
+}
+
+func GatherMQLObjects(tc *providers.Config, account string) ([]*asset.Asset, error) {
+	assets := []*asset.Asset{}
+	pCfg := providers.Config{Backend: providers.ProviderType_AWS}
```
imilchev: is there any reason why not doing `tc.Clone()`? If we go down that way we are not passing down options, for example, that might be required by the provider

motor/discovery/aws/mql_asset_objects.go
vjeffrey: sounds good to me!

motor/discovery/aws/mql_assets.go
vjeffrey: no reason, i can add that in

motor/discovery/aws/mql_asset_objects.go
vjeffrey: i also renamed the core platform to aws-core in one of the commits, need to undo that back to just "aws". mostly just noting that here so i remember

resources/packs/aws/aws.lr
```diff
@@ -1256,6 +1256,8 @@ private aws.ec2.volume @defaults("arn") {
   volumeType string
   // Time the volume was created
   createTime time
+  // region where the ec2 volume is stored
```
misterpantz:
```suggestion
  // Region where the EC2 volume is stored
```

misterpantz (commented):
One little suggestion from your fond nitpicker. :)

vjeffrey:
ok, i've undrafted the pr. i changed the platform names to be with `-` as discussed.

obviously more followups will come for more resources as assets, but this gets us a good chunk of em without breaking anything.

motor/discovery/aws/mql_asset_objects.go
```diff
@@ -0,0 +1,427 @@
...
+	case "redshift":
+		if awsObject.objectType == "cluster" {
+			return awsObjectPlatformInfo{title: "AWS Redshift Cluster", platform: "aws-redshift-cluster"}, nil
+		}
+	case "vpc":
+		if awsObject.objectType == "vpc" {
+			return awsObjectPlatformInfo{title: "AWS VPC", platform: "aws-vpc"}, nil
+		}
+	case "ec2":
+		if awsObject.objectType == "securitygroup" {
```
imilchev: you could us a `switch` here. I did that in my PR as well. Everywhere where there's more than 1 case I just use `switch`. Makes the code easier to read

motor/discovery/aws/mql_asset_objects.go
```diff
@@ -0,0 +1,427 @@
...
+				awsObject: awsObject{
+					account: account, region: region, arn: arn,
+					id: id, service: "vpc", objectType: "vpc",
+				},
+			}, tc))
+	}
+	return assets
+}
+
+func securitygroups(m *MqlDiscovery, account string, tc *providers.Config) []*asset.Asset {
```
imilchev: nit: using camel case in the function names makes them a bit easier to read. As an extra benefit, they will also match the casing of the MQL resources

motor/discovery/aws/resolver.go
```diff
@@ -15,23 +15,57 @@ import (
...
-	DiscoveryECS          = "ecs"
+	// api-only scan
+	DiscoveryAccounts       = "accounts"
+	DiscoveryResources      = "resources"
+	DiscoveryECS            = "ecs" // todo move to mql, add policy
+	DiscoveryS3             = "s3"
+	DiscoveryCloudtrail     = "cloudtrail"
+	DiscoveryRds            = "rds"
+	DiscoveryVPC            = "vpc"
+	DiscoverySecurityGroups = "security-groups"
```
imilchev: In my GCP PR I prefixed these with the service. For example, `ec2-security-groups`, `compute-firewalls`, `compute-networks`, etc... Perhaps good to align how we want to name the discovery flags so they have the same pattern across different clouds

vjeffrey:
```
~/go/src/go.mondoo.io/cnspec main* 1m 37s
[30/01/23 15:13:38] ❯ go run apps/cnspec/cnspec.go scan aws --config ../mondoo/demo.agent.credentials.json
# command-line-arguments
ld: warning: -no_pie is deprecated when targeting new OS versions
ld: warning: non-standard -pagezero_size is deprecated when targeting macOS 13.0 or later
→ loaded configuration from ../mondoo/demo.agent.credentials.json using source --config
Asset: AWS Account 172746783610
```
just to show that nothing is changed on the normal scan aws path

vjeffrey:
anything i can do to help get this one reviewed? i've got more work following up on this but i'm a bit blocked on it til this goes in

motor/discovery/aws/mql_assets.go
```diff
@@ -0,0 +1,191 @@
...
+	value, err := mqlExecutor.Exec(query, map[string]*llx.Primitive{})
+	if err != nil {
+		return nil
+	}
+
+	a := []interface{}{}
+	d, _ := mapstructure.NewDecoder(&mapstructure.DecoderConfig{
+		Result: &a,
+	})
+	d.Decode(value.Value)
```
imilchev: Not sure why the linter doesn't nag about this but `Decode` can actually return an error which we completely ignore now. There is also `value.Error` that  would be set for any MQL error (I think) which we do not process. I think we should allow this function to return errors otherwise we won't notice if it does not work.

vjeffrey: pushed a commit to check the error 👍

imilchev (approved):
LGTM. Ship it :shipit:

PR #810, opened 2023-01-21
⭐️ Add azure.subscription. Move all azure resources under azure.subscription

preslavgerchev:
Introduce a new `azure.subscription` resource:
```
cnquery> azure.subscription{name state}
azure.subscription: {
  name: "Mondoo Testing"
  state: "Enabled"
}
```

All the resources are now moved from `azure.*` to `azure.subscription.*`. This lets us build a resource hierarchy more clearly. Before:
```

cnquery> azure.compute.vms{name}
azure.compute.vms: [
  0: {
    name: "Security-Team-vm-linux-1-b9qa"
  }
  1: {
    name: "Security-Team-vm-linux-1-mjau"
  }
]
```

After:
```
cnuery> azure.subscription.compute.vms{name}
azure.subscription.compute.vms: [
  0: {
    name: "Security-Team-vm-linux-1-b9qa"
  }
  1: {
    name: "Security-Team-vm-linux-1-mjau"
  }
]
```

There are aliases introduced to keep the old resources untouched to avoid breaking changes so `azure.compute.vms` will still work


resources/packs/azure/azure.lr
```diff
@@ -133,13 +193,64 @@ private azuread.serviceprincipal @defaults("id") {
 // Azure Resources
 azure {
   // Get all the resources in a subscription
-  resources() []azure.resource
+  resources() []azure.subscription.resource
   // Get all the resource groups in a subscription
-  resourceGroups() []azure.resourcegroup
+  resourceGroups() []azure.subscription.resourcegroup
+  // Get the subscription
+  subscription() azure.subscription
```
chris-rock: You are shadowing the resource. It is not deterministic behaviour?

resources/packs/azure/azure.lr
```diff
@@ -133,13 +193,64 @@ private azuread.serviceprincipal @defaults("id") {
...
+  monitor() azure.subscription.monitor
+  // Get cloud defender resources inside the subscription
+  cloudDefender() azure.subscription.cloudDefender
+  // Get all the resources in a subscription
+  resources() []azure.subscription.resource
 }
 
 // Azure Resource
-private azure.resource @defaults("id name location") {
+private azure.subscription.resource @defaults("id name location") {
```
chris-rock, line 246: The overlay only works if the `azure.subscription` resource has only the plural terms.

chris-rock (changes requested):

imilchev:
This does mean that when we introduce the individual assets (not sure if that's already present for Azure) you will face the following issue:
- you will be listing VMs with:
  ```
  azure.subscription.compute.vms
  ```
- you will be querying 1 VM with:
  ```
  azure.subscription.computeService.vm
  ```
If you try to change `azure.subscription.computeService.vm` to `azure.subscription.compute.vm` it will result in an MQL runtime error. I have the same problem for GCP resources. Still need to look into how to make MQL work with both `azure.subscription.compute.vms()` and `azure.subscription.compute.vm`

imilchev (approved):
Looks similar to what we have for GCP

chris-rock (approved):
Thank you @preslavgerchev

PR #813, opened 2023-01-23
✨ add impact modifications to proto

arlimus:
After https://github.com/mondoohq/cnquery/pull/724

These are mostly used down the line in cnspec. The reason we have them in cnquery is the same we have impact in cnquery: as a way to express the impact any query has within the greater system. You can always ignore this entire structure and leave it empty. If used, however, we want to be able to share context around the impact value, which in this case is the way it gets scored and weighted.
    
Also restructure a few more things that needed cleanup from the previous
query v2 change, moving a few methods and names around from learnings
with cnspec's policy v2.


chris-rock (approved):
Woot. So great to see this in @arlimus

PR #771, opened 2023-01-12
🌈 Add multi-asset progress bars

czunker:
Signed-off-by: Christian Zunker <christian@mondoo.com>

PR #814, opened 2023-01-23
🧹 remove num-assets cli argument

chris-rock:
revert the accidental exposure of the progress bar settings in proto, introduced in  #771


arlimus (approved):
Thank you! +1, we want to be careful with the CLI args we expose. I do think making it user-configurable in the future is helpful, but it should be hidden in the config somewhere and not cluttering up CLI args.

PR #815, opened 2023-01-23
🧹 use correct underline for headlines in cli report

chris-rock:
Aligns the underline for headlines by using the cli theme H1 and H2

**before**

```
Asset: K8s Cluster minikube
========================

Gather ClusterRole Bindings with cluster-admin Permissions:
k8s.clusterrolebindings.where: [
  0: {
    roleRef: {
      apiGroup: "rbac.authorization.k8s.io"
      kind: "ClusterRole"
      name: "cluster-admin"
    }
    subjects: [
      0: {
        apiGroup: "rbac.authorization.k8s.io"
... 22 more lines ...
```

**after**

```
Asset: K8s Cluster minikube
===========================

Gather ClusterRole Bindings with cluster-admin Permissions:
k8s.clusterrolebindings.where: [
  0: {
    roleRef: {
      apiGroup: "rbac.authorization.k8s.io"
      kind: "ClusterRole"
      name: "cluster-admin"
    }
    subjects: [
      0: {
        apiGroup: "rbac.authorization.k8s.io"
... 22 more lines ...
```

PR #817, opened 2023-01-23
⭐️ print asset error in section

chris-rock:
after #815

**before**

```
Asset: kube-system/kube-proxy
=============================


Asset: kube-system/coredns-64897985d
====================================
````

**after**

```
Asset: kube-system/coredns-64897985d
------------------------------------

error: rpc error: code = InvalidArgument desc = asset does not match any of the activated query packs

Asset: kube-system/coredns
--------------------------

error: rpc error: code = InvalidArgument desc = asset does not match any of the activated query packs
```

### DIFF
--- a/cli/reporter/cli.go
+++ b/cli/reporter/cli.go
@@ -33,7 +33,7 @@ func (r *cliReporter) print() error {
 	r.bundle = r.data.Bundle.ToMap()
 
 	if !r.isSummary {
-		r.printQueries()
+		r.printQueryData()
 	}
 
 	r.printSummary()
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cli/reporter/reporter.go [hunk 1]
```diff
@@ -11,6 +12,44 @@ import (
 	"go.mondoo.com/cnquery/shared"
 )
 
+type Format byte
```
czunker: These changes were also part of https://github.com/mondoohq/cnquery/pull/815 are now in main.

czunker (approved):
LGTM. Thanks @chris-rock

cli/reporter/reporter.go
chris-rock: I'll rebase it on latest main. Github does not update the references once main changes

PR #819, opened 2023-01-23
⭐️ GCP resources as assets P0

imilchev:
The following GCP assets are added as potential discovery options when scanning GCP:
- [x] Compute images `./cnquery shell gcp --discover compute-images`
```coffee
cnquery> platform{name title runtime kind}
platform: {
  kind: "gcp-object"
  name: "gcp-compute-image"
  runtime: "gcp"
  title: "GCP Compute Image"
}
```
- [x] GKE clusters `./cnquery shell gcp --discover gke-clusters`
```coffee
cnquery> platform{name title runtime kind}
platform: {
  kind: "gcp-object"
  name: "gcp-gke-cluster"
  runtime: "gcp"
  title: "GCP GKE Cluster"
}
```
- [x] VPC firewalls `./cnquery shell gcp --discover compute-firewalls`
```coffee
cnquery> platform{name title runtime kind}
platform: {
  kind: "gcp-object"
  name: "gcp-compute-firewall"
  runtime: "gcp"
  title: "GCP Compute Firewall"
}
```
- [x] GCS buckets `./cnquery shell gcp --discover storage-buckets`
```coffee
cnquery> platform{name title runtime kind}
platform: {
  kind: "gcp-object"
  name: "gcp-storage-bucket"
  runtime: "gcp"
  title: "GCP Storage Bucket"
}
```
- [x] VPC networks `./cnquery shell gcp --discover compute-networks`
```coffee
cnquery> platform{name title runtime kind}
platform: {
  kind: "gcp-object"
  name: "gcp-compute-network"
  runtime: "gcp"
  title: "GCP Compute Network"
}
```
- [x] VPC subnetworks`./cnquery shell gcp --discover compute-subnetworks`
```coffee
cnquery> platform{name title runtime kind}
platform: {
  kind: "gcp-object"
  name: "gcp-compute-subnetwork"
  runtime: "gcp"
  title: "GCP Compute Subnetwork"
}
```
- [x] BigQuery datasets `./cnquery shell gcp --discover bigquery-datasets`
```coffee
cnquery> platform{name title runtime kind}
platform: {
  kind: "gcp-object"
  name: "gcp-bigquery-dataset"
  runtime: "gcp"
  title: "GCP BigQuery Dataset"
}
```


motor/discovery/gcp/mql_asset_objects.go
```diff
@@ -0,0 +1,232 @@
...
+		for k, v := range tags {
+			stringLabels[k] = v.(string)
+		}
+
+		assets = append(assets, MqlObjectToAsset(project,
+			mqlObject{
+				name: name, labels: stringLabels,
+				gcpObject: gcpObject{
+					project:    project,
+					region:     zone,
```
czunker, line 163: Is this correct?

Looking at one of our clusters, it looks like the cluster is in a region, and the nodes can be distributed across zones:
image

motor/discovery/gcp/mql_assets.go
```diff
@@ -0,0 +1,174 @@
...
+
+	a := []interface{}{}
+	d, _ := mapstructure.NewDecoder(&mapstructure.DecoderConfig{
+		Result: &a,
+	})
+	d.Decode(value.Value)
+	return a
+}
+
+func GatherMQLObjects(tc *providers.Config, project string) ([]*asset.Asset, error) {
```
czunker: nit: this functions returns assets. Perhaps rename it to `GatherAssets` to something similar?

motor/discovery/gcp/discovery.go
```diff
@@ -1,6 +1,13 @@
 package gcp
 
 const (
-	DiscoveryInstances = "instances"
-	DiscoveryProjects  = "projects"
+	DiscoveryInstances          = "instances"
```
czunker, line 4: I couldn't find this as an asset. Will it be part of another PR?

resources/packs/gcp/compute.go
```diff
@@ -1111,6 +1228,50 @@ func (g *mqlGcpProjectComputeServiceSubnetwork) id() (string, error) {
 	return "gcloud.compute.subnetwork/" + id, nil
 }
 
+func (g *mqlGcpProjectComputeServiceSubnetwork) init(args *resources.Args) (*resources.Args, GcpProjectComputeServiceSubnetwork, error) {
+	if len(*args) > 2 {
```
czunker: Don't we need `> 3` here because of the region?

motor/discovery/gcp/mql_assets.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	return nil
+}
+
+func addInformationalLabels(l map[string]string, o mqlObject) map[string]string {
+	if l == nil {
+		l = make(map[string]string)
+	}
+	l[RegionLabel] = o.gcpObject.region
+	l[common.ParentId] = o.gcpObject.project
```
czunker, line 172: Is this for the relations?

I have the example of project -> network -> sub-network in my mind. There the direct parent would not be the project but the network.

motor/discovery/gcp/mql_assets.go
```diff
@@ -0,0 +1,174 @@
...
+)
+
+const RegionLabel string = "mondoo.com/region"
+
+type MqlDiscovery struct {
+	rt *resources.Runtime
+}
+
+func NewMQLAssetsDiscovery(provider *gcpprovider.Provider) (*MqlDiscovery, error) {
+	m, err := motor.New(provider)
```
preslavgerchev, line 28: I am not yet sure if this will work if you pass credentials to authenticate, I had the same trouble with Azure.

Basically, the secret in the provider is unresolved and you need to resolve this by using the credentials query that the resolver gets passed.

This is the same reason why we use `motor.NewMotorConnection` in the resolver directly instead of `motor.New`.

Tldr: Should probably be using `motor.NewMotorConnection` here

Edit: It actually depends on how you're passing the provider down to these objects. If you're using the one that you get initially from calling `motor.NewMotorConnection` then it is probably fine 🤷

motor/discovery/gcp/discovery.go
imilchev: I didn't add this. That was already there and that's why I haven't touched it yet since I am not sure what it does or whether it's used

motor/discovery/gcp/mql_asset_objects.go
imilchev: There are a lot of mismatches in the GCP API. In some places this thing is called `region`, in others `zone` and in some `location`. That's what is causing the confusion. I did notice that zone is deprecated and shouldn't be used so I will add `location` which is the new field and use that instead. `Location` in the API maps to `Region` in the UI in this case

resources/packs/gcp/compute.go
imilchev: yup, that's valid and same applied for the GKE clusters too.

motor/discovery/gcp/mql_assets.go
imilchev: I took that from #804. @vjeffrey what was your idea with that label?

vjeffrey: i use that for general grouping -- e.g. if you scan two diff aws accounts, its hard to tell what's what in the ui. this allows us to filter by a single account - so its not so much about relationships and more about grouping

motor/discovery/gcp/mql_assets.go
vjeffrey: i'm not sure i understand -- can you give an example of what you mean when you say "pass credentials to authenticate", like the command you're using/thinking of?

motor/discovery/gcp/mql_assets.go
imilchev: sounds like the parent is then just a project (GCP)/account (AWS)

motor/discovery/gcp/mql_assets.go
imilchev: I can describe that. I have this inventory file for GCP:
```yaml
apiVersion: v1
kind: Inventory
metadata:
  name: inventory
spec:
  assets:
    - name: cool-stuff
      connections:
        - backend: 13
          credentials:
            - secret_id: storage/random-bucket2/foo
              type: 6
              secret_encoding: 3
          options:
          discover:
            targets:
              - compute-images
              - projects
  vault:
    name: gcp-berglas
    type: gcp-berglas
    options:
      project_id: mondoo-dev-262313
```

That breaks with discovering assets because the secret isn't resolved by default

vjeffrey: 👍  thanks for the example!

czunker (commented):
LGTM. Thanks for all the new assets :tada:

vjeffrey:
do you have a policy ready that we can test this with?

vjeffrey:
i'm gonna give this another local test after planning this morning so we can get to merging!

vjeffrey (approved):

PR #823, opened 2023-01-23
🧹 Follow-up to fix some issues in the multi asset progressbars

czunker:

### DIFF
--- a/cli/progress/multiprogress.go
+++ b/cli/progress/multiprogress.go
@@ -29,7 +29,7 @@ const (
 	padding                  = 0
 	maxWidth                 = 80
 	defaultWidth             = 40
-	defaultProgressNumAssets = 10
+	defaultProgressNumAssets = 1
 	overallProgressIndexName = "overall"
 )
 
--- a/cli/progress/multiprogress.go
+++ b/cli/progress/multiprogress.go
@@ -196,11 +196,6 @@ func (m modelMultiProgress) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			cmds = append(cmds, cmd)
 		}
 
-		// When all progress bars are completed, quit the program
-		// Also quit the Program when there is only one progress bar and it is completed
-		if overallPercent == 1.0 || (len(m.Progress) == 1 && m.Progress[msg.Index].model.Percent() == 1.0) {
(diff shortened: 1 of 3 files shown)


czunker:
This is a follow-up to https://github.com/mondoohq/cnquery/pull/771 and addresses the latest comments.

czunker:
With the changed behavior of the overall progress bar, I also changed the text for it:

before:
100% overall 11/16 assets 5/16 errors

after:
100% overall 11/16 scanned 5/16 errored

At least to me this is more clear. Please double check.

cli/progress/multiprogress.go
```diff
@@ -174,6 +175,7 @@ func (m modelMultiProgress) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			m.Progress[msg.Index].lock.Unlock()
 		}
 
+		overallPercent := 0.0
 		if _, ok := m.Progress[overallProgressIndexName]; ok {
 			m.Progress[msg.Index].lock.Lock()
```
chris-rock: why is the lock on the index and not the overallProgressIndexName

chris-rock:
On current branch this is where the progress bar is stuck. I had to ctrl+c to close the progress bar. Then the report rendered. As you can see the progress was not getting to 100. Is that caused by the fact that errored assets are not counted into the overall progress?

Screenshot 2023-01-23 at 14 20 10

cli/progress/multiprogress.go
czunker: Good catch. Thanks. That was a Copy&Paste error.

czunker:
> On current branch this is where the progress bar is stuck. I had to ctrl+c to close the progress bar. Then the report rendered. As you can see the progress was not getting to 100. Is that caused by the fact that errored assets are not counted into the overall progress?
> 
> Screenshot 2023-01-23 at 14 20 10

I changed the code to quit the tea program from inside the multi progress bar. That created some edge cases, because there are different situations when the program should end depending on whether it is used in cnquery or cnspec.

So, now the tea program is again closed when the scan ends. This will not work as soon as we want to run concurrent scans, but I'd like to change this in a separate PR. Giving it a quick try ended the program but left the CLI in a broken state.

chris-rock (approved):
Thank you @czunker

PR #826, opened 2023-01-23
⭐️ full width progress bar

chris-rock:
- move asset name the left
- full width progress bar
- percentage + score is on the right

https://user-images.githubusercontent.com/1178413/214160738-212eff31-154d-4fc3-a2d1-6cc72335134b.mov

### DIFF
--- a/cli/components/components.go
+++ b/cli/components/components.go
@@ -1,0 +1,16 @@
+package components
+
+import (
+	"errors"
+	"os"
+
+	"golang.org/x/term"
+)
+
+func TerminalWidth(f *os.File) (int, error) {
+	w, _, err := term.GetSize(int(f.Fd()))
+	if err != nil {
+		return 0, errors.New("can't query terminal size")
+	}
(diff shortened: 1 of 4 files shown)


jaym (approved):

czunker:
I changed how the overall stats are shown:
image

The picture does not show it, but the tests did.
Let's assume you have asset names which have a shorter name than "overall" or "overall 1/1000 scanned 555/1000 errored".
So, while the scan makes progress, we would need to move the progressbars further to the right.
To not reserve to much space which might not be needed later (" 555/1000 errored"), I moved the stats to the next line.

@chris-rock / @tas50 Please check, whether this is ok.

chris-rock:
I think I would remove the "overall" string and just move it into one line.

czunker:
> I think I would remove the "overall" string and just move it into one line.

Now, it looks like this:
Screenshot from 2023-01-30 06-08-44

PR #831, opened 2023-01-25
🧹 Refactor multiprogressbar

czunker:
This PR
- fixes a race condition
- changes the way how Quit for the tea.Program is called
- Abstracts the tea.Program behind a new interface

### DIFF
--- a/cli/progress/multiprogress.go
+++ b/cli/progress/multiprogress.go
@@ -349,7 +349,10 @@ func (m modelMultiProgress) updateOverallProgress() {
 		if k == overallProgressIndexName {
 			continue
 		}
-		if m.Progress[k].Errored {
+		m.Progress[k].lock.Lock()
+		errored := m.Progress[k].Errored
+		m.Progress[k].lock.Unlock()
+		if errored {
 			continue
 		}
 		m.Progress[k].lock.Lock()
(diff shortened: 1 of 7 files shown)


cli/progress/multiprogress.go
```diff
@@ -222,63 +328,101 @@ func (m modelMultiProgress) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+}
+
+func (m *modelMultiProgress) updateOverallProgress() {
 	if _, ok := m.Progress[overallProgressIndexName]; !ok {
-		return nil
+		return
 	}
 	overallPercent := 0.0
-	m.Progress[overallProgressIndexName].lock.Lock()
+	m.lock.Lock()
```
jaym, line 363: this method could probably be simplified by locking and the top and defer unlocking right after it

cli/progress/multiprogress.go
```diff
@@ -222,63 +328,101 @@ func (m modelMultiProgress) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
 }
 
-func (m modelMultiProgress) updateOverallProgress() tea.Cmd {
+func (m *modelMultiProgress) allDone() bool {
+	finished := 0
+	for k := range m.Progress {
+		if k == overallProgressIndexName {
+			continue
+		}
+		m.lock.Lock()
```
jaym, line 337: this method could probably be simplified by locking and the top and defer unlocking right after it instead of locking and unlocking for each individual thing

jaym (approved):
The interfaces look good. I think the locking in the multi progress bar implementation can still be simplified, but that doesn't need to happen here

czunker:
> The interfaces look good. I think the locking in the multi progress bar implementation can still be simplified, but that doesn't need to happen here

Ok, thanks for the feedback. I'll create a follow-up PR so we can move forward here.

czunker:
> > The interfaces look good. I think the locking in the multi progress bar implementation can still be simplified, but that doesn't need to happen here
> 
> Ok, thanks for the feedback. I'll create a follow-up PR so we can move forward here.

The follow-up PR is https://github.com/mondoohq/cnquery/pull/835

PR #834, opened 2023-01-25
✨ add query v2 examples + fix bugs

arlimus:
After https://github.com/mondoohq/cnquery/pull/813

This finalizes any of the missing changes we needed from initial testing of the query v2 updates.

- bugfix: run packs without specifying properties (it didnt work without it)
- bugfix: rename any occurance of 'filter' to 'filters' for consistency
      (and semantic accuracy, even if users only enter one filter)


chris-rock (approved):
Very helpful to have those examples @arlimus

PR #840, opened 2023-01-26
🐛 Remove secret encoding from credentials

imilchev:
**This change does not affect anything for the user**

It seems like when we convert a secret to `Credential` we don't set the encoding for that credential. This can cause issues further down the code since the encoding will remain `undefined` otherwise. In this PR the encoding is set based on the secret's encoding


imilchev:
@chris-rock please have a look since I don't know if it was a conscious decision to not set this value

chris-rock:
I am not sure I understand why the secret would need to know that. Is there a specific bug?

imilchev:
> @chris-rock please have a look since I don't know if it was a conscious decision to not set this value

I haven't logged a bug but it's related to what I am doing here https://github.com/mondoohq/cnquery/pull/819/commits/6fac0e46ef7c21b62e38aec2d7aae7bcb770de03 Basically, the code complains with: 
```
DBG fetch secret from vault secret-id=storage/random-bucket2/foo vault="Multi Vault"
DBG could not fetch secret for motor connection error="unknown secret format: unknown secret encoding" secret-id=storage/random-bucket2/foo
DBG connecting to asset mondoo-gcp-ubuntu-2004-secure-base-20220704145254 (gcp-object)
```
this is caused by the fact that I am copying a secret after it is resolved but when it gets copied it no longer has the secret encoding set. That is missing because the line that I added in this PR is not present. The issue is fixed when the changes from this PR are merged.

chris-rock:
@imilchev and I discussed this through. There is no good reason we could find to expose the secret encoding to the credential in the first place. Therefore we decided to remove that from credentials to avoid accidental changes.

The secret encoding is still relevant since the vault implementation want to decide how the secret is encoded for the specific vault.

chris-rock:
@vjeffrey Can you have a look to make sure we do not forget any edge cases for your use cases?

vjeffrey:
👍  tested it out, everything still working as expected for the aws cases

chris-rock (approved):

PR #845, opened 2023-01-27
🐛 Check GitHub RequiredPullRequestReviews before access

czunker:
Fixes https://github.com/mondoohq/cnspec/issues/289

### DIFF
--- a/resources/packs/github/github_repo.go
+++ b/resources/packs/github/github_repo.go
@@ -448,29 +448,36 @@ func (g *mqlGithubBranch) GetProtectionRules() (interface{}, error) {
 
 	var ghDismissalRestrictions *githubDismissalRestrictions
 
-	if branchProtection.RequiredPullRequestReviews.DismissalRestrictions != nil {
-		ghDismissalRestrictions = &githubDismissalRestrictions{
-			Users: []string{},
-			Teams: []string{},
-		}
-
-		for i := range branchProtection.RequiredPullRequestReviews.DismissalRestrictions.Teams {
-			ghDismissalRestrictions.Teams = append(ghDismissalRestrictions.Teams, branchProtection.RequiredPullRequestReviews.DismissalRestrictions.Teams[i].GetName())
-		}
-		for i := range branchProtection.RequiredPullRequestReviews.DismissalRestrictions.Users {
-			ghDismissalRestrictions.Users = append(ghDismissalRestrictions.Users, branchProtection.RequiredPullRequestReviews.DismissalRestrictions.Users[i].GetLogin())
-		}
-	}
-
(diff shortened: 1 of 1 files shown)


resources/packs/github/github_repo.go
```diff
@@ -448,29 +448,36 @@ func (g *mqlGithubBranch) GetProtectionRules() (interface{}, error) {
 
 	var ghDismissalRestrictions *githubDismissalRestrictions
 
-	if branchProtection.RequiredPullRequestReviews.DismissalRestrictions != nil {
-		ghDismissalRestrictions = &githubDismissalRestrictions{
-			Users: []string{},
-			Teams: []string{},
-		}
+	rprr, err := core.JsonToDict(githubRequiredPullRequestReviews{})
```
imilchev: why do you not initialize this as `nil`? It's making the code shorter and it won't do unnecessary work of serializing stuff.
```suggestion
	var rprr map[string]interface{}
```

resources/packs/github/github_repo.go
```diff
@@ -448,29 +448,36 @@ func (g *mqlGithubBranch) GetProtectionRules() (interface{}, error) {
...
+	if branchProtection.RequiredPullRequestReviews != nil {
 
-		for i := range branchProtection.RequiredPullRequestReviews.DismissalRestrictions.Teams {
-			ghDismissalRestrictions.Teams = append(ghDismissalRestrictions.Teams, branchProtection.RequiredPullRequestReviews.DismissalRestrictions.Teams[i].GetName())
-		}
-		for i := range branchProtection.RequiredPullRequestReviews.DismissalRestrictions.Users {
-			ghDismissalRestrictions.Users = append(ghDismissalRestrictions.Users, branchProtection.RequiredPullRequestReviews.DismissalRestrictions.Users[i].GetLogin())
+		if branchProtection.RequiredPullRequestReviews.DismissalRestrictions != nil {
+			ghDismissalRestrictions = &githubDismissalRestrictions{
+				Users: []string{},
```
imilchev: nit: when possible I try to initialize slices with the correct size beforehand. This makes memory allocation a bit more efficient since the underlying array doesn't need to be extended and copied in memory.
```suggestion
				Users: make([]string{}, 0, len(branchProtection.RequiredPullRequestReviews.DismissalRestrictions.Users)),
```
This means create a slice with length 0 with an underlying array of size `len(branchProtection...)`

czunker: Changed.

resources/packs/github/github_repo.go
czunker: Good point. Thanks.

vjeffrey (approved):

PR #846, opened 2023-01-27
🧹 Hide progressbars when debug logging is enabled

czunker:

### DIFF
--- a/explorer/scan/local_scanner.go
+++ b/explorer/scan/local_scanner.go
@@ -198,7 +198,7 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstreamConf
 		orderedKeys = append(orderedKeys, assetList[i].PlatformIds[0])
 	}
 	var multiprogress progress.MultiProgress
-	if isatty.IsTerminal(os.Stdout.Fd()) && !logger.Debug {
+	if isatty.IsTerminal(os.Stdout.Fd()) && !strings.EqualFold(logger.GetLevel(), "debug") && !strings.EqualFold(logger.GetLevel(), "trace") {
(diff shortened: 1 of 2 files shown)


imilchev (approved):
Nice! Works as expected!

logger/log.go
```diff
@@ -64,8 +64,10 @@ func Set(level string) {
 		zerolog.SetGlobalLevel(zerolog.InfoLevel)
 	case "debug":
 		zerolog.SetGlobalLevel(zerolog.DebugLevel)
+		Debug = true
```
chris-rock: Using this variable has disadvantages. Settings this to true also generates a log of log files that are currently only set via `DEBUG=1`. Can we instead change `isatty.IsTerminal(os.Stdout.Fd()) && !logger.Debug` to something that checks the log-level?

czunker: I exposed the log level and check for it.

explorer/scan/local_scanner.go
```diff
@@ -198,7 +198,7 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstreamConf
 		orderedKeys = append(orderedKeys, assetList[i].PlatformIds[0])
 	}
 	var multiprogress progress.MultiProgress
-	if isatty.IsTerminal(os.Stdout.Fd()) {
+	if isatty.IsTerminal(os.Stdout.Fd()) && !strings.EqualFold(logger.GetLevel(), "debug") {
```
chris-rock: just additional challenge: also do not show it in trace :-)

czunker: Fixed.

chris-rock (approved):
Thank you @czunker  Very clean.

PR #850, opened 2023-01-27
🐛 fix mql hangs with aliased and direct resource in the same policy

jaym:
Example policy that causes things to hang
```
policies:
  - uid: testpolicy3
    name: Test Policy 3
    version: 0.10.1
    is_public: false
    scoring_system: 2
    tags:
      stability: "testing"
    specs:
      - asset_filter:
          query: |
            true == true
        scoring_queries:
          jay-scoring-query-2:
          jay-scoring-query-3:
queries:
  - uid: jay-scoring-query-2
    title: test it 1
    severity: 10
    query: |
      file("foo").content
  - uid: jay-scoring-query-3
    title: test it 2
    severity: 10
    query: |
      os.base.file("foo").content
```

The issue lies in the fact that the cache key used by some of the observer code differs from the runtime code that caches the resources. The observer thinks its already seen a resource, so it must already be computed (or will be computed). However, the runtime code cached it using the aliased name. It then tries to ask the resources cache to look up a field, which is empty because its basically treated as a different resource that needs to be computed.

### DIFF
--- a/resources/runtime.go
+++ b/resources/runtime.go
@@ -148,26 +148,26 @@ func (ctx *Runtime) createMockResource(name string, cls *ResourceCls) (ResourceT
 	return res, nil
 }
 
-func (ctx *Runtime) lookupResource(name string) (*ResourceCls, error) {
+func (ctx *Runtime) lookupResource(name string) (*ResourceCls, string, error) {
 	for {
 		r := ctx.Registry.Resources[name]
 		if r == nil {
-			return nil, errors.New("cannot find resource '" + name + "'")
+			return nil, name, errors.New("cannot find resource '" + name + "'")
 		} else if r.Factory != nil {
-			return r, nil
+			return r, name, nil
 		} else if r.Name != name {
 			// A resource was given an alias. Look up through aliases
 			name = r.Name
 		} else {
 			// We found a resource with a factory
-			return nil, errors.New("cannot find resource factory for '" + name + "'")
+			return nil, name, errors.New("cannot find resource factory for '" + name + "'")
 		}
 	}
 }
 
 // CreateResourceWithID creates a new resource instance and force it to have a certain ID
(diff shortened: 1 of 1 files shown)


imilchev:
I built a cnspec binary with this version of cnquery and I tested it with your policy. That seems to be working as I get 2 errors. However, I tried the same with the BigQuery policy we tested initially and that is still hanging. Not sure what differs in the execution though. 

Here is the policy I tried:
```yaml
policies:
  - uid: sshd-server-policy
    name: SSH Server Policy
    version: "1.0.0"
    authors:
      - name: Jane Doe
        email: jane@example.com
    tags:
      key: value
      another-key: another-value
    specs:
      - asset_filter:
          query: "true"
        scoring_queries:
          sshd-score-01:
          sshd-score-02:
queries:
  - uid: sshd-score-01
    title: Compute image name not empty
    query: |
      gcp.bigquery.datasets
    docs:
      desc: asdas
      audit: Run the `sshd -T | grep maxauthtries` command and verify that output MaxAuthTries is 4 or less
      remediation: Open your `/etc/ssh/sshd_config` and set `MaxAuthTries` to `4`.
    refs:
      - title: CIS Distribution Independent Linux
        url: https://www.cisecurity.org/benchmark/distribution_independent_linux/
  - uid: sshd-score-02
    title: Compute image name not empty
    query: |
      gcp.project.bigquery.datasets
    docs:
      desc: dsadsa
      remediation: Open your `/etc/ssh/sshd_config` and set `MaxAuthTries` to `4`.
    refs:
      - title: CIS Distribution Independent Linux
        url: https://www.cisecurity.org/benchmark/distribution_independent_linux/
```

jaym:
ok, i pushed up another commit. maybe it works now

imilchev (approved):
I tested it with a couple of different policies. The problem is fixed and it doesn't seem to break anything else (at least I didn't see any weird behaviour)

chris-rock (approved):
Thank you @jaym and @imilchev

PR #856, opened 2023-01-31
🧹 fallback to noop progress bar if fail to create progress bar

vjeffrey:
i ran into an issue where i got

`FTL failed to run scan error="failed to create progress bars: number of elements and orderedKeys must be equal"` (i think an asset wasn't being initialised correctly). 

seems like we can just log and use the noop progress bar if something fails

### DIFF
--- a/explorer/scan/local_scanner.go
+++ b/explorer/scan/local_scanner.go
@@ -201,7 +201,8 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstreamConf
 	if isatty.IsTerminal(os.Stdout.Fd()) && !strings.EqualFold(logger.GetLevel(), "debug") && !strings.EqualFold(logger.GetLevel(), "trace") {
 		multiprogress, err = progress.NewMultiProgressBars(progressBarElements, orderedKeys)
 		if err != nil {
-			return nil, false, errors.Wrap(err, "failed to create progress bars")
+			log.Error().Err(err).Msg("failed to create progress bars; using noop progress bar")
+			multiprogress = progress.NoopMultiProgressBars{}
 		}
 	} else {
 		// TODO: adjust naming
--- a/motor/discovery/aws/resolver.go
+++ b/motor/discovery/aws/resolver.go
@@ -107,7 +107,7 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 		log.Debug().Int("instances", len(assetList)).Msg("completed ssm instance search")
 		for i := range assetList {
 			a := assetList[i]
-			if resolvedRoot != nil {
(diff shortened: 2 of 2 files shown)


explorer/scan/local_scanner.go
```diff
@@ -201,7 +201,8 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstreamConf
 	if isatty.IsTerminal(os.Stdout.Fd()) && !strings.EqualFold(logger.GetLevel(), "debug") && !strings.EqualFold(logger.GetLevel(), "trace") {
 		multiprogress, err = progress.NewMultiProgressBars(progressBarElements, orderedKeys)
 		if err != nil {
-			return nil, false, errors.Wrap(err, "failed to create progress bars")
+			log.Error().Err(err).Msg("failed to create progress bars; using noop progress bar")
+			multiprogress = progress.NoopMultiProgressBars{}
```
chris-rock, line 205: We should figure out why this error happened

motor/discovery/aws/resolver.go
```diff
@@ -107,7 +107,7 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 		log.Debug().Int("instances", len(assetList)).Msg("completed ssm instance search")
 		for i := range assetList {
 			a := assetList[i]
-			if resolvedRoot != nil {
+			if resolvedRoot != nil && a != nil {
```
chris-rock, line 110: This looks like a good improvement, but we should figure out why we have empty assets in the list in the first place. We should not have nil assets in the asset list.

explorer/scan/local_scanner.go
czunker: This error happened because of the problem with the asset.
I pass two lists to the progress bars. The first list contains the assets. The second list contains the keys in which order the progress bars should be shown. A key is a platformID: https://github.com/mondoohq/cnquery/blob/main/explorer/scan/local_scanner.go#L195

czunker: @vjeffrey Could you please add the command, which triggered the error? I'd give it a try to reproduce it.

czunker: This is another way to prevent the error: https://github.com/mondoohq/cnquery/pull/860/files
this way, we would at least get the progress bars for valid assets.

But it also boils down to the question below, why we have the nil asset.

vjeffrey: i think i know why, there was a nil ecs cluster that i had picked up, i'll work on a fix for that

vjeffrey: i'm pushing a fix to https://github.com/mondoohq/cnquery/pull/857

PR #858, opened 2023-01-31
🐛 GCP GKE cluster resource for clusters that are still provisioning

imilchev:
For GCP GKE clusters that are still provisioning some of the properties are not set. When trying to access them cnquery panics. This PR makes sure it accesses these fields only if they are set


chris-rock (approved):
Thank you @imilchev

PR #859, opened 2023-01-31
🐛 Fix issue with GCP storage bucket init

imilchev:
Since location was not set properly when scanning storage bucket assets, the resource was initialized as empty


chris-rock (approved):
Thank you @imilchev

PR #860, opened 2023-01-31
🐛 Check for empty asset/platformID

czunker:
Check before we add invalid data to the progress bars.

### DIFF
--- a/explorer/scan/local_scanner.go
+++ b/explorer/scan/local_scanner.go
@@ -194,6 +194,9 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstreamConf
 	progressBarElements := map[string]string{}
 	orderedKeys := []string{}
 	for i := range assetList {
+		if assetList[i] == nil || len(assetList[i].PlatformIds) == 0 {
+			continue
+		}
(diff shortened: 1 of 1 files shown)


czunker:
This draft is part of a discussion in https://github.com/mondoohq/cnquery/pull/856. If we use this, we would also need to implement it in cnspec.

explorer/scan/local_scanner.go
```diff
@@ -194,6 +194,9 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstreamConf
 	progressBarElements := map[string]string{}
 	orderedKeys := []string{}
 	for i := range assetList {
+		if assetList[i] == nil || len(assetList[i].PlatformIds) == 0 {
```
chris-rock, line 197: I would argue this needs to happen in the `im.GetAssets()` function. Having nil, also hints at a problem in the discovery code. Therefore we may need to add additional error handling when assets are added to the inventory.

imilchev: I agree with this. It seems strange to me that an asset can be `nil` or that it can have no platform ID. It seems like we are trying to "fix" discovery bugs with this code

PR #863, opened 2023-01-31
🐛 ensure we do not send double assets in

vjeffrey:


motor/discovery/aws/resolver.go
```diff
@@ -195,7 +195,17 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 		}
 	}
 
-	return resolved, nil
+	assetMap := make(map[string]*asset.Asset)
+	// ensure we don't return the same asset twice
+	for i := range resolved {
+		assetMap[resolved[i].PlatformIds[0]] = resolved[i]
+	}
+	new := []*asset.Asset{}
```
imilchev: nit: you could pre-initialize the array to the correct size here
```suggestion
	new := make([]*asset.Asset{}, 0, len(assetMap))
```

imilchev (approved):

PR #861, opened 2023-01-31
Renames terraform-configuration to terraform-hcl

scottford-io:
This PR renames `terraform-configuration` to `terraform-hcl`. The platform family will still rename `terraform` but now we will distinguish scans as assets between `terraform-hcl`, `terraform-plan`, and `terraform-state`.

## Test local Terraform working directory:

```
~/go/bin/cnquery shell terraform ~/go/src/github.com/scottford-io/cnspec-platform-engineering-demo-gcp/terraform 
→ no Mondoo configuration file provided. using defaults
→ discover related assets for 1 asset(s)
→ resolved assets resolved-assets=1
  ___ _ __   __ _ _   _  ___ _ __ _   _ 
 / __| '_ \ / _` | | | |/ _ \ '__| | | |
| (__| | | | (_| | |_| |  __/ |  | |_| |
 \___|_| |_|\__, |\__,_|\___|_|   \__, |
  mondoo™      |_|                |___/  interactive shell

cnquery> asset {*}
asset: {
  kind: "code"
  build: ""
  version: ""
  title: "Terraform HCL"
  arch: ""
  labels: {}
  platform: "terraform-hcl"
  vulnerabilityReport: mondoo upstream configuration is missing
  runtime: "terraform-hcl"
  ids: [
    0: "//platformid.api.mondoo.app/runtime/terraform/hash/b9635df010d8d404a8408660502e3f93a5ec608e21ef40f507458c6b7593f939"
  ]
  name: "Terraform Static Analysis directory terraform"
  fqdn: ""
  family: []
}
```

### Test queries

```
cnquery> terraform.resources
terraform.resources: [
  0: terraform.block id = terraform.block//Users/scottford/go/src/github.com/scottford-io/cnspec-platform-engineering-demo-gcp/terraform/main.tf/1/1
  1: terraform.block id = terraform.block//Users/scottford/go/src/github.com/scottford-io/cnspec-platform-engineering-demo-gcp/terraform/main.tf/12/1
]
cnquery> terraform.resources {*}
terraform.resources: [
  0: {
    type: "resource"
    end: terraform.fileposition id = file.position//Users/scottford/go/src/github.com/scottford-io/cnspec-platform-engineering-demo-gcp/terraform/main.tf/1/1
    labels: [
      0: "random_pet"
      1: "name"
    ]
    nameLabel: "random_pet"
    snippet: "     1 | resource \"random_pet\" \"name\" {}
     2 | 
     3 | locals {
     4 |   team              = var.team
"
[...]

PR #866, opened 2023-02-01
⭐️ introduce terraform family 

chris-rock:
different platforms `terraform-hcl`, `terraform-plan` and `terraform-state` also share the same family now

followup on  #861


scottford-io (approved):
🌟 @chris-rock Great add!

PR #868, opened 2023-02-01
🧹 extract arista into its own resource pack

chris-rock:
No functional changes, it just extracts the arista resoures out of the os resource pack since they are actually quite different than other operating systems.


imilchev (approved):
The code changes look good. Not sure how to test the specific resources but I see the code is just moved so I assume they work

PR #869, opened 2023-02-01
🧹 extract ipmi resources in its own resource pack

chris-rock:
after #868

No functional changes, it just extracts the ipmi resoures out of the core resource pack.


czunker (approved):
LGTM. Thanks @chris-rock

PR #870, opened 2023-02-01
✨ Allow caching credentials in memory 

imilchev:
Implemented a credentials resolver to replace the `GetCredentials` function that is being passed around. The resolver allows for caching credentials such that they don't need to be retrieved multiple times

### DIFF
--- a/motor/discovery/common/credentials.go
+++ b/motor/discovery/common/credentials.go
@@ -9,7 +9,7 @@ import (
 
 type (
 	// CredentialFn retrieves the credentials to connect to the platform
-	CredentialFn func(cred *vault.Credential) (*vault.Credential, error)
+	// CredentialFn func(cred *vault.Credential) (*vault.Credential, error)
 	// QuerySecretFn is used during discovery phase to identify a secret for an asset
 	QuerySecretFn func(a *asset.Asset) (*vault.Credential, error)
 )
--- a/motor/discovery/common/credentials.go
+++ b/motor/discovery/common/credentials.go
@@ -8,8 +8,6 @@ import (
 )
 
 type (
-	// CredentialFn retrieves the credentials to connect to the platform
-	// CredentialFn func(cred *vault.Credential) (*vault.Credential, error)
 	// QuerySecretFn is used during discovery phase to identify a secret for an asset
 	QuerySecretFn func(a *asset.Asset) (*vault.Credential, error)
 )
--- a/motor/inventory/manager.go
+++ b/motor/inventory/manager.go
@@ -63,6 +63,13 @@ func WithVault(v vault.Vault) Option {
 	}
 }
 
+func WithCachedCredsResolver() Option {
+	return func(im *inventoryManager) error {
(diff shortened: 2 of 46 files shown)


motor/inventory/manager.go
```diff
@@ -232,8 +185,8 @@ func (im *inventoryManager) QuerySecretId(a *asset.Asset) (*vault.Credential, er
 	return im.credentialQueryRunner.Run(a)
 }
 
-func (im *inventoryManager) Resolve(ctx context.Context) map[*asset.Asset]error {
-	resolvedAssets := discovery.ResolveAssets(ctx, im.assetList, im.GetCredential, im.QuerySecretId)
+func (im *inventoryManager) Resolve(ctx context.Context, credsResolver credentials_resolver.Resolver) map[*asset.Asset]error {
```
chris-rock: I am wondering if we should set the `credsResover` in `func New(opts ...Option) (*inventoryManager, error)`. By using the option we could always set the default resolver which makes calling the Resolve function easier.

chris-rock (commented):
This is a great addition. I am just wondering if we should set the credentials resolver when we create a new instance of the inventory. This would simplify the logic and avoid the setup spread.

motor/inventory/manager.go
imilchev: done

motor/discovery/common/credentials.go
```diff
@@ -9,7 +9,7 @@ import (
 
 type (
 	// CredentialFn retrieves the credentials to connect to the platform
-	CredentialFn func(cred *vault.Credential) (*vault.Credential, error)
+	// CredentialFn func(cred *vault.Credential) (*vault.Credential, error)
```
chris-rock: if we do not need this anymore lets remove it

motor/vault/credentials_resolver/cached_vault.go
```diff
@@ -0,0 +1,41 @@
+package credentials_resolver
+
+import (
+	"context"
+
+	"go.mondoo.com/cnquery/motor/vault"
+)
+
+type cachedVault struct {
```
chris-rock: The cached vault should move into `vault/cache` package since it can be used independently from the resolver

motor/vault/credentials_resolver/credentials_resolver.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"context"
+	"errors"
+
+	"github.com/rs/zerolog/log"
+	"go.mondoo.com/cnquery/motor/vault"
+)
+
+type Resolver interface {
```
chris-rock: Interface should move into the vault package into vault.go

chris-rock (commented):
Added a few more things that I saw with fresh eyes

motor/vault/credentials_resolver/credentials_resolver.go
imilchev: done

chris-rock (approved):
Thank you @imilchev

PR #872, opened 2023-02-01
🧹 Unify GCP compute instance discovery with the rest of the GCP asset discovery

imilchev:
**This change does not affect anything user-facing**

BUT

it does make sure that we no longer discover instances that are not running. Since we currently do ssh scanning on them it makes no sense to discover them. This is also what AWS does for EC2 instances. When we enable cloud scanning + ssh for a single asset we can adjust that logic.

The goal is to remove the code that does instance discovery at the moment since it doesn't fit the approach that is used for discovering any of the other GCP assets.

```
./cnquery scan gcp --discover instances 
```


motor/discovery/gcp/mql_asset_objects.go
```diff
@@ -230,3 +313,39 @@ func bigQueryDatasets(m *MqlDiscovery, project string, tc *providers.Config) []*
...
+		return asset.State_STATE_TERMINATED
+	default:
+		log.Warn().Str("state", state).Msg("unknown gcp instance state")
+		return asset.State_STATE_UNKNOWN
+	}
+}
+
+func disksContainWindows(disks []interface{}) bool {
+	for _, d := range disks {
+		feats := d.(map[string]interface{})["guestOsFeatures"].([]interface{})
```
imilchev: this kind of stuff is pretty nasty. Especially, if we need to be checking some deeply nested structure. I guess it's rarely needed on an asset level but still good to improve at some point

motor/discovery/gcp/mql_asset_objects.go
```diff
@@ -38,6 +45,82 @@ func getTitleFamily(o gcpObject) (gcpObjectPlatformInfo, error) {
 	return gcpObjectPlatformInfo{}, errors.Newf("missing runtime info for gcp object service %s type %s", o.service, o.objectType)
 }
 
+func computeInstances(m *MqlDiscovery, project string, tc *providers.Config, sfn common.QuerySecretFn) []*asset.Asset {
+	assets := []*asset.Asset{}
+	instances := m.GetList("return gcp.project.compute.instances { id name labels zone status networkInterfaces disks { guestOsFeatures } }")
```
imilchev: Apparently this thing `m.GetList` silently ignores errors. If you mess up the query it will just return an empty list and you won't even notice it actually does not work.
cc @vjeffrey

motor/discovery/gcp/mql_asset_objects.go
imilchev: This is fixed with #891 When that gets merged the problem will be gone

motor/discovery/gcp/resolver_project.go
```diff
@@ -103,32 +101,32 @@ func (r *GcpProjectResolver) Resolve(ctx context.Context, tc *providers.Config,
 		}
 	}
 
-	// discover compute instances
-	if tc.IncludesOneOfDiscoveryTarget(common.DiscoveryAll, DiscoveryInstances) {
-		client, err := provider.Client(compute.ComputeReadonlyScope)
-		if err != nil {
-			return nil, errors.Wrap(err, "use `gcloud auth application-default login` to authenticate locally")
-		}
+	// // discover compute instances
```
czunker: Why not completely remove this?

motor/discovery/gcp/mql_asset_objects.go
```diff
@@ -38,9 +45,92 @@ func getTitleFamily(o gcpObject) (gcpObjectPlatformInfo, error) {
...
+
+		tags := b["labels"].(map[string]interface{})
+		zone := b["zone"].(map[string]interface{})
+		zoneName := zone["name"].(string)
+		status := b["status"].(string)
+		stringLabels := make(map[string]string)
+		for k, v := range tags {
+			stringLabels[k] = v.(string)
+		}
+		stringLabels["mondoo.com/instance"] = id
```
czunker: nit: Other labels have a `const` in `motor/discovery/gcp/consts.go`. Would it help to also have one for this label?

motor/discovery/gcp/mql_asset_objects.go
```diff
@@ -38,9 +45,92 @@ func getTitleFamily(o gcpObject) (gcpObjectPlatformInfo, error) {
...
+				},
+			}, tc)
+		a.State = mapInstanceStatus(status)
+		a.Platform.Kind = providers.Kind_KIND_VIRTUAL_MACHINE
+		a.Platform.Runtime = providers.RUNTIME_GCP_COMPUTE
+		a.Connections = connections
+		// find the secret reference for the asset
+		common.EnrichAssetWithSecrets(a, sfn)
+
+		if a.State == asset.State_STATE_RUNNING {
```
czunker: Would it be possible to check this earlier in the code, so we can skip all the loops and data conversion when the instance isn't in a desired state?

motor/discovery/gcp/mql_asset_objects.go
```diff
@@ -38,9 +45,92 @@ func getTitleFamily(o gcpObject) (gcpObjectPlatformInfo, error) {
 	return gcpObjectPlatformInfo{}, errors.Newf("missing runtime info for gcp object service %s type %s", o.service, o.objectType)
 }
 
-func computeImages(m *MqlDiscovery, project string, tc *providers.Config) []*asset.Asset {
+func computeInstances(m *MqlDiscovery, project string, tc *providers.Config, sfn common.QuerySecretFn) ([]*asset.Asset, error) {
 	assets := []*asset.Asset{}
-	images := m.GetList("return gcp.project.compute.images { id name labels }")
+
+	instances, err := m.GetList("return gcp.project.compute.instances { id name labels zone { name } status networkInterfaces disks { guestOsFeatures } }")
```
vjeffrey: i think you could filter by status in the mql query, to only have to process the running ones

imilchev: That's a good idea. I'll do that.. completely forgot about the power of MQL

motor/discovery/gcp/mql_asset_objects.go
imilchev: Yeah, this one seemed to be a global label (not GCP-specific) but I will add it to that list too

imilchev:
I merged #891 here

czunker (approved):
LGTM. Thanks @imilchev

PR #876, opened 2023-02-02
🧹Drop deprecated fields in azure.sql.server resource.

preslavgerchev:
Drop deprecated azure fields that are no longer used


tas50:
Can we make sure to wait for v8 for this one

preslavgerchev:
> Can we make sure to wait for v8 for this one

Sounds good. I will fix the title a bit so we wait on it

resources/packs/azure/azure.lr
```diff
@@ -834,10 +828,6 @@ private azure.subscription.sqlService.database @defaults("id name") {
   elasticPoolName string
   // SQL database default secondary location
   defaultSecondaryLocation string
-  // Deprecated
-  serviceTierAdvisors() dict
-  // Deprecated. Use `advisor.recommendedActions` instead.
```
vjeffrey, line 839: none of our policies reference these fields, right?

vjeffrey (approved):

tas50:
Before we pull the trigger here can we double check on the process we wanted to take for deprecation. I opened a ticket to get formal deprecation going and Dom had mentioned removing this all in v9 after we have a deprecation going

tas50:
If it's broken we should go ahead and remove it. If a user could use it then we should deprecate it so they know to update.

resources/packs/azure/azure.lr
preslavgerchev: nope, these fields have been like this (not returning a value, only errors) for some time now.
I also double checked the policies so we're good

preslavgerchev:
I am going to go ahead and merge that, these fields havent been working for a while now so I think we're safe to cut them

PR #877, opened 2023-02-02
✨ Load ssh keys from home dir as a final ssh fallback

imilchev:
The last resort for loading ssh keys at the moment was ssh-agent. However, if that loads no keys then we stop. This PR also attempts to load the keys located in `~/.ssh` (if they exist).

### DIFF
--- a/motor/providers/ssh/session.go
+++ b/motor/providers/ssh/session.go
@@ -4,10 +4,13 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
+	"path/filepath"
 	"strconv"
 
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/cockroachdb/errors"
+	homedir "github.com/mitchellh/go-homedir"
 	"github.com/rs/zerolog/log"
 	"go.mondoo.com/cnquery/motor/providers"
 	"go.mondoo.com/cnquery/motor/providers/ssh/awsinstanceconnect"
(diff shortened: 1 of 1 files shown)


motor/providers/ssh/session.go
```diff
@@ -201,14 +204,48 @@ func prepareConnection(pCfg *providers.Config) ([]ssh.AuthMethod, []io.Closer, e
 		}
 	}
 
-	if len(sshSigners) > 0 {
```
imilchev, line 204: I moved this below loading the ssh-agent certs since anything that is appended to `sshSigners` after that is simply ignored. I feel like this was a bug.

motor/providers/ssh/session.go
```diff
@@ -201,14 +204,48 @@ func prepareConnection(pCfg *providers.Config) ([]ssh.AuthMethod, []io.Closer, e
...
-		auths = append(auths, ssh.PublicKeys(sshSigners...))
-	}
-
 	// if no credential was provided, fallback to ssh-agent and ssh-config
 	if len(pCfg.Credentials) == 0 {
 		sshSigners = append(sshSigners, signers.GetSignersFromSSHAgent()...)
 	}
 
+	// If no keys were loaded so far, then attempt loading the default keys in the home directory
+	if len(sshSigners) == 0 {
```
chris-rock, line 213: This is adding more implicit behavior. We should add a new credential type for ssh config and only load it when the credential type is set. The cli behaviour should stay implicit and therefore we need to add ssh_config and ssh_agent as default there.

motor/providers/ssh/session.go
chris-rock: This should be correct here. The ssh agents authentication is activated in the cli if the user has not provided anything. Therefore we can remove the fallback to ssh-agent here.

chris-rock (changes requested):

motor/providers/ssh/session.go
imilchev: so this code is not needed at all then?

motor/providers/ssh/session.go
imilchev: I am not sure whether ssh_config makes sense as a name. This is just using your default ssh settings for your user.

Issue #691, opened 2022-12-29
ports.listening not showing tls params under windows


atomic111:
Windows IIS is configured with port 443 and tls.

if you query the port with tls it shows no tls params:

```
ports.listening.where( port == 443 ) { tls { * } }
```

![iis-tls](https://user-images.githubusercontent.com/7220740/209968855-e6175207-ad89-430c-969a-799aec129934.png)

joelddiaz:
This should be addressed now that #878 has merged. Either building from source, or waiting for the next cnquery release should include the necessary changes.

PR #875, opened 2023-02-02
🐛 handle processes listening on ipv6

joelddiaz:
The process/ports list will show a process listening on "::". Trying to perform TLS testing against this "hostname" fails with a "too many colons in address" error. Detect this and format the host in an ipv6-friendly way.

having an environment where we have a process serving up TLS over ipv6 would show this problem:
```
[jdiaz@fedora ~]$ cnquery run local -c "ports.listening.where(port == 443)"
→ discover related assets for 1 asset(s)
→ resolved assets resolved-assets=1
Query encountered errors:
1 error occurred:
        * cannot read related processess: permission denied
ports.listening.where: [
  0: port port=443 protocol="tcp" address="0.0.0.0" process.executable=no data available
  1: port port=443 protocol="tcp" address="::" process.executable=no data available
]

### now list the ipv6 entries TLS certs and see that it is empty
[jdiaz@fedora ~]$ cnquery run local -c "ports.listening.where(port == 443)[1]{ tls.certificates.length }"
→ discover related assets for 1 asset(s)
→ resolved assets resolved-assets=1
ports.listening.where[1]: {
  tls.certificates.length: 0
}
```

With the changes in this PR, we now get our TLS certs data:

```
[jdiaz@fedora cnquery (joel/ipv6-tls *$%)]$ ./cnquery run local -c "ports.listening.where(port == 443)[1]{ tls.certificates.length }"
→ discover related assets for 1 asset(s)
→ resolved assets resolved-assets=1
ports.listening.where[1]: {
  tls.certificates.length: 1
}
```

Additionally, there were issues with the ipv6 address parsing, so detect little-endian, and do the byte swapping needed to convert the output of /proc/net/tcp6 to big-endian/network-endian bytes before reading it as an ipv6 address.

PR #878, opened 2023-02-02
🐛 pass down a golang-recognized protocol for TLS discovery

joelddiaz:
ipv4 and ipv6 are not recognized as protocols when being parsed through
golang's net package. While something like 'ip6:tcp' is accepted, it
then causes further problems when opening the socket connection.

Just set the protocol to tcp as what happens on Linux.

But!, change the linux protocol field when parting /proc/net to match
what we do in other OSes (namely setting ipv4 and ipv6 as the protocol
when appropriate).

Before this PR's changes, you can see no TLS certificate date fetched
from an ipv4 endpoint (a non-Windows/Darwin-specific fix - that also affects
Windows/Darwin - on ipv6 is fixed in #875).
```
>c:\Users\canef\cnquery.exe run local -c "ports.listening.where(port == 11443)[1]{ tls.certificates.length }"
ports.listening.where[1]: {
  tls.certificates.length: 0
}
```

Now we can fetch TLS cert data:
```
>cnquery-fixed.exe run local -c "ports.listening.where(port == 11443)[1]{ tls.certificates.length }"
ports.listening.where[1]: {
  tls.certificates.length: 2
}
```

### DIFF
--- a/resources/packs/core/tls.go
+++ b/resources/packs/core/tls.go
@@ -177,6 +177,11 @@ func (s *mqlTls) GetParams(socket Socket, domainName string) (map[string]interfa
 	if err != nil {
 		return nil, err
 	}
+	// Windows and Darwin will list a port's "protocol" as "ipv4/6" which
+	// is an unrecognized protocol for golang's net package.
+	if proto == "ipv6" || proto == "ipv4" {
+		proto = "tcp"
+	}
 
 	tester := tlsshake.New(proto, domainName, host, int(port))
 	if err := tester.Test(tlsshake.DefaultScanConfig()); err != nil {
--- a/resources/packs/core/tls.go
+++ b/resources/packs/core/tls.go
@@ -177,8 +177,8 @@ func (s *mqlTls) GetParams(socket Socket, domainName string) (map[string]interfa
 	if err != nil {
 		return nil, err
 	}
-	// Windows and Darwin will list a port's "protocol" as "ipv4/6" which
-	// is an unrecognized protocol for golang's net package.
+
+	// a proto of ipv6 or ipv4 isn't usable by the golang net package
 	if proto == "ipv6" || proto == "ipv4" {
 		proto = "tcp"
 	}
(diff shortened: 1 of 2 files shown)


joelddiaz:
This PR should resolve #691

joelddiaz:
This PR builds on top of #875 , and should be held until 875 merges.

resources/packs/core/tls.go
```diff
@@ -177,6 +177,11 @@ func (s *mqlTls) GetParams(socket Socket, domainName string) (map[string]interfa
 	if err != nil {
 		return nil, err
 	}
+	// Windows and Darwin will list a port's "protocol" as "ipv4/6" which
+	// is an unrecognized protocol for golang's net package.
+	if proto == "ipv6" || proto == "ipv4" {
```
czunker, line 182: We already have another place, where we handle this specific case: https://github.com/mondoohq/cnquery/blob/main/resources/packs/core/port.go#L680

Perhaps, we can also handle this problem there? I think, close to L680 is the point where this value originates.

czunker: I found the lines, where the protocol is set.

mac: https://github.com/mondoohq/cnquery/blob/main/resources/packs/core/port.go#L669

windows: https://github.com/mondoohq/cnquery/blob/main/resources/packs/core/port.go#L593

For Linux it is only set to tcp or udp: https://github.com/mondoohq/cnquery/blob/main/resources/packs/core/port.go#L477

joelddiaz: I added a commit that changes the linux protocol field to be set to ipv4 or ipv6 similar to how things show up in Windows/MacOS. Still leaving the change of protocol to "tcp" when going down the TLS codepath, since the rest of the TLS code is not written to work with anything except for "tcp" AFAICT (maybe it makes sense to not even check the protocol, and just set it to "tcp" no mater what...).

czunker: Agree. This seems like the rest of the code was intended.

czunker (approved):

PR #879, opened 2023-02-03
🧹 Change GCP root asset when scanning projects to gcp-project

imilchev:
Make sure the root asset has platform kind `gcp-object` and platform name `gcp-project`. This should be merged only after we have updated the existing policies to match the new asset filter

### DIFF
--- a/motor/discovery/gcp/resolver_project.go
+++ b/motor/discovery/gcp/resolver_project.go
@@ -73,6 +73,7 @@ func (r *GcpProjectResolver) Resolve(ctx context.Context, tc *providers.Config, 
 
 	var resolvedRoot *asset.Asset
 	if tc.IncludesOneOfDiscoveryTarget(common.DiscoveryAuto, common.DiscoveryAll, DiscoveryProjects) {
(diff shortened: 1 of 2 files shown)


vjeffrey:
https://github.com/mondoohq/cnspec-policies/pull/117

motor/discovery/gcp/resolver_project.go
```diff
@@ -73,6 +73,7 @@ func (r *GcpProjectResolver) Resolve(ctx context.Context, tc *providers.Config,
 
 	var resolvedRoot *asset.Asset
 	if tc.IncludesOneOfDiscoveryTarget(common.DiscoveryAuto, common.DiscoveryAll, DiscoveryProjects) {
+		pf.Name = "gcp-project"
```
chris-rock, line 76: This should really not be done here

imilchev: The plan was to fix it after all assets are migrated (basically when a gcp asset no longer exists). Currently, it is still returned for a gcp org. Once that is fixed I can refactor the platform code

PR #880, opened 2023-02-03
⭐️ add aws-ecs id-detector, ensure the iddetectors get passed thru

vjeffrey:
adding aws-ecs id-detector to allow for ecs container id detection when running a local scan on an ecs container.

also, it looks like we dropped support for passing thru the user-id-detector on the local scan path at least at some point, it's back in there now

`scan local --id-detector aws-ecs`

### DIFF
--- a/motor/motorid/awsecs/metadata_cmd.go
+++ b/motor/motorid/awsecs/metadata_cmd.go
@@ -1,0 +1,80 @@
+package awsecsid
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/cockroachdb/errors"
+	"github.com/rs/zerolog/log"
+	"go.mondoo.com/cnquery/motor/motorid/containerid"
+	"go.mondoo.com/cnquery/motor/platform"
+	"go.mondoo.com/cnquery/motor/providers/os"
+)
+
+const (
+	identityUrl = "${ECS_CONTAINER_METADATA_URI_V4}"
+)
+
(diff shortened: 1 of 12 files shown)


motor/motorid/clouddetect/providers/aws/aws.go
```diff
@@ -57,15 +58,31 @@ func Detect(provider os.OperatingSystemProvider, p *platform.Platform) (string,
 				return "", "", nil
 			}
 			id, err := mdsvc.Identify()
-			if err != nil {
+			if err == nil {
+				return id.InstanceID, id.InstanceName, []string{id.AccountID}
+			} else {
```
imilchev: nit: the `else` shouldn't be needed since the `if` has a return statement

motor/motorid/awsecs/metadata_cmd.go
```diff
@@ -0,0 +1,80 @@
...
+	data, err := ioutil.ReadAll(cmd.Stdout)
+	if err != nil {
+		return "", err
+	}
+
+	return strings.TrimSpace(string(data)), nil
+}
+
+func (m *ContainerMetadata) containerIdentityDocument() (string, error) {
+	return m.curlDocument(identityUrl)
```
jaym, line 73: you're relying on RunCommand being run through a shell. Will that always be the case?

jaym (approved):

motor/motorid/awsecs/metadata_cmd.go
vjeffrey: i think so, at least for the expected case. because to start the session on the container aws run command opens up an interactive shell

PR #885, opened 2023-02-05
🧹 semantically split up mquery Merge and AddBase

arlimus:
The current call only allowed the overwriting of the query that was trying to merge with a base. This is problematic in places where we do not want to alter the original query. This is not the case in cnquery, but it is the case in cnspec (which will be updated shortly after this merged to accomodate the change for v8).


chris-rock (approved):
Thank you @arlimus

PR #891, opened 2023-02-06
🧹 Refactor GCP GetList to return typed results

imilchev:
**Does not change anything user-facing**
This makes our code simpler since we no longer need to cast every field we query.

### DIFF
--- a/motor/discovery/gcp/mql_asset_objects.go
+++ b/motor/discovery/gcp/mql_asset_objects.go
@@ -1,8 +1,6 @@
 package gcp
 
 import (
-	"encoding/json"
-
 	"github.com/cockroachdb/errors"
 	"github.com/rs/zerolog/log"
 	"go.mondoo.com/cnquery/motor/asset"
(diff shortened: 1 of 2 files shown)


vjeffrey (approved):
nice!

motor/discovery/gcp/mql_asset_objects.go
```diff
@@ -47,49 +45,49 @@ func getTitleFamily(o gcpObject) (gcpObjectPlatformInfo, error) {
...
 			continue
 		}
 
-		var networkIfaces []*compute.NetworkInterface
-		if err := json.Unmarshal(data, &networkIfaces); err != nil {
-			log.Error().Msgf("failed to unmarshal network interfaces for gcp compute instance %s", name)
-			continue
-		}
+		stringLabels := i.Labels
+		stringLabels["mondoo.com/instance"] = i.Id
```
vjeffrey, line 84: i know it's not new in this pr, but should this label key be a const?

imilchev: yup, i fixed it in the base branch

PR #899, opened 2023-02-07
🐛Make inventoryManager implement the InventoryManager interface

jaym:
We also don't expose public things on inventoryManager and instead expose a method on the interface

### DIFF
--- a/motor/inventory/manager.go
+++ b/motor/inventory/manager.go
@@ -16,6 +16,8 @@ import (
 	"go.mondoo.com/cnquery/motor/vault/multivault"
 )
 
+var _ InventoryManager = (*inventoryManager)(nil)
+
 type InventoryManager interface {
 	// GetAssets returns all assets under management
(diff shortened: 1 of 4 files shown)


motor/inventory/manager.go
```diff
@@ -65,7 +68,7 @@ func WithVault(v vault.Vault) Option {
 
 func WithCachedCredsResolver() Option {
 	return func(im *inventoryManager) error {
-		im.CredsResolver = credentials_resolver.New(im.GetVault(), false)
```
jaym, line 68: This branch has enableCaching always false. Not sure what's going on

chris-rock: That does not make any sense. But we should find occurrences of WithCachedCredsResolver and remove them for now

chris-rock: and set isCached to true here

chris-rock (approved):
Thank you @jaym

PR #916, opened 2023-02-09
Small text improvements to azure provider cmd.

preslavgerchev:
- Remove the `pfx` word from the error we return if the certificate cannot be parsed, we do not know what's the format of the file from where the data came in
 - Add .pem as a supported format for the `--certificate-path` arg of `azure shell`


vjeffrey (approved):

tas50:
Great fix. Thanks @preslavgerchev

PR #920, opened 2023-02-10
🐛Add scopes to fetching org units and roles for a google workspace. Make report resource private

preslavgerchev:
This PR fixes 2 issues:

1. Missing scopes for fetching org units and roles.
Before:
image
After:
image
Before:
image
After:
image

2. Make all `googleworkspace.report` resources private. Querying `googleworkspace{*}` before would lead up to this:
image
After making all the `report` resources private, this no longer shows up when doing `googleworkspace{*}`.


preslavgerchev:
I checked our docs, these scopes are already in there, we were just missing requesting them in the code.

tas50 (approved):

PR #922, opened 2023-02-10
🐛 jobpool and aws creds error handling

joelddiaz:
First commit tries to handle the case where the jobpool wasn't completely set up b/c of a permissions/credentials error (for example when building the job pool and hitting https://github.com/mondoohq/cnquery/blob/main/resources/packs/aws/aws_vpcs.go#L58 ). Just check the job function for nil before calling it.

Second commit doesn't try to bury the initial credentials errors, and just errors early before we get further down the discovery codepaths.

### DIFF
--- a/motor/providers/aws/provider.go
+++ b/motor/providers/aws/provider.go
@@ -94,14 +94,18 @@ func New(pCfg *providers.Config, opts ...ProviderOption) (*Provider, error) {
 	// gather information about the aws account
 	identity, err := CheckIam(t.config)
 	if err != nil {
+		log.Debug().Err(err).Msg("could not gather details of AWS account")
 		// try with govcloud region
 		t.config.Region = "us-gov-west-1"
 		identity, err = CheckIam(t.config)
 		if err != nil {
-			log.Warn().Err(err).Msg("could not gather details of AWS account")
-			// do not error since this break with localstack
+			log.Debug().Err(err).Msg("could not gather details of AWS account")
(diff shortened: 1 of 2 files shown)


motor/providers/aws/provider.go
```diff
@@ -94,14 +94,18 @@ func New(pCfg *providers.Config, opts ...ProviderOption) (*Provider, error) {
 	// gather information about the aws account
 	identity, err := CheckIam(t.config)
 	if err != nil {
+		log.Warn().Err(err).Msg("could not gather details of AWS account")
```
vjeffrey: i think we're going to double warn with this?  
basically the logic here is 
no region was provided -> try us-east-1 -> that doesn't work, then try gov region. (a region is required by the api) invalid credentials won't work in either case

vjeffrey: we could probably just remove both warns if we're returning the err

joelddiaz: yeah, i wasn't sure how best to handle the two errors. it would definitely show two log messages, but i figured there was the remote possibility that the error for us-east-1 was different than govcloud (maybe some DNS or something for one but an actual credentials error for the other...).

what do you think about burying the messages behind debug, and returning the "last" error encountered?

vjeffrey: 👍  that sounds good!

vjeffrey (approved):

PR #924, opened 2023-02-12
⭐️ structured errors for asset results

chris-rock:
This change introduces structured errors for asset results and allows us show n/a for assets with no applicable query pack.

**before**
Screenshot 2023-02-12 at 22 46 03
Screenshot 2023-02-12 at 22 46 09

**after**

Screenshot 2023-02-12 at 22 44 48
Screenshot 2023-02-12 at 22 44 58


cli/progress/multiprogress.go
```diff
@@ -504,7 +564,15 @@ func (m *modelMultiProgress) View() string {
 			stats += fmt.Sprintf(" %d/%d errored", erroredAssets, len(m.Progress)-1)
 		}
 
-		pad := strings.Repeat(" ", m.maxNameWidth-len(stats))
+		if notApplicableAssets > 0 {
+			stats += fmt.Sprintf(" %d/%d n/a", notApplicableAssets, len(m.Progress)-1)
```
czunker, line 568: nit: Here, we have "n/a" in lowercase, for the individual progress bar, it's "N/A". I think we should decide on lower or upper case, but not use both.

cli/progress/multiprogress_test.go
```diff
@@ -225,12 +228,42 @@ func TestMultiProgressBarOrdering(t *testing.T) {
...
+		multiprogress.Score("3", "U")
+		multiprogress.NotApplicable("3")
+	}()
+	err = multiprogress.Open()
+	require.NoError(t, err)
+
+	assert.Contains(t, buf.String(), "test1                           ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% score: F")
+	assert.Contains(t, buf.String(), "test2                           ───────────────────────────────────    X score: X")
+	assert.Contains(t, buf.String(), "test3                           ───────────────────────────────────  N/A score: U")
+	assert.Contains(t, buf.String(), "1/3 scanned 1/3 errored 1/3 n/a ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100%")
```
czunker, line 268: Now, that I see this output and independent of this PR:
Should we perhaps change "scanned" to "scored"? Also not 100% correct, as "X" is also a score, but perhaps more intuitive?

cli/reporter/reporter.go
```diff
@@ -76,6 +78,7 @@ func New(typ string) (*Reporter, error) {
 }
 
 func (r *Reporter) Print(data *explorer.ReportCollection, out io.Writer) error {
+	logger.DebugDumpYAML("report_collection", data)
```
czunker, line 81: Left-over from debugging?

explorer/error_status.go
```diff
@@ -0,0 +1,121 @@
...
+	case NotApplicable:
+		return "NotApplicable"
+	case NoQueries:
+		return "NoQueries"
+	default:
+		return "Unknown"
+	}
+}
+
+// cnquery codes start at 100 to avoid conflicts with GRPC and Ranger RPC codes
```
czunker, line 43: Copy&Paste error? I do not see the 100 below in the `const` block.

czunker (commented):
Nice! Let's you focus on the real errors. Thanks @chris-rock 

Tested it locally and works like a charm.

cli/progress/multiprogress.go
chris-rock: I agree, lets go with lower case n/a

explorer/error_status.go
chris-rock: This comment is correct, it refers to https://github.com/mondoohq/ranger-rpc/blob/main/codes/codes.go. I want to avoid conflicting error codes. Updated the comment to refer to the ranger codes.

cli/reporter/reporter.go
chris-rock: This is intentional. It allows us to debug the reporting. It only writes down the reports when DEBUG=1 is used.

cli/progress/multiprogress_test.go
chris-rock: I think scored is not the correct term for cnquery. Lets change this as soon as we find a better term

chris-rock:
@czunker ready for rereview

czunker (approved):
Thanks @chris-rock

PR #930, opened 2023-02-13
🐛 Fix arch detection of scratch containers

czunker:
Before this change, we couldn't show an architecture for running containers build from scratch images:
```
cnquery run container 1ff13294e780 -c 'platform{ name arch }'
platform: {
  arch: ""
  name: "scratch"
}
```
The reason was, we couldn't execute any command or read any file inside the container.

But we can query the docker daemon for the architecture:
```
cnquery run container 1ff13294e780 -c 'platform{ name arch }'
platform: {
  arch: "amd64"
  name: "scratch"
}
```


imilchev (approved):
Nice!

PR #931, opened 2023-02-14
🐛 fix empty iddetector

arlimus:
Without this change, the default value for `iddetector` can be `""` empty string, which is then handed down all the way to `motor/motorid`, where it then fails, because it falls through the empty check and then tries to run with an empty string ID-detector, not finding anything.

Fixes the issue introduced in e806cb11033f058b7c584bea8389fafd07bbc28c.


vjeffrey (approved):
🤦‍♀️ thank you!

PR #933, opened 2023-02-14
🧹Compress windows amd64 binary with upx

jaym:
This takes the size from 200MB to 50MB. It doesn't work for arm.

Some downsides may be that the binaries now take longer to start. Its also possible that this is more likely to trigger antivirus software.


chris-rock:
An alternative would be if we had an option to compress MSI more then this found not affect the binary but still reduce the download size

arlimus (approved):
let's make it tiny!

PR #934, opened 2023-02-14
🌙 Add GCP compute address MQL resource

### LINKED ISSUE #871: Retrieve GCP subnetworks for a network via MQL
Currently, the GCP subnetworks that are retrieved for a network are being retrieved via the GCP API. This means that we will be calling the API many times when resolving the subnetworks. See [here](https://github.com/mondoohq/cnquery/blob/main/resources/packs/gcp/compute.go#L1093-L1111)

**Describe the solution you'd like**
A better approach is to get all the GCP subnetworks via MQL. This would allow us to cache them so all subsequent calls will retrieve them from the cache. This is already done for other resources so the approach is consistent with the rest of the implementation.

imilchev:
`./cnquery run gcp -c "gcp.project.compute.addresses"`:
```coffee
gcp.project.compute.addresses: [
  0: gcp.project.computeService.address name="gke-mondoo-gke-cluster-2-c255f8bc-73b71c8f-pe" address="10.10.0.2" addressType="INTERNAL"
  1: gcp.project.computeService.address name="nexus-globalservices" address="10.0.0.19" addressType="INTERNAL"
  2: gcp.project.computeService.address name="mondoo-cloud-loadbalancer-ip" address="***" addressType="EXTERNAL"
  3: gcp.project.computeService.address name="mondoo-gke-cluster-2-lb-api" address="***" addressType="EXTERNAL"
  4: gcp.project.computeService.address name="mondoo-gke-cluster-2-lb-console" address="***" addressType="EXTERNAL"
]
```

`./cnquery run gcp -c "gcp.project.compute.addresses[0]{*}"`:
```coffee
gcp.project.compute.addresses[0]: {
  ipv6EndpointType: ""
  created: 2022-12-15 12:45:25.62 -0800 -0800
  address: "10.10.0.2"
  network: data is not a map to auto-expand
  networkTier: "PREMIUM"
  id: "2700460578865297802"
  userUrls: [
    0: "https://www.googleapis.com/compute/v1/projects/mondoo-edge/regions/us-central1/forwardingRules/gke-mondoo-gke-cluster-2-c255f8bc-73b71c8f-pe"
  ]
  ipVersion: ""
  name: "gke-mondoo-gke-cluster-2-c255f8bc-73b71c8f-pe"
  status: "IN_USE"
  subnetworkUrl: "https://www.googleapis.com/compute/v1/projects/mondoo-edge/regions/us-central1/subnetworks/mondoo-gke-cluster-2-subnet"
  prefixLength: 0
  networkUrl: ""
  regionUrl: "https://www.googleapis.com/compute/v1/projects/mondoo-edge/regions/us-central1"
  addressType: "INTERNAL"
  purpose: "GCE_ENDPOINT"
  description: ""
  subnetwork: gcp.project.computeService.subnetwork name="mondoo-gke-cluster-2-subnet"
}
```

Also fixes #871


resources/packs/gcp/gcp.lr
```diff
@@ -172,6 +172,44 @@ private gcp.project.computeService {
...
+  // Address description
+  description string
+  // IP version
+  ipVersion string
+  // Endpoint type
+  ipv6EndpointType string
+  // Address name
+  name string
+  // URL of the network in which to reserve the address
+  networkUrl string
```
chris-rock, line 198: I would expect that `.network()` returns the nextwork resource

resources/packs/gcp/gcp.lr
```diff
@@ -172,6 +172,44 @@ private gcp.project.computeService {
...
+  // Prefix length if the resource represents an IP range
+  prefixLength int
+  // Address purpose
+  purpose string
+  // Region URL
+  regionUrl string
+  // Address status
+  status string
+  // URL of the subnetwork in which to reserve the address
+  subnetworkUrl string
```
chris-rock, line 212: I would expect that `subnetwork()` would return the subnetwork resource

resources/packs/gcp/gcp.lr
```diff
@@ -172,6 +172,44 @@ private gcp.project.computeService {
...
+  // URLs of the resources that are using this address
+  userUrls []string
```
chris-rock: What is the type here. `userUrl` is confusing to me. Is it returning different resource types or just a list of one type?

resources/packs/gcp/gcp.lr
imilchev: I have not added these intentionally for now since we don't have the use-case for it. It's missing on a lot of the resources, not just on that one

resources/packs/gcp/gcp.lr
imilchev: Same for this. Do we want this included already?

resources/packs/gcp/gcp.lr
imilchev: Yes, this returns a list of resource URLs. It can be mixed types according to the GCP documentation

resources/packs/gcp/gcp.lr
chris-rock: I do not understand why it is complicated to add this. This should be just five lines since all we need to init the resource is in the url

resources/packs/gcp/gcp.lr
chris-rock: Then it should be called `resourcesUrls` or probably better `resourceIds`

imilchev: That will deviate from the GCP API. They called it `users`. I try to have the properties closely related to the original API because then people can cross-reference them with the GCP docs if needed. I can rename that but not sure if it's a good idea

resources/packs/gcp/gcp.lr
imilchev: It was not a straight-forward implementation. There were other open issues that needed to be addressed first. Here's the commit that fixes that https://github.com/mondoohq/cnquery/pull/934/commits/abcce78c7ae8d90934c103240733ef76ecbb18c8

PR #937, opened 2023-02-15
🧹 remove v8 platform deprecations

arlimus:
Release is replaced by Version, so remove it.


motor/platform/platform.proto
```diff
@@ -7,8 +7,6 @@ option go_package = "go.mondoo.com/cnquery/motor/platform";
 
 message Platform {
   string name = 1;
-  // FIXME: remove in v8.0 vv
```
arlimus, line 10: removed due to this indicator, the rest is cleanup

mql/mql_test.go
```diff
@@ -63,12 +67,12 @@ func TestMqlSimple(t *testing.T) {
 		query     string
 		assertion interface{}
 	}{
-		{"platform.name", "arch"},
-		{"platform { name release }", map[string]interface{}{
-			"name":    "arch",
-			"release": "",
+		{"asset.platform", "arch"},
```
arlimus, line 70: we have been switching from platform to asset in v7; this is coming late

mql/mql_test.go
```diff
@@ -63,12 +67,12 @@ func TestMqlSimple(t *testing.T) {
...
+		{"asset { platform version }", map[string]interface{}{
+			"platform": "arch",
+			"version":  "",
 		}},
-		{"users.list { name uid }", []interface{}{
+		{"users { name uid }", []interface{}{
```
arlimus, line 75: another leftover v7 change that was missing

motor/platform/platform.proto
jaym: does this get stored as proto somewhere. Do you need to mark 2 as unusable?

jaym (approved):

PR #941, opened 2023-02-15
 cnquery can read config file from aws ssm parameter store

vjeffrey:
```
[15/02/23 12:27:02] ❯ go run apps/cnspec/cnspec.go scan aws --config aws-ssm-ps://region/us-east-1/parameter/MondooAgentConfig
→ loaded configuration from aws-ssm-ps://region/us-east-1/parameter/MondooAgentConfig using source --config
```

cnspec can read the service account credentials stored in AWS ssm parameter store

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -173,6 +173,15 @@ func initConfig() {
 	} else {
 		Source = "default"
 	}
+	if strings.HasPrefix(Path, AWS_SSM_PARAMETERSTORE_PREFIX) {
+		err := loadAwsSSMParameterStore(Path)
+		if err != nil {
+			LoadedConfig = false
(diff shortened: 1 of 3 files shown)


cli/config/config.go
```diff
@@ -173,6 +173,17 @@ func initConfig() {
 	} else {
 		Source = "default"
 	}
+	if len(Path) > 0 {
```
arlimus: nit: do we need to check `len(Path) > 0`? wouldn't the next statement suffice

arlimus (approved):
thank you!

cli/config/config.go
vjeffrey: oh, good point, we can remove that

PR #943, opened 2023-02-16
🧹Parse okta profile correctly. Use OKTA_CLIENT_TOKEN env var over --token if set

preslavgerchev:
The okta provider now supports using the `OKTA_CLIENT_TOKEN` env var to get a token. This makes the --token arg optional.

Using an env var, token is optional
```
OKTA_CLIENT_TOKEN=<token> cnquery shell okta --organization <org>
```
Using a token
```
cnquery shell okta --organization <org> --token <token>
```

Using both, the env var will have precedence
```
OKTA_CLIENT_TOKEN=<token> cnquery shell okta --organization <org> --token <token>
```

Furthermore, this PR also fixes retrieving an okta user profile by making sure it's a correct dictionary, allowing queries on it.
Before:
```
cnquery> okta.users{profile}
okta.users: [
  0: {
    profile: &map[email:preslav@mondoo.com firstName:Preslav lastName:Gerchev login:preslav@mondoo.com mobilePhone:<nil> secondEmail:<nil>]
  }
]
```

After:
```
cnquery> okta.users{profile}
okta.users: [
  0: {
    profile: {
      email: "preslav@mondoo.com"
      firstName: "Preslav"
      lastName: "Gerchev"
      login: "preslav@mondoo.com"
      mobilePhone: null
      secondEmail: null
    }
  }
]
```

PR #944, opened 2023-02-16
Move env var check for slack to the cmd parser.

preslavgerchev:
Similar to #943, move the env var check to the command parser for Slack. Add a FIXME to remove a deprecated inventory file option.


czunker (approved):
LGTM. Thanks @preslavgerchev

PR #949, opened 2023-02-17
🧹 Add --credentials-path param for gcp/googleworkspace commands.

preslavgerchev:
This PR adds `--credentials-path` arg to both `gcp` and `googleworkspace` shell/scan commands.

Furthermore it moves out the env var reading to the commands, the provider now only works with credentials, making it easy to do a inventory-file scan.

Environment variables will have precedence over the new `--credentials-path` arg

Examples:

With env var:
```
GOOGLE_APPLICATION_CREDENTIALS=~/mondoo-dev-1111111.json cnquery shell gcp
```

With `--credentials-path`:
```
cnquery shell gcp --credentials-path ~/mondoo-dev-1111111.json
```

### DIFF
--- a/apps/cnquery/cmd/builder/parse.go
+++ b/apps/cnquery/cmd/builder/parse.go
@@ -389,6 +389,29 @@ func ParseTargetAsset(cmd *cobra.Command, args []string, providerType providers.
 	case providers.ProviderType_GCP:
 		connection.Backend = providerType
 
+		// env vars have precedence over the --credentials-path arg
+		credsPath := readEnvs("GOOGLE_APPLICATION_CREDENTIALS", "GOOGLE_CREDENTIALS")
+		if credsPath != "" {
+			serviceAccount, err := os.ReadFile(credsPath)
+			if err == nil {
+				connection.Credentials = append(connection.Credentials, &vault.Credential{
+					Type:   vault.CredentialType_json,
+					Secret: serviceAccount,
+				})
+			}
+		} else {
+			if credsPath, err := cmd.Flags().GetString("credentials-path"); err != nil {
+				log.Fatal().Err(err).Msg("cannot parse --credentials-path value")
+			} else if credsPath != "" {
+				serviceAccount, err := os.ReadFile(credsPath)
+				if err == nil {
+					connection.Credentials = append(connection.Credentials, &vault.Credential{
+						Type:   vault.CredentialType_json,
+						Secret: serviceAccount,
+					})
+				}
+			}
+		}
 		switch assetType {
 		case DefaultAssetType:
 			// deprecated, remove in v8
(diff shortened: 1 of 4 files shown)


apps/cnquery/cmd/builder/parse.go
```diff
@@ -389,6 +389,29 @@ func ParseTargetAsset(cmd *cobra.Command, args []string, providerType providers.
 	case providers.ProviderType_GCP:
 		connection.Backend = providerType
 
+		// env vars have precedence over the --credentials-path arg
+		credsPath := readEnvs("GOOGLE_APPLICATION_CREDENTIALS", "GOOGLE_CREDENTIALS")
```
preslavgerchev: I added `GOOGLE_APPLICATION_CREDENTIALS` as per the [ADC](https://cloud.google.com/docs/authentication/application-default-credentials) docs

apps/cnquery/cmd/builder/parse.go
```diff
@@ -612,14 +635,38 @@ func ParseTargetAsset(cmd *cobra.Command, args []string, providerType providers.
 		}
 	case providers.ProviderType_GOOGLE_WORKSPACE:
 		connection.Backend = providerType
+
+		// env vars have precedence over the --credentials-path arg
+		credsPath := readEnvs("GOOGLEWORKSPACE_CREDENTIALS", "GOOGLEWORKSPACE_CLOUD_KEYFILE_JSON", "GOOGLE_CREDENTIALS")
```
preslavgerchev: These env vars seem to be used for terraform (at least that's what I saw thru a quick google search). Do we want to keep them? The google docs do not mention those

cc @chris-rock @imilchev

apps/cnquery/cmd/builder/parse.go
```diff
@@ -612,14 +635,38 @@ func ParseTargetAsset(cmd *cobra.Command, args []string, providerType providers.
...
+		if credsPath != "" {
```
imilchev: this seems to be identical for google workspace and gcp. Why not extract it in a function?

apps/cnquery/cmd/builder/parse.go
chris-rock: we should only use env vars that either Google or terraform is using

apps/cnquery/cmd/builder/parse.go
chris-rock: Maybe not the full part but we could restructure it in somehthing like:

```go
var credsPaths []string
credsPaths = readEnvs("GOOGLEWORKSPACE_CREDENTIALS", "GOOGLEWORKSPACE_CLOUD_KEYFILE_JSON", "GOOGLE_CREDENTIALS"))

if credsPath, err := cmd.Flags().GetString("credentials-path"); err != nil {
    log.Fatal().Err(err).Msg("cannot parse --credentials-path value")
} else if credsPath != "" {
  credsPaths = append(credsPaths, credPath)
}

for i := range credsPaths {
   credPath := credsPaths[i]
   
  serviceAccount, err := os.ReadFile(credsPath)
  if err == nil {
    connection.Credentials = append(connection.Credentials, &vault.Credential{
	Type:   vault.CredentialType_json,
	Secret: serviceAccount,
     })
     break // abort loop since we found credential
  }
}
```

chris-rock (approved):
Great addition @preslavgerchev I added a few nit picks to make the code more readable

imilchev (approved):
I tested it and it seems to work correctly! :shipit:

PR #951, opened 2023-02-18
✨ use variable names in labels

arlimus:
Some queries can get very hard to read and understand, if there is a long preamble before the actual assertion. For example:

```
sshd.config('/etc/ssh/my_sshd_config').params["Protocol"] == 2
```

A way to simplify this is to use named variables. This allows users to write:

```
sshProto = sshd.config('/etc/ssh/my_sshd_config').params["Protocol"]
sshProto == 2
```

This is done with the intention to have the named variable show up in the result like this:

```
[failed] sshProto == 21
  expected: == 21
  actual:   "22"
```

However, up until this PR we would always bind the source of the reference, which lost track of the variable's name, and led to this:

```
[failed] sshd.config.params[Protocol] == 21
  expected: == 21
  actual:   "22"
```

This PR remedies it by retaining the original reference.

FYI @scottford-io @atomic111

### DIFF
--- a/cli/printer/printer_test.go
+++ b/cli/printer/printer_test.go
@@ -200,7 +200,7 @@ func TestPrinter(t *testing.T) {
 		},
 		{
 			"a = 3\n if(true) {\n a == 3 \n}",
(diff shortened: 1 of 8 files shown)


chris-rock (approved):
Great improvement @arlimus

PR #953, opened 2023-02-21
🐛 fix incorrect (and recursive) filter+docs unmarshal

arlimus:
1. unmarshaling into `filter.Items` doesn't make any sense, since we don't want users to set their own keys; they are computed and only used for deduplication/lookup
2. if we don't do the temporary type the `UnmarshalJSON` will be called recursively without end. This is true for both Filter and Remediation


chris-rock (approved):
Thank you @arlimus

PR #962, opened 2023-02-22
🐛 Add upstream config to run command

### LINKED ISSUE #956: `run` command is missing the upstream config
**Describe the bug**
When executing `cnquery run` with queries that require an upstream config, this config isn't present. But running the same query with `cnquery shell` works.

**To Reproduce**

1. Run `cnquery run local -c 'platform.eol'`
2. Note the error `failed to create resource 'platform.eol': mondoo upstream configuration is missing`

**Expected behavior**
`cnquery` should connect upstream and the query should succeed.

**Screenshots or CLI Output**
```
cnquery run local -c 'platform.eol' 
Query encountered errors:
failed to create resource 'platform.eol': mondoo upstream configuration is missing
platform.eol: no data available
```

In contrast to:
```
cnquery shell local 

cnquery> platform.eol
platform.eol: platform.eol date=2027-04-01 00:00:00 +0000 UTC

```

czunker:
This also adds the upstream config to all commands executed as plugin.

Fixes #956

### DIFF
--- a/apps/cnquery/cmd/plugin.go
+++ b/apps/cnquery/cmd/plugin.go
@@ -7,6 +7,8 @@ import (
 	"github.com/hashicorp/go-plugin"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
+	cnquery_config "go.mondoo.com/cnquery/apps/cnquery/cmd/config"
+	"go.mondoo.com/cnquery/cli/config"
 	"go.mondoo.com/cnquery/cli/printer"
 	"go.mondoo.com/cnquery/cli/reporter"
 	"go.mondoo.com/cnquery/cli/shell"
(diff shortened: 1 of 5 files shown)


czunker:
@arlimus / @chris-rock It works, but I'm not sure this is the correct place to integrate with https://github.com/mondoohq/cnquery/pull/71 as intended.

apps/cnquery/cmd/plugin.go
```diff
@@ -109,6 +121,20 @@ func (c *cnqueryPlugin) RunQuery(conf *proto.RunQueryConfig, out shared.OutputHe
 		out.WriteString("[")
 	}
 
+	var upstreamConfig *resources.UpstreamConfig
+	serviceAccount := opts.GetServiceCredential()
+	if serviceAccount != nil {
+		certAuth, _ := upstream.NewServiceAccountRangerPlugin(serviceAccount)
```
joelddiaz: Is it really not possible for this to error because of the ReadConfig() above, or should this potential error case be handled?

joelddiaz (commented):
Looks good. Just a question about the potential for an error codepath.

apps/cnquery/cmd/plugin.go
czunker: Nice catch. Thanks.

That was a problem in multiple commands. I added a check.

joelddiaz (approved):

PR #965, opened 2023-02-23
🧹 externalize action in proto

arlimus:
1. stop burying actions inside of mqueries; they are universal for queries, packs, and filters (not to mention policies in cnspec)
2. rename the fields to better reflect what they do (not add/delete, nothing is removed, just turned on or off; also unknown is wrong, it's really just unspecified)


chris-rock (approved):
Thank you @arlimus

PR #980, opened 2023-02-25
Add support for detecting services on Kali Linux

tas50:
It's Debian under the hood and it's been systemd since 2015.


scottford-io (approved):
Great stuff @tas50 ! 

image

PR #988, opened 2023-02-27
🧹 Use generic func to convert slices of a type to slices of interface

imilchev:
**doesn't change anything for the end user**

Just replaced the `core.StrSliceToInterface` func with a generic function that works for slices of any type. We discussed this in https://github.com/mondoohq/cnquery/pull/981#discussion_r1117954998


resources/packs/core/core.utils.go
```diff
@@ -35,15 +35,15 @@ func ToBool(b *bool) bool {
...
+	ssI := make([]interface{}, 0, len(ss))
+	for _, s := range ss {
+		ssI = append(ssI, s)
 	}
-	return res
+	return ssI
 }
 
-func PtrStrSliceToInterface(value []*string) []interface{} {
+func PtrSliceToInterfaceSlice(value []*string) []interface{} {
```
chris-rock: Wouldn't that need a type as input instead of string?

chris-rock (commented):
Great harmonization @imilchev

resources/packs/core/core.utils.go
imilchev: That's a different existing function. I haven't touched it. It looks confusing when you have a unified diff. If you compare the files side-by-side you will see the function above is the generic one. The one you are commenting on is a different func that works with slice of string pointers.

I think that use-case can be captured by the generic func as well. I will give it a try

imilchev: Yup, that works. Got rid of it!

chris-rock: Even better. I was just triggered by the rename of the function. The name indicated a generic implementation

resources/packs/core/core.utils.go
```diff
@@ -35,22 +35,12 @@ func ToBool(b *bool) bool {
 	return *b
 }
 
-func StrSliceToInterface(value []string) []interface{} {
-	res := make([]interface{}, len(value))
-	for i := range value {
-		res[i] = value[i]
+func SliceToInterfaceSlice[T any](ss []T) []interface{} {
```
chris-rock, line 39: Can we add a simple test case that checks:

- slice of strings to interface
- slide of strings with pointers and nil elements (this has worked before and nil elements where filtered, the new implementation behaves slightly different)

imilchev: I added tests. Also made sure the behaviour remains the same although I don't think it's needed. The ptr function was used only in 1 place for azure sql resources. I don't think they return nil entries in slices. It could be that this check was added during development for some testing. Anyways, now the behaviour is identical

chris-rock: If it is just one place, lets test it without the old behavior and simplify the implementation

preslavgerchev: If we want this to be a generic function, I think we should keep the nil checks in place. The signature of that function allows you to pass in an array with nils in it, so we should assume that this can happen, otherwise we're making assumptions about the consumers of that function. 

Right now it's used only in the azure stuff which *probably* will never give you back a nil pointer amongst the other stuff, but maybe 20 new API providers down the road some of those have weird APIs that also give you back nil entries in a list of pointers.

Instead, if you don't wanna keep that generic function then we can get rid of it and just do that conversion in the azure-related code directly

chris-rock:
@imilchev is that one still up to date. If we want to get this improvement in, we should rebase and add tests for the cases.

preslavgerchev:
We can probably close this?

PR #989, opened 2023-02-27
🧹 mark azure and gcp related assets

vjeffrey:
ensure that gcp and azure assets are correctly assigned as related in the console ui


imilchev:
I added a commit. We probably want the related assets to be just 1 level deep. With the current code it would add compute instances as related assets to the org. The hierarchy is org -> project -> compute instances. I would assume the project is related to the org and the compute instances are related to the project in this case

imilchev (approved):

preslavgerchev (approved):

PR #992, opened 2023-02-28
🐛 fix parsing for google workspace domain timestamps

chris-rock:
fixes an issue where we parse the time wrongly


resources/packs/googleworkspace/domain.go
```diff
@@ -30,7 +30,7 @@ func (g *mqlGoogleworkspace) GetDomains() ([]interface{}, error) {
 }
 
 func newMqlGoogleWorkspaceDomain(runtime *resources.Runtime, entry *directory.Domains) (interface{}, error) {
-	unixTimeUTC := time.Unix(entry.CreationTime, 0)
+	unixTimeUTC := time.Unix(0, entry.CreationTime*int64(time.Millisecond))
```
preslavgerchev: nit: if this is ms, `unixTimeUTC := time.UnixMilli(entry.CreationTime)` should also produce the same result

preslavgerchev (approved):

resources/packs/googleworkspace/domain.go
chris-rock: Thank you `time.UnixMilli` is much better. Love it. I adapted the code.

PR #995, opened 2023-02-28
🧹Add old_filters field, bring back is_public for v7 compatability.

preslavgerchev:
This is how the `filters` field looks like in [v7](https://github.com/mondoohq/cnquery/blob/v7/explorer/cnquery_explorer.proto#L27):
```
repeated string filters = 43; // user-defined
```

These are also used when building the asset filters, so to avoid breaking that for v7 clients, this is now brought back under the same number (43) with a different name.

I also brought back `is_public`, that is used in one place:
https://github.com/mondoohq/cnquery/blob/v7/explorer/querypack.go#L42
We can remove the `is_public`, it will just default to false, I just want to verify that this is fine, seeing as it's used to generate a content checksum.


explorer/cnquery_explorer.proto
```diff
@@ -28,15 +28,24 @@ message QueryGroup {
...
   map<string,Mquery> asset_filters = 7; // key == checksum
   // ^^
 
+  // FIXME: DEPRECATED, remove in v9.0 vv and reserve number
+  // This is replaced by the field "filters"
+  repeated string old_filters = 43;
+  // ^^
+
+  // FIXME: DEPRECATED, remove in v9.0 vv and reserve number
+  bool is_public = 20;
```
chris-rock: this should not be there and should be a reserved field

explorer/cnquery_explorer.proto
```diff
@@ -28,15 +28,24 @@ message QueryGroup {
...
 message QueryPack {
-  // FIXME: DEPRECATED, remove in v9.0 vv
+  // FIXME: DEPRECATED, remove in v9.0 vv and reserve number
   // This is replaced by the field "filters"
   map<string,Mquery> asset_filters = 7; // key == checksum
   // ^^
 
+  // FIXME: DEPRECATED, remove in v9.0 vv and reserve number
+  // This is replaced by the field "filters"
+  repeated string old_filters = 43;
```
chris-rock: I think this should be called `deprecated_filters`

chris-rock (commented):
Thank you @preslavgerchev for aligning the fields

explorer/cnquery_explorer.proto
preslavgerchev: much better, thank you

chris-rock (approved):
Thank you @preslavgerchev

PR #1002, opened 2023-03-01
🐛 align the ids for remediations with the old struct

chris-rock:
We accidentally switched the type and kept the proto id


explorer/cnquery_explorer.proto
```diff
@@ -185,10 +185,12 @@ message QueryPackDocs {
 }
 
 message MqueryDocs {
+  reserved 3;
```
preslavgerchev, line 188: do v7 clients expect that field being filled in? if so, we should make a deprecated field here with the old type that we can fill in when sending back mquery docs.

jaym (approved):

explorer/cnquery_explorer.proto
chris-rock: remediation is not important for the client

PR #1005, opened 2023-03-01
🧹 add v7 proto for backwards-compatibility

chris-rock:


arlimus (approved):
lgtm, thank you! let's carry it for the next major version for backwards compatibility with v7 :)

PR #1007, opened 2023-03-02
🐛 Scan images only once per digest

czunker:
This deduplicates images in a registry by their digest. Images with multiple tags, but the same digest will only be scanned once.

It stores all tags found for the image with the same digest.
In the UI, it looks like this:
image

Fixes https://github.com/mondoohq/cnspec/issues/399


tas50:
From the product side this looks really great. Fixes a bug and actually improves the product experience while doing so.

czunker:
Failed test because of https://github.com/mondoohq/cnquery/issues/1026

chris-rock (approved):
Thank you @czunker

PR #1010, opened 2023-03-02
🧹 Make Impact value and weight nullable

jaym:
Zero is a valid value for these. When you use zero, they get serialized as `{}` because the proto struct says these are `omitempty`. At this point, you cannot distinguish between zero and unset


arlimus:
After discussion with Jay we agreed to modify this:
1. impact value stays the way it is in the PR
2. weight value has been reverted and instead replaced with a new action: ignore. Unlike deactivate (which turns the execution off), ignore simply indicates that any scores from this query are ignored (only relevant for cnspec, but needs to be set here, since actions are defined here).

arlimus (approved):
lgtm, i made a few adjustments as discussed

explorer/mquery.go
```diff
@@ -324,33 +324,43 @@ func (v *Impact) Merge(base *Impact) {
...
+
 	if v.Scoring == Impact_SCORING_UNSPECIFIED {
 		v.Scoring = base.Scoring
 	}
+	if v.Value == nil {
+		v.Value = base.Value
+	}
+	if v.Weight < 1 {
+		v.Value = base.Value
+	}
```
arlimus, line 336: these had been missing from merge ^^

explorer/mquery.go
```diff
@@ -324,33 +324,53 @@ func (v *Impact) Merge(base *Impact) {
...
 	var res int32
-	v.Value = -1
-	v.Weight = -1
+	if err := json.Unmarshal(data, &res); err == nil {
+		v.Value = &ImpactValue{Value: res}
+		return nil
+	}
 
+	type tmp Impact
+	return json.Unmarshal(data, (*tmp)(v))
```
arlimus, line 347: @jaym this part is still necessary for straightforward unmarshaling e.g. `impact: 70` in yaml. The complex struct (impact value + weight + scoring) is an advanced variant with more fine-grained control.

PR #1011, opened 2023-03-02
🧹 add pass in labels to aws discovery

vjeffrey:
follow up to the discovery refactor i just did, we need the pass in labels to make sure we add in the integration mrn label when discovering from the context of lambda

### DIFF
--- a/motor/discovery/aws/ec2_instances.go
+++ b/motor/discovery/aws/ec2_instances.go
@@ -26,6 +26,7 @@ type Ec2Instances struct {
 	mqlDiscovery   *MqlDiscovery
 	providerConfig *providers.Config
 	account        string
+	PassInLabels   map[string]string
 }
 
 type Ec2InstanceState int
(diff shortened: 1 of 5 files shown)


preslavgerchev (approved):

motor/discovery/aws/ec2_instances.go
```diff
@@ -26,6 +26,7 @@ type Ec2Instances struct {
 	mqlDiscovery   *MqlDiscovery
 	providerConfig *providers.Config
 	account        string
+	PassInLabels   map[string]string
```
chris-rock, line 29: Why is this not just called labels?

vjeffrey: named it that way because it's labels that will be passed in/added to the discovered assets. we can rename tho 🤷

chris-rock: Lets leave it at that for now

chris-rock (approved):
Thank you @vjeffrey

PR #1013, opened 2023-03-03
✨ Link serviceAccounts to multiple GCP resources

imilchev:
```typescript
cnquery> gcp.project.gke.clusters{ nodePools.map(config.serviceAccount) }
gcp.project.gke.clusters: [
  0: {
    nodePools.map: [
      0: gcp.project.iamService.serviceAccount id = 116085562322346626211
      1: gcp.project.iamService.serviceAccount id = 116085562322346626211
    ]
  }
]
```

Done the same for Dataproc clusters and CloudRun services and jobs


chris-rock:
That is the right change but breaking. @mm-weber do we use this in our policies?

imilchev:
> That is the right change but breaking. @mm-weber do we use this in our policies?

We already discussed that change with Manuel. He requested it for GKE clusters specifically but I looked up all other places where this is used and updated them as well. These properties aren't used in existing policies so we should be safe. Also the new state of the resources now makes more sense than before.

chris-rock (approved):
Thank you @imilchev

PR #1016, opened 2023-03-03
🧹 Remove adapter over msgraph-sdk-go, bump msgraph-sdk-go

preslavgerchev:
We've had a layer over the msgraph-sdk-go, which used some code that was copied directly from the library itself.

As per https://github.com/microsoftgraph/msgraph-sdk-go/issues/129#issuecomment-1453538318, it seems that this library is now improved enough that it doesn't bloat up our binary size and builds fast

This PR removes all the adapter code and bumps the msgraph-sdk to the latest production (0.56) version

The rest is mostly changing struct names where needed


chris-rock (approved):
Thank you @preslavgerchev

PR #1020, opened 2023-03-03
⭐️ python packages

joelddiaz:
Implement a new python resource that will fill out `python.packages` will the information found while searching for python packages in various standard locations (/usr/local/lib(64)/python*, /usr/lib(64)/python*, /opt/homebrew/lib/python*, C:\Python*, and a couple of darwin-specific locations).

a python.package resource should look like:

```
cnquery> python.packages[0]{ * }
python.packages[0]: {
  author: "Armin Ronacher"
  name: "Babel"
  dependencies: [
  0: python.package name="pytz" version="2022.7.1"
  ]
  summary: "Internationalization utilities"
  id: "/usr/lib/python3.11/site-packages/Babel-2.10.3-py3.11.egg-info/PKG-INFO"
  file: file path="/usr/lib/python3.11/site-packages/Babel-2.10.3-py3.11.egg-info/PKG-INFO" size=1283 permissions.string="-rw-r--r--"
  license: "BSD"
  version: "2.10.3"
}
```

There is also an effort made to identify child/leaf packages by noticing when pip marks an installed package with the empty `REQUESTED` filename. These can be found with `python.children{ * }`.

Lastly, a single python package can be parsed by pointing to the metadata file with:

```
cnquery> python.package("/home/jdiaz/.local/lib/python3.11/site-packages/python_ftp_server-1.3.17.dist-info/METADATA"){ * }
python.package: {
  version: "1.3.17"
  summary: "Command line FTP server tool designed for performance and ease of use."
  file: file path="/home/jdiaz/.local/lib/python3.11/site-packages/python_ftp_server-1.3.17.dist-info/METADATA" size=1186 permissions.string="-rw-r--r--"
  author: "Vadym Stupakov"
  dependencies: data is not a map to auto-expand
  id: "/home/jdiaz/.local/lib/python3.11/site-packages/python_ftp_server-1.3.17.dist-info/METADATA"
  license: "MIT"
  name: "python-ftp-server"
}
```

^ this will not be able to fill out dependency info as we've explicitly scanned a single file in this case.


resources/packs/python/python.lr
```diff
@@ -0,0 +1,22 @@
+
+import "../core/core.lr"
+
+// Python packages resource
```
arlimus: nit: this resource describes the current operating system's Python installation, its packages and path etc

resources/packs/python/python.lr
```diff
@@ -0,0 +1,22 @@
...
+python {
+  init(path? string)
+  path string
+  packages() []python.package
+  children() []python.package
```
arlimus: it feels weird to me to type `python.children` and know that it is about immediate packages. Semantically it could be child installations of python, possibly variants, runtimes, as well as packages. 

I do like `python.packages` for showing all installed packages

arlimus: A few ideas:
`python.packages.explicit` for explicitly installed ones (this is how pacman works in arch linux)
`python.packages.toplevel` for toplevel packages maybe
`python.packages.userinstalled` which is how dnf does it

resources/packs/python/python.lr
joelddiaz: Yes, I went and added all the comments for the auto-docs while I was there as well.

resources/packs/python/python.lr
joelddiaz: I ended up changing it to `toplevel`.

chris-rock (approved):
Super great addition @joelddiaz

PR #1031, opened 2023-03-06
🐛 Fix subscription platform id. Do not link assets to empty root.

preslavgerchev:
Assign the right platform id to subscriptions, until now it was using the wrong provider.

Removed links from root to assets. It seems that the root right now is a completely blank asset so that's of no use. The assets already have to links to the subscriptions assets so that does not change


chris-rock (commented):
Thank you @preslavgerchev

chris-rock (approved):

PR #1024, opened 2023-03-04
✨Add public IP addresses resources for a VM and for the whole subscription

preslavgerchev:
This PR adds the `azure.subscription.network.ipAddress` resource. 

Public IP addresses can now be retrieved on
  - the entire subscription, lets a user scan throughout all public IPs in the sub
 ```
cnquery> azure.subscription.network.publicIpAddresses{*}
azure.subscription.network.publicIpAddresses: [
  0: {
    name: "super-cool-public-ip"
    id: "/subscriptions/<redacted>/resourceGroups/<redacted>/providers/Microsoft.Network/publicIPAddresses/<redacted>"
    location: "uksouth"
    tags: {}
    ipAddress: "255.255.255.255"
  }
]
```

 - a specific VM, lets an user see which IPs are attached to it
 ```
cnquery> azure.subscription.compute.vms{publicIpAddresses{*}}
azure.subscription.compute.vms: [
  0: {
    publicIpAddresses: [
      0: {
        name: "super-cool-public-ip"
        id: "/subscriptions/<redacted>/resourceGroups/<redacted>/providers/Microsoft.Network/publicIPAddresses/<redacted>"
        location: "uksouth"
        tags: {}
        ipAddress: "255.255.255.255"
      }
    ]
    name: "vm-name"
  }
]
```

PR #1033, opened 2023-03-06
🧹 Use MQL to discover azure instances

preslavgerchev:
After #1031 

This builds on top of #1024, as we can now fetch all the ip addresses for a VM


motor/discovery/azure/mql_asset_objects.go
```diff
@@ -92,7 +96,19 @@ func computeInstances(m *MqlDiscovery, subscription string, tc *providers.Config
 					service:      "compute",
 					objectType:   "vm",
 				},
-			}, tc))
+			}, tc)
+		for _, ip := range vm.PublicIpAddresses {
```
vjeffrey, line 100: i think if you add the ip connections here then you'll run into weirdness later when we do resources as assets.

because one is the inside (os) scan - the ones we need the ip for, and the other side of this asset is the outside scan (configuration, from the api)

vjeffrey: i may be wrong, i'm staring at the code still. haven't tested it myself

motor/discovery/azure/resolver.go
```diff
@@ -140,18 +143,15 @@ func (r *Resolver) Resolve(ctx context.Context, root *asset.Asset, tc *providers
 				},
 			}
 			subsAssets[*sub.Sub.SubscriptionID] = subAsset
-			// if there's a root, link the sub to it
```
vjeffrey, line 143: did the related assets stuff get moved somewhere else/gets picked up elsewhere in the file?

preslavgerchev: i removed the links to the roots. right now the root is an empty azure object which brings no value. the assets are already being linked to the sub they belong to. i think we should fix the root to make it more sensible (or just have no root?) and then we can bring the root linking

motor/discovery/azure/mql_asset_objects.go
preslavgerchev: i dont fully understand the problem

do you mean that it would show up as potentially different assets?

vjeffrey: ok, sorry, new day brain, i can explain now:

- when we have inside/outside scan available for an asset, we want to leave the inside and outside scans as two different assets (we'll do some work with asset relationships to show them as one in console)
- so one asset does the configuration (outside scan)
- other asset does the os (inside) scan

so you would have two discovery targets for the compute instances, "instances" and "instances-api" (or something like that)

that said, we can also just change that later on 🤷

vjeffrey: code looks good! i'm gonna hit approve on this, we can always move the connections logic around later

vjeffrey (approved):

PR #1034, opened 2023-03-06
🧹 Make listing k8s namespaces optional

### LINKED ISSUE #1022: k8s scan tries to list all namespaces despite namespace filter
**Describe the bug**
a user reported that the api was erroring because we try to request all namespaces despite the filter. it appears we aren't passing the filters through to the api call, and are instead filtering listing all namespaces. this causes an issue for users for whom the namespace filter is more of an rbac requirement than a desire
https://github.com/mondoohq/cnquery/blob/main/motor/discovery/k8s/resolver.go#L108

imilchev:
Fixes #1022

This does not make running queries against a cluster for which we have no namespace rights possible. We need to figure out what to do with the platformID generation for a cluster. At the moment the UID of the kube-system namespace is used. However, it is not always the case that we have access to that

### DIFF
--- a/motor/discovery/k8s/resolver.go
+++ b/motor/discovery/k8s/resolver.go
@@ -17,6 +17,7 @@ import (
 	"go.mondoo.com/cnquery/motor/providers/local"
 	"go.mondoo.com/cnquery/motor/vault"
 	"go.mondoo.com/cnquery/resources/packs/os/kubectl"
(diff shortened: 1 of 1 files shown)


imilchev:
We can merge this already. However, it won't fix #1022 for that we will need a follow up. There is no need to delay this PR for now

vjeffrey (approved):

PR #1038, opened 2023-03-07
🧹 rewrite ipv6 addresses before connecting over ssh

joelddiaz:
trying to connect to an ipv6 address w/o the surrounding []s around the ipv6 address will error.

detect when we're dealing with an ipv6 address and add the []s as needed.

This will solve it for cases like when dealing with ansible inventory files with ipv6 addresses for some of the target hosts.

Specifying an ipv6 host via the CLI still would require the []s be added for example `cnquery shell ssh user@[fe80::0010]`

Fixes: https://github.com/mondoohq/cnspec/issues/403

### DIFF
--- a/apps/cnquery/cmd/builder/parse.go
+++ b/apps/cnquery/cmd/builder/parse.go
@@ -1,8 +1,10 @@
 package builder
 
 import (
+	"fmt"
 	"net/url"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -22,6 +24,32 @@ type target struct {
 	Port     int32
 }
 
+// match [leading_protocol:]//[user@]EVERYTHING_ELSE
+var ipv6Regexp = regexp.MustCompile(`(.*//)(.*@)?(.*)`)
+
+func addIPv6Brackets(uri string) string {
+	if strings.ContainsAny(uri, "[]") {
+		// already has surrounding []s
+		return uri
+	}
+
+	matched := ipv6Regexp.FindStringSubmatch(uri)
+
+	if matched == nil {
+		return uri
+	}
+
+	host := matched[3]
+
+	// if there is just one : (or less) then we're not dealing with ipv6
+	if strings.Count(host, ":") <= 1 {
+		return uri
+	}
+
+	// at this point we assume we're dealing with ipv6 w/o []s
+	return fmt.Sprintf("%s%s[%s]", matched[1], matched[2], host)
+}
+
 // parseTarget parses the specified target, which may be specified as either:
 // - [user@]hostname or
 // - a URI of the form ssh://[user@]hostname[:port]
@@ -30,6 +58,8 @@ func parseTarget(uri string) (target, error) {
 	if !strings.Contains(uri, "://") {
 		uri = "//" + uri
 	}
+
+	uri = addIPv6Brackets(uri)
 
 	u, err := url.Parse(uri)
 	if err != nil {
--- a/apps/cnquery/cmd/builder/parse_test.go
+++ b/apps/cnquery/cmd/builder/parse_test.go
@@ -36,3 +36,52 @@ func TestTargetParser(t *testing.T) {
 		assert.Equal(t, tests[i].target, res, tests[i].value)
 	}
 }
+
+func TestParseIPv6(t *testing.T) {
+	tests := []struct {
+		input    string
+		expected string
+	}{
+		{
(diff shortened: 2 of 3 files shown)


chris-rock:
I can easily login with ssh to an ipv6 target but it does not work with cnquery:

```
$ ssh chris@fd00::20c:29ff:fe8a:9da0
chris@fd00::20c:29ff:fe8a:9da0's password: 
Linux debian 5.10.0-21-arm64 #1 SMP Debian 5.10.162-1 (2023-01-21) aarch64

The programs included with the Debian GNU/Linux system are free software;
the exact distribution terms for each program are described in the
individual files in /usr/share/doc/*/copyright.

Debian GNU/Linux comes with ABSOLUTELY NO WARRANTY, to the extent
permitted by applicable law.
Last login: Tue Mar  7 05:12:33 2023 from 2002:b2be:66e7:0:5852:f1b9:65a9:1922
chris@debian:~$ exit
logout
Connection to fd00::20c:29ff:fe8a:9da0 closed.

$ cnquery shell ssh chris@fd00::20c:29ff:fe8a:9da0 --ask-pass                                                                                          x1 19h1m chris-rock/osr-dev[bbb20bf3f]
→ loaded configuration from /Users/chris/.config/mondoo/mondoo.yml using source default
Enter password:            
x cannot parse target error="parse \"//chris@fd00::20c:29ff:fe8a:9da0\": invalid port \":9da0\" after host"
x could not connect to asset error="dial tcp :22: connect: connection refused" asset=
FTL could not resolve assets
```

joelddiaz:
> I can easily login with ssh to an ipv6 target but it does not work with cnquery:

Yes, for specifying the connection right on the CLI (not through ansible inventory), you need to add the []s around the ipv6 address. To allow ipv6 w/o []s on the CLI, we would need additional parsing to be able to tell the difference between user@fe80::1234:22 and user@[fe80::1234]:22 (to be able to distinguish 22 as part of the IP and 22 as the port number

joelddiaz:
Actually, I'm not sure that we would reliably be able to distinguish between `fe80::1234:22` being an IP+port number and an ipv6 address alone w/o port information...

chris-rock:
Do yo think it would be possible to support the following schemes:

- cnquery shell ssh chris@fd00::20c:29ff:fe8a:9da0
- cnquery shell ssh chris@[fd00::20c:29ff:fe8a:9da0]:22

So essentially users have to use brackets if they want to use a port. This would be consistent with what I see being used for url https://softwareengineering.stackexchange.com/questions/311413/how-should-we-represent-an-ipv6-address-with-port-number-in-text

chris-rock:
I was able to connect to ipv6 with `[]`

```
cnquery shell ssh "chris@[fd00::20c:29ff:fe8a:9da0]" --ask-pass                                                                                     x1 
Enter password:            

cnquery> 

```

joelddiaz:
> Do yo think it would be possible to support the following schemes:
> 
>     * `cnquery shell ssh chris@fd00::20c:29ff:fe8a:9da0`
> 
>     * `cnquery shell ssh chris@[fd00::20c:29ff:fe8a:9da0]:22`
> 
> 
> So essentially users have to use brackets if they want to use a port. This would be consistent with what I see being used for url https://softwareengineering.stackexchange.com/questions/311413/how-should-we-represent-an-ipv6-address-with-port-number-in-text

What I think we could support is a kind of approach where we will take anything with more than 1 `:` in the address as an IPv6 address, and if you want to specify/override a port number, then you must add the []s.

This could end up surprising someone who is trying to connect to host `fe80::1234` on port `22`, but we can try to add some text to the error output to say something like "we see you're specifying an IPv6 address. if you want to add a port number, remember to surround the IPv6 address in []s".

I'll start working on in.

joelddiaz:
@chris-rock ok, you can try it now by using ipv6 w/o the []s as long as you don't need to specify the port information

If we see a connection error when trying to connect to an ipv6 address, we'll print some extra output to suggest that it might be related to the port being used as part of the ipv6 address:

`→ ensure proper []s when combining IPv6 with port numbers host=[2605:a601:a619:de00:d4ea:e446:e993:22] port=22`

chris-rock:
Works like a charm:

```
cnquery shell ssh chris@fd00::20c:29ff:fe8a:9da0 --ask-pass
Enter password:            

cnquery> 
```

chris-rock (approved):
Great improvement @joelddiaz

PR #1040, opened 2023-03-07
🐛 Use filters from groups

czunker:
Filters from querypacks are converted to filters inside a group when saved as a policy. when we get the policy back as a querypack, we need to query the group for filters.


arlimus:
@czunker I made a few adjustments to work with local filters for querypacks as well.

This will need a follow-up after v8 to correctly layer filters, i.e. local filters (pack.Filters) vs group filters vs computed filters. I will add a proto change to clarify the computed filters as well. Local filters will also be added to policies, since they are missing there. More to come in v8...

arlimus (approved):
Looks great :)

PR #1042, opened 2023-03-08
🧹 include pure container name on ecs container resource

vjeffrey:
we need the container name for the scan cmd


motor/discovery/aws/ecs.go
```diff
@@ -134,7 +134,7 @@ func (ecs *ECSContainers) addConnectionInfoToECSContainerAsset(asset *asset.Asse
 			Host:    publicIp,
 			Options: map[string]string{
 				"region":      region,
-				ContainerName: asset.Name,
+				ContainerName: asset.Labels[ContainerName],
```
chris-rock, line 137: What happens if asset.Labels is nil? Do we need a fallback? Why are we not setting the asset.Name to the container name in the first place?

vjeffrey: labels can't be nil here, the steps before this check that.

asset.Name is currently set to the container name + the ip, to make them uniquely recognizable, or else you end up with just the name of the container and it's hard to tell which one is which (e.g. i have 5 "fargate-app" containers running, but i know i've only set one up for the scanning)

chris-rock (approved):
Thank you @vjeffrey

PR #1046, opened 2023-03-09
postV8: 🌟 query variants + computed filters

arlimus:
1. **Separated `ComputedFilters` from `Filters` in QueryPacks.** So far we only supported setting filters, which served as both an aggregation for all filters of the pack as well as user-defined filters. This caused problems because either too many filters were stored (making it impossible to retrieve the querypack correctly) or user filters were overwritten.

   With the separation into `ComputedFilters` (just like in cnspec) it is now clear that this field is dynamically calculated and not user-defined. It gets reset at every compile step in the bundle. This behavior is in line with cnspec's Policy behavior and helps us separate user- from system-defined values.

2. **Query variants** The proto already had the fields defined, but we haven't used them yet. With this change it is now possible to use query variants. An example can be seen in the `complex.mql.yaml`. Here is the relevant portion:

   ```yaml
   - uid: uname 
     title: Collect uname info 
     variants: 
       - uid: unix-uname 
       - uid: windows-uname
       
   - uid: unix-uname 
     mql: command("uname -a").stdout 
     filters: asset.family.contains("unix")
     
   - uid: windows-uname 
     mql: command("uname").stdout 
     filters: asset.family.contains("windows") 
   ```

   Now it is possible to take one query (`uname`) and have it test
   differently against different targets. All child queries that match
   the `filters` criteria will be executed.

   Generally we aim to use this as a selector for systems that are
   supported. It is also much easier to override aggregate queries for
   entirely different target systems.


explorer/filters.go
```diff
@@ -5,6 +5,7 @@ import (
 	"sort"
 	"strconv"
 
+	"github.com/pkg/errors"
```
chris-rock: should probably just the errors package

chris-rock (approved):

explorer/filters.go
arlimus: great catch, no idea why it didn't default to it...

jaym (approved):

PR #1047, opened 2023-03-09
🧹 Catch asset problems before they reach the progressbar

czunker:
This checks whether we have duplicate platformID across the assets.

When not checked here, we would see this:
FTL failed to run scan error='could not create progress bar: number of elements and orderedKeys must be equal'

### DIFF
--- a/explorer/scan/local_scanner.go
+++ b/explorer/scan/local_scanner.go
@@ -2,6 +2,7 @@ package scan
 
 import (
 	"context"
+	"fmt"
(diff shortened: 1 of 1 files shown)


czunker:
This hopefully makes it more clear what the underlying problem in cases like https://github.com/mondoohq/cnspec/issues/399 is. We also saw this with other providers.

chris-rock (approved):
Thank you @czunker

PR #1048, opened 2023-03-09
✨Okta improvements

preslavgerchev:
- Bump okta sdk to 2.17
 - Fix `okta.trustedOrigin.scopes` parsing
 - Add `okta.role` and `user.roles`, allowing to query which roles are assigned to an user
 - Add custom logic to fetch access policy rules. The okta SDK do not provide us with the tools to extract that data, even though it's being returned from their API, there's a linked GH issue for context


vjeffrey (approved):
works great! thank you!

PR #1053, opened 2023-03-10
🧹 ensure we do not fail the whole resource when there are region perm restrictions

vjeffrey:
some users have region restrictions on different services, so that they can access us-east-1 for a specific service, but not, for example, us-west-1

a user reported running into this behavior when attempting to query their aws resources

before:
```
 [9/03/23 22:10:14] ❯ AWS_PROFILE="restricted" cnquery shell aws
→ no Mondoo configuration file provided. using defaults

cnquery> aws.cloudwatch.logGroups { * }
Query encountered errors:
job err: : could not gather aws cloudwatch log groups: operation error CloudWatch Logs: DescribeLogGroups, https response error StatusCode: 400, RequestID: 87000c84-59e8-4aae-bec2-b1caba97254d, api error AccessDeniedException: User: arn:aws:iam::172746783610:user/restricted-test is not authorized to perform: logs:DescribeLogGroups on resource: arn:aws:logs:us-east-2:172746783610:log-group::log-stream: with an explicit deny in an identity-based policy
aws.cloudwatch.logGroups: no data available
cnquery> exit
```

AFTER:
```
cnquery> aws.cloudwatch.logGroups { * }
! error accessing region for AWS API region=us-east-2
aws.cloudwatch.logGroups: [
  0: {
    kmsKey: data is not a map to auto-expand
    arn: "arn:aws:logs:eu-west-1:172746783610:log-group:/aws/lambda/MondooLambda:*"
    region: "eu-west-1"
    metricsFilters: []
    name: "/aws/lambda/MondooLambda"
  }
  1: {
    kmsKey: data is not a map to auto-expand
    arn: "arn:aws:logs:eu-central-1:172746783610:log-group:/aws/lambda/MondooLambda:*"
    region: "eu-central-1"
    metricsFilters: []
    name: "/aws/lambda/MondooLambda"
  }
...
```

### DIFF
--- a/resources/packs/aws/aws.go
+++ b/resources/packs/aws/aws.go
@@ -2,8 +2,10 @@ package aws
 
 import (
 	"errors"
+	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws/arn"
+	"github.com/aws/smithy-go/transport/http"
 	"go.mondoo.com/cnquery/motor/providers"
 	aws_provider "go.mondoo.com/cnquery/motor/providers/aws"
 	"go.mondoo.com/cnquery/resources/packs/aws/info"
@@ -51,4 +53,14 @@ func GetRegionFromArn(arnVal string) (string, error) {
 		return "", err
 	}
 	return parsedArn.Region, nil
+}
+
+func Is400AccessDeniedError(err error) bool {
+	var respErr *http.ResponseError
+	if errors.As(err, &respErr) {
+		if respErr.HTTPStatusCode() == 400 && strings.Contains(respErr.Error(), "AccessDeniedException") {
+			return true
+		}
+	}
+	return false
 }
--- a/resources/packs/aws/aws_access_analyzer.go
+++ b/resources/packs/aws/aws_access_analyzer.go
@@ -59,6 +59,10 @@ func (a *mqlAwsAccessAnalyzer) getAnalyzers(provider *aws_provider.Provider) []*
 			for nextToken != nil {
 				analyzers, err := svc.ListAnalyzers(ctx, params)
 				if err != nil {
+					if Is400AccessDeniedError(err) {
+						log.Warn().Str("region", regionVal).Msg("error accessing region for AWS API")
+						return res, nil
+					}
 					log.Error().Err(err).Str("region", regionVal).Msg("error listing analyzers")
 					return nil, err
 				}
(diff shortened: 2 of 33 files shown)


preslavgerchev:
This code lgtm but it raises a question to me on whether we may want to tackle this problem elsewhere. If we fix the `GetRegions` in the provider to only return regions we have access to then we wouldnt have to check every AWS call going through. Here's what I am thinking (all of this could be completely wrong or simply not work):
  - We fetch all regions
  - For each region we do some kind of API call that requires permissions. I think most of the AWS calls support a `DryRun` to test our permissions. If valid we add to the regions list
  - We could go as far as having an `aws.region` mql resource or something with a property that indicates if that region can be queried.

resources/packs/aws/aws.go
```diff
@@ -52,3 +54,13 @@ func GetRegionFromArn(arnVal string) (string, error) {
 	}
 	return parsedArn.Region, nil
 }
+
+func Is400AccessDeniedError(err error) bool {
+	var respErr *http.ResponseError
+	if errors.As(err, &respErr) {
+		if respErr.HTTPStatusCode() == 400 && strings.Contains(respErr.Error(), "AccessDeniedException") {
```
preslavgerchev, line 61: Bit weird it's a 400, i'd expect this to be a 403

vjeffrey:
> We fetch all regions
For each region we do some kind of API call that requires permissions

we already get the list of regions from the ec2 aws api. the problem is this isn't global region restriction. it's service specific. for example, my restricted user i tried this with cannot do anything with cloud watch log groups in us-east-2. my restricted user can do everything else just fine

preslavgerchev:
> > We fetch all regions
> > For each region we do some kind of API call that requires permissions
> 
> we already get the list of regions from the ec2 aws api, it's the aws.regions resource and the list we get from the provider.the problem is this isn't global region restriction. it's service specific. for example, my restricted user i tried this with cannot do anything with cloud watch log groups in us-east-2. my restricted user can do everything else just fine

ooh i see, thanks for explaining that

then it looks like we can't determine in advance the permissions

preslavgerchev (approved):

resources/packs/aws/aws.go
vjeffrey: ya, i think it's because access denied, not authorization denied, that its a 400

PR #1057, opened 2023-03-10
Add instances-api azure discovery target

preslavgerchev:
Problem: Trying to scan an instance right now with an azure policy does not work since the instances have an attached SSH connection. This means that a query like `azure.compute.subscription.vm{...}` will not work on the ssh provider.

To fix this, this PR introduces a new target `instances-api`.
  - `instances` is meant to instantiate the instances as proper VMs with attached SSH connection to them.
  - `instances-api` is meant to instantiate the instances as API objects, solely looking at what the azure API returns on them

Discovering both will produce 2 assets for 1 VM, one on the 'API' (external) side and one for the 'OS' (inside) side
```
➜ ~/go/bin/cnquery shell azure --discover instances-api,instances
! could not determine credentials for asset name=ms365-windows
→ resolved assets resolved-assets=2

    Available assets                       
                                           
  > 1. ms365-windows (azure-compute-vm)    
    2. ms365-windows (azure-compute-vm-api)
```

This will allow us to rewrite policies by using the new platform name for a target so that azure policies only target API instance assets.

### DIFF
--- a/motor/discovery/azure/resolver.go
+++ b/motor/discovery/azure/resolver.go
@@ -15,6 +15,12 @@ import (
 	"go.mondoo.com/cnquery/motor/vault"
 )
 
+var ResourceDiscoveryTargets = []string{
+	DiscoveryInstances, DiscoverySqlServers, DiscoveryPostgresServers,
+	DiscoveryMySqlServers, DiscoveryMariaDbServers, DiscoveryStorageAccounts,
+	DiscoveryStorageContainers, DiscoveryKeyVaults, DiscoverySecurityGroups, DiscoveryInstancesApi,
+}
+
 type Resolver struct{}
 
 func (r *Resolver) Name() string {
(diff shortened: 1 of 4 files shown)


vjeffrey:
aw man, your explanation of that is so much more clear than the way i tried to explain that last week, well done ⭐️

motor/discovery/azure/resolver.go
```diff
@@ -23,9 +29,10 @@ func (r *Resolver) Name() string {
 
 func (r *Resolver) AvailableDiscoveryTargets() []string {
 	return []string{
-		common.DiscoveryAuto, common.DiscoveryAll, DiscoverySubscriptions, DiscoveryInstances,
-		DiscoverySqlServers, DiscoveryPostgresServers, DiscoveryMySqlServers, DiscoveryMariaDbServers,
-		DiscoveryStorageAccounts, DiscoveryStorageContainers, DiscoveryKeyVaults, DiscoverySecurityGroups,
+		common.DiscoveryAuto, common.DiscoveryAll, DiscoverySubscriptions,
```
vjeffrey: i think u could replace some of this by adding 'ResourceDiscoveryTargets...' to the list

vjeffrey (approved):

PR #1058, opened 2023-03-10
reduce potential dependencies in the command builder

joelddiaz:
This should allow building a wrapper tool w/o pulling in all the real dependencies, so that the wrapper can mimmick the CLI subcommands and parameters of cnquery.

Essentially we move the functions for building the subcommands out into a separate package to allow the current cmd/builder to wrap the function that receives the ProvidersType and AssetType info around a standard cobra.Command function signature. For example:

```
func(cmd *cobra.Command, args []string) {
   callCmdBuilderFunction(cmd, args, providers/ProviderType_AZURE, defaultAssetType)
}
```

The new "slim builder" will allow building the subcommands w/o needing to pull in the `cnquery/motors/providers` package as a dependency (and avoiding all the sub-dependencies that come along with it (like azure, aws, gcp, etc...)).

### DIFF
--- a/apps/cnquery/cmd/builder/builder.go
+++ b/apps/cnquery/cmd/builder/builder.go
@@ -7,7 +7,8 @@ import (
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	"go.mondoo.com/cnquery/motor/discovery/common"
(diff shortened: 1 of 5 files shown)


arlimus (approved):
Awesome, thank you @joelddiaz !

I'll leave this note here for when we get to v9:
1. We will probably want to further clean up/structure the way the builder composes the CLI and
2. The v7 shim this is used for will be removed once we get to v9, which gives us more opportunity to optimize.

PR #1060, opened 2023-03-13
🧹 Keep our base image updated with dependabot

czunker:


czunker:
This should detect our `Dockerfile` and bump alpine from 3.16 to 3.17.

Docs: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem

czunker:
When it works here, we can add it to other repos as well.

czunker:
Superseded by #1066

PR #1071, opened 2023-03-14
⭐️ k8s namespace-scoped scanning

### LINKED ISSUE #1022: k8s scan tries to list all namespaces despite namespace filter

imilchev:
- [x] Split k8s cluster and namespace resolver
- [x] When `--namespaces` and listing namespaces is not working, attempt getting those specific namespaces
- [x] Adjust the k8s provider to be able to work with a cluster and namespace root (requires adjusting the platformID)
- [x] Make sure k8s MQL resources for a namespace root, list only the resources in that namespace (not globally)
- [x] Test with the mondoo-operator (node scan, cluster scan, admission controller)

Fixes #1022 

```
./cnquery run k8s --namespaces default -c "k8s.pods" --discover namespaces                                                                                                                                                                             
→ loaded configuration from /Users/ivanmilchev/.config/mondoo/mondoo.yml using source default
k8s.pods: [
  0: k8s.pod namespace="default" name="nginx" created=2023-03-14 16:59:07 +0100 CET
]
```

Best is to test this by creating a new k8s user and giving it access only to 1 namespace. I used [this tutorial](https://medium.com/@HoussemDellai/rbac-with-kubernetes-in-minikube-4deed658ea7b) for creating a new minikube user. You can use this role and bind it to the test user:
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: ns-admin
  namespace: default
rules:
- apiGroups:
  - '*'
  resources:
  - '*'
  verbs:
  - '*'
```

### DIFF
--- a/motor/discovery/k8s/resolver_namespace.go
+++ b/motor/discovery/k8s/resolver_namespace.go
@@ -2,6 +2,8 @@ package k8s
 
 import (
 	"context"
+	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/gobwas/glob"
@@ -67,6 +69,11 @@ func (r *NamespaceResolver) Resolve(ctx context.Context, root *asset.Asset, tc *
 	clusterNamespaces, err := p.Namespaces()
 	if err != nil {
 		if errors.IsForbidden(err) {
+			// If the user does not have permissions to list the cluster namespaces, we cannot do glob matching.
+			// We can only work with exact matching in that case
+			if containsGlob, _ := regexp.MatchString(`[\*\/\\\[\]\{\}\?]`, includeNamespaces); containsGlob {
+				return nil, fmt.Errorf("glob patterns are not allowed for k8s users with no list namespace permissions")
+			}
 			log.Warn().Msg("cannot list cluster namespaces, skipping check for non-existent namespaces...")
 		} else {
 			return nil, err
--- a/motor/discovery/k8s/list_workloads_test.go
+++ b/motor/discovery/k8s/list_workloads_test.go
@@ -139,7 +139,8 @@ func TestListCronJobs(t *testing.T) {
 	require.Equal(t, []string{k8s.NewPlatformWorkloadId(clusterIdentifier,
 		strings.ToLower(parent.Kind),
 		parent.Namespace,
-		parent.Name)},
+		parent.Name,
+		"")},
 		ownershipDir.OwnedBy(expectedAssetPlatformIds[0]))
 
 	var assetNames []string
@@ -373,7 +374,8 @@ func TestListDaemonsets(t *testing.T) {
 	require.Equal(t, []string{k8s.NewPlatformWorkloadId(clusterIdentifier,
 		strings.ToLower(parent.Kind),
 		parent.Namespace,
-		parent.Name)},
+		parent.Name,
+		"")},
 		ownershipDir.OwnedBy(expectedAssetPlatformIds[0]))
 
 	var assetNames []string
@@ -580,7 +582,8 @@ func TestListDeployments(t *testing.T) {
 	require.Equal(t, []string{k8s.NewPlatformWorkloadId(clusterIdentifier,
 		strings.ToLower(parent.Kind),
 		parent.Namespace,
-		parent.Name)},
+		parent.Name,
+		"")},
 		ownershipDir.OwnedBy(expectedAssetPlatformIds[0]))
 
 	var assetNames []string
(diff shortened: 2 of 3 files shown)


scottford-io:
@imilchev thanks for adding this. Given the flag is labeled `--namespaces`, are we going to support scanning multiple namespaces? 🤔  If not, should the flag be `--namespace` instead?

imilchev:
> @imilchev thanks for adding this. Given the flag is labeled `--namespaces`, are we going to support scanning multiple namespaces? 🤔 If not, should the flag be `--namespace` instead?

We support scanning multiple namespaces at a time. The flag has been there for a while. It just worked only if you have global cluster access. With this change it works also if you have access to just those namespaces

motor/providers/k8s/platform.go
```diff
@@ -17,3 +24,11 @@ func NewPlatformWorkloadId(clusterIdentifier, workloadType, namespace, name stri
 	platformIdentifier += "/" + workloadType + "s" + "/name/" + name
 	return platformIdentifier
 }
+
+func NewNamespacePlatformId(clusterIdentifier, name string) string {
+	if clusterIdentifier == "" {
+		return fmt.Sprintf("//platformid.api.mondoo.app/runtime/k8s/namespace/%s", name)
```
czunker, line 30: I think this will prevent us from scanning a namespace with the same name in different clusters. The platformID needs to be unique across assets.

I'm quite sure, but I'll try this out.

motor/providers/k8s/api_provider.go
```diff
@@ -201,9 +197,9 @@ func (t *apiProvider) SupportedResourceTypes() (*resources.ApiResourceIndex, err
 
 func (t *apiProvider) Resources(kind string, name string, namespace string) (*ResourceResult, error) {
 	ctx := context.Background()
-	ns := t.namespace
+	// ns := t.namespace
```
czunker: nit: We can remove this completely.

motor/discovery/k8s/resolver_namespace.go
```diff
@@ -0,0 +1,123 @@
...
+		DiscoveryClusters,
+		DiscoveryPods,
+		DiscoveryJobs,
+		DiscoveryCronJobs,
+		DiscoveryStatefulSets,
+		DiscoveryDeployments,
+		DiscoveryReplicaSets,
+		DiscoveryDaemonSets,
+		DiscoveryContainerImages,
+		DiscoveryAdmissionReviews,
```
czunker, line 42: What about these non-namespaced discovery options?

How does the namespace filter here work in combination with the operator? I'll have to double-check that.

motor/providers/k8s/platform.go
czunker: Ok, it does not error, but the assets get overridden.

image

resource version and the uuid change in the backend.

motor/discovery/k8s/resolver_namespace.go
czunker: I currently suspect this would break the admission webhook: https://github.com/mondoohq/mondoo-operator/blob/main/cmd/mondoo-operator/webhook/cmd.go#L80

I'm not sure how to test this.

motor/providers/k8s/platform.go
imilchev: Will need to add the ns UID to the platform too to make this work

motor/discovery/k8s/resolver_namespace.go
imilchev: It shouldn't break the admission controller. It doesn't use ns filtering. I can test it once we have settled on the implementation. Let's resolve all comments on this PR and as a last step before merging I will check it with the operator. Will add this to the list of tasks at the top

joelddiaz:
> @imilchev thanks for adding this. Given the flag is labeled `--namespaces`, are we going to support scanning multiple namespaces? thinking If not, should the flag be `--namespace` instead?

Should just be a comma-separated list of namespaces https://github.com/mondoohq/cnquery/blob/9177e86c4a0add14bbb8bd7fb86c532e2ca4db14/motor/discovery/k8s/resolver.go#L90-L93

motor/discovery/k8s/resolver_namespace.go
czunker: I managed to test it with a local minkube. Works as you expected. Nice!

Webhook filters objects before sending them to the scan API when they match the namespaces filter.

imilchev: I also ran a few tests and it seems to be working as expected. Waiting on @joelddiaz to triple-check 😄

czunker (approved):

PR #1073, opened 2023-03-15
Avoid sudo on self-hosted runners

benr:
Jobs on self hosted runners do not run in containers, they run on the host.  Allowing a workflow to install software or otherwise manipulate the system as root is a major security risk.  Instead lets just check that it's installed.  (Sadly, there isn't an 'if' expression to determine if we're on a shared or self-hosted runner, which would be ideal.)


benr:
This is the job failure that prompted this change, for the record: https://github.com/mondoohq/cnquery/actions/runs/4425666048/workflow

imilchev (approved):
LGTM

PR #1075, opened 2023-03-15
🐛 fix remediation parsing

arlimus:
it expected the `items` structure to be created in non-simple remediations, which is annoying. We intended for remediations to either be simple string or to be an array of remediation items (but without the need for an addition `items` field).

### DIFF
--- a/explorer/filters.go
+++ b/explorer/filters.go
@@ -139,6 +139,9 @@ func (s *Filters) RegisterChild(child *Filters) {
 	}
 }
 
+// RegisterQuery attempt to take a query (or nil) and register all its filters.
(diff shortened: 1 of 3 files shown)


explorer/filters.go
```diff
@@ -139,6 +139,9 @@ func (s *Filters) RegisterChild(child *Filters) {
 	}
 }
 
+// RegisterQuery attempt to take a query (or nil) and register all its fitlers.
```
joelddiaz: s/fitlers/filters

joelddiaz (approved):
Other than the comment typo LGTM

PR #1076, opened 2023-03-16
🧹 externalize scoring system

arlimus:
We have currently only had it built into the impact field. Elevate it as a shared enum across the proto. Nothing else changes. Downstream dependencies will be udpated. This also helps us in removing the scoring spec in cnspec (given that the impact is already here).

Please note that this will need additional work once we get to v9. Ideally we'd want to avoid having scoring bits in cnquery altogether, but it is tied to the impact field for this major version and migrating that to a better structure will have to wait for the next major release. In the meantime this is still useful - even if counter-intuitive - because it removes a source of duplication for this enum.

Additionally we migrated the methods for unmarshaling json+yaml from cnspec and created a new method to marshal to yaml which uses words (since it's user-facing content and the words are documented).


chris-rock (approved):
Great improvement @arlimus

PR #1077, opened 2023-03-16
Allow looking up query through a function

jaym:
1. Add a versatile method to add child filters with a query lookup (instead of just a map of queries). This is necessary when we don't have all queries at hand (like requesting queries outside of a bundle).
2. Rename filter helpers for more clarity. Also add docs.


arlimus (approved):
Great addition @jaym !

PR #1078, opened 2023-03-16
⭐️ allow specifying a proxy just for Mondoo API connections

joelddiaz:
Allow cnquery to be told to use a specific proxy for connections to the Mondoo API. In high-to-low order of precedence:

1) MONDOO_API_PROXY env var
2) --api-proxy CLI param (global param used for all subcommands where a connection to the Mondoo API is needed)
3) api_proxy value in config file

Update the various places in cnquery where connections are set up for the Mondoo API to use the new shared/rangerclient package.

This depends on the changes in https://github.com/mondoohq/ranger-rpc/pull/25


shared/rangerclient/client.go
```diff
@@ -0,0 +1,53 @@
...
+
+	return rangerClient, nil
+}
+
+// getMondooAPIProxy will in order of precedence use the proxy info found in
+// 1) MONDO_API_PROXY env var
+// 2) the --api-proxy CLI parameter
+// 3) the api_proxy setting in the config file
+func getMondooAPIProxy() string {
+	proxy, envSet := os.LookupEnv("MONDOO_API_PROXY")
```
jaym: [nit] this should probably be handled by viper

shared/rangerclient/client.go
```diff
@@ -0,0 +1,53 @@
...
+	if envSet {
+		return proxy
+	}
+
+	proxy = viper.GetString("api-proxy")
+	if proxy != "" {
+		return proxy
+	}
+
+	opts, optsErr := cnquery_config.ReadConfig()
```
jaym: [nit] It might be better to inject the configuration this thing needs into the constructor than to go digging around global memory. I feel like we end up reading the config multiple times if we do it here

jaym (approved):

shared/rangerclient/client.go
joelddiaz: I looked into this, but it got complicated when I tried to preserve the MONDOO_API_PROXY -> --api-proxy CLI param -> mondoo.yaml precedence.

shared/rangerclient/client.go
```diff
@@ -0,0 +1,53 @@
...
+	"go.mondoo.com/ranger-rpc"
+)
+
+type RangerClientOpts struct {
+	Proxy string
+}
+
+// NewRangerClient will set up the underlyig ranger client
+// with the appropriate proxy if needed.
+func NewRangerClient() (*http.Client, error) {
```
chris-rock: This name is strange, it is just a new http client

shared/rangerclient/client.go
```diff
@@ -0,0 +1,53 @@
+package rangerclient
```
chris-rock: this belongs purely into the `upstream` package

resources/packs/core/asset_advisories.go
```diff
@@ -39,7 +39,11 @@ func (a *mqlAsset) GetVulnerabilityReport() (interface{}, error) {
 
 	// get new advisory report
 	// start scanner client
-	scannerClient, err := newAdvisoryScannerHttpClient(mcc.ApiEndpoint, mcc.Plugins, ranger.DefaultHttpClient())
+	rangerClient, err := rangerclient.NewRangerClient()
```
chris-rock: this should not be determined here but when the mcc object is created

resources/packs/core/mondoo_eol.go
```diff
@@ -29,7 +29,11 @@ func (p *mqlMondooEol) GetDate() (*time.Time, error) {
 
 	// get new advisory report
 	// start scanner client
-	scannerClient, err := newAdvisoryScannerHttpClient(mcc.ApiEndpoint, mcc.Plugins, ranger.DefaultHttpClient())
+	rangerClient, err := rangerclient.NewRangerClient()
```
chris-rock: this should not be created here but when the mcc object is created. Then it is just `mcc.HttpClient`

explorer/services.go
```diff
@@ -41,16 +42,19 @@ func NewLocalServices(datalake DataLake, uuid string) *LocalServices {
 
 // NewRemoteServices initializes a services struct with a remote endpoint
 func NewRemoteServices(addr string, auth []ranger.ClientPlugin) (*Services, error) {
-	client := ranger.DefaultHttpClient()
```
chris-rock, line 44: This is way too much work, we should create a ranger.Client plugin that set the http client. Then we can just pass a list

shared/rangerclient/client.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	return rangerClient, nil
+}
+
+// getMondooAPIProxy will in order of precedence use the proxy info found in
+// 1) MONDO_API_PROXY env var
+// 2) the --api-proxy CLI parameter
+// 3) the api_proxy setting in the config file
+func getMondooAPIProxy() string {
```
chris-rock: It feels strange to have this here, this belongs into the apps/cnquery/cmd package

chris-rock: It should be a helper in the opts instead similar to the upstream helper

chris-rock (dismissed):
This is adding too much bloat and needs to be simplified

resources/packs/core/mondoo_eol.go
joelddiaz: gotcha...i added the httpclient so that it can be set up way up at the scan/shell/etc subcommands

apps/cnquery/cmd/login.go
```diff
@@ -63,10 +62,12 @@ func register(token string) {
 	apiEndpoint := viper.GetString("api_endpoint")
 	token = strings.TrimSpace(token)
 
-	httpClient, err := httpclient.NewClient()
+	// NOTE: login is special because we do not have a config yet
+	proxy, err := cnquery_config.GetAPIProxy()
```
joelddiaz, line 66: Avoiding reading the config file means the ansible and chef stuff will need to work a different way. I suppose we can just do something like:

`if API_PROXY != nil {
  cnspec login --api-proxy=API_PROXY ...
}

if API_PROXY == nil {
  cnspec login ...
}`

in the ansible/chef logic paths

chris-rock: So you want to read the config during login? We can do that but then I think we should properly parse the opts earlier and move the code from 125

```go
httpClient, err = opts.GetHttpClient()
		if err != nil {
			log.Fatal().Err(err).Msg("could not create http client")
		}
```

up and just use that one

joelddiaz: Because it was reading the config file, and it non-destructively adds the extra fields (after successful login), it was/is an easy way to set the api_proxy and have it stick around permanently after `login`

joelddiaz: Although, thinking about it, it is a bit surprising that `login` reads the config file. Perhaps this behavior is less surprising, and we can just add the `--api-proxy` param as part of `login`, and add the yaml `api_proxy` field right after successful login.

PR #1079, opened 2023-03-17
🧹 check for expected results after parsing process output

joelddiaz:


chris-rock:
In general we should avoid those defensive approaches since we want to find the failure quick. In this case I think it helps us to identify the missing error quicker

chris-rock (approved):

chris-rock:
related to https://github.com/mondoohq/cnspec/issues/453

PR #1080, opened 2023-03-19
🔨 filters.summarize

arlimus:
For an easier access to all filters in the struct


chris-rock (approved):
Nice addition @arlimus

PR #1086, opened 2023-03-20
🐛 handle empty COMMAND ps column

joelddiaz:
We have seen that the 'ps' output returned can have an empty field for the COMMAND column. Update the regexp to allow for the possibility of it being empty.

Extend the test case to explicitly cover this scenario.

Upgrade the checking for unexpected regexp results to a Fatal().


chris-rock (approved):
Thank you @joelddiaz

PR #1087, opened 2023-03-20
🐛 Fix the property merging in props cache

jaym:
PropsCache seems to get filled top down. So asset policy, space policy, actual policies. This means that the bases go in last. If something is in the cache, then the thing thats being merged into it is more base than what was there


arlimus (approved):
Thank you for fixing @jaym !

PR #1095, opened 2023-03-23
🐛 fix cvss aggregated score calculation

chris-rock:


atomic111 (approved):
@chris-rock thanks for the fix

PR #1100, opened 2023-03-24
🐛 Compile Query Packs from upstream

### LINKED ISSUE #1099: `cnquery scan` does not apply query packs from upstream
**Describe the bug**
When running `cnquery scan` with upstream config, the asset is not scanned against a query pack, although a matching query pack is enabled.

**To Reproduce**

1. Enable query pack (Linux, Mac, ...) in the UI
2. Run scan `cnquery scan local` with a valid upstream config
3. Note the error

**Expected behavior**
The asset should get scanned.

**Screenshots or CLI Output**
image

But:
```
asset doesn't support any policies
```

This works with cnquery v8.0.0, but fails starting with v8.1.0

**Additional context**
Running this with debug output, the empty filters array seems to be the problem:
```
DBG query.hub> fetched bundle from upstream querypack=//assets.api.mondoo.app/spaces/test-infallible-taussig-796596/assets/2NOz1z0JpLSyPV6clLDPYoCuxeg req-id=global
DBG client> got bundle
DBG client> got filters asset=//assets.api.mondoo.app/spaces/test-infallible-taussig-796596/assets/2NOz1z0JpLSyPV6clLDPYoCuxeg
DBG client> shell update filters asset=//assets.api.mondoo.app/spaces/test-infallible-taussig-796596/assets/2NOz1z0JpLSyPV6clLDPYoCuxeg
[]
```

czunker:
This populates the `ComputedFilters` field, which is later needed to match the pack for the asset.

Fixes #1099

### DIFF
--- a/explorer/query_hub.go
+++ b/explorer/query_hub.go
@@ -245,7 +245,12 @@ func (s *LocalServices) cacheUpstreamQueryPackBundle(ctx context.Context, mrn st
 		return nil, errors.New("failed to retrieve upstream query pack " + mrn + ": " + err.Error())
 	}
 
-	bundleMap := bundle.ToMap()
+	bundleMap, err := bundle.Compile(ctx)
+	if err != nil {
+		logCtx.Error().Err(err).Str("querypack", mrn).Msg("query.hub> failed to compile query pack bundle")
+		return nil, errors.New("failed to retrieve upstream query pack " + mrn + ": " + err.Error())
+	}
+
 	if err = s.setBundleFromMap(ctx, bundleMap); err != nil {
 		logCtx.Error().Err(err).Str("querypack", mrn).Msg("query.hub> failed to set query pack retrieved from upstream")
 		return nil, err


explorer/query_hub.go
```diff
@@ -245,7 +245,12 @@ func (s *LocalServices) cacheUpstreamQueryPackBundle(ctx context.Context, mrn st
 		return nil, errors.New("failed to retrieve upstream query pack " + mrn + ": " + err.Error())
 	}
 
-	bundleMap := bundle.ToMap()
+	bundleMap, err := bundle.Compile(ctx)
```
czunker, line 248: This is what we also do with default packs in case no upstream is configured.

But for cnspec, we do this server-side when loading the policies. So, is this the correct place?

arlimus: cnquery uses `GetBundle` in its datalake, which should likely be renamed to `GetValidatedBundle` to align with cnspec. That said, it `retrieves and if necessary updates the pack.`. From this definition alone, this change requires the bundle to be returned in a fully computed state - including computed filters. 

Thus we don't want to compile the bundle here. It must be returned in a compiled state.

One thing to do would be to migrate over a few of the tests that are missing to validate this behavior (which I missed when I originally split these methods out).

arlimus (changes requested):
I think this is an upstream problem because the endpoint breaks the contract. I recommend closing this PR.

PR #1102, opened 2023-03-24
Remove words in the expect file

tas50:
Moving funky stuff in all the repos into the main dictionary


chris-rock (approved):
Thank you @tas50

PR #1114, opened 2023-03-28
🐛 change viper binding from api-proxy to api_proxy

joelddiaz:
this way, when the 'login' subcommand goes to save the config file, the --api-proxy that was set via CLI param just goes along for the ride and ends up in the config file.

presently it ends up in the config file as 'api-proxy' which wouldn't be picked up when reading the config file.


chris-rock (approved):
Thank you @joelddiaz

PR #1118, opened 2023-03-30
🧹 add Ingresses to the auto-discovered k8s resources

joelddiaz:
also alphabetize the order of the list of resources that get auto-discovered


chris-rock (approved):
Thank you @joelddiaz

PR #1136, opened 2023-04-06
📄 add client-specific proxy setting documentation

misterpantz:


apps/cnquery/cmd/root.go
```diff
@@ -65,6 +65,7 @@ func init() {
 	rootCmd.PersistentFlags().BoolP("verbose", "v", false, "Enable verbose output")
 	rootCmd.PersistentFlags().String("log-level", "info", "Set log level: error, warn, info, debug, trace")
 	rootCmd.PersistentFlags().String("api-proxy", "", "Set proxy for communications with Mondoo API")
+	rootCmd.PersistentFlags().String("api-https-proxy", "", "Set proxy for communications with Mondoo Platform")
```
chris-rock, line 68: Why is that a new property? We only have `api-proxy`

chris-rock (changes requested):

apps/cnquery/cmd/root.go
misterpantz: Oh dear. I must be misunderstanding this opportunity: https://www.notion.so/mondoo/cnquery-cnspec-Client-Specific-Proxy-a0070fc63a4c44158b4044938bde4816 

Sorry to waste your time.

misterpantz: No, I re-read it and I didn't misunderstand. It looks like it just didn't get implemented as specified.

chris-rock: we decided to make it just simpler in the implementation

PR #1147, opened 2023-04-12
🧹 Remove deprecatedV5 parts

czunker:


czunker:
Looks like we missed this during one of the migrations.

chris-rock (approved):
Thank you @czunker

PR #1148, opened 2023-04-12
🐛 fix up --api-proxy handling

joelddiaz:
When doing something like:

cnquery login --api-proxy http://INVALID_PROXY:111 --token XYZ...

The login actually works even through it should have failed b/c of the bad proxy details.

This is due to the way we need to fetch the proxy CLI parameter through the viper package. Since the binding of this parameter is actually "api_proxy" with an underscore, we should reference it exactly that way.

With this change, we now get the expected failure/behavior when given the bad proxy info:

$ ./cnquery login --api-proxy http://192.168.86.3:111 --token "SOME_TOKEN_DATA"
→ token description: test
→ token will expire at Wed, 12 Apr 2023 09:51:44 EDT FTL failed to log in client error="failed to do request: Post \"https://api.edge.mondoo.com/AgentManager/RegisterAgent\": proxyconnect tcp: dial tcp 192.168.86.3:111: connect: connection refused"


preslavgerchev (approved):

apps/cnquery/cmd/config/proxy.go
```diff
@@ -42,7 +42,7 @@ func (c *CommonCliConfig) GetAPIProxy() (*url.URL, error) {
 	return nil, nil
 }
 
-func (c *CommonCliConfig) GetHttpClient() (*http.Client, error) {
+=func (c *CommonCliConfig) GetHttpClient() (*http.Client, error) {
```
chris-rock: is that intentional?

chris-rock (commented):
Great catch @joelddiaz

apps/cnquery/cmd/config/proxy.go
joelddiaz: Nope. A total typo. Fixed. Thanks

PR #1150, opened 2023-04-13
Add support for Eurolinux 7/8/9

tas50:
This is another RHEL-deriv

cnquery> platform { name version family }
platform: {
  name: "eurolinux"
  family: [
    0: "redhat"
    1: "linux"
    2: "unix"
    3: "os"
  ]
  version: "9.1"
}


chris-rock (approved):
Thank you @tas50

PR #1154, opened 2023-04-17
Give queries with variants a code id

jaym:
This simplifies things in a few way. There are many places where we assume there is a code id. Sometimes if code is missing, a mrn is put in its place. But this is confusing. We shouldn't mix these 2 concepts


arlimus (approved):
Looks good, let's keep an eye on it tho for side-effects...

arlimus:
Please note my comment added to the related:
https://github.com/mondoohq/cnspec/pull/518#pullrequestreview-1393382326

I'm merging these to move forward, but I don't understand what problem we are trying to solve.

PR #1161, opened 2023-04-19
Mark client-id and tenant-id as required flags when scanning ms365.

### LINKED ISSUE #1112: tenant-id and client-id flags should not be optional with ms365
**Describe the bug**
To scan a ms365 tenant you need to provide a tenant id and client id, but these are both optional. This results in really confusing output when you don't provide the necessary data.

**To Reproduce**

1. Run `cnquery scan ms365 --certificate-path 'foo'
3. See cryptic error message about the backend

**Expected behavior**
These should be required fields so our error messages are clear that you need to pass in this required data.

**Screenshots or CLI Output**
```
cnquery scan ms365 --certificate-path ~/dev/combined.pem
x could not resolve asset error="ms365 backend requires a tenant-id" asset=
FTL failed to run scan error="failed to resolve multiple assets"
```

preslavgerchev:
Fixes #1112 

image


imilchev (approved):
LGTM

PR #1165, opened 2023-04-24
🧹 add user-friendly err when using unauthenticated cnquery

imilchev:
For the cases when authentication is actually needed but not provided:
```
cnquery> platform.cves
Query encountered errors:
This resource requires data from the Mondoo Platform.
For instructions on using cnspec with the Mondoo platform, see:
platform.cves.list: []
```

Fixes mondoohq/cnspec#364

### DIFF
--- a/resources/packs/core/core.go
+++ b/resources/packs/core/core.go
@@ -2,6 +2,10 @@ package core
 
 import "go.mondoo.com/cnquery/resources/packs/core/info"
 
+const MissingUpstreamErr = `To use this resource, you must authenticate with the Mondoo Platform.
+To learn how, read: 
+https://mondoo.com/docs/cnspec/cnspec-adv-install/registration/`
+
 var Registry = info.Registry
 
 func init() {
--- a/resources/packs/core/asset_advisories.go
+++ b/resources/packs/core/asset_advisories.go
@@ -14,8 +14,8 @@ import (
 func (a *mqlAsset) GetVulnerabilityReport() (interface{}, error) {
 	r := a.MotorRuntime
 	mcc := r.UpstreamConfig
-	if mcc == nil {
-		return nil, errors.New("mondoo upstream configuration is missing")
+	if mcc == nil || mcc.ApiEndpoint == "" {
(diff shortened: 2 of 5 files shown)


resources/packs/core/core.go
```diff
@@ -2,6 +2,10 @@ package core
 
 import "go.mondoo.com/cnquery/resources/packs/core/info"
 
+const MissingUpstreamErr = `This resource requires data from the Mondoo Platform.
```
misterpantz:
```suggestion
const MissingUpstreamErr = `To use this resource, you must authenticate with the Mondoo Platform.
```

resources/packs/core/core.go
```diff
@@ -2,6 +2,10 @@ package core
...
+For instructions on using cnspec with the Mondoo platform, see:
```
misterpantz:
```suggestion
To learn how, read: 
```

resources/packs/core/core.go
```diff
@@ -2,6 +2,10 @@ package core
 
 import "go.mondoo.com/cnquery/resources/packs/core/info"
 
+const MissingUpstreamErr = `To use this resource, you must authenticate with the Mondoo Platform.
+To learn how, read: 
+https://mondoo.com/docs/cnspec/cnspec-adv-install/registration/`
```
czunker, line 7: I find it strange, to see a cnspec URL, when using cnquery. We have the same page also for cnquery: https://mondoo.com/docs/cnquery/cnquery-platform/

Is it possible to show the URL depending on which client is used?

czunker (commented):
I gave it a try with cnquery and cnspec and both worked. Thanks!

resources/packs/core/core.go
imilchev: I don't know how to do that since this code is directly used into cnspec. Not sure if it's worth it too since the plans are to not require auth for CVEs and advisories too. Might be the case that this err becomes obsolete soon

czunker (approved):

PR #1166, opened 2023-04-24
🧹 Add defaults to two GCP resources

czunker:


tas50 (approved):

resources/packs/gcp/gcp.lr
```diff
@@ -1199,7 +1199,7 @@ private gcp.project.bigqueryService.dataset @defaults("id name") {
 }
 
 // GCP BigQuery dataset access entry
-private gcp.project.bigqueryService.dataset.accessEntry {
+private gcp.project.bigqueryService.dataset.accessEntry @defaults("id") {
```
imilchev: I think the ID isn't a good choice since this is some random thing that we generate for ourselves. It's not a GCP ID of any sort. Maybe set as defaults role + entity + entityType?

czunker: Good point. Changed it.

PR #1167, opened 2023-04-24
Fix wording in bundle command

tas50:
This was fixed in the docs, but we need to get it fixed here to avoid reverting things.


misterpantz (approved):
🎉

PR #1169, opened 2023-04-25
🐛 discover k8s assets with no namespace when scanning manifests

### LINKED ISSUE #1157: Discovery for a k8s manifest without a namespace
**Describe the bug**
Discovery for a k8s manifest, containing an object w/o a namespace yields no results

**To Reproduce**
`cnquery shell k8s <path>`, using the provided deployment at https://kubernetes.io/docs/concepts/workloads/controllers/deployment/

Note it's missing the namespace in the yaml file

I would expect we try and fill in the namespace the same way kubectl does it when applying a file so that these k8s objects can then be discovered as assets and connected to.

imilchev:
fixes #1157

also added a test for manifests containing resources with no namespace

### DIFF
--- a/motor/discovery/k8s/list_workloads.go
+++ b/motor/discovery/k8s/list_workloads.go
@@ -167,7 +167,7 @@ func ListNamespacedObj[T runtime.Object](
 		}
 	}
 
-	assets := []*asset.Asset{}
+	assetsIdx := map[string]*asset.Asset{}
 	for i := range workloads {
 		od.Add(workloads[i])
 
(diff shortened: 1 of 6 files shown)


motor/discovery/k8s/list_workloads.go
```diff
@@ -180,7 +180,15 @@ func ListNamespacedObj[T runtime.Object](
 		obj, _ := meta.Accessor(workloads[i])
 		log.Debug().Str("name", obj.GetName()).Str("connection", asset.Connections[0].Host).Msgf("resolved %s", workloadType)
 
-		assets = append(assets, asset)
+		assetsIdx[asset.PlatformIds[0]] = asset
+	}
+
+	// Return a unique list of assets. Manifests can contain a namespaces that is an empty string. When we try to list k8s
+	// resources for the empty namespace, that actually means list all resources. Therefore we can have duplicate entries in the list.
+	// Here we just return only the unique assets to make sure the code works correctly with both manifests and k8s API.
```
czunker, line 188: Because of this, I was curious to give this command a try: `cnquery run k8s /tmp/2cronjobs.yaml --namespaces default -c 'k8s.cronjob.name'`. It fails with:
```
x could not resolve asset error="failed to create asset from namespace: could not determine object kind " asset=
FTL failed to run query error="could not find an asset that we can connect to"
```
Without the namespace filter, cnquery discovers both CronJobs.

I'm not sure whether this is really a valid use-case for the manifest scan. Perhaps something, we could fix in a follow up?

For reference the file I used: https://gist.github.com/czunker/ee29f077a0babcf6f2ce847c29823e8c

czunker (approved):

motor/discovery/k8s/list_workloads.go
imilchev: Ah yeah, I need to set the `Kind` too now that we have a namespace asset. Nice catch! I will fix this

imilchev:
I found quite a few bugs in the manifest provider for k8s so I added multiple tests and fixed what I found. We need to figure out a way to disable gitguardian for the file that I added. It's intentionally containing a secret for the sake of testing loading k8s secrets from manifests.

PR #1170, opened 2023-04-26
🧹 remove k8s runtime

imilchev:
seems like this is no longer used in the code


czunker (approved):
Thanks for the cleanup @imilchev

PR #1183, opened 2023-05-03
🐛 Fix failures in GitHub pagination when fetching org repos

tas50:
We were passing in a new repo options data structure each time that set the pagination back to the initial value.


chris-rock (approved):
Thank you @tas50

PR #1196, opened 2023-05-09
🧹 make dynamodb table tags optional

vjeffrey:
the api on dynamodb is, by default, very restricted. making the tags an optional argument means we wont call the tags api (which has a 10 times per second limit)
 everytime we create a table resource


chris-rock (approved):
Thank you @vjeffrey

PR #1200, opened 2023-05-10
🧹 Add isProtected to GitHub repo branch

### LINKED ISSUE #882: Rename GitHub branch protection resource from "protected" to "isProtected"
**What is not working as you expected it?**
The `github.repository.branch` resource provides the `protected` field to check whether a branch has branch protection on. The same resource also provides the `isDefault` field as a `bool` to check whether the branch is the default branch. We should be consistent and rename `protected` to `isProtected`

https://github.com/mondoohq/cnquery/blob/main/resources/packs/github/github_repo.go#L381

**Where on the platform does it happen?**
Eg. Mondoo Platform, Mondoo Client, Mondoo Docs

**How do we replicate the issue?**
Please be specific as possible. Use dashes (-) or numbers (1.) to create a list of steps

**Expected behavior (i.e. solution)**
What should have happened?

**Other Comments**

czunker:
isProtected replaces protected.
protected is marked as deprecated.

Fixes #882


czunker:
It looks like this:
```
cnquery run github repo mondoohq/cnquery -c 'github.repository.branches{ name protected isProtected }'                                            
github.repository.branches: [
  0: {
    protected: false
    name: "benr/rpm-signing"
    isProtected: false
  }
...
  42: {
    protected: true
    name: "main"
    isProtected: true
  }
...
```

chris-rock:
Talked also with @arlimus and he agrees with you @czunker. `is` prefix makes it easier to understand. Then lets go with it.

PR #1204, opened 2023-05-10
🧹 Improve headers in OCI docs

tas50:
Capitalize some things there for better MQL docs


resources/packs/oci/oci.lr
```diff
@@ -20,7 +20,7 @@ oci.tenancy @defaults("name") {
   retentionPeriod() time
 }
 
-// OCI region
+// OCI Region
```
chris-rock, line 23: Oh I thought we decided on not doing that.

chris-rock (approved):
Thank you @tas50

PR #1211, opened 2023-05-15
✨ Show default fields for data inside '.all()' queries

czunker:
This will show the specified default fields instead of our internal id.

Partial fix for https://github.com/mondoohq/cnspec/issues/532


czunker:
This shows the results with the default fields. E.g. GKE:
```
[failed] [].all()
  actual:   [
    0: {
      name: "test-gke-cluster"
    }
  ]
```
Or for users:
```
c run local -c 'users.all( uid < 1000 )' 
[failed] users.all()
  actual:   [
    0: {
      uid: 65534
      gid: 65534
      name: "nobody"
    }
...
```

cli/printer/mql.go
```diff
@@ -187,10 +187,24 @@ func (print *Printer) array(typ types.Type, data []interface{}, codeID string, b
...
+		listType += " "
+		// newIndent = ""
+	}
+
 	for i := range data {
 		res.WriteString(fmt.Sprintf(
-			indent+"  %d: %s\n",
+			indent+"  %d: %s%s\n",
 			i,
+			listType,
```
czunker, line 206: I added this, because in nested queries, the resource type might not show up in the shortened query after "failed":
```
...
[failed] [].all()
  actual:   [
    0: gcp.project.gkeService.cluster {
      name: "test-cluster-2"
    }
  ]
```

mqlc/mqlc.go
```diff
@@ -1831,6 +1840,29 @@ func (c *compiler) updateEntrypoints(collectRefDatapoints bool) {
 		return res[i] < res[j]
 	})
 	c.block.Datapoints = append(c.block.Datapoints, res...)
+	// In the case of .all(...) queries, we have two datapoints for the list:
```
jaym: I found this comment a little confusing. It only mentions all, however the implementation below applies to more cases than all.

llx/code.go
```diff
@@ -399,7 +399,25 @@ func (l *CodeV2) entrypoint2assessment(bundle *CodeBundle, ref uint64, lookup fu
...
+				for j := len(code.Blocks[i].Datapoints) - 1; j >= 0; j-- {
+					// skip the resource ids datapoint
+					if code.Blocks[i].Datapoints[j] == listRef {
+						continue
+					}
+					chunk := code.Chunk(code.Blocks[i].Datapoints[j])
+					// this contains the default values
+					if chunk.Function.Binding == listRef {
+						listRef = code.Blocks[i].Datapoints[j]
+						break
```
jaym: does this break only the inner loop? should it break the outer loop?

llx/code.go
```diff
@@ -399,7 +399,25 @@ func (l *CodeV2) entrypoint2assessment(bundle *CodeBundle, ref uint64, lookup fu
 
 		if !truthy {
 			listRef := chunk.Function.Binding
-			list, ok := lookup(code.Checksums[listRef])
+			for i := range code.Blocks {
```
jaym, line 408: I think it would be nice to have a comment on what the intent for this block of code is

llx/code.go
czunker: Good catch. It can actually break the outer loop. Changed it.

llx/code.go
czunker: Added a comment to explain what the loop should accomplish.

mqlc/mqlc.go
czunker: I tried to clarify the comment.

But I'm not sure whether I understood your comment entirely correctly. Is the comment confusing because we have the additional datapoint also for `.none(...)` and alike, or because the loop also runs for other functions (`.where(...)`, ...)? Should I also add a switch statement as I did for the other code changes in this PR?

jaym: I think the changes you made are good. When I initially read it, it sounded like the logic should have only been applied for all.

PR #1212, opened 2023-05-15
🧹 move sns topic tags to be fetched only when needed

vjeffrey:


czunker (approved):
LGTM. Thanks @vjeffrey

PR #1215, opened 2023-05-16
🧹 gracefully handle missing keypairs for aws

vjeffrey:


scottford-io (approved):
@vjeffrey looks great! Thank you!!! ⭐ 

```
cnquery> aws.ec2.instances.where( region == "us-east-1" && keypair.name == "key-pair-to-delete") { keypair.name }
! key key-pair-to-delete does not exist in us-east-1 region
aws.ec2.instances.where: []
```

PR #1216, opened 2023-05-17
🐛 support containsOnly for dicts

arlimus:
We never ported this from arrays to dicts, which caused some common checks to surprisingly fail.

```
parse.json('dummy.json').params['string-array'].containsOnly(['b'])
```

```
[failed] [].containsOnly()
  expected: == []
  actual:   [
    0: "a"
    1: "c"
  ]
```

Vs

```
parse.json('dummy.json').params['string-array'].containsOnly(['a','c','b'])
```

```
[ok] value: []
```

with dummy.json that is found in the tests, simplified to just the line we want:

```json
{
  "string-array": ["a", "b", "c"],
}
```


chris-rock (approved):
Thank you @arlimus

PR #1217, opened 2023-05-17
🐛 fix the recording

chris-rock:
This issue was reported by @mm-weber. The recordings where working for `cnquery shell` but not for `cnquery scan`. This brings back the ability to record the data (os provider only) for a query pack.


mm-weber (approved):
Works as expected. LGTM

PR #1219, opened 2023-05-18
✨ Show actual value for failed assessments

czunker:
This adds the actual value, which was used for the assessment, to the output of the failed resources.

Partial fix for https://github.com/mondoohq/cnspec/issues/532


czunker:
Some examples of how the output now looks like:
```
cnquery run gcp project mondoo-edge -c 'gcp.project.gke.clusters.all(databaseEncryption["state"] == "NOT_ENCRYPTED")'  
...
[failed] [].all()
  actual:   [
    0: gcp.project.gkeService.cluster {
      databaseEncryption[state]: "ENCRYPTED"
      name: "mondoo-gke-cluster-2"
    }
  ]
```

```
cnquery run local -c 'users.all(shell == "/bin/zsh")' 
...
[failed] users.all()
  actual:   [
    0: user {
      gid: 0
      name: "root"
      shell: "/bin/bash"
      uid: 0
    }

```

Before, it looked like this for GCP:
```
Result:
    [failed] [].all()
      actual:   [
        0: gcp.project.gkeService.cluster id = gcp.project.gkeService.cluster/7c50e440aa5c41cf8eff749a4f313953c4c974b985ab43d1b44871e7dbf3e9a7
        1: gcp.project.gkeService.cluster id = gcp.project.gkeService.cluster/db3328e173c84de49d92229c02378c9f59b69e0a568a4448b52b3d7ff2f201f6
      ]
```

czunker:
This needs some more testing and test cases, but is basically working.

czunker:
It also works for queries like this:
```
cnquery run local --sudo -c 'users.all(sshkeys.length > 2)' 
...
[failed] users.all()
  actual:   [
    0: user {
      name: "root"
      gid: 0
      sshkeys.length: 0
      uid: 0
    }
...
```

mqlc/mqlc.go
```diff
@@ -1670,16 +1670,88 @@ func (c *compiler) postCompile() {
...
+				err := c.expandResourceFields(chunk, typ, ref)
+				// when no defaults are definied or query isn't about a resource, no block was added
+				if err == nil {
+					block.Datapoints = append(block.Datapoints, block.TailRef(ref))
+
+					c.addValueFieldChunks()
+				}
 			default:
-				c.expandResourceFields(chunk, typ, ref)
+				// ignore errrors here, log messages are already printed
```
jaym: i'm not seeing where they are printed

mqlc/mqlc.go
```diff
@@ -1710,25 +1782,25 @@ func (c *compiler) expandListResource(chunk *llx.Chunk, ref uint64) (*llx.Chunk,
 	return newChunk, newType, newRef
 }
 
-func (c *compiler) expandResourceFields(chunk *llx.Chunk, typ types.Type, ref uint64) {
+func (c *compiler) expandResourceFields(chunk *llx.Chunk, typ types.Type, ref uint64) error {
```
jaym: from what I can tell, it doesn't look like we care about error anywhere. It might be better to return a bool indicating success

mqlc/mqlc.go
```diff
@@ -1670,16 +1670,88 @@ func (c *compiler) postCompile() {
...
+}
+
+// addValueFieldChunks takes the value fields of the assessment and adds them to the
+// block forthe default fields
+// This way, the actual data of the assessment automatically shows up in the output
+// of the assessment that failed the assessment
+func (c *compiler) addValueFieldChunks() {
+	var assessmentBlock *llx.Block
+OUTER:
+	for i := range c.Result.CodeV2.Blocks {
```
jaym: I think i'm lost here. This function gets called whenever we compile an all/any/etc. But this is always finding the assessment block of the first matching chunk.

Take this for example:
```
go run apps/cnquery/cnquery.go run local --ast -c 'users.all(sshkeys.length > 2) users.none(sshkeys.length > 20)'
```

This produces the following code:
```
-> block 1
   entrypoints: [<1,5> <1,10>]
   1: users
   2: list bind: <1,1> type:[]user
   3: $whereNot bind: <1,1> type:users (ref<1,2>, => <2,0>)
   4: list bind: <1,3> type:[]user
   5: $all bind: <1,4> type:bool
   6: groups
   7: list bind: <1,6> type:[]group
   8: where bind: <1,6> type:groups (ref<1,7>, => <3,0>)
   9: list bind: <1,8> type:[]group
   10: $none bind: <1,9> type:bool
   11: {} bind: <1,4> type:[]block (=> <4,0>)
   12: {} bind: <1,9> type:[]block (=> <5,0>)
-> block 2
   entrypoints: [<2,4>]
   1: user id = context
   2: sshkeys bind: <2,1> type:[]privatekey
   3: length bind: <2,2> type:int
   4: > bind: <2,3> type:bool (2)
-> block 3
   entrypoints: [<3,3>]
   1: group id = context
   2: gid bind: <3,1> type:int
   3: == bind: <3,2> type:bool (0)
-> block 4
   entrypoints: [<4,2> <4,3> <4,4> <4,6>]
   1: user id = context
   2: name bind: <4,1> type:string
   3: uid bind: <4,1> type:int
   4: gid bind: <4,1> type:int
   5: sshkeys bind: <4,1> type:[]privatekey
   6: length bind: <4,5> type:int
-> block 5
   entrypoints: [<5,2> <5,3> <5,5>]
   1: group id = context
   2: name bind: <5,1> type:string
   3: gid bind: <5,1> type:int
   4: sshkeys bind: <5,1> type:[]privatekey
   5: length bind: <5,4> type:int
```

The last block looks incorrect to me. `ssh_keys` shouldn't be on groups

mqlc/mqlc.go
czunker: They were printed inside `expandResourceFields`. But with the bool as return value, this changed anyways.

mqlc/mqlc.go
czunker: Fixed. Thanks for the suggestion.

mqlc/mqlc.go
czunker: Nice catch. Thanks @jaym .

I used a different example to test the AST:
```
cnquery run local --sudo -c 'users.all(sshkeys.length > 2) groups.none(members.length > 20)' --ast
```
This now results in the following:
```
-> block 1
   entrypoints: [<1,5> <1,10>]
   1: users 
   2: list bind: <1,1> type:[]user
   3: $whereNot bind: <1,1> type:users (ref<1,2>, => <2,0>)
   4: list bind: <1,3> type:[]user
   5: $all bind: <1,4> type:bool
   6: groups 
   7: list bind: <1,6> type:[]group
   8: where bind: <1,6> type:groups (ref<1,7>, => <3,0>)
   9: list bind: <1,8> type:[]group
   10: $none bind: <1,9> type:bool
   11: {} bind: <1,4> type:[]block (=> <4,0>)
   12: {} bind: <1,9> type:[]block (=> <5,0>)
-> block 2
   entrypoints: [<2,4>]
   1: user id = context
   2: sshkeys bind: <2,1> type:[]privatekey
   3: length bind: <2,2> type:int
   4: > bind: <2,3> type:bool (2)
-> block 3
   entrypoints: [<3,4>]
   1: group id = context
   2: members bind: <3,1> type:[]user
   3: length bind: <3,2> type:int
   4: > bind: <3,3> type:bool (20)
-> block 4
   entrypoints: [<4,2> <4,3> <4,4> <4,6>]
   1: user id = context
   2: name bind: <4,1> type:string
   3: uid bind: <4,1> type:int
   4: gid bind: <4,1> type:int
   5: sshkeys bind: <4,1> type:[]privatekey
   6: length bind: <4,5> type:int
-> block 5
   entrypoints: [<5,2> <5,3> <5,5>]
   1: group id = context
   2: name bind: <5,1> type:string
   3: gid bind: <5,1> type:int
   4: members bind: <5,1> type:[]user
   5: length bind: <5,4> type:int

```

czunker:
The panic happens with this query: `'aws.ec2.volumes.all( attachments.length > 0 ) if (true) { aws.ec2.instances.all( deviceMappings.all(deleteOnTermination == true )) }'`

mqlc/mqlc.go
```diff
@@ -1684,18 +1684,111 @@ func (c *compiler) postCompile() {
...
+			return
+		}
+		if chunk.Function.Args != nil {
+			whereChunk = chunk
+			break
+		}
+		ref = chunk.Function.Binding
+	}
+
+	// find the referenced blcok for the where function
```
jaym: s/blcok/block

mqlc/mqlc.go
```diff
@@ -1684,18 +1684,111 @@ func (c *compiler) postCompile() {
...
+	for i := len(whereChunk.Function.Args) - 1; i >= 0; i-- {
+		arg := whereChunk.Function.Args[i]
```
jaym, line 1733: is the argument you're looking for now always at the same index?

mqlc/mqlc.go
```diff
@@ -1684,18 +1684,111 @@ func (c *compiler) postCompile() {
...
+		if types.Type(arg.Type).Underlying() == types.FunctionLike {
+			raw := arg.RawData()
+			blockRef := raw.Value.(uint64)
+			assessmentBlock = c.Result.CodeV2.Block(blockRef)
+			break
+		}
+	}
+
+	defaultFieldsBlock := c.Result.CodeV2.Blocks[len(c.Result.CodeV2.Blocks)-1]
+	defaultFieldsRef := defaultFieldsBlock.Chunks[1].Function.Binding
```
jaym: isn't this the first chunk in the block `defaultFieldsBlock | 1`?

mqlc/mqlc.go
```diff
@@ -1684,18 +1684,111 @@ func (c *compiler) postCompile() {
...
+				}
 			default:
 				c.expandResourceFields(chunk, typ, ref)
 			}
 		}
 	}
 }
 
+// addValueFieldChunks takes the value fields of the assessment and adds them to the
+// block forthe default fields
```
jaym: for the

mqlc/mqlc.go
```diff
@@ -1684,18 +1684,111 @@ func (c *compiler) postCompile() {
...
+		ref = chunk.Function.Binding
+	}
+
+	// find the referenced blcok for the where function
+	for i := len(whereChunk.Function.Args) - 1; i >= 0; i-- {
+		arg := whereChunk.Function.Args[i]
+		if types.Type(arg.Type).Underlying() == types.FunctionLike {
+			raw := arg.RawData()
+			blockRef := raw.Value.(uint64)
+			assessmentBlock = c.Result.CodeV2.Block(blockRef)
```
jaym, line 1737: From what I can tell, it references the predicate part of the expression. Maybe i'm missing something?

mqlc/mqlc.go
```diff
@@ -1684,18 +1684,111 @@ func (c *compiler) postCompile() {
...
+		if newChunk.Function != nil {
+			newChunk.Function.Binding = chunkRef
+		}
+
+		defaultFieldsBlock.AddChunk(c.Result.CodeV2, chunkRef, newChunk)
+		chunkRef = defaultFieldsBlock.TailRef(chunkRef)
+	}
+	// Only add last chunk as entrypoint, to not get data multiple times
+	// e.g. for dict["key"] we would otherwise get the entry and the whole dict
+	defaultFieldsBlock.Entrypoints = append(defaultFieldsBlock.Entrypoints, chunkRef)
```
jaym: This doesn't seem right to me. If you dont have each thing being collected as a entrypoint, its going to get skipped. That led me to try this example:
```
dlv debug --headless --api-version=2 --listen=:2345 apps/cnquery/cnquery.go -- run local --ast -c 'users.all(sshkeys.length > 2 && enabled == true)'
```

```
-> block 1
   entrypoints: [<1,5>]
   1: users
   2: list bind: <1,1> type:[]user
   3: $whereNot bind: <1,1> type:users (ref<1,2>, => <2,0>)
   4: list bind: <1,3> type:[]user
   5: $all bind: <1,4> type:bool
   6: {} bind: <1,4> type:[]block (=> <3,0>)
-> block 2
   entrypoints: [<2,7>]
   1: user id = context
   2: sshkeys bind: <2,1> type:[]privatekey
   3: length bind: <2,2> type:int
   4: > bind: <2,3> type:bool (2)
   5: enabled bind: <2,1> type:bool
   6: == bind: <2,5> type:bool (true)
   7: && bind: <2,4> type:bool (ref<2,6>)
-> block 3
   entrypoints: [<3,2> <3,3> <3,4> <3,9>]
   1: user id = context
   2: name bind: <3,1> type:string
   3: uid bind: <3,1> type:int
   4: gid bind: <3,1> type:int
   5: sshkeys bind: <3,1> type:[]privatekey
   6: length bind: <3,5> type:int
   7: > bind: <3,6> type:bool (2)
   8: enabled bind: <3,7> type:bool
   9: == bind: <3,8> type:bool (true)
```

block 3, chunk 8 seems incorrect to me: <3,7> is of type bool. We are calling `enabled` on it

czunker: Thanks for catching this bug. I have to look into this. For some queries, we only need the last chunk as entrypoint. This prevents, e.g, for `.all(sshkeys.length == 0)` the complete data for `sshkeys` to show up. Otherwise we would see `.length` and
```
sshkeys: [
        0: user privatekey id = privatekey:/home/christian/.ssh/xyz
...
      ]
```
But in the example you found, the data for the 2nd part after `&&` is missing completely. This should show up.

mqlc/mqlc.go
czunker: Yes, it is.
But I'm not sure what you are suggesting. Do you suggest renaming the variable, to make it more precise?

mqlc/mqlc.go
czunker: It's the reference to the first chunk. I didn't find a way to get the "address" of a chunk. I only found it in the binding.

mqlc/mqlc.go
czunker: I'm not sure about which index it actually is. That's why I iterate over the args.

mqlc/mqlc.go
czunker: Fixed.

mqlc/mqlc.go
czunker: Fixed.

mqlc/mqlc.go
jaym: I think it gets done by `blockRef << 32 | 1`. You'll see that a lot in the tests. Probably should be a function to do that

mqlc/mqlc.go
jaym: maybe. Or maybe a comment about the meaning of assessmentBlock

mqlc/mqlc.go
czunker: I opted for the solution you suggested. I couldn't find a way to make this work with all sorts of queries and, at the same time, keep the output minimal.

The query you mentioned now also works, and also some more. But we display more data. I don't know whether all this data is needed. I'll keep this in the back of my head. Maybe I'll find a solution in a follow-up PR.

mqlc/mqlc.go
czunker: Fixed.

mqlc/mqlc.go
czunker: I don't have the `blockRef`.

What do you think about adding the `ref` to a block as an exposed field? The moment we create a chunk or block, we know the ref. That would be something for a different PR.

jaym: `blockRef` in this example is `len(c.Result.CodeV2.Blocks) << 32` and then fieldref would be that ` | 1`

czunker: Ah, thanks.

I added some functions and changed the above code accordingly.

mqlc/mqlc.go
```diff
@@ -1724,25 +1839,26 @@ func (c *compiler) expandListResource(chunk *llx.Chunk, ref uint64) (*llx.Chunk,
 	return newChunk, newType, newRef
 }
 
-func (c *compiler) expandResourceFields(chunk *llx.Chunk, typ types.Type, ref uint64) {
+func (c *compiler) expandResourceFields(chunk *llx.Chunk, typ types.Type, ref uint64) bool {
+	expanded := false
```
arlimus: If we only ever use this variable `expanded` once, I'd avoid introducing it, since we now have to track where and how it changed. The only time it does is when we already return this.

llx/code.go
```diff
@@ -98,6 +104,11 @@ func (l *CodeV2) Block(ref uint64) *Block {
 	return l.Blocks[uint32(ref>>32)-1]
 }
 
+// Retrieve the ref for the last block
+func (l *CodeV2) LastBlockRef() uint64 {
```
arlimus, line 108: Great job on these methods, they are nice and helpful and can easily be inlined by the compiler.
A nit on the comment for LastBlockRef: Make sure the documentation starts with the name of the function, e.g. in this case `// LastBlockRef returns the ref for the last block in the code`

mqlc/mqlc.go
```diff
@@ -1684,18 +1684,133 @@ func (c *compiler) postCompile() {
...
+	// find chunk with where/whereNot function
+	// it holds the reference to the block with the predicate(s) for the assessment
+	for {
+		chunk := c.Result.CodeV2.Chunk(ref)
+		if chunk.Function == nil {
+			// this is a safe guard and shouldn't happen
+			log.Error().Msg("failed to find where function for assessment")
+			return
+		}
+		if chunk.Function.Args != nil {
```
arlimus: I'm curious to better understand why we look for the Args in the chunk, instead of looking for either of the methods where/whereNot?

mqlc/mqlc.go
arlimus: yeah i think it's fine for now, since the where calls have arguments on different places depending on their underlying type. but it really shows that we may want to clean that up and harmonize it to avoid having to guess it here :thinking:

arlimus (changes requested):
Great progress @czunker !!

One more bug, I have a feeling it might have to do with the way you positionally go through the blocks, is when you try to run this (eg in shell):

```coffee
> users.none( "/root" == home ); users.all( name != "root" )
...
  [failed] users.all()
    error: cannot find value for assessment (all)
```

it works one at a time.

Also we'll probably want to work on the output, but I'm happy to move that to a separate follow-up ticket so we can get this one merged first. That is, when we run things with the default expansions, we get things like this:

```
  0: user name="root" uid=0 gid=0
  1: user name="bin" uid=1 gid=1
  2: user name="daemon" uid=2 gid=2
  3: user name="mail" uid=8 gid=12
  4: user name="ftp" uid=14 gid=11
  5: user name="http" uid=33 gid=33
  6: user name="nobody" uid=65534 gid=65534
```

I really want to have expansions like this ^^ :grin: For example if we look for `uid` in our checks, that expansions is perfect already! Imagine you looked for something that is not in `@defaults`, then we can add the expansion easily:

```
  0: user name="root" uid=0 gid=0 {
      home: "/root"
  }
  1: user name="bin" uid=1 gid=1{
     home: "/"
  }
  ...
``` 

Again, happy to do this separately, that's going to make this really awesome!

mqlc/mqlc.go
czunker: Fixed.

llx/code.go
czunker: Fixed.

mqlc/mqlc.go
czunker: Good point. There is no reason :) I made it more complicated than it needed to be. Fixed this now.

czunker:
> Great progress @czunker !!
> 
> One more bug, I have a feeling it might have to do with the way you positionally go through the blocks, is when you try to run this (eg in shell):
> 
> ```coffee
> > users.none( "/root" == home ); users.all( name != "root" )
> ...
>   [failed] users.all()
>     error: cannot find value for assessment (all)
> ```
> 
> it works one at a time.
> 
> Also we'll probably want to work on the output, but I'm happy to move that to a separate follow-up ticket so we can get this one merged first. That is, when we run things with the default expansions, we get things like this:
> 
> ```
>   0: user name="root" uid=0 gid=0
>   1: user name="bin" uid=1 gid=1
>   2: user name="daemon" uid=2 gid=2
>   3: user name="mail" uid=8 gid=12
>   4: user name="ftp" uid=14 gid=11
>   5: user name="http" uid=33 gid=33
>   6: user name="nobody" uid=65534 gid=65534
> ```
> 
> I really want to have expansions like this ^^ grin For example if we look for `uid` in our checks, that expansions is perfect already! Imagine you looked for something that is not in `@defaults`, then we can add the expansion easily:
> 
> ```
>   0: user name="root" uid=0 gid=0 {
>       home: "/root"
>   }
>   1: user name="bin" uid=1 gid=1{
>      home: "/"
>   }
>   ...
> ```
> 
> Again, happy to do this separately, that's going to make this really awesome!

Thanks for finding this bug. I'll have a look at it.

I also prefer to move the output formatting to a separate PR. This will spark other discussion and I'd like to get this PR done.

czunker:
> Great progress @czunker !!
> 
> One more bug, I have a feeling it might have to do with the way you positionally go through the blocks, is when you try to run this (eg in shell):
> 
> ```coffee
> > users.none( "/root" == home ); users.all( name != "root" )
> ...
>   [failed] users.all()
>     error: cannot find value for assessment (all)
> ```
> 
> it works one at a time.
> 
> Also we'll probably want to work on the output, but I'm happy to move that to a separate follow-up ticket so we can get this one merged first. That is, when we run things with the default expansions, we get things like this:
> 
> ```
>   0: user name="root" uid=0 gid=0
>   1: user name="bin" uid=1 gid=1
>   2: user name="daemon" uid=2 gid=2
>   3: user name="mail" uid=8 gid=12
>   4: user name="ftp" uid=14 gid=11
>   5: user name="http" uid=33 gid=33
>   6: user name="nobody" uid=65534 gid=65534
> ```
> 
> I really want to have expansions like this ^^ grin For example if we look for `uid` in our checks, that expansions is perfect already! Imagine you looked for something that is not in `@defaults`, then we can add the expansion easily:
> 
> ```
>   0: user name="root" uid=0 gid=0 {
>       home: "/root"
>   }
>   1: user name="bin" uid=1 gid=1{
>      home: "/"
>   }
>   ...
> ```
> 
> Again, happy to do this separately, that's going to make this really awesome!

The bug is now fixed:
```
[failed] users.none( "/root" == home ); users.all( name != "root" )
  [failed] users.none()
    actual:   [
      0: user {
        gid: 0
        home: "/root"
        name: "root"
        uid: 0
      }
    ]
  [failed] users.all()
    actual:   [
      0: user {
        name: "root"
        uid: 0
        gid: 0
      }
    ]
```

arlimus (approved):
Thank you Christian, the fixes look good. Let's follow up on the improved output as well :)

czunker:
> Great progress @czunker !!
> 
> One more bug, I have a feeling it might have to do with the way you positionally go through the blocks, is when you try to run this (eg in shell):
> 
> ```coffee
> > users.none( "/root" == home ); users.all( name != "root" )
> ...
>   [failed] users.all()
>     error: cannot find value for assessment (all)
> ```
> 
> it works one at a time.
> 
> Also we'll probably want to work on the output, but I'm happy to move that to a separate follow-up ticket so we can get this one merged first. That is, when we run things with the default expansions, we get things like this:
> 
> ```
>   0: user name="root" uid=0 gid=0
>   1: user name="bin" uid=1 gid=1
>   2: user name="daemon" uid=2 gid=2
>   3: user name="mail" uid=8 gid=12
>   4: user name="ftp" uid=14 gid=11
>   5: user name="http" uid=33 gid=33
>   6: user name="nobody" uid=65534 gid=65534
> ```
> 
> I really want to have expansions like this ^^ grin For example if we look for `uid` in our checks, that expansions is perfect already! Imagine you looked for something that is not in `@defaults`, then we can add the expansion easily:
> 
> ```
>   0: user name="root" uid=0 gid=0 {
>       home: "/root"
>   }
>   1: user name="bin" uid=1 gid=1{
>      home: "/"
>   }
>   ...
> ```
> 
> Again, happy to do this separately, that's going to make this really awesome!

The follow-up for the output formatting: https://github.com/mondoohq/cnquery/issues/1277

PR #1222, opened 2023-05-21
⭐️ add filesystem provider to command builder

chris-rock:
This allows you to test mounted firmware files:

```
cnquery shell fs /mnt/firmware/
→ discover related assets for 1 asset(s)
→ resolved assets resolved-assets=1
  ___ _ __   __ _ _   _  ___ _ __ _   _ 
 / __| '_ \ / _` | | | |/ _ \ '__| | | |
| (__| | | | (_| | |_| |  __/ |  | |_| |
 \___|_| |_|\__, |\__,_|\___|_|   \__, |
  mondoo™      |_|                |___/  interactive shell

cnquery> asset {name version platform }
asset: {
  version: "v3.1.17"
  platform: "poky-iot2000"
  name: "iot2000"
}
```

You can easily test this locally via:

```
cnquery shell fs motor/providers/fs/testdata/centos8 
```

PR #1223, opened 2023-05-21
⭐️ Improve opkg parsing for yocto-based linux or openwrt

chris-rock:
- fixes an issue where not all running packages where shown
- adds support for static analysis of opkg packages
- improves opkg detection for `packages` resource, since yocto based linux environments commonly do not share a family

*Static Analysis*

The static analysis is extremely helpful to test yocto based images without starting a VM. Just mount the files system and that point cnquery at it. Works in conjunction with #1222:

```coffeescript
cnquery shell fs /mnt/firmware/
cnquery> packages
packages.list: [
  0: package name="shadow-securetty" version="4.6-r3"
  1: package name="os-release-dev" version="1.0-r0"
  2: package name="iptables-module-xt-cgroup" version="1.8.4-r0"
  3: package name="libwebsockets-dev" version="4.0.1-r0"
...
]
```

*Runtime package detection*

```coffeescript
cnquery shell docker openwrt
cnquery> packages
packages.list: [
  0: package name="base-files" version="1531-r22995-09d0643c18"
  1: package name="busybox" version="1.36.0-1"
  2: package name="ca-bundle" version="20211016-1"
  3: package name="dnsmasq" version="2.89-4"
...
]
```


resources/packs/core/packages/opkg_packages.go
```diff
@@ -33,6 +36,78 @@ func ParseOpkgPackages(input io.Reader) []Package {
...
+	const STATE_RESET = 0
+	const STATE_DESC = 1
+	pkgs := []Package{}
+
+	add := func(pkg Package) {
+		// do sanitization checks to ensure we have minimal information
+		if pkg.Name != "" && pkg.Version != "" {
+			pkgs = append(pkgs, pkg)
+		} else {
+			log.Debug().Msg("ignored deb packages since information is missing")
```
czunker: Does this package manager handle "deb" packages?

resources/packs/core/packages/opkg_packages.go
```diff
@@ -33,6 +36,78 @@ func ParseOpkgPackages(input io.Reader) []Package {
...
+		case key == "Package":
+			pkg.Name = strings.TrimSpace(m[2])
+		case key == "Version":
+			pkg.Version = strings.TrimSpace(m[2])
+		case key == "Architecture":
+			pkg.Arch = strings.TrimSpace(m[2])
+		case key == "Status":
+			pkg.Status = strings.TrimSpace(m[2])
+		case key == "Source":
+			o := DPKG_ORIGIN_REGEX.FindStringSubmatch(m[2])
```
czunker: `OPKG_ORIGIN_REGEX` instead of `DPKG...`?

resources/packs/core/packages/opkg_packages.go
```diff
@@ -33,6 +36,78 @@ func ParseOpkgPackages(input io.Reader) []Package {
...
+			if o != nil && len(o) >= 1 {
+				pkg.Origin = strings.TrimSpace(o[1])
+			} else {
+				log.Error().Str("origin", m[2]).Msg("cannot parse dpkg origin")
```
czunker: I think the same here, `opkg` instead of `dpkg`.

resources/packs/core/packages/opkg_packages.go
```diff
@@ -33,6 +36,78 @@ func ParseOpkgPackages(input io.Reader) []Package {
...
+		case key == "Description":
+			pkg.Description = strings.TrimSpace(m[2])
+			state = STATE_DESC
+		// next desc line, append to previous one
+		case state == STATE_DESC:
+			pkg.Description += "\n" + strings.TrimSpace(line)
+		}
+	}
+
+	// if the last line is not an empty line we have things in flight, lets check it
```
czunker, line 105: What if the last line is an empty line? Don't then have the package twice?

chris-rock: In this case the loop calls:

```golang
if len(line) == 0 {
  add(pkg)
  pkg = Package{Format: OpkgPkgFormat}
}
```

Then the package information is empty. Now it leaves the loop and calls `add` again.

```
add := func(pkg Package) {
  // do sanitization checks to ensure we have minimal information
  if pkg.Name != "" && pkg.Version != "" {
    pkgs = append(pkgs, pkg)
  } else {
    log.Debug().Msg("ignored opkg packages since information is missing")
  }
}
```

Since the package information is empty, the only thing that happens is that you see a debug message.

chris-rock:
Thank you @czunker I updated the PR to update the code. The opkg format is a derivative of the debian status file format therefore the implementations are very similar.

resources/packs/core/packages/opkg_packages.go
czunker: Thanks for the explanation. I missed the empty part.

czunker (approved):
Thanks @chris-rock

PR #1234, opened 2023-05-23
✨ dict.containsNone support

arlimus:
After: https://github.com/mondoohq/cnquery/pull/1233 !!

the counter-part for containsOnly for dicts (parsed json etc)


resources/packs/core/core_test.go
```diff
@@ -1101,6 +1101,26 @@ func TestDict_Methods_Contains(t *testing.T) {
...
+		{
+			p + "params['string-array'].containsOnly(['a', 'b'])",
+			1, false,
+		},
+		// {
+		// 	p + "params['string-array'].containsOnly('a')",
+		// 	1, false,
+		// },
+		{
+			p + "params['string-array'].containsNone(['d','e'])",
```
czunker, line 1117: Do we have the `dummy.json` somewhere? I'm just wondering whether this test data is suited for this test.

mqlc/builtin_map.go
```diff
@@ -203,6 +203,67 @@ func compileDictContainsOnly(c *compiler, typ types.Type, ref uint64, id string,
...
+		_, err := llx.BuiltinFunctionV2(typ, chunkId)
+		if err != nil {
+			return types.Nil, errors.New("called '" + id + "' with wrong type; either provide a type " + typ.Label() + " value or write it as an expression (e.g. \"_ == 123\")")
+		}
+	}
+
+	// .containsNone
+	c.addChunk(&llx.Chunk{
+		Call: llx.Chunk_FUNCTION,
+		Id:   "containsNone",
```
czunker, line 238: When do we set a `$` like in `$all` and when not?

czunker (approved):

mqlc/builtin_map.go
arlimus: Generally when we don't want the given method to be exposed. It signifies internal methods, often helper methods to support other methods

PR #1236, opened 2023-05-25
🛑 change the behavior of first/last to support empty lists

arlimus:
`first` and `last` both were introduced as helper methods to easily access the first and last element of an array. As such, having to deal with empty lists and the consequences of `null` is really annoying.

To better fit their behavior, these methods will now return `null` if you access elements from an empty or null array. For the latter case we already had this behavior, so now it better fits together.

Users can still use `[0]` to access the first and `[-1]` for the last element and have it throw an error if those elements could not be found.

```
> [1,2,3].where(_>4).first
where.first: null
```


chris-rock (approved):
Thank you @arlimus

PR #1238, opened 2023-05-26
✨ Add validity timestamp to queries

czunker:


explorer/cnquery_explorer.proto
```diff
@@ -194,6 +194,8 @@ message Mquery {
   repeated ObjectRef variants = 39;
   // Action is used for all query overrides (eg: in packs, policies, APIs etc)
   Action action = 41;
+  // ValidUntil is used to specify a date until which the IGNORE action is valid.
+  string valid_until = 42; // RFC3339
```
jaym, line 198: In cnspec, I think we do this at the group level https://github.com/mondoohq/cnspec/blob/main/policy/cnspec_policy.pb.go#L442-L443. 

Is this explicitly supposed to be different than that?

czunker: With https://github.com/mondoohq/cnspec/pull/588/files#diff-9c28b6dea039d5f4a7b352b47fc811523d578a95e0755e252b2190d9308c7d7bR884, we will be able to set exception on a query level. This is where I need this.

At least, this is how I understood it should work.

arlimus: @czunker Same as Jay, let's keep this focused on the group definitions and not establish this as something that is done for individual queries. The reason is that the moment that queries contain individualized timestamps, they would need the intirety of the exception process all associate with them: creation, validitiy, authors, reviewers, justifications and much more. We want to keep these focused on groups, even if we end up creating multiple groups in e.g. cnspec to manage these exceptions. Happy to go into a discussion on this next week in person OR feel free to chat since I'm in your timezone! Until then, let's try to avoid managing these on the query level

arlimus (changes requested):
Added my comment to the proto, let's handle this differently due to the slew of other changes that'd be necessary if we took this route. I'd love to learn more and happy to have an open discussion

czunker:
Thanks for your feedback @jaym / @arlimus . I'll close this PR and use the cnspec parts.

PR #1244, opened 2023-05-26
🧹 Memoize the results of GetKindResources

jaym:
This function was getting called any time we initialized a kubernetes resource. From what I can tell, this caused a lot of load for deserialization, and was also calling out to kubernetes each time.

This knocked off about 30 seconds in scan time, going from 124 seconds to 96 seconds (sample size 1)


imilchev (approved):
nice improvement! Just tested it and seems to be working as expected

scottford-io:
@jaym @imilchev is this ready? We have workload scanning off in their environment for now. Any improvements are greatly welcomed.

PR #1246, opened 2023-05-27
:bug: fix fail if config file not found

### LINKED ISSUE #1056: Fail scan if config file is not found?
This is not necessarily a bug but more of a question around the intended behaviour here

Running a cnquery scan and providing a config file that does not exist will end up with cnquery falling back to incognito.
```
cnquery scan --config bogus.json
! could not load configuration file bogus.json
! No credentials provided. Switching to --incognito mode.
```

I've personally been caught by this a few times where I specify a path I think is correct which turns out to be the wrong one. Since cnquery doesn't exit but performs a scan, I end up looking for the asset in the space I'd expect it to be only to come back and find out that cnquery's been running in incognito mode.

Should we fail if the scan if a config file is explicitly provided but it does not exist?

mariuskimmina:
closes #1056

New behavior: 
image

I believe that if the user provides the `--config` flag he does not want us to run the scan without his config if we can't find the file, so it makes sense to just fail here and let the user fix his command.

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"bytes"
 	"encoding/base64"
+	"fmt"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -36,7 +37,10 @@ var (
 	Features             cnquery.Features
 )
 
-const configSourceBase64 = "$MONDOO_CONFIG_BASE64"
+const (
+	configSourceBase64 = "MONDOO_CONFIG_BASE64"
+	configSourcePath   = "MONDOO_CONFIG_PATH"
+)
 
 // Init initializes and loads the mondoo config
 func Init(rootCmd *cobra.Command) {
@@ -157,7 +161,7 @@ func initConfig() {
 
 	Path = strings.TrimSpace(UserProvidedPath)
 	// base 64 config env setting has always precedence
-	if len(os.Getenv("MONDOO_CONFIG_BASE64")) > 0 {
+	if len(os.Getenv(configSourceBase64)) > 0 {
 		Source = configSourceBase64
 		decodedData, err := base64.StdEncoding.DecodeString(os.Getenv("MONDOO_CONFIG_BASE64"))
 		if err != nil {
@@ -166,8 +170,8 @@ func initConfig() {
 		viper.ReadConfig(bytes.NewBuffer(decodedData))
 	} else if len(Path) == 0 && len(os.Getenv("MONDOO_CONFIG_PATH")) > 0 {
 		// fallback to env variable if provided, but only if --config is not used
-		Source = "$MONDOO_CONFIG_PATH"
-		Path = os.Getenv("MONDOO_CONFIG_PATH")
+		Source = configSourcePath
+		Path = os.Getenv(configSourcePath)
 	} else if len(Path) != 0 {
 		Source = "--config"
 	} else {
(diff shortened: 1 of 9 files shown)


cli/config/config.go
```diff
@@ -229,7 +229,8 @@ func initConfig() {
 func DisplayUsedConfig() {
 	// print config file
 	if !LoadedConfig && len(UserProvidedPath) > 0 {
-		log.Warn().Msg("could not load configuration file " + UserProvidedPath)
+		log.Error().Msg("could not load configuration file " + UserProvidedPath)
+		os.Exit(1)
```
chris-rock: We cannot exit here. There are a couple of reasons:

- the semantics of the method name indicate it is displaying the used config. consumers of that function would not expect to exit the program
- I think there should be a display function, and a separate function that validates the config

chris-rock (commented):
Thank you @mariuskimmina for the first iteration to kick-start the discussion. 

Config validation is quite complex. We have multiple cases that we need to handle:

- `MONDOO_CONFIG_BASE64`
- use `--config`
- `MONDOO_CONFIG_PATH`
- `AWS_SSM_PARAMETERSTORE_PREFIX`
- path auto detect

The user expects the following behavior:

1. It is ok if the user does not provide any config, since cnquery / cnspec work without any
2. We only want to error out if the user provided an explicit value eg. via `MONDOO_CONFIG_BASE64`, `--config`, `MONDOO_CONFIG_PATH` or  `AWS_SSM_PARAMETERSTORE_PREFIX`
3. We do not want to error out if we could not find a config via auto-detect

We also want to be able to have quite a bit of control about the error. Therefore I propose to:

- introduce a new method `IsValidConfigLocation` in `cli/config` package
- since `cli/config` does not know anything about the contents of the config the `IsValidConfigLocation` should only verify the cases mentioned abobe
- the method should return an error, so that the caller can still decide if the program needs to exit or not

Once that is implemented we replace 

```
config.DisplayUsedConfig()
```

used in the commands with something like:

```
err := config.IsValidConfigLocation()
if err != nil {
  log.Fatal().Err(err).Msg("could not load user-provided config")
}
config.DisplayUsedConfig()
```

mariuskimmina:
Thanks for the review and good points @chris-rock - I'll convert this to a draft PR for now and think about further improvements

mariuskimmina:
I made some progress on this - `ValidateConfigPath` currently returns an error if
1. --config is used but not a valid file
2. MONDOO_CONFIG_PATH is set but not a valid file

I haven't yet looked at the `AWS_SSM_PARAMETERSTORE_PREFIX` and I haven't covered all cases where `DisplayUsedConfig` is called but I assume this is going in the direction you had in mind @chris-rock ?

image

mariuskimmina:
Note: I just saw that this `DisplayUsedConfig` function is also used in cnspec, so I would have to open a second PR there

mariuskimmina:
I have tested this with 

* `--config` 
* `MONDOO_CONFIG_PATH`
* `MONDOO_CONFIG_BASE64` 
* no user provided config

Each case works if the provided config exists and fails if the path is invalid.
If multiple options are used `--config` seems to take precedence over `MONDOO_CONFIG_PATH` which takes precedence over `MONDOO_CONFIG_BASE64` and only the option with the highest precedence will be checked for a valid config. 

I have not tested this change with `AWS_SSM_PARAMETERSTORE_PREFIX`, mostly because it requires a lot more setup.

cli/config/config.go
```diff
@@ -157,7 +161,7 @@ func initConfig() {
 
 	Path = strings.TrimSpace(UserProvidedPath)
 	// base 64 config env setting has always precedence
-	if len(os.Getenv("MONDOO_CONFIG_BASE64")) > 0 {
+	if len(os.Getenv(configSourceBase64)) > 0 {
```
chris-rock, line 171: I like the extraction

cli/config/config.go
```diff
@@ -228,13 +232,31 @@ func initConfig() {
...
 		log.Info().Msg("no Mondoo configuration file provided. using defaults")
 	}
 }
+
+func ValidateUserProvidedConfigPath() error {
+	configPath := os.Getenv(configSourcePath)
+	if !LoadedConfig && len(UserProvidedPath) > 0 {
+		_, err := os.Stat(UserProvidedPath)
+		if err != nil {
+			return errors.New(fmt.Sprintf("Could not load config: %v - file does not exist", UserProvidedPath))
```
chris-rock: I think we want to introduce specific error types and then attach the text to the errors. This makes it easier to check validation programmatically and we can make the log.Fatal output more nice since we know it is an expected error. https://go.dev/blog/error-handling-and-go is a good reading for that.

cli/config/config_test.go
```diff
@@ -96,3 +96,56 @@ func Test_inventoryPath(t *testing.T) {
 	assert.Equal(t, systemInventory, path)
 	assert.True(t, ok)
 }
+
+func Test_ValidateUserProvidedConfigPath(t *testing.T) {
```
chris-rock, line 100: It is great to see tests for configuration checks being added.

chris-rock (commented):
@mariuskimmina This looks great and moves in the right direction. Lets do the following improvements:

- introduce custom error types with a common `ConfigValidationError` interface
- improve the output for the validation errors for the user so that it is easier to understand the next steps for the user to take (if required)

mariuskimmina:
I've added a custom error type `FileNotFoundError`

For the user it would look like this:
image

and the check would be 
```
	if err != nil {
		if serr, ok := err.(*config.FileNotFoundError); ok {
			log.Error().Err(serr).Msg("Couldn't find user provided config file")
			log.Info().Msgf("Ensure that %s provided through %s is a valid file path", serr.Path(), serr.Source())
			os.Exit(1)
		}
		log.Fatal().Err(err).Msg("Failed to load user provided config")
	}
```

I have for now only added it to the `scan` command.
Is this what you had in mind?

mariuskimmina:
I found that in some cases we have been using `https://github.com/cockroachdb/errors` instead of the std errors package - which one should I be using here? @chris-rock

mariuskimmina:
Outdated, if we work on this again it might as well be a new PR

PR #1249, opened 2023-05-28
🧹 update bubbletea rendering go dependencies

chris-rock:
superseeds #1106


cli/components/list.go
```diff
@@ -39,8 +39,9 @@ func (d itemDelegate) Render(w io.Writer, m list.Model, index int, listItem list
 
 	fn := itemStyle.Render
 	if index == m.Index() {
-		fn = func(s string) string {
-			return selectedItemStyle.Render("> " + s)
+		fn = func(strs ...string) string {
+			strs = append([]string{"> "}, strs...)
```
czunker: The renderer changed. It now looks like this:
image

The selected item is now intended one space. When we want to keep it as it was before, we would need to remove the space in this line.

chris-rock: We should remove that and keep the current behavior.

chris-rock: Thank you @czunker I updated and fixed this issue

czunker (approved):
LGTM. Thanks @chris-rock

PR #1250, opened 2023-05-29
:bug: Fix display config when using MONDOO_CONFIG_PATH

mariuskimmina:
While working on #1246   I noticed that `DisplayUsedConfig` would not work with `$MONDOO_CONFIG_PATH` - this PR fixes that.

Previous behavior:
image

New behavior:
image

Note that this shows as successful no matter if the path for `MONDOO_CONFIG_PATH` is a valid one. The same is already true for `MONDOO_CONFIG_BASE64` can also be set to e.g. "bogus" and it will say it loaded it just fine.
So I think it's fine to for now and I'll continue working on validation improvements in #1246

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -37,6 +37,7 @@ var (
 )
 
 const configSourceBase64 = "$MONDOO_CONFIG_BASE64"
+const configSourcePath = "$MONDOO_CONFIG_PATH"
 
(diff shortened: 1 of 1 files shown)


mariuskimmina:
Actually I was wrong here, if I set $MONDOO_CONFIG_PATH to a valid file it already shows up properly

PR #1255, opened 2023-05-30
🦘 bump to go v1.20

arlimus:
Check if we can have all the blockers removed for letting this run on OSX, which seems to have been the blocker the last time we tried to bump it.

Also: YAML is hard :grin:  I replaced every version that was set as a number with an explicit string, otherwise it tries to pull go v1.2 (instead of v1.20)


arlimus:
We merged https://github.com/mondoohq/cnquery/pull/1254 before this landed. I'm closing this and pushing a separate update for the minimum version that was requested.

PR #1259, opened 2023-05-31
:bug: fix --context option ignored in cnspec

mariuskimmina:
Closes https://github.com/mondoohq/cnspec/issues/593

Working example of the `--context` flag
image

If the flag is not provided it still falls back to the default cluster
image


imilchev (changes requested):
Thank you for the contribution @mariuskimmina! You are correct that the `--context` flag isn't working. The code that you wrote so far is correct. However, you are only retrieving the cluster name based on the provided context. The actual context that is used for talking to the kubernetes API is still the default context in your environment.

The code that retrieves the kube config is [here](https://github.com/mondoohq/cnquery/blob/main/motor/providers/k8s/api_provider.go#L84). You need to make sure the right context is used when the config is built

imilchev (approved):
LGTM! Thanks for the contribution!

PR #1264, opened 2023-06-03
:broom: remove unused pager and no-pager arguments

mariuskimmina:
I belive that `--pager` and `--no-pager` arguments are currently not having any effect, I was not able to make them work, neither on cnquery nor on cnspec.

I found that there was some effort to introduce paging in https://github.com/mondoohq/cnquery/pull/279 but that code is still not used anywhere. So I think we should remove these flags until we have a proper implementation. 

I left these commented out since I belive this feature is still planned.

See also: https://github.com/mondoohq/cnspec/issues/602

If these flags are indeed used somewhere please let me know

### DIFF
--- a/apps/cnquery/cmd/scan.go
+++ b/apps/cnquery/cmd/scan.go
@@ -301,8 +301,8 @@ This example connects to Microsoft 365 using the PEM formatted certificate:
 		// output rendering
 		cmd.Flags().StringP("output", "o", "compact", "Set output format: "+reporter.AllFormats())
 		cmd.Flags().BoolP("json", "j", false, "Set output to JSON (shorthand).")
-		cmd.Flags().Bool("no-pager", false, "Disable interactive scan output pagination.")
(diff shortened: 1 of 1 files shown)


chris-rock:
I had a discussion with @arlimus We think we do not need the pager anymore since 

1. We have web view for cnspec reports and we want to ship this for query packs too
2. We switched to a much better default output which does not require a pager anymore

In order to get this implemented, we need to:

1. remove pager capabilities from cnspec, see https://github.com/mondoohq/cnspec/pull/609
2. also remove the detection of the pager in `cli/pager`

mariuskimmina:
Okay, since the pager is not planned I have removed them now completely and also the whole `cli/pager` package should be safe to remove, I couldn't find an import of this package anywhere.

chris-rock (approved):
@mariuskimmina thank you for the clean up. This is save to be merged once https://github.com/mondoohq/cnspec/pull/609 is merged.

PR #1272, opened 2023-06-05
⚡ reduce # of lines printed in shell

arlimus:
When running soemthing in shell that generates thousands of lines of output, the shell will block and take a long time to process (and pipe) all the output to STDOUT. We can't speed up STDOUT a lot, but we can make sure we don't push an insane amount of data to it, that would slow it down.

With this change we are introducing a maximum number of lines. The 1k lines is the default that many CLI shells come with. This is a ton of entries, even if we block-extract 8 fields, accounting for 2 extra lines per block, we would be able to display 100 items. This is a very sane amount for a regular REPL environment.

We will make this configurable again down the line. Feedback is very welcome!

For now, it will stop the shell from almost locking up when a ton of lines/items are extracted.

### DIFF
--- a/cli/shell/shell.go
+++ b/cli/shell/shell.go
@@ -25,6 +25,7 @@ import (
 	"go.mondoo.com/cnquery/resources"
 	"go.mondoo.com/cnquery/resources/packs/all"
 	"go.mondoo.com/cnquery/sortx"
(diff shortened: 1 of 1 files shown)


chris-rock (approved):
Thank you @arlimus

PR #1275, opened 2023-06-09
🐛 Fix memory usage when scanning large images

czunker:
This fix prevents a copy of the image in memory. But that has a performance impact.

### DIFF
--- a/motor/providers/container/container.go
+++ b/motor/providers/container/container.go
@@ -1,6 +1,9 @@
 package container
 
 import (
+	"strings"
+	"time"
+
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/rs/zerolog/log"
 	"go.mondoo.com/cnquery/motor/asset"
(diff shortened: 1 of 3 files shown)


czunker:
This keeps the memory low, but the image scan is 6x/7x slower than before.

motor/providers/container/image/docker.go
```diff
@@ -35,7 +35,8 @@ func (r ShaReference) Scope(scope string) string {
 }
 
 func LoadImageFromDockerEngine(sha string) (v1.Image, io.ReadCloser, error) {
-	img, err := daemon.Image(&ShaReference{SHA: strings.Replace(sha, "sha256:", "", -1)})
+	opts := []daemon.Option{daemon.WithUnbufferedOpener()}
+	img, err := daemon.Image(&ShaReference{SHA: strings.Replace(sha, "sha256:", "", -1)}, opts...)
```
czunker, line 39: By default, this is created with a bufferedReader: https://github.com/google/go-containerregistry/blob/main/pkg/v1/daemon/image.go#L62

motor/providers/container/container.go
```diff
@@ -110,14 +113,28 @@ func NewDockerEngineImage(endpoint *providers.Config) (ContainerProvider, error)
...
 		return nil, err
 	}
 
 	var identifier string
+	// This is not the digest used to pull the image from the registry.
+	// And it takes a lot of time on large images... >2min for 1.8GB image on my laptop
+	// because it actually calculates the hash of the image and does not fetch it from metadata
+	log.Debug().Msg("starting to calculate digest")
+	start := time.Now()
 	hash, err := img.Digest()
```
czunker, line 136: We could remove this to speed things up. Not sure whether we really need the hash or could also use the image ID.

From what I understood, it could be the case the image ID is different when the same image is pulled from different registries.

motor/providers/container/container.go
```diff
@@ -110,14 +113,28 @@ func NewDockerEngineImage(endpoint *providers.Config) (ContainerProvider, error)
...
 	}
 
+	labelImageIdFull := strings.Split(ii.Labels["docker.io/digests"], ",")[0]
+	labelImageId := strings.Split(labelImageIdFull, "@")[1]
+
 	log.Debug().Msg("found docker engine image " + ii.ID)
+	// This is the image id that is used to pull the image from the registry.
+	log.Debug().Msg("found docker engine image " + labelImageId)
+	if ii.Size > 1024 { // > 1GB
+		log.Warn().Int64("size", ii.Size).Msg("image is bigger than 1GB, this will take some time")
```
czunker, line 123: This is only to demonstrate, that we could act upon the image size. We could create a bufferedReader for images which are smaller than X. We would than need to talk about how large X is.

motor/providers/container/container.go
```diff
@@ -110,14 +113,28 @@ func NewDockerEngineImage(endpoint *providers.Config) (ContainerProvider, error)
...
+	}
 	img, rc, err := image.LoadImageFromDockerEngine(ii.ID)
 	if err != nil {
 		return nil, err
 	}
 
 	var identifier string
+	// This is not the digest used to pull the image from the registry.
```
imilchev, line 131: IIRC I used that to filter out duplicate container images in k8s. If that's not the digest used to pull the image, then the function is used in the wrong way. If we can use the actual digest it would be good. I can't see a reason to use this `img.Digest()`

PR #1276, opened 2023-06-09
🐛 Fix error messages for `.none()` queries

czunker:
Queries like `users.list.duplicates(uid).none()` produced the error message `failed to find where function for assessment`. These errors were wrong.

In cases where no field is checked for the assessment, there is no `where` function


arlimus (approved):
Thank you!

PR #1279, opened 2023-06-12
🐛 where/contains/... + vars

arlimus:
... was incorrectly interpreted as a non-standalone block. That is wrong. The `standalone` attribute in compilers referes to the contents of the compiler not to reference its bindings, i.e. everything in the expressions to stand on their own (or said differently: the expressions to be standalone). We had a bug where standalone was set to `false` when variables were used. The use of vars does not make something non-standalone. Only the values inside of vars can do that, which are handled in their assignment.

Example:

```
allowed_roles = ["roles/owner", "roles/editor", "roles/viewer"]
role = "roles/owner2"
allowed_roles.contains(role)
```

Since `role` is not in allowed roles, this should fail. Before, this was interpreted incorrectly and returned true. This bug is now fixed.

### DIFF
--- a/mqlc/mqlc.go
+++ b/mqlc/mqlc.go
@@ -1239,7 +1239,6 @@ func (c *compiler) compileIdentifier(id string, callBinding *variable, calls []*
 			Call:      llx.Chunk_PRIMITIVE,
 			Primitive: llx.RefPrimitiveV2(variable.ref),
 		})
-		c.standalone = false
 		checksum := c.Result.CodeV2.Checksums[c.tailRef()]
 		c.Result.Labels.Labels[checksum] = variable.name
 		return restCalls, variable.typ, nil
--- a/mqlc/mqlc.go
+++ b/mqlc/mqlc.go
@@ -21,9 +21,10 @@ import (
 )
 
 type variable struct {
-	name string
-	ref  uint64
-	typ  types.Type
+	name     string
+	ref      uint64
+	typ      types.Type
+	callback func()
 }
 
 type varmap struct {
@@ -571,6 +572,9 @@ func (c *compiler) blockOnResource(expressions []*parser.Expression, typ types.T
 	v := variable{
 		ref: blockCompiler.blockRef | 1,
 		typ: typ,
+		callback: func() {
+			blockCompiler.standalone = false
+		},
 	}
 	blockCompiler.vars.add("_", v)
 	blockCompiler.Binding = &v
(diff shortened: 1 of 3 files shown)


mqlc/mqlc.go
```diff
@@ -21,9 +21,10 @@ import (
 )
 
 type variable struct {
-	name string
-	ref  uint64
-	typ  types.Type
+	name     string
+	ref      uint64
+	typ      types.Type
+	callback func()
```
jaym, line 32: its not clear to me what callback on variable would mean. can we add a comment

jaym (approved):

mqlc/mqlc.go
arlimus: added, good point :+1:

resources/packs/core/core_test.go
```diff
@@ -1057,6 +1061,16 @@ func TestListResource(t *testing.T) {
...
 			0, []interface{}([]interface{}{"root", "chris", "christopher", "chris", "bin"}),
 		},
+		{
+			// outside variables cause the block to be standalone
+			"n=false; users.contains(n)",
+			1, false,
+		},
+		{
+			// variables do not override local fields in blocks
+			"name=false; users.contains(name)",
```
czunker, line 1071: I do not understand, why this test passes.

When I run this on my local Linux machine, I also get an "ok". But I would expect something else:
```
cnquery> name=false; users.contains(name)
[ok] value: 56
```
"56" is the length of the users array. But what does it check against, that it returns "ok"? The last element in the users array is:
```
55: user name="snap_daemon" uid=584788 gid=584788
```

When I change the query, I get the result I'd expect:
```
cnquery> name=false; users.map(name).contains(name)
[failed] [].contains()
  expected: > 0
  actual:   0
```

chris-rock: It seems that this issue happens when the slice is a list of resources. We probably need to handle this separately. Lets get this merged anyway since it fixes an active issue.

arlimus: @czunker Actually this is behaving correctly right now: it uses `user.name` inside the block, instead of the variable you declared. That's because the variable is overwritten by the block's own fields. However, we may want to change this behavior, and feedback very warmly welcome on this! I can see how this is confusing and we may want the outer variable to win this race, instead of the field. Users can still address the field via `_`.

The best way to see it is to try this:

```coffee
> name="root"; users.contains(name == "root")
[ok] value: 1
```

The resulting list is 1, because it only applies to 1 name ^^

```coffee
> name2="root"; users.contains(name2 == "root")
[ok] value: 37
```

The resulting list is 37, because it now applies to everything, because it uses the variable

chris-rock: I completely have not thought about that we overshadow the variable here. I think it is standard that the deepest level overrides the level above. Instead of fixing the problem, only to discover than then some users need exact the opposite, I recommend we add a compiler warning to cases where we see variable shadowing happening. This would be similar to `go build` and `go vet`

czunker: Thanks for the explanation. That helped to understand what happens here.

PR #1280, opened 2023-06-12
✨ string.contains(regex)

arlimus:
**AFTER https://github.com/mondoohq/cnquery/pull/1279**
Feel free to remove the `DRAFT` from the title once the above PR is merged!

... and a handful of derivates like string.contains([]regex), dict.contains(regex) and dict.contains([]regex).
    
Now you 

```
"hello world".contains(/wor.d/)
```
    
Please note: if the regex has beginning or end conditions (i.e. `^` or `$`) these will be respected by `contains`, i.e. it will not magically remove these.
    
We are introducing this operation to deprecated the old approach of `x == regex` with v9. It currently leads to very misleading results with e.g. `1.2.3.4abc == regex.ipv4` returning `true`.
    
### DIFF
--- a/llx/builtin.go
+++ b/llx/builtin.go
@@ -303,6 +303,8 @@ func init() {
 			string("contains" + types.Array(types.String)): {f: stringContainsArrayStringV2, Label: "contains"},
 			string("contains" + types.Int):                 {f: stringContainsIntV2, Label: "contains"},
 			string("contains" + types.Array(types.Int)):    {f: stringContainsArrayIntV2, Label: "contains"},
+			string("contains" + types.Regex):               {f: stringContainsRegex, Label: "contains"},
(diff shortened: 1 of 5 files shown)


charlesjohnson:
This was linked in the 8.14 release notes, punted the link to the 8.15 release notes instead.

czunker:
I'm not sure about the regex example for IPv4:
```
"1.2.3.4abc".contains(regex.ipv4)
[ok] value: true
```
And this is true, as the string starts with an IPv4 address.

How does `contains` help in identifying IPv4 addresses?

arlimus:
> I'm not sure about the regex example for IPv4:

By doing:

> We are introducing this operation to deprecated the old approach of x == regex with v9.

It currently behaves like `contains`. We have an option to address the behavior in v9 and decide from there. We could change it to only match the entire regex. Or introduce other flags to help solve this problem.

PR #1289, opened 2023-06-16
🐛 Fix memory usage when scanning large images

czunker:
This fix prevents a copy of the image in memory. But that has a performance impact.

### DIFF
--- a/motor/providers/container/container.go
+++ b/motor/providers/container/container.go
@@ -1,6 +1,9 @@
 package container
 
 import (
+	"strconv"
+	"strings"
+
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/rs/zerolog/log"
 	"go.mondoo.com/cnquery/motor/asset"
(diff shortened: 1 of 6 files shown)


czunker:
This replaces https://github.com/mondoohq/cnquery/pull/1275

czunker:
This fix is only for CLI scans. This does not fix the slight memory increase when scanning inside a container runtime.

motor/providers/container/container.go
```diff
@@ -110,17 +121,27 @@ func NewDockerEngineImage(endpoint *providers.Config) (ContainerProvider, error)
...
+		splitFullLabel := strings.Split(labelImageIdFull, "@")
+		if len(splitFullLabel) > 1 {
+			labelImageId = strings.Split(labelImageIdFull, "@")[1]
+		}
+	}
+
+	// This is the image id that is used to pull the image from the registry.
+	log.Debug().Msg("found docker engine image " + labelImageId)
+	if ii.Size > 1024 && !disableInmemoryCache { // > 1GB
+		log.Warn().Int64("size", ii.Size).Msg("The image is bigger than 1GB, this will require a lot of memory. Consider disabling the in-memory cache with `--disable-cache=true`.")
```
imilchev:
```suggestion
		log.Warn().Int64("size", ii.Size).Msg("The image is bigger than 1GB, this will require a lot of memory. Consider disabling the in-memory cache with `--disable-cache=true`.")
```

The message reads a bit weird. Not sure how to re-formulate it to be better though

czunker: @misterpantz Do you have a suggestion?

misterpantz: How about...
```suggestion
		log.Warn().Int64("size", ii.Size).Msg("Because the image is larger than 1 GB, this task will require a lot of memory. Consider disabling the in-memory cache by adding this flag to the command: `--disable-cache=true`")
```

czunker: Thanks @misterpantz

imilchev (approved):
LGTM

PR #1291, opened 2023-06-16
🐛 fix map values correct type

arlimus:
Now you can call this and get the right results:

```coffee
> pam.conf.entries.values.flat{*}
pam.conf.entries.values.flat: [
  0: {
    options: []
    control: "sufficient"
    module: "pam_rootok.so"
    pamType: "auth"
    lineNumber: 1
    service: "/etc/pam.d/chsh"
  },
  ...
]
```


czunker (approved):
LGTM. Thanks @arlimus

PR #1300, opened 2023-06-20
🧹 add ostype tag to discovered azure vms

vjeffrey:
so we can later determine is the machine is linux/vm for run command scan purposes


motor/discovery/azure/mql_asset_objects.go
```diff
@@ -74,6 +74,17 @@ func convertInstance(vm instance, subscription string, tc *providers.Config, obj
 			vm.Tags["azure.mondoo.com/computername"] = osProfileDict["computerName"].(string)
 		}
 	}
+	storageProfile, ok := vm.Properties["storageProfile"]
+	if ok {
+		if storageProfile, ok := storageProfile.(map[string]interface{}); ok {
+			osDisk, ok := storageProfile["osDisk"]
+			if ok {
+				if osDisk, ok := osDisk.(map[string]interface{}); ok {
+					vm.Tags["azure.mondoo.com/ostype"] = osDisk["osType"].(string)
```
preslavgerchev: is it guaranteed that `osDisk` will have an `osType` property? I am not sure if the conversion won't panic if such a prop is missing

vjeffrey: 🤷  may as well add a check to make sure! ill do that now

preslavgerchev (approved):

PR #1314, opened 2023-06-27
⭐️ add support for GCP snapshot scanning

chris-rock:
follow up on https://github.com/mondoohq/cnquery/pull/1311

This change allows users to scan GCP compute snapshots directly:

```bash
cnquery shell gcp snapshot suse12 --project-id my-project-id
→ discover related assets for 1 asset(s)
→ resolved assets resolved-assets=1
→ found target volume device name=/dev/sdb3
  ___ _ __  ___ _ __   ___  ___ 
 / __| '_ \/ __| '_ \ / _ \/ __|
| (__| | | \__ \ |_) |  __/ (__ 
 \___|_| |_|___/ .__/ \___|\___|
   mondoo™     |_|              
cnspec> asset.platform
asset.platform: "sles"
cnspec> asset.version
asset.version: "12.5"
cnspec> packages
packages.list: [
  0: package name="release-notes-sles" version="12.5.20200504-3.11.1"
  1: package name="libqrencode3" version="3.4.3-1.31"
  2: package name="lifecycle-data-sle-module-toolchain" version="1-3.15.1"
  3: package name="yast2-firewall" version="3.4.0-6.3.2"
  4: package name="recode" version="3.6-663.62"
  5: package name="sle-module-legacy-release-POOL" version="12-10.10.1"
  6: package name="SuSEfirewall2" version="3.6.312.333-3.13.1"
  7: package name="gamin-server" version="0.1.10-11.19"
...
```

PR #1315, opened 2023-06-27
⭐️ allow users to use latest GCP instance snapshots instead of creating new ones

chris-rock:
after #1314

With this change we introduce a new flag `--use-latest-snapshot` and allow users o use existing snapshots of instances if they exist. If no snapshot exists we still fallback to create a instance disk clone.

```bash
cnspec scan gcp instance sles12 --project-id my-project-id --zone us-central1-a --use-latest-snapshot
→ no Mondoo configuration file provided. using defaults

 sles12 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% score: C


Asset: sles12
-------------

Checks:
✕ Fail:  D  20  Ensure auditing for processes that start prior to auditd is enabled
✕ Fail:  D  20  Ensure successful file system mounts are collected
✕ Fail:  C  40  Ensure Advanced Intrusion Detection Environment (AIDE) is installed
✓ Pass:  A 100  Ensure rsh server is stopped and not enabled
✕ Fail:  F   0  Ensure secure permissions on /etc/group- are set
✓ Pass:  A 100  Ensure Avahi server is stopped and not enabled
✕ Fail:  D  20  Ensure system accounts are non-login
✓ Pass:  A 100  Ensure secure permissions on /etc/group are set
! Error:        Ensure rsyslog default file permissions configured
✓ Pass:  A 100  Ensure prelink is disabled
✓ Pass:  A 100  Ensure auditd is installed
✓ Pass:  A 100  Ensure X Window System is not installed
! Error:        Ensure access to the su command is restricted
✕ Fail:  D  20  Ensure session initiation information is collected
✕ Fail:  F   0  Ensure broadcast ICMP requests are ignored
✕ Fail:  D  20  Ensure login and logout events are collected
...
```

### DIFF
--- a/motor/providers/gcpinstancesnapshot/snapshot.go
+++ b/motor/providers/gcpinstancesnapshot/snapshot.go
@@ -10,6 +10,8 @@ import (
 
 	"github.com/rs/zerolog/log"
 	"go.mondoo.com/cnquery/motor/motorid/gce"
+	"go.mondoo.com/ranger-rpc/codes"
+	"go.mondoo.com/ranger-rpc/status"
 	googleoauth "golang.org/x/oauth2/google"
 	"google.golang.org/api/cloudresourcemanager/v3"
 	"google.golang.org/api/compute/v1"
@@ -134,15 +136,15 @@ func (sc *SnapshotCreator) SnapshotInfo(projectID, snapshotName string) (snapsho
 }
 
 // searchLatestSnapshot looks for the latest available snapshot for the instance
-func (sc *SnapshotCreator) searchLatestSnapshot(projectID, sourceDiskUrl string) (string, error) {
+func (sc *SnapshotCreator) searchLatestSnapshot(projectID, sourceDiskUrl string) (string, time.Time, error) {
 	ctx := context.Background()
+	latestSnapshotTimestamp := time.UnixMilli(0)
 
 	computeService, err := sc.computeServiceClient(ctx)
 	if err != nil {
-		return "", err
+		return "", latestSnapshotTimestamp, err
 	}
 
-	latestSnapshotTimestamp := time.UnixMilli(0)
 	var latestSnapshot *compute.Snapshot
 
 	req := computeService.Snapshots.List(projectID)
@@ -168,10 +170,14 @@ func (sc *SnapshotCreator) searchLatestSnapshot(projectID, sourceDiskUrl string)
 		}
 		return nil
 	}); err != nil {
-		return "", err
+		return "", latestSnapshotTimestamp, err
 	}
 
-	return latestSnapshot.SelfLink, nil
+	if latestSnapshot == nil {
+		return "", latestSnapshotTimestamp, status.Error(codes.NotFound, "no snapshot found")
+	}
+
+	return latestSnapshot.SelfLink, latestSnapshotTimestamp, nil
 }
 
 // createDisk creates a new disk
--- a/motor/providers/gcpinstancesnapshot/snapshot.go
+++ b/motor/providers/gcpinstancesnapshot/snapshot.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"time"
 
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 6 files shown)


motor/providers/gcpinstancesnapshot/snapshot.go
```diff
@@ -97,28 +98,92 @@ func (sc *SnapshotCreator) InstanceInfo(projectID, zone, instanceName string) (i
...
+				latestSnapshot = snapshot
+				latestSnapshotTimestamp = snapshotCreated
+			}
+		}
+		return nil
+	}); err != nil {
+		return "", err
+	}
+
+	return latestSnapshot.SelfLink, nil
```
czunker: This panics, when there is no snapshot:
```
cnquery scan gcp instance cz-test-cnquery-pr2 --zone europe-west1-b --project-id mondoo-dev-262313 --use-latest-snapshot --verbose
...
DBG run resolver resolver="GCP Compute Instance Resolver" resolver-id=gcp-compute-instance
DBG connecting to asset cz-test-cnquery-pr2
DBG establish connection to asset connection=gcp-compute-instance insecure=false
DBG establish motor connection
...
DBG platform> detected os family=["debian","linux","unix","os"] platform=debian
DBG local> run command curl --noproxy '*' -H Metadata-Flavor:Google http://metadata.google.internal/computeMetadata/v1/project/project-id?alt=json
DBG local> run command curl --noproxy '*' -H Metadata-Flavor:Google http://metadata.google.internal/computeMetadata/v1/instance/id?alt=json
DBG local> run command curl --noproxy '*' -H Metadata-Flavor:Google http://metadata.google.internal/computeMetadata/v1/instance/name?alt=json
DBG local> run command curl --noproxy '*' -H Metadata-Flavor:Google http://metadata.google.internal/computeMetadata/v1/instance/zone?alt=json
[signal SIGSEGV: segmentation violation code=0x1 addr=0xe8 pc=0x80ebfce]

goroutine 35 [running]:
go.mondoo.com/cnquery/motor/providers/gcpinstancesnapshot.(*SnapshotCreator).searchLatestSnapshot(0x8fa4360?, {0x7ffd0a73b75e, 0x11}, {0xc000a84a10, 0x6f})
	/home/christian/workspace/mondoo/github.com/cnquery/motor/providers/gcpinstancesnapshot/snapshot.go:174 +0x1ce
go.mondoo.com/cnquery/motor/providers/gcpinstancesnapshot.New(0xc00059fc80)
	/home/christian/workspace/mondoo/github.com/cnquery/motor/providers/gcpinstancesnapshot/provider.go:132 +0x75e
...
```

With the above command, I tried to scan an instance which has no snapshots.

vjeffrey:
if a user doesn't specify this new flag, do you create a snapshot every time, even if there's a recent one that exists?

default behavior with the aws ebs scanning is that if there's a snapshot that's less than...i think 8 hours old, we just use that instead of creating one. we should align on expected functionality for vm scanning via snapshots so we don't make it too confusing

chris-rock:
> if a user doesn't specify this new flag, do you create a snapshot every time, even if there's a recent one that exists?
> 
> default behavior with the aws ebs scanning is that if there's a snapshot that's less than...i think 8 hours old, we just use that instead of creating one. we should align on expected functionality for vm scanning via snapshots so we don't make it too confusing

I agree. I like the approach you described. Lets do the following:

- we use a recent snapshot when it is newer than 8 hours by default
- we introduce a flag to force the creation of a new snapshot eg. `--generate-snapshot`

motor/providers/gcpinstancesnapshot/snapshot.go
chris-rock: Great catch, will adjust the code.

chris-rock:
@vjeffrey @czunker This PR has been updated and is ready for re-review. The logic changed to the behavior we have for ebs snapshots. We only create a new snapshot when its older than 8 hours.

vjeffrey (approved):

PR #1323, opened 2023-06-28
🧹 A few other minor tweaks to help

tas50:
More syncs from cnspec this time so we can get these docs generating nicely


czunker (approved):
Some nice catches. Thanks @tas50

PR #1327, opened 2023-06-30
🧹 warn, do not fail, on additional resources when creating ec2 instance resource

vjeffrey:
if a user does not have permissions to describe vpcs, they should still be able to see the rest of the ec2 instance configuration

### DIFF
--- a/resources/packs/aws/aws_ec2.go
+++ b/resources/packs/aws/aws_ec2.go
@@ -701,26 +701,8 @@ func (s *mqlAwsEc2) gatherInstanceInfo(instances []types.Reservation, imdsvVersi
 				}
 				mqlDevices = append(mqlDevices, mqlInstanceDevice)
 			}
-			sgs := []interface{}{}
-			for i := range instance.SecurityGroups {
-				// NOTE: this will create the resource and determine the data in its init method
-				mqlSg, err := s.MotorRuntime.CreateResource("aws.ec2.securitygroup",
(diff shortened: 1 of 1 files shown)


czunker (approved):
Nice! Thanks @vjeffrey

resources/packs/aws/aws_ec2.go
```diff
@@ -735,30 +717,52 @@ func (s *mqlAwsEc2) gatherInstanceInfo(instances []types.Reservation, imdsvVersi
...
+
+			for i := range instance.SecurityGroups {
+				// NOTE: this will create the resource and determine the data in its init method
+				mqlSg, err := s.MotorRuntime.CreateResource("aws.ec2.securitygroup",
+					"arn", fmt.Sprintf(securityGroupArnPattern, regionVal, account.ID, core.ToString(instance.SecurityGroups[i].GroupId)),
+				)
+				if err == nil {
+					sgs = append(sgs, mqlSg)
+				} else {
+					log.Warn().Err(err).Msg("unable to create security groups resource for instance")
```
chris-rock, line 740: This does not feel correct to me. 

I think we need to change:

```
// List of security groups (ids) associated with the instance
securityGroups []aws.ec2.securitygroup
```

to lazy load:

```
// List of security groups (ids) associated with the instance
securityGroups() []aws.ec2.securitygroup
```

This would still show the error but users can see the rest of ec2 instance. I suggest we store the security groups in the cache of the resources and just use that one when MQL calls the GetSecurityGroups() methods

chris-rock (changes requested):

vjeffrey:
this is outdated, closing

PR #1329, opened 2023-07-02
:broom: replace cockroachdb/errors with standard library errors

mariuskimmina:
Replacing the cockroachdb `errors.Wrap` with `errors.Join` which is available since go 1.20.

Related discussion: https://github.com/orgs/mondoohq/discussions/1328#discussioncomment-6335615

### DIFF
--- a/apps/cnquery/cmd/config/config.go
+++ b/apps/cnquery/cmd/config/config.go
@@ -1,7 +1,7 @@
 package config
 
 import (
-	"github.com/cockroachdb/errors"
+	"errors"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/viper"
 	"go.mondoo.com/cnquery"
@@ -15,7 +15,7 @@ func ReadConfig() (*CliConfig, error) {
 	var opts CliConfig
 	err := viper.Unmarshal(&opts)
 	if err != nil {
-		return nil, errors.Wrap(err, "unable to decode into config struct")
+		return nil, errors.Join(err, errors.New("unable to decode into config struct"))
 	}
 
 	return &opts, nil
--- a/apps/cnquery/cmd/plugin.go
+++ b/apps/cnquery/cmd/plugin.go
@@ -5,7 +5,7 @@ import (
 	"github.com/spf13/viper"
 	"os"
 
-	"github.com/cockroachdb/errors"
+	"errors"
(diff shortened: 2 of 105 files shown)


mariuskimmina:
Currently seeing an error in the tests of the Terraform provider

```
--- FAIL: TestLoadHclBlocks (0.00s)
    hcl_parser_test.go:24:
                Error Trace:    /home/marius/projects/mondoo/cnquery/motor/providers/terraform/hcl_parser_test.go:24
                Error:          Not equal:
                                expected: 5
                                actual  : 4
                Test:           TestLoadHclBlocks
```

Not yet sure how it can be related to this change but the tests passes on main so it has to be.

mariuskimmina:
This test is confusing me, here I tried a couple of different variations but only the `errors.Wrap()` seems to work (`gerrors` refers to the standard library errors in this case).
Just not returning anything there would also result in a passing test.

```
				if !d.IsDir() {
					if strings.HasSuffix(path, ".terraform/modules/modules.json") {
						modulesManifest, err = ParseTerraformModuleManifest(path)
						if errors.Is(err, os.ErrNotExist) {
							log.Debug().Str("path", path).Msg("no terraform module manifest found")
						} else {
							//return gerrors.Join(err, fmt.Errorf("could not parse terraform module manifest %s", path)) --> test fails
							//return gerrors.New(fmt.Sprintf("could not parse terraform module manifest %s", path))  --> test fails
							//return fmt.Errorf("could not parse terraform module manifest %s: %w", path, err) --> test fails
							return errors.Wrap(err, fmt.Sprintf("could not parse terraform module manifest %s", path)) // --> test passes
						}
					}
```

tas50:
@mariuskimmina Can you rebase this PR? There's a good number of conflicts from main.

mariuskimmina:
Hi @tas50 yea my fork was falling quite far behind, I rebased and that also fixed the weird failing test it seems

mariuskimmina:
As this touches so many parts of the codebase it will keep getting conflicts all time if not quickly merged. I'd thus close this for now as I still try to figure out what's going on in that failing test and then try to make this change and more but smaller PRs, potentially going one package at a time.

chris-rock:
I recommend doing this once v9 is released since we currently restructure the code quite a bit.

PR #1342, opened 2023-07-06
Detect services on raspbian Linux

tas50:
It's a systemd distro.

Fixes https://github.com/mondoohq/cnquery/issues/1341


czunker (approved):
Thanks @tas50

PR #1344, opened 2023-07-11
🧹 Unify default output for assessments and regular queries

### LINKED ISSUE #1277: Change output of default values for assessment queries
This is a follow up to https://github.com/mondoohq/cnquery/pull/1219#pullrequestreview-1453088832 .

https://github.com/mondoohq/cnquery/pull/1219 and https://github.com/mondoohq/cnquery/pull/1211 added additional output to assessment queries. But this output looks different than the output for other queries, especially the output for the default values.

We want to align the output for the default values with other queries:
```
0: user name="root" uid=0 gid=0 {
      home: "/root"
  }
```

czunker:
This outputs the resource defaults for assessment queires in the same way as for regular queries, The data of the assessment is appended as before, as the datatypes can vary.

Fixes #1277


czunker:
Only tested this against v8. Need to check how this works with the new resources.

czunker:
Before assessment queries and regular queries displayed the default fields of a resource in different ways:
```
cnquery run -c "users.all(home=='/tmp')" 
[failed] users.all()
  actual:   [
    0: user {
      gid:0
      uid: 0
      name: "root"
      home: "/root"
    }
...
```
vs.
```
cnquery run -c "users.list.where(home=='/root')" 
users.list.where: [
  0: user name="root" uid=0 gid=0
]
```

Not the assessment queries display the default fields in the same way:
```
cnquery run -c "users.all(home=='/tmp')" 
[failed] users.all()
  actual:   [
    0: user gid=0 uid=0 name="root" {
      home: "/root"
    }
....
```
`gid`, `uid`, and `name` are the default fields. `home` the fields checked for the assessment and appended in a generic way, as this might also be a dict, or a field of a dict, or ...

cli/printer/printer.go
```diff
@@ -52,6 +54,10 @@ var DefaultPrinter = Printer{
 	},
 }
 
+func (p *Printer) SetSchema(schema *llx.Schema) {
```
arlimus: given llx.Schema is an interface, do we need this to be a pointer?

cli/printer/printer_test.go
```diff
@@ -267,16 +270,8 @@ func TestPrinter_Assessment(t *testing.T) {
...
-				"      gid: 1000",
-				"      uid: 1001",
-				"    }",
-				"    1: user {",
-				"      name: \"chris\"",
-				"      gid: 1000",
-				"      uid: 1000",
-				"    }",
+				"    0: user name=\"christopher\" gid=1000 uid=1001 ",
+				"    1: user name=\"chris\" gid=1000 uid=1000 ",
```
arlimus, line 274: these are so much nicer, great job!

cli/printer/printer_test.go
```diff
@@ -381,140 +341,154 @@ func TestPrinter_Assessment(t *testing.T) {
...
-		// 		},
+		// {
+		// 	"users.all(groups.none(gid==0))\n",
+		// 	strings.Join([]string{
+		// 		"[failed] users.all()",
+		// 		"  actual:   [",
+		// 		"    0: user uid=0 gid=0 name=\"root\" {",
+		// 		"      groups.list: [",
+		// 		"        0: user group id = group/0/root",
+		// 		"        1: user group id = group/1001/chris",
```
arlimus, line 352: Am I right to assume that this child list still prints them in the old way? (feel free to resolve this comment if so, I can see that this is a follow-up if that's the case)

arlimus (changes requested):
Just one quick element with changes left

cli/printer/printer.go
czunker: I can't remember why I implemented it this way. It also works without a pointer and the code is nicer to read.

Thanks for the hint. I removed the pointer.

cli/printer/printer_test.go
czunker: Thanks

cli/printer/printer_test.go
czunker: No, the child list is completely missing:
```
cnquery run -c "users.all(groups.none(gid==0))" 
[failed] users.all()
  actual:   [
    0: user uid=0 gid=0 name="root" 
    1: user uid=1 gid=1 name="daemon"
```

But it looks like this query isn't evaluated correctly, at least not as I read it. When you take a look at the `daemon` user:
```
id daemon 
uid=1(daemon) gid=1(daemon) groups=1(daemon)
```

And cnquery v9.0.2 also does not show the list:
```
cnquery run -c "users.all(groups.none(gid==0))" 
[failed] users.all()
  actual:   [
    0: user {
      uid: 0
      gid: 0
      name: "root"
    }
    1: user {
      uid: 1
      gid: 1
      name: "daemon"
    }
...
```

There is something strange here, but it seems to be independent of this PR.

Ok, with this PR, I changed the expected output. But the tests were commented out with https://github.com/mondoohq/cnquery/pull/1374. So, how we return these child lists has changed during the v9 migration. I'd opt for solving this in a follow-up.

vjeffrey:
@czunker is this ready for re-review?

czunker:
> @czunker is this ready for re-review?

Yes. Already pinged @arlimus as he requested changes.

cli/printer/printer_test.go
arlimus: I think I mentioned this before: The `users` resource does not have a `groups` field. This is what is likely causing the behavior

arlimus (approved):

arlimus:
Let's move forward and continue this track! This is a long-running tricky problem to resolve effectively that will continue to evolve for us. Thank you for pushing it forward!

PR #1348, opened 2023-07-12
✨ auto-generate id+init on resources is optional

arlimus:
relevant for provider developers, makes things a lot more comfy


arlimus:
I'm closing this for the time, since the structure for v9 is still in flux.
As of today, we have decided on an update to the providers structure, which will make this change a lot easier. Once the overhaul is done, it will already include this change, so it still applies.

PR #1349, opened 2023-07-12
🐛 support variants for querypacks

arlimus:
I ran the experiment with https://github.com/mondoohq/cnquery-packs on the `scottford/terraform-asset-inventory-variants` branch. In this example, we have a querypack with queries in a group that use variants. The way the syntax is set up was not originally support. This PR enables it.


scottford-io (approved):
Just tested with new variant query pack and it worked great!

Thanks @arlimus

PR #1353, opened 2023-07-13
🐛 fix variant query lookup

arlimus:
When used locally, we miss to store the bundle-level queries, so that the variant lookup fails. This fixes the issue and allows variant queries to be looked up correctly.


imilchev (approved):
Tested it and it's working. Thanks! 🎉

Issue #1321, opened 2023-06-28
run a specific query pack​ fails


michaelkrieg:
**Describe the bug**
The documentation states that I can run a specific query by setting `--query-id`:
https://mondoo.com/docs/cnquery/cnquery-run-pack/#run-a-specific-query-pack.
This seems not to work (anymore?).

**To Reproduce**

`cnquery scan aws -f mondoo-aws-inventory.mql.yaml --query-id mondoo-asset-inventory-aws-enabled-regions`

**Expected behavior**
Only the given query gets executed.

**Screenshots or CLI Output**
Use "cnquery scan aws [command] --help" for more information about a command.
unknown flag: --query-id

**Desktop (please complete the following information):**

- OS: macOS
- OS version:  13.4.1 (22F82)

**Additional context**
cnquery 8.15.0

mariuskimmina:
Started looking into this. 
First the option `--query-id` seems to have been renamed to '--querypack` so that the command should be something like this:

```
./cnquery scan local -f core/mondoo-linux-inventory.mql.yaml --querypack mondoo-linux-platform
```

I'll update the docs here later. This `--querypack` option doesn't seem to work as expected either tho as I get the following errors:

```
FTL failed to run scan error="all available packs filtered out. nothing to do."
```

Even tho the uid I am providing should be valid. 

Will update once I found out more.

michaelkrieg:
thanks for your feedback @mariuskimmina. I can confirm that `--querypack` does not do the trick.

PR #1357, opened 2023-07-15
remove non-existing query-id option from README

mariuskimmina:
#1321 
I can't find any reference for the option `query-id` other than the README. 
I have also been trying to find another option to run a single query from a querpack but I don't think we currently have an option for that. I think that would a good option to have and would leave the above mentioned issue open until we can actually support this case.

I have also briefly looked through older versions of the code base but wasn't able to find any signs of `query-id` existing in the past.


chris-rock (approved):
Thank you @mariuskimmina

PR #1358, opened 2023-07-16
🧹 decouple inventory manager from providers 

arlimus:
Shoving the entire MQL runtime into providers is not very helpful, we should keep them slim. This change moves the asset resolution with credentials to the calling runtime (cnquery/cnspec) instead of the plugin. This keeps plugins more focused and slim, fewer things for developers to know.

get closer to turning the tests green

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -140,36 +140,6 @@
         "--log-level",
         "debug"
       ],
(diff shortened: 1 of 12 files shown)


chris-rock (approved):
super nice

PR #1360, opened 2023-07-16
✨ migrate os.file resource

arlimus:
Updating the testing next, the current toml-based system does not work well anymoer since we have replaced the underlying connection. The recording also cannot cross-link resources yet...


chris-rock (approved):
Thank you!

PR #1361, opened 2023-07-16
✨🐛 pretty json recording + cross-connecting resources

arlimus:
After #1360 

Add a hidden flag to record to pretty json, because they are easier to modify for testing (and track for diffing). Additionally, add the ability to cross-connect resources in recordings, which was not supported.

One more step to fix the test utilities on the way to turning tests green...


chris-rock (approved):
Thank you @arlimus

PR #1362, opened 2023-07-17
⭐ file tests + testutils + 🐛 recordingdata

arlimus:
A few things in this one:
1. Migrate file tests to the provider system
2. Switch from TOML to JSON files and align with the new recording format
3. Build out testutils for shared helpers across all test suites
4. 🐛 fix an issues with int vs float from loading recording data
5. 🐛 fix incorrect propagation of errors (which crashed on trying to call on nil)

I suspect we will rewrite providers/mock. I added one more fix on file while still working on TOML, but after the latest change it may all go out and get replaced.

It was also incredibly easy using the new tests with the json recordings. For this example, I ran (and later modified a bit) the recording via:

```
cnquery shell --record --prettyjson --recording providers-sdk/v1/testutils/testdata/arch.json
```


chris-rock (approved):
Thank you @arlimus

PR #1371, opened 2023-07-17
⭐ migrate os.users+groups

arlimus:
- this includes file.user+group to be connected
- allow recordings to load lists of resources (and embedded references)
- prevent deadlocks in users+groups while allowing them to load efficiently


chris-rock (approved):
Thank you @arlimus

PR #1374, opened 2023-07-18
✨ migrate where for resources

arlimus:
- restore resourceWhere to the new plugin system
- only set resource IDs when necessary (in LR generator)
- create resource with ID support
- sort resources in recording

only 7 remaining, but we need more os resources


chris-rock (approved):
Thank you @arlimus

PR #1375, opened 2023-07-18
⭐ migrate packages to plugin system

arlimus:
After https://github.com/mondoohq/cnquery/pull/1374

Additionally, this forces LR builds to avoid init-lockups because the global methods are now initialized during the `init()` call and not before.


chris-rock (approved):
Nice nice nice

PR #1346, opened 2023-07-11
🐛 maintain correct binding chain when adding assessment fields to defaults

jaym:
This fixes an issue where the call binding chain was not being maintained when adding fields from the assessment block to the default fields block

For example:
```
go run apps/cnquery/cnquery.go run --ast -c '
entires = ["/etc/hosts"]
entires {
    files.find(from: _, type: "file").all( permissions.other_readable == false )
}
'
```
was producing:
```
-> block 1
   entrypoints: [<1,3>]
   1: [
     0: "/etc/hosts"
   ]
   2: ref<1,1>
   3: {} bind: <1,2> type:[]block (=> <2,0>)
-> block 2
   entrypoints: [<2,6>]
   1: _
   2: files.find bind: <0,0> type:files.find ("from", ref<2,1>, "type", "file")
   3: list bind: <2,2> type:[]file
   4: $whereNot bind: <2,2> type:files.find (ref<2,3>, => <3,0>)
   5: list bind: <2,4> type:[]file
   6: $all bind: <2,5> type:bool
   7: {} bind: <2,5> type:[]block (=> <4,0>)
-> block 3
   entrypoints: [<3,4>]
   1: file id = context
   2: permissions bind: <3,1> type:file.permissions
   3: other_readable bind: <3,2> type:bool
   4: == bind: <3,3> type:bool (false)
-> block 4
   entrypoints: [<4,2> <4,3> <4,5> <4,6>]
   1: file id = context
   2: path bind: <4,1> type:string
   3: size bind: <4,1> type:int
   4: permissions bind: <4,1> type:file.permissions
   5: string bind: <4,4> type:string
   6: other_readable bind: <4,1> type:bool
```

In this example, `<4,6>` shohuld bind to `<4,4>` but incorrectly assumed it should bind to `<4,1>`.

As part of this, I've filtered out any function calls that have references. This should keep https://github.com/mondoohq/cnquery/issues/1339 from panicking

Fixes https://github.com/mondoohq/cnquery/issues/1316

PR #1376, opened 2023-07-18
🐛 Remove printer tests that fail

jaym:
These tests started failing with #1346. The issue here is that code only copies over things from the assessment block to the defaults block that ultimately depend on the context. It ignores everything else as it would not always be possible to merge the blocks. In this case, we could technically merge those correctly, but the code as written assumes there is only one root, the context. Since the queries as written don't make sense, I'm removing them. I will come back and make changes to allow for that case in another change


czunker:
@jaym Is this still needed? Tests were passing for the PRs the last weeks.

jaym:
no, looks like these are gone

PR #1377, opened 2023-07-19
✨ migrate authorizedkeys

arlimus:
Also:
- 🐛 fix recording loading, so that asset information is set on the connection
- ✨ allow users to be loaded via user(name: ...); this is more of a backport


chris-rock (approved):
Thank you @arlimus

arlimus:
merged, the overhaul will lay the foundation to fix the tests

PR #1379, opened 2023-07-20
🐛 fix discovery for private github repos

imilchev:
The wrong function was used to list the repos. Now it's using the same function that is used by the MQL resources


chris-rock (approved):
Thank you @imilchev

PR #1381, opened 2023-07-24
🧹 remove tags from aws global dynamodb table discovery

vjeffrey:
global dynamo db tables do not have tags


chris-rock (approved):
Thank you @vjeffrey

PR #1382, opened 2023-07-24
Dom/connection overhaul

arlimus:
After https://github.com/mondoohq/cnquery/pull/1377

🪓 overhaul provider connections
    
This changes a few more core aspects:
1. ParseCLI never connects to the asset. As such, there is no need to track any possible connections anymore. This also means it cannot do things that would require a connected asset, like platform detection, discovery, and resolution. All gone. This behavior was adding complexity, breaking responsibilities, and was exploding in code with what comes next...
2. Connections now only take on asset. No more inventory. Same point as above, it was adding unnecessary amount of complexity that was breaking the primariy purpose and colliding with the next point...
3. Asset connections now hold a callback. The primary use in this PR is to pull recording data from the caller. This means that we don't go through the old call-stack anymore (host:getData => plugin:create => host:recognizeID?->storeData => plugin:store) and instead this is now directed through the plugin (host:getData => plugin:create->getRecording? => host:giveRecording => plugin:store). This is critically important, because any plugin provider at any point can call CreateResource, and whenever that happens we need to be able to (a) feed it from cache, (b) record it, and (c) cache it in the plugin. This removes complexity from the caller.
    
As a consequence, the resource generator has been overhauled. It now generates 2 methods in every resource: NewResource and CreateResource.
- Only NewResource goes through argument parsing
- argument parsing is now disconnected from the resource object (you can't just write to the resource anymore). See providers-sdk/v1/plugin/plugin.go for a full explanation
- NewResource is used when users write something in MQL and call it (e.g. user('bob'), asset{*} etc)
- CreateResource is used when resources create other resources (and thus don't want to go through all the trouble of argument parsing). It is optimized for e.g. resource.list (eg users, services, etc)


chris-rock (approved):
Yeah @arlimus

arlimus:
This still won't fix the tests, but it lays the foundation to have much more meaningful tests now.

PR #1385, opened 2023-07-25
🐛 load resource async from recording

arlimus:
- 🐛 fix the detection of new init-style argument parsers for resources
- ✨ add the provider that generated an error message to a few of the auto-generated messages (more to come)
- 🐛 handle errors from resource ID-generation
- 🧹 stop exposing data setters for call-arguments in resources
- 🐛 avoid loading resource fields from within providers when calling recordings (needs recursive loading in a follow-up, for now it prevents a crash)
- and most importantly: 🟢 fix 3/4 authorized keys tests (without actually storing the authorized keys resource or fields!) (which caused this giant rewrite)


chris-rock (approved):
Woot

PR #1386, opened 2023-07-26
🧹 enable linters

imilchev:
I was just working on a bug and I discovered that all linters were disabled. We are barely checking for any issues at the moment. If those were enabled, we wouldn't have had the bug I was working on, since it was related to an error that was never checked


imilchev:
will circle back to this later

PR #1389, opened 2023-07-27
🐛 allow bi-directional loading of recordings

arlimus:
- re-expose SetAllData because we need it in calling providers to set resources from recordings
- types can now check if they contain a resource; this is important because loading resources from recordings requires proper initialization
- 🐛 a resource that cannot be found in a provider may have not been initialized from the caller; this happens when we load data from a recording, but are missing a number of fields in the recording. This is actually a great case where the recording only needs the data it cannot compute on the fly

This makes it 4/4 for authorized keys tests


chris-rock (approved):
Thank you @arlimus

PR #1390, opened 2023-07-27
🐛 load embedded recording resources inside of plugins

arlimus:
After https://github.com/mondoohq/cnquery/pull/1389
    
Sometimes a plugin provider will decide to load a resource that comes from a recording, which has embedded child resources. This does not pass the regular flow where the host (cnquery) would be able to detect that such a resource was requested and could intervene. Instead, all of this plays inside of the plugin. Thus, we need to know when embedde resources are requested and load them on the fly.

Note: we don't want to load it all in the beginning, because (1) it could mean a lot of data which is never used and (2) (more importantly) it can lead to a chain-reaction of resources that are all connected (which we love to do because it's awesome) and then lock the provider into a loading loop (which needs additional handling to be terminated). This mechanism just does it async when the field is requested at the cost of a bit of overhead when the request is done, but saving a lot of effort especially as resources are getting more and more connected.

A few more notes:
- 🐛 fix incorrect initResource override hint to developers
- 🐛 do a much better job at detecting embedded resources in LR
- ✨ tie CreateResource to the plugin Runtime
- 🐛 avoid broken casting when we can't load a resource field from records (it tried to cast nil to the result type)
- 🐛 fix the user init call (it still used the old syntax)

### DIFF
--- a/cli/printer/printer_test.go
+++ b/cli/printer/printer_test.go
@@ -185,7 +185,7 @@ func TestPrinter_Assessment(t *testing.T) {
 				"    expected: == 1",
 				"    actual:   \"development\"",
 				"  [failed] user.authorizedkeys.file",
-				"    error: failed to create resource 'user': user 'notthere' does not exist",
+				"    error: cannot find user with name 'notthere'",
 				"",
 			}, "\n"),
 		},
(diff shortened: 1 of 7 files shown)


chris-rock (approved):
Great @arlimus

PR #1396, opened 2023-07-27
⭐ migrate files.find + pam

arlimus:
After https://github.com/mondoohq/cnquery/pull/1394

In addition to migrating these two heavyweights, it also:
- 🐛 fix circular dependency in mqlc
- NOTE: deactivate a few mqlc tests until the final resources come in (to be done by me)
- 🐛 add names to generated schema for implicit resources

Also note: since we are using the core provider for testing, I pulled in the generated schema JSON


chris-rock (approved):
Thank you @arlimus

PR #1398, opened 2023-07-28
✨ add min version to schema

arlimus:
After https://github.com/mondoohq/cnquery/pull/1396

This is migrating a few fields from the manifest directly into the schema. I'm not sure if we want to distribute the manifest, we certainly don't want to load like we load the schema (which is broadly all of them when we search for things). That said, the version info is useful and should be contained (plus we have most of the structure there already, so it's more efficient this way anyway).

Also this squashes a few more tests and turns them 🟢


chris-rock (approved):
Great stuff @arlimus

PR #1399, opened 2023-07-28
✨ migrate parsers (json, yaml, date)

arlimus:
After https://github.com/mondoohq/cnquery/pull/1398

I'm not fully happy with this yet, in fact I think I'm going to split this in a follow-up PR. Right now, the json+yaml parsers are in os, because they depend on core and I didn't put file into core anymore. Technically, if your read from the OS then that's the dependency... however, we could view file as a more generic concept that goes into core, but doesn't get implemented fully there, i.e. you cannot read the file from core alone. Another option is to put parse.json into core, but then we'd have to overload the init method into the os package so that it works with file... the more I think about this, the more the former approach might be better, where a baseline file interface lives in core, without full implementation. Just have to get the overlays right for that approach


chris-rock:
I like the idea to have a file interface, think about a github provider that allows you to read a file from the repository. It is still a file and I should be able to load a file directly from the git repo.

chris-rock (approved):
Yeah @arlimus

PR #1400, opened 2023-07-28
add testdata for files.find

arlimus:
After https://github.com/mondoohq/cnquery/pull/1399

This was just missing some testdata in the recording, so I picked a few files from a local record


chris-rock (approved):
Thank you @arlimus

PR #1405, opened 2023-07-29
🐛 ensure schema has provider field set

arlimus:
Before this it was possible for the field to not be set, depending on the order that the map was iterated over. Fix it by setting provider fields first, then setting them on implicit resource fields wherever necessary.


chris-rock (approved):
Thank you @arlimus

PR #1406, opened 2023-07-29
✨ make providers/install for convenience

arlimus:
After https://github.com/mondoohq/cnquery/pull/1405 (since it stops the schema from flip-flopping)


chris-rock (approved):
Nice @arlimus

PR #1407, opened 2023-07-29
✨ structure provider plugins into folders

arlimus:
In the old model, all plugin files were flat in a plugins folder. That can get out of hand because a ton of files will be all over the place. 

Instead, put all plugin files into a folder with the plugin name, e.g.:

```
.config/mondoo/providers
└── os
    ├── os
    ├── os.json
    └── os.resources.json
```


chris-rock (approved):
go go go @arlimus

PR #1412, opened 2023-07-31
⭐ migrate eol, advisories, and cves resources

arlimus:
This not only migrates a number of pivotal resources, it also establishes schema-merging.

**Resources migration** 

The `platform` keyword had previously been deprecated, but we hadn't yet migrated some of its critical child resources, namely eol, advisories, and cves. These retain their previous position, despite the fact that most of platform has now been removed (in favor of asset). 

We will tackle these remaining fields shortly. Despite the fact that we can't remove them in v9 (due to our maintenance policy) we will target them for an eventual v10 removal, if the changes can make it in time for the upcoming release.

Also, these resources have now moved from core to os, which is not intuitive and not where we want them to be long-term. In fact asset.vulnerabilityReport is also found in os! All of this is explained when we look at the next section:

**Schema merging**

While `core` may establish some of these principles, they cannot be implemented universally this way. To clarify: Core is the foundation to all other providers. Other providers can know about core, but core cannot know about other providers. Thus, implementing any type of cve/advisory/eol code into core is not possible, since it only knows about itself and not about anything else.

However, other providers can implement the affected resources and they can even implement the same resource fields. For example, os can implement the corpus to generate EOL information about an operating system. The code for that - which is entirely OS-specific(!) - now only lives in the OS resource.

We will create a few foundations in core which will serve as a base for overwriting (which has not yet happened in this PR). That said: the mechanism for overrides is there today.


providers/os/resources/os.lr
```diff
@@ -10,6 +10,95 @@ alias os.base.service = service
...
+  // Product Name
+  product string
+  // Product Version
+  version string
+  // End-of-Life date for the product
+  date() time
+}
+
+// Information about the platform end-of-life. Deprecated, will be removed in v10.
+// Please use asset.eol instead.
```
chris-rock, line 28: Lets define this in follow-up PRs

chris-rock (approved):
Thank you @arlimus

PR #1413, opened 2023-07-31
✨ update recording cli options 

arlimus:
After https://github.com/mondoohq/cnquery/pull/1412

... to:

```bash
> cnquery --record file.json

> cnquery --use-recording file.json

> cnquery --record file.json --pretty
```


chris-rock (approved):
Thank you for the harmonization of the flags

PR #1414, opened 2023-07-31
🐛 pre-load all providers

arlimus:
After https://github.com/mondoohq/cnquery/pull/1413

The mechanisms where it was only loaded from disk when necessary breaks because we need the CLI options to be loaded before we de-dupe any providers. This is done via name, so we have to have their metadata loaded. This makes the code simpler and removes any surprises with having a provider that has no metadata.


chris-rock (approved):
Thank you @arlimus

PR #1418, opened 2023-08-01
⭐️ add suse 11 platform detection

chris-rock:
This adds platform detection of suse systems without /etc/os-release


atomic111 (approved):
@chris-rock thanks for the enhancement

PR #1420, opened 2023-08-02
⭐ support loading fields across multiple providers

arlimus:
This way calls like `asset{*}` will fundamentally work. Parts of the resource resides with the original provider for it, parts reside with the override.

On the example we still have to fix the `eol` call within the asset resource, since it is forwarded...


chris-rock (approved):
Nice

PR #1422, opened 2023-08-02
✨ release to ghcr

imilchev:
We always release to ghcr. For prod releases we add an extra job that copies the images from ghcr to dockerhub

### DIFF
--- a/.github/.goreleaser-edge.yml
+++ b/.github/.goreleaser-edge.yml
@@ -39,7 +39,7 @@ dockers: # https://goreleaser.com/customization/docker/
     dockerfile: Dockerfile-ubi
     image_templates:
       - "ghcr.io/mondoohq/{{ .ProjectName }}:edge-{{ .Version }}-ubi-amd64"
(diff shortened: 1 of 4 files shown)


.github/.goreleaser-edge.yml
```diff
@@ -38,7 +38,7 @@ dockers: # https://goreleaser.com/customization/docker/
     goarch: amd64
     dockerfile: Dockerfile-ubi
     image_templates:
-      - "mondoo/{{ .ProjectName }}:edge-{{ .Version }}-ubi-amd64"
+      - "ghcr.io/mondoohq/{{ .ProjectName }}:edge-{{ .Version }}-ubi-amd64"
       - "mondoo/{{ .ProjectName }}:edge-latest-ubi-amd64"
```
chris-rock: do we need to update the ubi container as well?

imilchev: we do. Nice catch!

chris-rock:
I think we do not need this for now, lets re-open when we need it

PR #1423, opened 2023-08-02
🐞 add non-interactive flag to the zypper command

atomic111:
During a cnspec scan on a system without access to the suse repositories, zypper waits for user input. This results in an unresponsive cnspec scan.


mick-lue (approved):
this will work as expected in the tested case.

chris-rock (approved):
Thank you @atomic111

PR #1424, opened 2023-08-02
🐛 fix sysv running service

chris-rock:


atomic111 (approved):
@chris-rock for the incredible fast fix

PR #1425, opened 2023-08-02
🐞 add non-interactive flag to the zypper command

chris-rock:
port of https://github.com/mondoohq/cnquery/pull/1423 to v8


atomic111 (approved):
@chris-rock thanks for the fix

PR #1430, opened 2023-08-03
🐛 attach cli PreRun on command-generation

arlimus:
Without this, PreRun for these new commands is not called. This means that things like `viper.BindPFlag` is not executed, which in term breaks later execution.


chris-rock (approved):
Thank you @arlimus

PR #1431, opened 2023-08-03
🐛 inject runtime into services

arlimus:
After https://github.com/mondoohq/cnquery/pull/1430

Currently, when creating a new set of local services, we are actually creating a new runtime. So far that never caused problems, because cnquery was monolithic. Now that is changing, and we want the correct runtime to be injected and used (ie not created anew) into services. This popped up as a bug when running scans, where part of the queries were not executable because providers were missing.

This is also bad behavior in the previous iteration, since we never used the correct runtime.


chris-rock (approved):
Nice catch @arlimus

PR #1433, opened 2023-08-03
🐛 use correct schema + conditional upstream config

arlimus:
After https://github.com/mondoohq/cnquery/pull/1432

1. Like previously stated, in the past we were taking a lot of shortcuts by instantiating schemas multiple times. Now that we have runtime-bound schemas (with specificly loaded plugin providers) we have to do this properly, i.e. schemas are handed over.
2. We do not always have an upstream config, so for the cases where we don't, we have to work around it. We don't want to set an upstream config when we didn't load one, so it needs a check now.


chris-rock (approved):
Thank you @arlimus

PR #1437, opened 2023-08-05
🌟 add network provider

arlimus:
Another major migration!

- migrate `tls`, `socket`, and `certificates` from `core` to a new `network` provider
- make `os` depend on both `network` and `core` (it doesn't really use `core` yet, so we may kick it out later...)
- migrate `parse.certificates` from `core` to `os`. It doesn't belong in `network`, since it operates on the OS. Which gets us into the fun part of this:
- **cross-call resources and fields between providers**. This allows us to `parse.certificates` inside of `os` and then call into `network` to create the actual `certificate` objects
- create a new `certificates` resource which is initialized from PEM and creates multiple `certificate` resources
- migrate `process` and `port` to OS. It's a bit debatable if port should remain here or if we split it between `network` and `os`...

There is still more resources to migrate to network, which are coming up next...


chris-rock (approved):
Great to see the cross-resource work running

PR #1443, opened 2023-08-05
🧹 Set the title of AWS accounts to be AWS Account

tas50:
Right now an AWS Account passes into this functionw with the name of "" and we set the name to "aws" and the title to "Amazon Web Services". The "aws" platform is actually the account so we should properly name it. This is important as time goes on with fine grained scanning where we need to distinguish between the resources and the main account.


tas50:
@vjeffrey tagging you on this one since you have the best background on any gotchas here

chris-rock (approved):
Thank you @tas50

PR #1444, opened 2023-08-05
✨ migrate dns, openpgp, and parse.openpgp

arlimus:
As with certificates: `parse.openpgp` goes to `os` now, while `openpgp` moves to `network`. There is also a new resource: `openpgp.entities` which is private (because at the moment not user-facing, but this could change) and it takes the content of OpenPGP and then does all the parsing / entity creation.


chris-rock (approved):
Yeah!

PR #1446, opened 2023-08-06
🐛 fix time conversion + core test migrations

arlimus:
Discovered a very nasty bug in the time conversion, that caused a few seconds of drift whenever larger timestamps were converted. This was caused by the way we handled nanoseconds in time conversions and led to 3-7 seconds of difference to what the time should have been. Tests have been added to guard against this.

A ton of core tests have been migrated to both `os` and `core` respectively. Wherever possible we moved them to `core`, since most are language tests. Long-term we expect to move to a more dedicated mock provider instead of abusing `os`, but for now it has the added benefit of being closer to integration testing.

Also - 🎉 - this is the first major removal of data from the original repo, since `core.lr.go` is now migrated. Just in case you were wondering why the last few migration PRs seemed to add thousands of lines, this one removes 17k lines ;)


chris-rock (approved):
Love more tests 🤣

PR #1452, opened 2023-08-07
✨ deduplicate errors

arlimus:
We print too many duplicate errors, it is a really bad user experience.
    
To fix this, a few steps had to be taken:
1. We had a number of utilitis at a non-scalable interface. As much as I hate having a generic `utils` folder, it seemed like the best idea. Moved sortx and stringx there (without renaming them any further, because it conflicts with string and sort in go native)
2. This repo https://github.com/pkg/errors has been archived for over a year. Remove all references to it.
3. That said, it has better methods for simple flows, like `errors.Wrap` which requires far less processing than `fmt.Errorf` (unless you really need that). For that reason, the method for `Wrap` has been copied over (without the callstack).
4. Using this new errors utility, we then added `Deduplicate` to take care of embedded errors. It can certainly be further improved, but it gets us the desired results.
5. This allowed us to migrate all cases where errors get wrapped to
   `multierr.Wrap`, while the rest moved to the standard go library.

This does leave us with a question of whether to follow Go's std `errors` or wrap to something like `cockroachdb/errors` instead. The latter already provides utility like `Wrap`, but is also crammed with other contents we don't necessarily need. Still, it's something we want to evaluate and unify.

Before:

image

After:

image


tas50:
This may be my favorite PR

chris-rock (approved):
Nice small wrap 🎁

PR #1454, opened 2023-08-07
✨ migrate logindefs, lsblk, mount, ntp.conf, shadow, yum

arlimus:
a few more windows and mac resources remain


chris-rock (approved):
more os resources!

PR #1456, opened 2023-08-07
🐛 Fix conversion of primitives with NO type

### LINKED ISSUE #957: `RawData` fails for some resources
**Describe the bug**
When I try to get the RawData of a resource/field, the data isn't converted because of typing problems. But the resource/field has a type.

I came across this problem while exposing query pack data:
```
{
              "name": "date",
              "description": "End-of-Life date",
              "path": "platform.eol.date",
              "value": "{\"error\":\"cannot convert primitive with NO type information\"}",
              "type": "time",
              "numElements": 1
            }
```

This happens here: https://github.com/mondoohq/cnquery/blob/a082a88998535a39ad991a4cdd5e73fd1bb472e4/llx/data_conversions.go#L678

Strange is, that it works for other fields of the same type:
```
{
              "name": "uptime",
              "description": "Current uptime",
              "path": "os.uptime",
              "value": "\"0000-01-05T04:29:31.423622144+00:53\"",
              "type": "time",
              "numElements": 1
            },
```

**To Reproduce**
Have to find out, how to isolate this.

2. Select '....'
4. Note the error

**Expected behavior**
The data should get converted.

**Screenshots or CLI Output**
Some more examples:
```
{
              "name": "vulnerabilityReport",
              "description": "Full advisory & vulnerability report",
              "path": "platform.vulnerabilityReport",
              "value": "{\"error\":\"cannot convert primitive with NO type information\"}",
              "type": "dict",
              "numElements": 1
            },
```

```
{
              "name": "jobEnvironment",
              "description": "Returns the agent execution environment",
              "path": "mondoo.jobEnvironment",
              "value": "{\"error\":\"cannot convert primitive with NO type information\"}",
              "type": "dict",
              "numElements": 1
            }
```

```
{
              "name": "params",
              "description": "Configuration values of this SSH server",
              "path": "sshd.config.params",
              "value": "{\"error\":\"cannot convert primitive with NO type information\"}",
              "type": "map[string]string",
              "numElements": 1
            }
```

czunker:
In case we have a resource without data, it also does not have a type. We need to catch this to prevent the error: `cannot convert primitive with NO type information`

This happens, e.g., when we test for the exists of a file before doing anything else with the resource.

Partially-Fixes #957

### DIFF
--- a/llx/data_conversions.go
+++ b/llx/data_conversions.go
@@ -461,7 +461,8 @@ func (r *Result) RawData() *RawData {
 
 	data := &RawData{}
 	if r.Data != nil {
-		if r.Data.IsNil() {
+		// The type can be empty, when we do not have data
+		if r.Data.IsNil() || types.Type(r.Data.Type).IsEmpty() {
 			data.Type = types.Nil
 		} else {
 			data = r.Data.RawData()


czunker:
Before:
Screenshot from 2023-08-07 10-12-02


and after in the UI:
Screenshot from 2023-08-07 10-22-20

llx/data_conversions.go
```diff
@@ -461,7 +461,8 @@ func (r *Result) RawData() *RawData {
 
 	data := &RawData{}
 	if r.Data != nil {
-		if r.Data.IsNil() {
+		// The type can be empty, when we do not have data
+		if r.Data.IsNil() || types.Type(r.Data.Type).IsEmpty() {
```
czunker, line 465: At this point, als `r.Data.(Map|Array|Value)` are empty. We could also add these as checks.

czunker:
I used this query as a test: https://github.com/mondoohq/cnspec-policies/blob/main/core/mondoo-linux-security.mql.yaml#L1084

PR #1459, opened 2023-08-07
🐛 fix hashi vault

chris-rock:
To use hashi-vault, use the following config:

```yaml
apiVersion: v1
kind: Inventory
metadata:
  name: mondoo-inventory
  labels:
    environment: production
spec:
  assets:
    - name: vsphere
      connections:
        - backend: vsphere
          host: 192.168.1.24
          credentials:
            - secret_id: test-key # uses the secret kv backend automatically
          discover:
            targets:
              - host-machines
  vault:
    name: local
    type: hashicorp-vault
    options:
      url: http://127.0.0.1:8200 # dev-server url
      token: <token>
```

### DIFF
--- a/motor/vault/hashivault/hashivault.go
+++ b/motor/vault/hashivault/hashivault.go
@@ -8,12 +8,14 @@ import (
 	"strings"
 
 	"github.com/hashicorp/vault/api"
+	"github.com/rs/zerolog/log"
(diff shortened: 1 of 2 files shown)


atomic111 (approved):
@chris-rock thanks for the quick fix

PR #1461, opened 2023-08-08
🧹 do not error on filepath.SkipDir when reading linux bios

imilchev:
When running a snapshot scan we can run into the issue that we log a `SkipDir` error when reading bios. That error is irrelevant and we should just return empty data

### DIFF
--- a/resources/packs/os/smbios/linux.go
+++ b/resources/packs/os/smbios/linux.go
@@ -95,5 +95,10 @@ func (s *LinuxSmbiosManager) Info() (*SmBiosInfo, error) {
 		return nil
 	})
 
-	return &smInfo, wErr
+	// If the error is SkipDir we can safely ignore it
+	if wErr != nil && wErr != filepath.SkipDir {
+		return nil, wErr
+	}
+
+	return &smInfo, nil
 }


resources/packs/os/smbios/linux.go
```diff
@@ -26,7 +26,7 @@ func (s *LinuxSmbiosManager) Info() (*SmBiosInfo, error) {
 	root := "/sys/class/dmi/id/"
 
 	wErr := afs.Walk(root, func(path string, info os.FileInfo, fErr error) error {
-		if fErr != nil {
+		if fErr != nil || os.IsPermission(fErr) {
```
czunker: Should we somehow log the error, so users know why there is no data?
This is how it looks like for ports:
```
cnquery run -c "ports.listening" 
! mql[processes]> could not retrieve process information error="process 782341 does not exist: /proc/782341" pid=782341
Query encountered errors:
1 error occurred:
	* cannot read related processess: permission denied
ports.listening: [
  0: port port=53 protocol="tcp4" address="127.0.0.53" process.executable=no data available
  1: port port=2222 protocol="tcp4" address="127.0.0.1" process.executable=no data available
  2: port port=53 protocol="tcp4" address="192.168.122.1" process.executable=no data available
  3: port port=631 protocol="tcp4" address="127.0.0.1" process.executable=no data available
  4: port port=6463 protocol="tcp4" address="127.0.0.1" process.executable="Discord"
  5: port port=111 protocol="tcp4" address="0.0.0.0" process.executable=no data available
  6: port port=631 protocol="tcp6" address="[::1]" process.executable=no data available
  7: port port=111 protocol="tcp6" address="[::]" process.executable=no data available
]
```

imilchev: nvm... we don't want to do this

imilchev:
> Tried this on my local machine. It now looks like this:
> 
> ```
> Query encountered errors:
> failed to create resource 'machine.baseboard': could not retrieve smbios info for platform: open /sys/class/dmi/id/board_serial: permission denied
> machine.baseboard: no data available
> ```

that's the correct behaviour. We want to fix another error here
image

czunker (approved):
Thanks @imilchev

PR #1469, opened 2023-08-09
🐛 fixes for suse linux

chris-rock:
This fixes the following reported issues:

- cnspec scan unresponsive - "zypper --xmlout list-updates" missing -n flag
- handles suse system services specially

Followup on https://github.com/mondoohq/cnquery/pull/1424


atomic111 (approved):
@chris-rock thanks it is working!

PR #1483, opened 2023-08-10
🐛 fix arch specific provider builds

chris-rock:
This fixes the build of the plugins. All current available archs where accidentally just compiled as linux, therefore plugins do not work on macOS or other platforms.


scripts/provider_bundler.sh
```diff
@@ -76,13 +84,13 @@ build_bundle darwin arm64
...
-build_bundle linux armv6
-build_bundle linux armv7
+build_bundle linux arm 6
+build_bundle linux arm 7
 build_bundle linux ppc64le
 
 # Build Windows Architectures
-build_bundle windows amd64
-build_bundle windows arm64
+# build_bundle windows amd64
```
chris-rock, line 92: windows requires more fixes, coming in follow up PRs

imilchev (approved):

PR #1487, opened 2023-08-10
⭐️ slack provider for v9

chris-rock:
This PR ports over the slack provider to the new v9 format:

```coffeescript
cnquery> slack.users
slack.users.list: [
  0: slack.user id="USLACKBOT"
  ...
]
```

conversations also work as expected:

```
cnquery> slack.conversations
slack.conversations: [
  0: slack.conversation id="CEKDKSH7A" name="allgemein"
  ... 
]
```


Makefile
```diff
@@ -306,19 +310,18 @@ test: test/go test/lint
 test/go: cnquery/generate test/go/plain
 
 test/go/plain:
-	# TODO /motor/docker/docker_engine cannot be executed inside of docker
-	go test -cover $(shell go list ./... | grep -v '/motor/discovery/docker_engine')
+	go test -cover $(shell go list ./... | grep -v '/providers/slack')
 
 test/go/plain-ci: prep/tools providers/build
-	gotestsum --junitfile report.xml --format pkgname -- -cover $(shell go list ./... | grep -v '/vendor/' | grep -v '/motor/discovery/docker_engine')
+	gotestsum --junitfile report.xml --format pkgname -- -cover $(shell go list ./... | grep -v '/vendor/' | grep -v '/providers/slack')
```
chris-rock: this excludes providers from global testing, need to come up with something that does testing for each provider

providers/slack/resources/conversations.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	created := conversation.Created.Time()
+
+	var creator plugin.Resource
+
+	if conversation.Creator != "" {
+		creator, err = CreateResource(runtime, "slack.user", map[string]*llx.RawData{
```
imilchev, line 104: This won't work. I just ran into the same problem. `CreateResource` is used for creating new resources with pre-processed data. If you want to initialize a resource with "user" input and effectively call the `Init` function, you need tou use `NewResource`

providers/slack/resources/users.go
```diff
@@ -0,0 +1,259 @@
...
+		"isOwner":        llx.BoolData(user.IsOwner),
+	})
+}
+
+func (x *mqlSlackEnterpriseUser) id() (string, error) {
+	return "slack.enterpriseUser/" + x.EnterpriseId.Data + "/" + x.Id.Data, nil
+}
+
+// init method for user
+func (s *mqlSlackUser) init(args map[string]interface{}) (map[string]interface{}, *mqlSlackUser, error) {
```
imilchev, line 230: this is not correct and won't do anything. The signature of the function must be:
```go
func initSlackUser(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error) {
```
then you need to rebuild the provider

imilchev (commented):
Check if there are other cases where you used init functions or tried to call other mql resources. This is a bit confusing 😢

providers/slack/resources/users.go
chris-rock: great catch 🤣

providers/slack/provider/provider.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+}
+
+func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error) {
+	flags := req.Flags
+	if flags == nil {
+		flags = map[string]*llx.Primitive{}
+	}
+
+	conn := &inventory.Config{
```
preslavgerchev, line 37: this seems to be discarded, it's never put in the asset's connection

preslavgerchev: means that the token flag is likely ignored

vjeffrey (approved):

PR #1495, opened 2023-08-11
⭐️ migrate container.image to v9

imilchev:
This is needed for migrating the k8s provider. You can instantiate the resources like this:
```coffee
cnquery> container.image(reference: 'mondoohq/cnspec:latest'){*}
container.image: {
  identifierType: "tag"
  name: "index.docker.io/mondoohq/cnspec:latest"
  repository: container.repository id = index.docker.io/mondoohq/cnspec
  reference: "mondoohq/cnspec:latest"
  identifier: "latest"
}
```

The `reference` field is new. I added it so I can easily instantiate the resource from another provider.


providers/os/resources/os.lr
```diff
@@ -953,3 +953,29 @@ yum.repo {
   // Indicates if this repository is used as package source
   enabled() bool
 }
+
+// Container Image
+container.image {
+  // Image reference
+  reference string
```
czunker, line 960: I do not understand the difference between `reference` and `name`. Do you have an example?

The comment in the code shows the same value in my understanding.

imilchev: The reference is what you type in docker, for example, when you run a container. So `mondoohq/cnspec:latest`. However, the name is actually the resolved value of that reference, which is `index.docker.io/mondoohq/cnspec:latest`

czunker: Thanks!

czunker (approved):
LGTM. Thanks @imilchev

PR #1359, opened 2023-07-16
⭐️ native windows registry resource

chris-rock:
This PR adds native support for Windows registry keys when cnquery is executed locally. Specifically cnspec is going to profit from this change with a 3.8 times faster execution for CIS Windows 2022 benchmark.

Besides the performance improvements, this PR also adds support for multi-line string and binary data support for registry keys.

We deprecate `registrykey.properties` 

```
cnspec> registrykey(path: 'HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\LanManServer\Parameters').properties
registrykey.properties: {
  EnableAuthenticateUserSharing: "0"
  Guid: ""
  NullSessionPipes: ""
  ServiceDll: "%SystemRoot%\\system32\\srvsvc.dll"
  ServiceDllUnloadOnStop: "1"
  autodisconnect: "15"
  enableforcedlogoff: "1"
  enablesecuritysignature: "0"
  requiresecuritysignature: "0"
  restrictnullsessaccess: "1"
}
```

in favor of the new `registrykey.items`. This returns the full property MQL resources and also allows inspection of type:

```
cnspec> registrykey(path: 'HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\LanManServer\Parameters').items { * }
registrykey.items: [
  0: {
    value: "0"
    type: "dword"
    name: "EnableAuthenticateUserSharing"
    path: "HKEY_LOCAL_MACHINE\\SYSTEM\\CurrentControlSet\\Services\\LanManServer\\Parameters"
    data: 0
    exists: true
  }
  1: {
    value: ""
    type: "multistring"
    name: "NullSessionPipes"
    path: "HKEY_LOCAL_MACHINE\\SYSTEM\\CurrentControlSet\\Services\\LanManServer\\Parameters"
    data: [
      0: ""
    ]
    exists: true
  }
  2: {
    value: "%SystemRoot%\\system32\\srvsvc.dll"
    type: "expandstring"
    name: "ServiceDll"
    path: "HKEY_LOCAL_MACHINE\\SYSTEM\\CurrentControlSet\\Services\\LanManServer\\Parameters"
    data: "%SystemRoot%\\system32\\srvsvc.dll"
    exists: true
  }
  3: {
    value: "1"
    type: "dword"
    name: "ServiceDllUnloadOnStop"
    path: "HKEY_LOCAL_MACHINE\\SYSTEM\\CurrentControlSet\\Services\\LanManServer\\Parameters"
[...]

PR #1498, opened 2023-08-13
✨ migrate registry keys

arlimus:


_resources/packs/os/os.lr
```diff
@@ -23,18 +23,6 @@ alias os.unix.mount = mount
...
-  path string
-  exists() bool
-  // Registry key properties
-  properties() map[string]string
-  // Registry key children
-  children() []string
-}
-
 // Windows registry key property
 registrykey.property @defaults("path name") {
```
czunker: This can also be removed.

providers/os/resources/registrykey.go
```diff
@@ -0,0 +1,177 @@
...
+
+	for i := range children {
+		child := children[i]
+		res = append(res, child.Path)
+	}
+
+	return res, nil
+}
+
+type mqlRegistrykeyPropertyInternal struct {
```
czunker, line 114: I don't see this type referenced. What do we need it for?

providers/os/resources/registrykey.go
```diff
@@ -0,0 +1,177 @@
...
+		"path": llx.StringData(p.Path.Data),
+	})
+	if err != nil {
+		p.key = plugin.TValue[*mqlRegistrykey]{Error: err, State: plugin.StateIsSet}
+		return p.key.Data, p.key.Error
+	}
+
+	registryKey := obj.(*mqlRegistrykey)
+	p.key = plugin.TValue[*mqlRegistrykey]{Data: registryKey, State: plugin.StateIsSet}
+	return p.key.Data, p.key.Error
```
czunker, line 138: This looks like a Copy&Paste error to me.
I'd expected:
```
return p.key.Data, nil
```
Or is `p.key.Error` initialized with `nil`?

providers/os/resources/registrykey.go
czunker: Came across this somewhere else. It's used for caching.

vjeffrey:
@czunker since @arlimus is at the conference this week could you take over the needed changes here?

czunker:
> Reminder to self that we need to port #1359 on top of this change

I added an issue for that: https://github.com/mondoohq/cnquery/issues/1604

PR #1500, opened 2023-08-13
🐛 fix vault sub-command

chris-rock:
The sub-command has not worked properly when there were permission errors for the local keyring. That features was barely used since you can embedded vault configuration in inventory files.

Lets assume you have an inventory file like:

 ```yaml
apiVersion: v1
kind: Inventory
spec:
  assets:
    - name: vsphere
      connections:
        - backend: vsphere
          host: 192.168.5.24
          credentials:
            - secret_id: my-secret
          discover:
            targets:
              - host-machines
  credentials:
    my-secret:
      type: password
      user: chris@vsphere.local
      password: myPassword
```

You can easily scan the inventory via `cnquery scan --inventory-file inventory.yaml`. `cnquery` allows you to use external vault systems to store secrets. To generate the correct config for vault, you can use:

```
$ cnquery vault configure mondoo-client-vault --type encrypted-file --option=password='superdupapassword' --option path='vault.enc'
→ add the following vault configuration to your inventory file
spec:
  vault:
    name: mondoo-client-vault
    options:
      password: superdupapassword
      path: vault.enc
    type: encrypted-file
```

With this config you change the inventory to:

```yaml
apiVersion: v1
kind: Inventory
spec:
  assets:
    - name: vsphere
      connections:
        - backend: vsphere
          host: 192.168.5.24
          credentials:
            - secret_id: my-secret
          discover:
            targets:
              - host-machines
  vault:
    name: mondoo-client-vault
    options:
      password: superdupapassword
      path: vault.enc
    type: encrypted-file
```

To add the secret, you can call the add secret option:

```
cnquery vault add-secret  my-secret '{   "type": "password",   "user": "chris@vsphere.local",   "password": "myPassword" }' --inventory-file inventory.yaml     
→ load vault configuration from inventory
→ stored secret successfully
```

Now the secret is stored in the configured vault.

### DIFF
--- a/apps/cnquery/cmd/vault.go
+++ b/apps/cnquery/cmd/vault.go
@@ -3,37 +3,32 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
+
+	v1 "go.mondoo.com/cnquery/motor/inventory/v1"
 
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"go.mondoo.com/cnquery/motor/vault"
-	"go.mondoo.com/cnquery/motor/vault/config"
 )
 
 func init() {
-	vaultListCmd.Flags().Bool("show-options", false, "displays configured options")
-	vaultCmd.AddCommand(vaultListCmd)
-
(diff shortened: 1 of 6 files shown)


atomic111 (approved):
@chris-rock thank you for the improvement

PR #1509, opened 2023-08-15
🐛 SUSE11 ignore all *boot* services

nfsprodriver:
... as it hangs on at "boot" during any service check. Applies to #1469

### DIFF
--- a/resources/packs/os/services/sysv.go
+++ b/resources/packs/os/services/sysv.go
@@ -34,7 +34,7 @@ func (s *SysVServiceManager) List() ([]*Service, error) {
 	}
 
 	// eg. we ignore the following run levels since `service halt status` may shutdown the system
-	ignored := []string{"functions", "killall", "halt", "reboot", "shutdown"}
+	ignored := []string{"functions", "killall", "halt", "boot", "shutdown"}
 	statusServices := []string{}
 	for i := range services {
 		service := services[i]
--- a/resources/packs/os/services/sysv.go
+++ b/resources/packs/os/services/sysv.go
@@ -34,7 +34,7 @@ func (s *SysVServiceManager) List() ([]*Service, error) {
 	}
 
 	// eg. we ignore the following run levels since `service halt status` may shutdown the system
-	ignored := []string{"functions", "killall", "halt", "boot", "shutdown"}
+	ignored := []string{"boot", "boot.local", "functions", "halt", "halt.local", "killall", "rc", "reboot", "shutdown", "single", "skeleton", ".depend.boot", ".depend.start", ".depend.stop" }
 	statusServices := []string{}
 	for i := range services {
 		service := services[i]


nfsprodriver:

resources/packs/os/services/sysv.go
```diff
@@ -34,7 +34,7 @@ func (s *SysVServiceManager) List() ([]*Service, error) {
 	}
 
 	// eg. we ignore the following run levels since `service halt status` may shutdown the system
-	ignored := []string{"functions", "killall", "halt", "reboot", "shutdown"}
+	ignored := []string{"functions", "killall", "halt", "boot", "shutdown"}
```
chris-rock: I think we need both, "reboot" and "boot" since we just ignore the specific services, if I remember correctly "reboot" is used on older ubuntu machines.

chris-rock (commented):
Thank you for the contribution @nfsprodriver. Great addition!

resources/packs/os/services/sysv.go
mick-lue: instead of using `stringx.Contains` you could use a for loop with `strings.Contains` together with a label on the outside loop. This would also catch files like **boot.local** or **halt.local** (and many more like this, that may not work as expected when running 'service x status'). This is how we thought the current implementation would handle the ignored services, but looking at the documentation, that seems to not be the case.
```
	ignored := []string{"functions", "killall", "halt", "boot", "shutdown"}
	statusServices := []string{}
	SERVICE_ITERATION:
	for i := range services {
		service := services[i]
		for j := range ignored {
			if strings.Contains(service, ignored[j]) {
				continue SERVICE_ITERATION
			}
		}
		statusServices = append(statusServices, service)
	}
```
This does open up the possibility of ignoring services that coincidentally contain these strings though. I don't know how many that might be, if any.

chris-rock: I think we should be explicit in the list, it leads to less ambiguity. Lets just list all the known services that we want to ignore. This makes it also easier for cnquery users to reason why a service shows up or not.

mick-lue: That is true.
I added the files mentioned in the SUSE boot readme file. Most of these would not handle a call like `service x status` properly and are used for booting, switching run levels, or killing processes. (The list is also sorted now because it became quite long)
```suggestion
	ignored := []string{"boot", "boot.local", "functions", "halt", "halt.local", "killall", "rc", "reboot", "shutdown", "single", "skeleton", ".depend.boot", ".depend.start", ".depend.stop" }
```

chris-rock (approved):
Thank you @nfsprodriver @mick-lue

chris-rock:
we released https://github.com/mondoohq/cnquery/releases/tag/v8.23.1 which includes your fix! Thank you again.

PR #1516, opened 2023-08-16
✨ disable provider auto-update

arlimus:
- Add an optional flag `--auto-update=false` to disable provider install or updates
- Move the pre-loader to a new model where a dummy command is being used instead of manual parsing. This is done so that we can inject all the crazy config handling we have to do, because the option could be set in the config.

Follow-ups:
1. There is still a bug in loading this from config...
2. It currently hard-codes the provider updates on indirect calls from the runtime. This needs to respect config/CLI options. (even global is fine...)


chris-rock (approved):
Thank you @arlimus

PR #1517, opened 2023-08-16
🐛 Fix manager test, gofmt a file.

preslavgerchev:
Make the v8 build green again, looks like we now ignore more services due to https://github.com/mondoohq/cnquery/commit/b4047b4a8b16576b95b0818ac2932fa0cfba4a9d


chris-rock (approved):
thank you @preslavgerchev

PR #1518, opened 2023-08-16
✨ migrate k8s kubelet resource

imilchev:
There is an open issue here where the `k8s.lr.go` contains errors because it is trying to create Go structs for resources that are actually located in other providers - `file` and `process`. I am not sure why that happens


providers/k8s/resources/k8s.lr.go
```diff
@@ -6749,3 +6778,62 @@ func (c *mqlK8sUserinfo) GetUsername() *plugin.TValue[string] {
 func (c *mqlK8sUserinfo) GetUid() *plugin.TValue[string] {
 	return &c.Uid
 }
+
+// mqlK8sKubelet for the k8s.kubelet resource
+type mqlK8sKubelet struct {
+	MqlRuntime *plugin.Runtime
+	__id string
+	// optional: if you define mqlK8sKubeletInternal it will be used here
+	ConfigFile plugin.TValue[*mqlFile]
```
imilchev, line 6787: I think this and the line below aren't correct because `file` and `process` are part of `os` provider

providers/k8s/resources/k8s.lr
```diff
@@ -843,4 +844,14 @@ private k8s.userinfo @defaults("username") {
   username string
   // The UID of the user
   uid string
+}
+
+// Kubernetes Kubelet configuration
+private k8s.kubelet {
+  // Kubelet config file
+  configFile file
```
arlimus, line 852: If you want to use `file` here, you'll have to cross-reference the OS provider: `os.file` is the right way to call it

imilchev:
I will fix this in the other PR

PR #1531, opened 2023-08-18
🐛 Assign default MRN to incognito assets.

preslavgerchev:
We now assign the asset id during the connection to the provider so we don't have to check that anymore


imilchev (approved):

chris-rock (approved):
Nice catch, was just looking for the same issue.

PR #1532, opened 2023-08-18
⭐️ Azure V9

preslavgerchev:


providers/azure/go.mod
```diff
@@ -0,0 +1,121 @@
...
+require (
+	github.com/Azure/azure-sdk-for-go/sdk/internal v1.3.0 // indirect
+	github.com/Azure/azure-sdk-for-go/sdk/keyvault/internal v0.7.1 // indirect
+	github.com/Azure/go-ansiterm v0.0.0-20230124172434-306776ec8161 // indirect
+	github.com/AzureAD/microsoft-authentication-library-for-go v1.0.0 // indirect
+	github.com/Masterminds/semver v1.5.0 // indirect
+	github.com/Microsoft/go-winio v0.6.1 // indirect
+	github.com/aws/aws-sdk-go-v2 v1.18.1 // indirect
+	github.com/aws/aws-sdk-go-v2/config v1.18.19 // indirect
+	github.com/aws/aws-sdk-go-v2/credentials v1.13.26 // indirect
```
preslavgerchev: i do not understand why these are being imported indirectly here, need to check it out

providers/azure/resources/subscription.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	subscriptions "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armsubscriptions"
+	"go.mondoo.com/cnquery/llx"
+	"go.mondoo.com/cnquery/providers-sdk/v1/plugin"
+	"go.mondoo.com/cnquery/providers-sdk/v1/util/convert"
+	"go.mondoo.com/cnquery/providers/azure/connection"
+	"go.mondoo.com/cnquery/types"
+)
+
+// TODO: we need to make this NOT go through init, this is heavy on the API
```
preslavgerchev, line 18: to fix: #1537 (can be done in a later PR)

providers/azure/go.mod
preslavgerchev: fixed

providers/azure/resources/compute.go
```diff
@@ -0,0 +1,316 @@
...
+		return args, nil, nil
+	}
+
+	conn := runtime.Connection.(*connection.AzureConnection)
+	args["subscriptionId"] = llx.StringData(conn.SubId())
+
+	return args, nil, nil
+}
+
+func (a *mqlAzureSubscriptionCompute) vms() ([]interface{}, error) {
```
vjeffrey, line 35: i'm surprised to not see init functions for the vms, like initAzureSubscriptionComputeVm ?

preslavgerchev: there's no discovery yet in this PR so that's why there are no init funcs for all of those assets, i will do this is a follow up with the discovery

imilchev (approved):

PR #1536, opened 2023-08-18
🐛 Fix output for the build provider make target.

preslavgerchev:
This ensures that we end up with the artifacts undet the right `dist` folder so that `providers/install` can work correctly after that

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,8 @@ cnquery/generate: clean/proto llx/generate shared/generate providers explorer/ge
 define buildProvider
 	$(eval $@_HOME = $(1))
 	$(eval $@_NAME = $(shell basename ${$@_HOME}))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -145,8 +146,8 @@ providers/build/os: providers/lr
 
 providers/install:
 #	@$(call installProvider, providers/core)
-	@$(call installProvider, providers/network)
-	@$(call installProvider, providers/os)
+#	@$(call installProvider, providers/network)
```
imilchev: why are these commented out?

chris-rock: Once that is fixed LGTM

chris-rock (changes requested):

chris-rock (approved):
Thank you @preslavgerchev

PR #1540, opened 2023-08-18
🐛 fix inventory resolution and shell requirement on valid configuration

### LINKED ISSUE #1524: Cannot use cnquery without a Mondoo config
If I try to use cnquery v9 without a Mondoo config present, I get:
```
root@podkrepibg01:~# ./cnquery shell local
FTL failed to connect to asset error="could not initialize client authentication: agent credentials must be set"
```
I think the config is used to check the providers versions or something?

chris-rock:
fixes #1524


arlimus (approved):
lgtm!

PR #1546, opened 2023-08-20
⭐️ provider scaffolding

chris-rock:
Quick start to get a new provider with the base files:

```bash
go install apps/provider-scaffold/provider-scaffold.go
provider-scaffold --path providers/vsphere --provider-id vsphere --provider-name "VMware vSphere" --go-package go.mondoo.com/cnquery/providers/vsphere
```

### DIFF
--- a/apps/provider-scaffold/template/config/config.go.template
+++ b/apps/provider-scaffold/template/config/config.go.template
@@ -1,0 +1,18 @@
+package config
+
+import "go.mondoo.com/cnquery/providers-sdk/v1/plugin"
+
+var Config = plugin.Provider{
+	Name:    "{{ .ProviderID }}",
+	ID:      "{{ .GoPackage }}",
+	Version: "9.0.0",
+	Connectors: []plugin.Connector{
+		{
+			Name:      "{{ .ProviderID }}",
+			Use:       "{{ .ProviderID }}",
+			Short:     "{{ .ProviderName }}",
+			Discovery: []string{},
+			Flags:     []plugin.Flag{},
+		},
+	},
+}
--- a/apps/provider-scaffold/template/provider/provider.go.template
+++ b/apps/provider-scaffold/template/provider/provider.go.template
@@ -32,6 +32,7 @@ func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error)
 
 	conn := &inventory.Config{
 		Type: req.Connector,
+		Options: map[string]string{},
 	}
 
 	// Do custom flag parsing here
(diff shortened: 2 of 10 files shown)


apps/provider-scaffold/template/resources/providerid.go.template
```diff
@@ -0,0 +1,9 @@
+package resources
+
+func (k *mql{{ .CamelcaseProviderID }}) id() (string, error) {
+	return "{{ .ProviderID }}", nil
+}
+
+func (s *mql{{ .CamelcaseProviderID }}) field() (string, error) {
```
czunker: Shouldn't this be the same as above? So, either `(s *mql...)` or `(k *mql...)`?

czunker:
Thanks for this @chris-rock I'm currently giving it a try with the GitHub provider.

apps/provider-scaffold/template/resources/providerid.go.template
chris-rock: definitely

chris-rock: fixed

apps/provider-scaffold/template/provider/provider.go.template
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error) {
+	flags := req.Flags
+	if flags == nil {
+		flags = map[string]*llx.Primitive{}
+	}
+
+	conn := &inventory.Config{
+		Type: req.Connector,
```
czunker, line 34: Only a suggestion, as we use this in some of our providers.

Add:
```
Options: map[string]string{},
```

apps/provider-scaffold/template/config/config.go.template
```diff
@@ -0,0 +1,18 @@
+package config
+
+import "go.mondoo.com/cnquery/providers-sdk/v1/plugin"
+
+var Config = plugin.Provider{
+	Name:    "{{ .ProviderID }}",
+	ID:      "{{ .GoPackage }}",
+	Version: "9.0.0",
```
czunker, line 8: Only a suggestion, as we use this in some of our providers.

Add:
```
MinArgs: 1,
MaxArgs: 1,
```

czunker (approved):
Added only some suggestions.

Otherwise looking good and working.

apps/provider-scaffold/template/config/config.go.template
chris-rock: If we add this, this provider would not work after compile since we need an argument.

apps/provider-scaffold/template/provider/provider.go.template
chris-rock: That is a great point, lets initialize it

PR #1551, opened 2023-08-21
⭐️ vsphere provider

chris-rock:
This is pretty complete. It misses the discovery and some special credential handling.


providers/vsphere/go.mod
```diff
@@ -0,0 +1,62 @@
+module go.mondoo.com/cnquery/providers/vsphere
+
+replace go.mondoo.com/cnquery => ../..
```
czunker, line 3: Is this intended?

chris-rock: yes, this allows us to always use the latest cnquery version

Makefile
```diff
@@ -144,19 +144,25 @@ providers/build/network: providers/lr
...
 providers/bundle:
 	@$(call bundleProvider, providers/network)
 	@$(call bundleProvider, providers/os)
+	@$(call bundleProvider, providers/vsphere)
 
 providers/test:
 	@$(call testProvider, providers/core)
 	@$(call testProvider, providers/network)
 	@$(call testProvider, providers/os)
+	@$(call testGpModProvider, providers/vsphere)
```
czunker, line 201: What is a `GpMod` provider?

providers/vsphere/provider/provider.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	conn := &inventory.Config{
+		Type: req.Connector,
+	}
+
+	// parse args
+	user := ""
+	port := 0
+	if len(req.Args) != 0 {
```
czunker, line 47: What happens when `len` is zero?

providers/vsphere/provider/provider.go
```diff
@@ -0,0 +1,218 @@
...
+
+	asset.Platform = &inventory.Platform{
+		Name:   "vsphere",
+		Family: []string{"vsphere"},
+		Kind:   "api",
+		Title:  "VMware vSphere",
+	}
+
+	// TODO: Add platform IDs
+	asset.PlatformIds = []string{"//platformid.api.mondoo.app/runtime/oci/"}
```
czunker: That's the wrong platform ID.

providers/vsphere/provider/provider.go
```diff
@@ -0,0 +1,218 @@
...
+		flags = map[string]*llx.Primitive{}
+	}
+
+	conn := &inventory.Config{
+		Type: req.Connector,
+	}
+
+	// parse args
+	user := ""
+	port := 0
```
czunker, line 46: Has vsphere a default port we could use?

providers/vsphere/provider/provider.go
```diff
@@ -0,0 +1,218 @@
...
+		HasRecording:   req.HasRecording,
+		CreateResource: resources.CreateResource,
+		Upstream:       upstream,
+	}
+
+	return conn, err
+}
+
+func (s *Service) detect(asset *inventory.Asset, conn *connection.VsphereConnection) error {
+	// TODO: adjust asset detection
```
czunker: Is this comment still needed?

providers/vsphere/resources/host.go
```diff
@@ -0,0 +1,451 @@
...
+
+func (v *mqlVsphereHost) acceptanceLevel() (string, error) {
+	esxiClient, err := v.esxiClient()
+	if err != nil {
+		return "", err
+	}
+	return esxiClient.SoftwareAcceptance()
+}
+
+func (v *mqlVsphereHost) kernelModules() ([]interface{}, error) {
```
czunker, line 283: Out of curiosity: Why do we not need a safe type here but in other func signatures?

providers/vsphere/resources/vsphere.lr
```diff
@@ -1,3 +1,6 @@
+option provider = "go.mondoo.com/cnquery/providers/vsphere"
```
czunker, line 4: As discussed today, this would now also need a license header.

providers/vsphere/provider/provider.go
```diff
@@ -0,0 +1,218 @@
...
+		Type: req.Connector,
+	}
+
+	// parse args
+	user := ""
+	port := 0
+	if len(req.Args) != 0 {
+		target := req.Args[0]
+		if !strings.Contains(target, "://") {
+			target = "ssh://" + target
```
czunker: Something doesn't match for me. Here, we expect an ssh connection string. At other places in the code, we use an API. I'm not sure how this should look like here.

providers/vsphere/resources/host.go
arlimus: arrays are still following the interface-list signature, which we can tackle in a future version. The change requires a bit more than just returning the safe type

Makefile
arlimus: Those are providers that have their own go mods. It requires us to switch into the path before running commands. We will unify this around this pattern to deduplicate this case again.

providers/vsphere/provider/provider.go
arlimus: For the cli parsing, if there are no args provided, then the connection will not have any information about it. The caller can still work around this fact and insert their own asset information, but for now we don't do that.

providers/vsphere/provider/provider.go
arlimus: definitely a good follow-up! i know this code has mostly just migrated what was already there

arlimus (approved):
awesome, thank you for migrating!

providers/vsphere/provider/provider.go
chris-rock: We do not care in cli parsing, defaults should happen in the connection itself. Otherwise missing values in inventories would not work. The handling for 0 ports is here https://github.com/mondoohq/cnquery/pull/1551/files#diff-719e7f1b1168518ce8f65b0279ac8c2892e64752122488e50d36241a7c0199abR28-R30

providers/vsphere/provider/provider.go
chris-rock: This does not use a ssh or the scheme at all. It is just a helper to use Go''s built-in URL-parse for the following:

```
cnquery shell vsphere chris@vsphere.local@192.168.5.24
```

In vSphere land we specifically need to tackle cases with double-@ which works with the built-in one super neat. I add a comment to make this more clear and rename it to something more neutral

providers/vsphere/provider/provider.go
chris-rock: Great catch, will update

providers/vsphere/provider/provider.go
chris-rock: This still needs to be updated with the full discovery, which is not implemented in this PR

PR #1553, opened 2023-08-21
🐛 Don't let parseValue panic on bad input

jaym:
For example, if you use the number 089, that will cause a panic because 9 is not valid octal.


chris-rock (approved):
Great improvement @jaym

PR #1554, opened 2023-08-21
⭐️ okta provider

chris-rock:


providers/okta/connection/connection.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, errors.New("provider type does not match") // TODO: switch to plugin.ErrProviderTypeDoesNotMatch
+	}
+
+	if conf.Options == nil || conf.Options["organization"] == "" {
+		return nil, errors.New("okta provider requires an organization id. please set option `organization`")
+	}
+
+	org := conf.Options["organization"]
+
+	// if a secret was provided, it always overrides the env variable since it has precedence
```
czunker: To me, this comment was a bit misleading as the actual handling of env var or token flag happens somewhere else.

providers/okta/provider/provider.go
```diff
@@ -0,0 +1,205 @@
...
+		Options: make(map[string]string),
+	}
+
+	// flag parsing
+	token := ""
+	if x, ok := flags["token"]; ok && len(x.Value) != 0 {
+		token = string(x.Value)
+	}
+	if token == "" {
+		token = os.Getenv("OKTA_TOKEN")
```
czunker, line 54: This env var is different from the one mentioned at https://github.com/mondoohq/cnquery/pull/1554/files#diff-f41f27f2f43d22fdc7bd16377e3897c34df572efddc892df7ec7d59373ff8c63R59

providers/okta/resources/okta.lr
```diff
@@ -1,4 +1,5 @@
-option go_package = "go.mondoo.com/cnquery/resources/packs/okta"
```
czunker, line 4: This now needs also a license header.

providers/okta/resources/organization.go
```diff
@@ -0,0 +1,180 @@
...
+		ranger.DefaultHttpClient(),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return convert.JsonToDict(emails)
+}
+
+// GetThreatInsightSettings returns the Threat Insight settings for the organization
```
czunker: nit: this should now be:
```suggestion
// threatInsightSettings returns the Threat Insight settings for the organization
```

providers/okta/resources/policies.go
```diff
@@ -146,57 +145,44 @@ func newMqlOktaPolicy(runtime *resources.Runtime, entry *sdk.PolicyWrapper) (int
...
-	if err != nil {
-		return nil, err
+	ctx := context.Background()
+	if o.Id.Error != nil {
+		return nil, o.Id.Error
 	}
 
-	if policyType == ACCESS_POLICY {
-		return getAccessPolicyRules(ctx, o.MotorRuntime, policyId, op.OrganizationID(), op.Token())
+	if o.Type.Data == ACCESS_POLICY {
```
czunker, line 181: Could `o.Type.Error` be not nil?

Above, the code previously checked for `o.Type()` returning an error.

providers/okta/provider/provider.go
arlimus: Good catch! fixing

providers/okta/resources/organization.go
arlimus: +1, since this is now internal i'm removing it

providers/okta/resources/policies.go
arlimus: Since the field is static, we don't expect errors. 

In fact this is a pattern we will want to ensure down the line, because it leads to a lot of overhead code every provider dev has to do this manually.

arlimus (approved):
Kudos for migrating this @chris-rock 
Updated based on feedback, thank you @czunker

providers/okta/resources/okta.lr
chris-rock: done

providers/okta/connection/connection.go
chris-rock: Great catch, definitely a copy error. The new model is more clear and this issue does not apply here anymore. I am going to remove this comment

providers/okta/provider/provider.go
chris-rock: I am going to harmonize it to `OKTA_API_TOKEN` as used by the terraform provider https://registry.terraform.io/providers/okta/okta/latest/docs. I keep `OKTA_TOKEN` as fallback. I think `OKTA_CLIENT_TOKEN` was never working, so I remove that.

PR #1559, opened 2023-08-22
✨ GitHub provider for v9

czunker:


Makefile
```diff
@@ -144,19 +144,26 @@ providers/build/network: providers/lr
 providers/build/os: providers/lr
 	@$(call buildProvider, providers/os)
 
+providers/build/github: providers/lr
+	@$(call buildProvider, providers/github)
+	cp providers/github/resources/github.resources.json providers/github.resources.json
```
chris-rock: This should not be required anymore

providers/github/go.mod
```diff
@@ -0,0 +1,280 @@
+module go.mondoo.com/cnquery/providers/github
+
+go 1.20
+
+require (
+	github.com/cockroachdb/errors v1.9.1
+	github.com/google/go-github/v49 v49.1.0
+	github.com/rs/zerolog v1.30.0
+	github.com/stretchr/testify v1.8.4
+	go.mondoo.com/cnquery v0.0.0-20230822120357-4b3c486e0c35
```
chris-rock, line 12: this should be a relative dependency since we we always build them from the same git commit sha

providers/github/resources/github_user.go
```diff
@@ -1,37 +1,37 @@
...
-		return "", err
+	if g.Id.Error != nil {
+		return "", g.Id.Error
 	}
+	id := g.Id.Data
 	return "github.user/" + strconv.FormatInt(id, 10), nil
 }
 
+/*
 func (g *mqlGithubUser) init(args *resources.Args) (*resources.Args, GithubUser, error) {
```
chris-rock: the implementation of `cnquery shell github user chris-rock` does not work without this init method

providers/builtin.go
```diff
@@ -21,15 +21,17 @@ import (
...
 )
 
 var BuiltinCoreID = coreconf.Config.ID
 
 //go:embed core/resources/core.resources.json
 var coreInfo []byte
 
-// //go:embed os/resources/os.resources.json
-// var osInfo []byte
+// //go:embed github.resources.json
```
chris-rock: I think we should not add providers with go mod this way. I am going to remove it form here fore now

chris-rock:
Also discovered a bug in cnquery https://github.com/mondoohq/cnquery/issues/1565

providers/builtin.go
chris-rock: fixed

providers/github/resources/github_user.go
chris-rock: fixed

providers/github/go.mod
chris-rock: update the go.mod

Makefile
chris-rock: removed from the makefile

chris-rock (approved):
Thank you @czunker

PR #1561, opened 2023-08-22
🧹 Add new ptr funcs for bool and int

czunker:


chris-rock (approved):
Great addition @czunker

PR #1572, opened 2023-08-23
🧹 remove error from generator struct

arlimus:
It was never returning an error, since errors are collected in the internal multi-error field.


chris-rock (approved):
Thank you @arlimus

PR #1573, opened 2023-08-23
🐛 avoid building dynamic `id` fields in resources

arlimus:
After https://github.com/mondoohq/cnquery/pull/1572

tldr: use static `id` fields in resources, dynamic fields would need the same `id()` method we use for internal IDs

Long explanation: It is possible for a resource to define itself an `id` field. We have done this before (eg in yum repos) where the field returns a custom ID that users can access.

All MQL resources have their own internal ID. The confusion happens, when we have a semantic `id` field we expose to users and an internal `id` field that we do not expose. At the time of writing, we have no scenario where we would ever need an ID field that is different from the internal ID of the resource. Ideally we can have both use the same ID, i.e. if users ever expose an `id` field it should semantically provide everything we would expect from the ID of a resource.

For the time being, we only need to solve one problem: Dynamic `id()` fields would force users to implement an `func id()` call, that conflicts with the call to create internal IDs. This PR will not deduplicate this concept (we can do this in a follow-up), but it at least needs to prevent the scenario from happening and give users some guidance.

For all users that expose an `id` field, we solution is to create it as a static field, so that at the time of resource creation it is set. This will also allow us to deduplicate this field if we choose to down the line. Semantically too, this makes sense, since the ID of the object should be known at creation time, whether internal or user-facing.

### DIFF
--- a/providers-sdk/v1/lr/go.go
+++ b/providers-sdk/v1/lr/go.go
@@ -442,6 +442,11 @@ func (b *goBuilder) goField(r *Resource, field *Field) {
 		return
 	}
 
+	if field.BasicField.ID == "id" {
+		b.errors.Add(errors.New("cannot create a dynamically computed `id` field, please turn it into a static field (ie: `id(..)` => `id`)"))
+		return
+	}
+
 	goName := field.BasicField.goName()
 	goType := field.BasicField.Type.goType(b)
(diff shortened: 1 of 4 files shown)


chris-rock (approved):
Thank you @arlimus That helps to make the id generation more stable

PR #1580, opened 2023-08-25
🧹 Refactor conn var to conf

czunker:
We used conn for a config and a connection.
This confused me multiple times during reviews.


chris-rock:
I agree, this confused me a couple of times too

chris-rock (approved):
Great improvement @czunker

PR #1588, opened 2023-08-27
🧹 remove old github files

chris-rock:
removes files that were not properly removed in #1559 

after #1589


czunker (approved):
Thanks @chris-rock

PR #1589, opened 2023-08-27
🧹 update go dependencies of providers

chris-rock:
When we update dependencies in cnquery (eg. via dependabot), the build of the plugins will fail since they depend on latest cnquery with the statement of `replace go.mondoo.com/cnquery => ../..`. This is still the desired behavior since when I check out a commit, all code references should only depend on that specific commit in cnquery. Otherwise debugging issues will become very painful going forward.

This PR is a first step to automate go.mod update for all the providers via a `make providers/gomodtidy`. Once that runs well, we probably want to run this as an automatic task for PRs that change go.mod in cnquery base.


czunker (approved):
Thanks @chris-rock

PR #1590, opened 2023-08-27
⭐️ Discovery for GCP resources

chris-rock:
This PR builds on top of #1587

This changes ports over the old mql based approach from v8 to v9. Instead of using MQL, we use the resources directly. This new approach has a lot of advantages compared to the previous one. We have full type support when we use resources and we can directly use every field defined in lr. On top the code is much simplified since we just have one file where all the discovery happens.


providers/gcp/provider/provider.go
```diff
@@ -197,6 +214,21 @@ func (s *Service) detect(asset *inventory.Asset, conn *connection.GcpConnection)
 	return nil
 }
 
+func (s *Service) discover(conn *connection.GcpConnection) (*inventory.Inventory, error) {
+
+	if conn.Conf.Discover != nil {
```
preslavgerchev: this should probably be `== nil`, else flags are ignored?

vjeffrey:
i pushed a commit with a couple small fixes, and it gets the list of assets displaying for selection. something is still off with the fine grained discovery (pretty sure it's the getAssetIdentifiers funcs in the init calls that aren't working) so we're not loading up the single assets well yet, but a step in the direction we wanna go

imilchev:
> hmm, i feel like there's still something acting up with the assets on discovery:
> 
> ```
> [29/08/23 12:52:11] ❯ DEBUG=1 go run apps/cnquery/cnquery.go shell gcp project mondoo-dev-262313 --discover compute-networks
> DBG using provider gcp with connector gcp
> ! using builtin provider for gcp
> → loaded configuration from /Users/vj/.config/mondoo/mondoo.yml using source default
> DBG fallback to default google sdk authentication
> DBG fallback to default google sdk authentication
> DBG fallback to default google sdk authentication
> DBG resolved 2 assets
> → selected asset asset=chris-ipv6-test selection=0
> DBG fallback to default google sdk authentication
> DBG initialize client authentication issuer=mondoo/ams kid=//agents.api.mondoo.app/spaces/funny-poitras-168395/serviceaccounts/2ULneVaceGcHlqqyeRZ0voWif78
> DBG fallback to default google sdk authentication
> DBG fallback to default google sdk authentication
> → connected to GCP Compute Network
>   ___ _ __   __ _ _   _  ___ _ __ _   _
>  / __| '_ \ / _` | | | |/ _ \ '__| | | |
> | (__| | | | (_| | |_| |  __/ |  | |_| |
>  \___|_| |_|\__, |\__,_|\___|_|   \__, |
>   mondoo™      |_|                |___/  interactive shell
> 
> cnquery> gcp.compute.network
> DBG starting query execution qrid=eAN13L8tnd0=
> DBG finished query execution qrid=eAN13L8tnd0=
> DBG o3S49cyBJR7QRtc9NTG4x0XmQeOYD32sT9ZOSLDGcyRoOyG/AETJHBmHIgTxHEfcQ+OmpsLEeChXmKwoS1epOQ== finished
> DBG graph has received all datapoints
> cannot find resource gcp.compute in this provider
> gcp.compute.network: no data available
> cnquery> exit
> ```

you used the wrong resource. It's called `gcp.project.computeService.network`. Technically, `gcp.compute.network` should work too but seems like aliasing is broken in this case

vjeffrey:
ah ok, got it

vjeffrey (approved):

PR #1591, opened 2023-08-27
🐛 gcp compute instances should be labelled as platform instance not image

vjeffrey:
also, i didn't touch the logic for this part right now, but why are we entirely skipping windows instances, like not discovering them at all? @preslavgerchev @imilchev does one of you know?
```
		if disksContainWindows(i.Disks) {
			log.Debug().Msgf("skipping windows instance %s", i.Name)
			continue
		}
```


tas50 (approved):
I was real confused why these were coming back as images. Great fix.

PR #1602, opened 2023-08-29
✨ tar/container connection for v9

czunker:


czunker:
I tested this with an image: `cnquery run container image nginx -c 'asset{ * }'`

A running container with the same image: `cnquery run container cc7f5aaaa246 -c 'asset{ * }'`

And the stopped container, which creates a snapshot.

What's open:
- Test registry
- Test registry with auth

vjeffrey (approved):

PR #1605, opened 2023-08-29
🧹 iterate over all discovered assets when using run

imilchev:
In v8 we discover all assets and then execute the requested command for all of them. Currently, v9 doesn't do this

### DIFF
--- a/apps/cnquery/cmd/plugin.go
+++ b/apps/cnquery/cmd/plugin.go
@@ -18,6 +18,7 @@ import (
 	"go.mondoo.com/cnquery/providers"
 	"go.mondoo.com/cnquery/providers-sdk/v1/inventory"
 	pp "go.mondoo.com/cnquery/providers-sdk/v1/plugin"
+	v1plugin "go.mondoo.com/cnquery/providers-sdk/v1/plugin"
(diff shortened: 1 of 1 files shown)


apps/cnquery/cmd/plugin.go
```diff
@@ -79,7 +80,16 @@ func (c *cnqueryPlugin) RunQuery(conf *run.RunQueryConfig, runtime *providers.Ru
 		return nil
 	}
 
-	assetList := conf.Inventory.Spec.Assets
+	res, err := runtime.Provider.Instance.Plugin.Connect(&v1plugin.ConnectReq{
+		Features: conf.Features,
+		Asset:    conf.Inventory.Spec.Assets[0],
```
preslavgerchev, line 85: that's the asset, coming back from `ParseCLI` ?

imilchev: no, this is hardcoded in the run command. https://github.com/mondoohq/cnquery/blob/main/apps/cnquery/cmd/run.go#L49-L53

czunker:
I think, we can close this PR. The changes already merged with one of the other PR: https://github.com/mondoohq/cnquery/blob/main/apps/cnquery/cmd/plugin.go#L82

PR #1616, opened 2023-08-31
✨ gcp snapshot connection for v9

czunker:


czunker:
That's how it looks like in v8:
```
asset: {
  kind: "api"
  build: ""
  version: "15.1"
  title: "SUSE Linux Enterprise Server 15 SP1, GCP Virtual Machine"
  arch: ""
  labels: {}
  platform: "sles"
  vulnerabilityReport: To use this resource, you must authenticate with Mondoo Platform.
To learn how, read: 
  runtime: "gcp-vm"
  ids: [
    0: "//platformid.api.mondoo.app/runtime/gcp/compute/v1/projects/mondoo-dev-262313/snapshots/suse15"
  ]
  name: "suse15"
  fqdn: ""
  family: [
    0: "suse"
    1: "linux"
    2: "unix"
    3: "os"
  ]
}
```
Seems like the platform information isn't set correctly.

czunker:
> It looks like we do not clean up afterward. The disk created from the snapshot is still present.
> 
> * [ ]  Double-check and fix

We do not clean up. We have a func to take care of this, but it isn't called. Looks like we are missing a point to hook this in: https://github.com/mondoohq/cnquery/issues/1628

czunker:
This now looks like v8:
```
DBG graph has received all datapoints
2 errors occured:
	* cannot find 'eol' in resource 'asset'
	* To use this resource, you must authenticate with Mondoo Platform.
To learn how, read:
asset: {
  kind: "api"
  build: ""
  version: "15.1"
  title: "SUSE Linux Enterprise Server 15 SP1, GCP Virtual Machine"
  arch: ""
  labels: {}
  platform: "sles"
  vulnerabilityReport: To use this resource, you must authenticate with Mondoo Platform.
To learn how, read:
  eol: cannot find 'eol' in resource 'asset'
  runtime: "gcp-vm"
  ids: [
    0: "//platformid.api.mondoo.app/runtime/gcp/compute/v1/projects/mondoo-dev-262313/snapshots/suse15"
  ]
  name: "suse15"
  fqdn: ""
  family: [
    0: "suse"
    1: "linux"
    2: "unix"
    3: "os"
  ]
}
```

But it needs some re-work, once the cross provider feature is implemented and the close func can be aplied.

czunker:
After https://github.com/mondoohq/cnquery/pull/1632.

The `Shutdown` is required to clean up.

apps/cnquery/cmd/plugin.go
```diff
@@ -102,6 +102,14 @@ func (c *cnqueryPlugin) RunQuery(conf *run.RunQueryConfig, runtime *providers.Ru
 		}
 	}
 
+	// FIXME: workaround for gcp-snapshot
+	// For a gcp-snapshot asset, we start with a GCP connection.
+	// This get's overriden by a filesystem connection. The fileswystem connection is what we need for the scan
+	// But later, we need the GCP runtime to cleanup the snapshot disk
+	if runtime.Provider.Instance.Name == "gcp" && runtime.Provider.Connection.Name == "filesystem" {
```
czunker, line 109: It should be possible to have both providers in the same runtime.

Caveat with the gcp-snapshot is, we can call Connect on the gcp connection only once. The 2nd connect will trigger an error because the disk is already attached.

providers/gcp/provider/provider.go
```diff
@@ -166,20 +192,35 @@ func (s *Service) Connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 // It is not necessary to implement this method.
 // If you want to do some cleanup, you can do it here.
 func (s *Service) Shutdown(req *plugin.ShutdownReq) (*plugin.ShutdownRes, error) {
+	for i := range s.runtimes {
+		runtime := s.runtimes[i]
+		// FIXME: I think, we might need the asset here to cleanup the correct connection
```
czunker, line 201: I think, I'll need this for the gcp instance scanning via snapshot. This will be the next PR, as the command still needs migration.

czunker:
This PR is a bit more complicated to test.

You need a GCP VM and run cnquery on this VM. The SA for the VM also needs extended permission for the GCP APIs:
image

The command I used to test this:
```
cnquery run gcp snapshot suse15 --project-id mondoo-dev-262313 -c "platform.eol" --verbose
```

apps/cnquery/cmd/plugin.go
imilchev: if we make sure that calling connect for an asset to which we have already connected is a no-op, that issue will be fixed. I feel like this might be needed in other places too

imilchev:
To test this also cnquery needs to run as root and for that specific command it needs to authenticate with mondoo. I, finally, managed to test this. It's working!

imilchev (approved):
I think this can be further improved in a followup

PR #1618, opened 2023-08-31
✨ migrate windows resources

imilchev:


providers/os/resources/windows.go
```diff
@@ -0,0 +1,184 @@
...
+
+	return windows.ParseComputerInfo(executedCmd.Stdout)
+}
+
+func (wh *mqlWindowsHotfix) id() (string, error) {
+	return wh.HotfixId.Data, nil
+}
+
+func initWindowsHotfix(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error) {
+	if len(args) > 2 {
```
czunker: From the code, I think, `> 1` would be sufficient.

providers/os/resources/windows.go
```diff
@@ -0,0 +1,184 @@
...
+
+	return mqlHotFixes, nil
+}
+
+func (wh *mqlWindowsFeature) id() (string, error) {
+	return wh.Path.Data, nil
+}
+
+func initWindowsFeature(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error) {
+	if len(args) > 2 {
```
czunker: Same here.

czunker (commented):
Code is LGTM. Thanks @imilchev 

I need to test this somehow.

czunker:
Looks like we need to fix cnquery on Windows before testing this:
image

Screenshot from 2023-09-04 13-55-26

I created a windows binary with builtin os provider. The screenshots are from a Windows 2022 VM.

imilchev:
I tested this in combination with #1682 everything seems to be working

imilchev:
it seems like there is an issue with the suggestions. When typing `ssh`, one of the suggestions returned is `windows.security.health` Not sure why

mqlc/mqlc_test.go
```diff
@@ -1783,7 +1783,7 @@ func TestSuggestions(t *testing.T) {
 		{
 			// resource suggestions
 			"ssh",
-			[]string{"os.unix.sshd", "sshd", "sshd.config"},
+			[]string{"os.unix.sshd", "sshd", "sshd.config", "windows.security.health"},
```
imilchev, line 1786: I want to merge this PR so I just added the thing that is returned as a suggestion. When I run cnquery and type `ssh` in the shell I don't see any windows stuff. Not sure where the difference is with the test

arlimus: that looks correct to me, you might not be on builtin or have the old provider installed?

arlimus:
> it seems like there is an issue with the suggestions. When typing `ssh`, one of the suggestions returned is `windows.security.health` Not sure why

This happens because `windowS.securityHealtH`, we have fuzzy searching ;)

czunker:
I gave it another try with builtin providers. Looking good:
image

Thanks @imilchev

PR #1620, opened 2023-09-01
⭐️ MS365 v9 provider

preslavgerchev:
After #1611


providers/ms365/provider/provider.go
```diff
@@ -0,0 +1,242 @@
...
+	if len(req.Asset.Connections) == 0 {
+		return nil, errors.New("no connection options for asset")
+	}
+
+	asset := req.Asset
+	conf := asset.Connections[0]
+	var conn *connection.Ms365Connection
+	var err error
+
+	switch conf.Type {
```
imilchev: can we remove this?

preslavgerchev: nice catch, also did the same with azure

imilchev (approved):

PR #1621, opened 2023-09-01
✨ k8s container image discovery

imilchev:


czunker:
Testing `run` works as much as it can currently work.

Testing `shell` looks like this:
image

But, I think, this is caused by https://github.com/mondoohq/cnquery/issues/1645.

When you use `run`, the name of the container image is also empty:
```
cnquery run k8s --discover container-images -c "container.image" 
cannot find connection type k8s
container.image: no data available
1 error occurred:
	* cannot convert primitive with NO type information
container.image: container.image name=null
...
```

providers/assets.go
```diff
@@ -0,0 +1,68 @@
...
+
+// filterUniqueAssets filters assets with duplicate platform IDs
+func filterUniqueAssets(assetCandidates []*inventory.Asset) []*inventory.Asset {
+	uniqueAssets := []*inventory.Asset{}
+	platformIds := map[string]struct{}{}
+	for _, asset := range assetCandidates {
+		found := false
+		for _, platformId := range asset.PlatformIds {
+			if _, ok := platformIds[platformId]; ok {
+				found = true
```
czunker, line 59: Should we add a debug message here?

Yes, platformIDs should be unique, but what if not? Then debugging will get hard.

czunker:
With providers as builtin, `shell` works:
image

czunker:
With the latest commit, it also works for installed providers:
image

czunker (approved):
Thanks @imilchev

PR #1623, opened 2023-09-01
Rename OCI to OCI Tenancy

tas50:
These are tenancies. We should be super clear since OCI does have a concept of an organization. Eventually we should really rename the actual platform from oci -> oci-tenancy, but for now this makes it more clear in the console.


resources/packs/oci/oci.lr
```diff
@@ -1,14 +1,14 @@
 option go_package = "go.mondoo.com/cnquery/resources/packs/oci"
 
-// Oracle Cloud Infrastructure Tenant
+// Oracle Cloud Infrastructure (OCI) Tenancy
```
tas50, line 3: This improves search results in our internal docs search as well as via Google. Otherwise, the MQL docs pages lack the word Oracle in them.

vjeffrey (approved):

PR #1630, opened 2023-09-04
🧹 add ConnectionTypes for all providers

imilchev:
This change is needed to make sure we can properly do cross-provider discovery of resources


czunker (approved):
Nice! Thanks @imilchev

PR #1632, opened 2023-09-04
🧹 Add shutdown to v9 plugins

### LINKED ISSUE #1628: v9 needs a plugin close mechanism
**Describe the bug**
We need a mechanism to clean up connections after they are no longer used.

E.g., when doing a GCP snapshot scan, we create disks in GCP, attach these to a VM, ...

But we never clean up. I think, we need to do this after cnquery is done with executing the queries. The shell already has an onCloseHandler which takes care of recordings. This might also be the point where we would need to hook into with a close method for the connection.

**To Reproduce**

1. Run cnquery snapshot scan
2. Wait for scan to finish successful.
3. Check the GCP disks
4. Note that the created disks aren't deleted.

**Expected behavior**
I would expect that cnquery cleans up after it finished the scan.

czunker:
This allows us to call clean-up funcs, when we close the connection.

Fixes #1628


czunker:
Here is an example how I used it to clean up GCP instance snapshot disks: https://github.com/mondoohq/cnquery/pull/1616/files#diff-23a98f8734bcd98d0e69e0c5439e7f5307bd6aff8c729caa90e7f34ca14d861dR191

providers-sdk/v1/plugin/interface.go
```diff
@@ -32,6 +33,7 @@ type ProviderCallback interface {
 type ProviderPlugin interface {
 	ParseCLI(req *ParseCLIReq) (*ParseCLIRes, error)
 	Connect(req *ConnectReq, callback ProviderCallback) (*ConnectRes, error)
+	Disconnect(req *emptypb.Empty) error
```
preslavgerchev: if I look at how providers are currently structured, we can assume one provider can hold multiple connections at at time.

Maybe we should let the disconnect function specify a connection id so we can shut down a specific one, rather than the whole provider

providers/arista/provider/provider.go
```diff
@@ -109,6 +110,10 @@ func (s *Service) Connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 	}, nil
 }
 
+func (s *Service) Disconnect(req *emptypb.Empty) error {
+	return nil
```
preslavgerchev: maybe we should throw a 'not implemented' err instead. else the caller might assume it's been a successful disconnect where in reality we don't do anything here

preslavgerchev: although returning nil might also be fine, else we will be getting errors every time we close something

apps/provider-scaffold/template/provider/provider.go.template
```diff
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"strconv"
 
+  "google.golang.org/protobuf/types/known/emptypb"
```
preslavgerchev: indentation

providers/arista/provider/provider.go
czunker: That was my intent. It will be called all the time, when the shell closes. In most cases, it will just do nothing.

But writing this, I should add a comment to make this clear.

apps/provider-scaffold/template/provider/provider.go.template
czunker: Fixed

providers-sdk/v1/plugin/interface.go
czunker: Currently, Disconnect is called short before cnspec exits. So, I'd assume none of the connections is needed any longer.

Do you think of other use cases for this func? Who would know which connection to shut down? My current understanding is, that we handle all connections inside a plugin. Shouldn't we then also shut down all these connections inside a plugin?

A lot of questions, I know. I implemented this for the use cases I could think of. Perhaps there are other ones I missed. That's why I asked so many questions.

vjeffrey: perhaps in a scenario like: 
open up aws connection, open up fs connection to scan filesystem, scan filesystem, close fs connection, do a couple more things, close aws connection

in that case we would want to close the fs connection before we're totally done with the aws connection, i think

preslavgerchev: I personally cannot think of a use case, i only based this on what the `ConnectRes` looks like. I think it's fine for now to leave it like this, perhaps we could replace the google's `Empty` struct with our own that we can then extend in the future, if that's needed

arlimus: Semantics first: Disconnect being the opposite of Connect, I'd love for it to behave in a similar way. Where connect opens a connection, it would be great for disconnect to also only target one connection.

Thank you for sharing the use-case in https://github.com/mondoohq/cnquery/issues/1628 @czunker , this helped a lot! In that case, I'd rather we add a Shutdown call that is used to clean up anything the provider plugin wants to before terminating. 

I think Disconnect for specific connections is also a great idea. For now, it looks like the Shutdown of the provider is all we need to fulfill the use-case? That would make things a bit simpler.

(A note on this down the line: The use-case where we have a ton on assets discovered that create a ton of connections is one where something like disconnect for specific connections will be useful. We don't optimize for this case just yet, but after v9 is finished I think it's something we can start to look into, especially when memory issues start to rise. There is also a case for starting individual providers for connections, which will slow down some of the exchange, but rapidly accelerate memory cleanup on shutdown ;) )

One more thing on Shutdown: Please make sure to include a sane timeout for cleanup actions, so we don't get stuck forever. It's fine to do this in a follow-up, doesn't have to be right away...

czunker: Thanks al for the feedback.

I renamed the func to Shutdown, added a timeout, and incorporated some other feedback. Please have another look.

I tried it with the gcp snapshot PR and it worked there.

About the sane timeout:
"sane" is a bit difficult here. I opted for 2 minutes. The only timeout I found in the code is for VMWare, which is 5 minutes. I didn't want to go too low, because in the GCP example we are calling cloud APIs. So, hopefully 2 minutes is fine.

It might be a solution to add it as an option to providers in one of the upcoming iterations. Or use a CLI flag to set it.

vjeffrey (approved):

PR #1633, opened 2023-09-04
🧹 Generate YAML LR manifest

preslavgerchev:
We don't seem to be generating the .yaml manifests for the LR definitions, so I've added that in

Additionally, I've ran that for ms365.


czunker (approved):
Nice. Thanks @preslavgerchev 
I was already wondering why these files were missing.

PR #1639, opened 2023-09-05
🐛 add missing ms365 in make providers/build

arlimus:
also fix the runtime initialization in the provider to work again


providers/ms365/provider/provider.go
```diff
@@ -128,7 +128,6 @@ func (s *Service) connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 	asset.Connections[0].Id = conn.ID()
 	s.runtimes[conn.ID()] = &plugin.Runtime{
 		Connection:     conn,
-		Resources:      syncx.Map[string]plugin.Resource{},
```
arlimus, line 131: not necessary anymore for the `syncx.Map`, since it uses `sync.Map` under the hood

vjeffrey (approved):

PR #1646, opened 2023-09-06
🐛 avoid crashing when no additional assets are discovered

arlimus:


arlimus:
> I'm lacking a bit of context here. The code looks valid, but I'm unsure which issue this PR should solve.
> 
> I tried different `cnquery run` with `local`, `ssh`, and `aws`. I see the same results with main and this PR.

```
..go/src/go.mondoo.com/cnquery ±> cnquery shell host https://mondoo.com -c 'asset'                                                  x130 20h25m main[4b80d5fc]
! using builtin provider for network
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x1d84da7]

goroutine 1 [running]:
go.mondoo.com/cnquery/apps/cnquery/cmd.StartShell(0xc0004f5a40, 0xc001020180)
	/pub/go/src/go.mondoo.com/cnquery/apps/cnquery/cmd/shell.go:102 +0x107
go.mondoo.com/cnquery/apps/cnquery/cmd.glob..func23(0x0?, 0xc001020120?, 0x27?)
	/pub/go/src/go.mondoo.com/cnquery/apps/cnquery/cmd/shell.go:43 +0x28
go.mondoo.com/cnquery/cli/providers.setConnector.func2(0xc0006ea500?, {0xc00066bb90?, 0x1, 0x3})
	/pub/go/src/go.mondoo.com/cnquery/cli/providers/providers.go:323 +0xeb5
github.com/spf13/cobra.(*Command).execute(0xc0006f8300, {0xc00066bb30, 0x3, 0x3})
	/pub/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:944 +0x863
github.com/spf13/cobra.(*Command).ExecuteC(0x3921120)
	/pub/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:1068 +0x3a5
github.com/spf13/cobra.(*Command).Execute(...)
	/pub/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:992
go.mondoo.com/cnquery/apps/cnquery/cmd.Execute()
	/pub/go/src/go.mondoo.com/cnquery/apps/cnquery/cmd/root.go:73 +0x211
main.main()
	/pub/go/src/go.mondoo.com/cnquery/apps/cnquery/cnquery.go:9 +0xf
```

Specifically: 

```assetList := res.Inventory.Spec.Assets```

The connect call doesn't need to return an inventory, if no additional assets are discovered.

File: apps/cnquery/cmd/plugin.go [hunk 1]
```diff
@@ -88,20 +87,6 @@ func (c *cnqueryPlugin) RunQuery(conf *run.RunQueryConfig, runtime *providers.Ru
 		return err
 	}
 
-	assetList := runtime.Provider.Connection.Inventory.Spec.Assets
-	log.Debug().Msgf("resolved %d assets", len(assetList))
-
-	assetCandidates := []*inventory.Asset{}
-	if len(assetList) > 1 && conf.PlatformId != "" {
-		filteredAsset, err := filterAssetByPlatformID(assetList, conf.PlatformId)
```
arlimus, line 96: @imilchev Afaics we need to run your new detection code before we can filter assets by platform ID, namely the part: `detectAssets connects to all assets that do not have a platform ID yet`. That's why I removed all platformID handling here and put it into `ProcessAssetCandidates` instead, so that this call results in the final assets after detection, deduplication, and platformID selection

czunker:
I did some more tests and can confirm that this PR fixes the mentioned issues. Thanks for the examples.

I also tried some other queries and discoveries. There are still some issues: https://github.com/mondoohq/cnquery/issues/1654

But these issues are independent of this PR. They already exist in main. I think we should solve them in a follow-up PR.

czunker (approved):
Thanks @arlimus

PR #1647, opened 2023-09-06
🐛 gcp discovery fixes

imilchev:
- add `auto` and `all` discovery
- fix bigquery dataset discovery issue
- fix bigquery dataset platform ID

### DIFF
--- a/providers/gcp/config/config.go
+++ b/providers/gcp/config/config.go
@@ -20,6 +20,8 @@ var Config = plugin.Provider{
 			Use:   "gcp",
 			Short: "GCP Cloud",
 			Discovery: []string{
+				resources.DiscoveryAll,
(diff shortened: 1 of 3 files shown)


czunker:
Discovery all is looking good:
image

Auto does not discover anything, but that's the same with v8.

Bigquery discovery is also looking good:
image

czunker:
But something is wrong when I select a bigquery dataset:
image

Perhaps something for a follow-up PR.

czunker (approved):
Thanks @imilchev

imilchev:
> But something is wrong when I select a bigquery dataset: image
> 
> Perhaps something for a follow-up PR.

aliases aren't working at the moment. If you want to query this you need to use the resource directly `gcp.project.bigqueryService.datasets{*}`

PR #1649, opened 2023-09-06
Attempt at fixing the scan runtime problem.

preslavgerchev:
Not necessarily the fix, but an attempt to get a proper solution for this.

There are 2 problems I am trying to solve here:
1. Do not rely on two separate `assets` and `runtime` lists to get the scan to work. This will crash if we do not have an equal amount of assets and runtimes. Instead use a private struct to tie those two together
2. Running filters (queries) on a provider uses the currently attached asset, see https://github.com/mondoohq/cnquery/blob/main/providers/runtime.go#L419. At this moment we can only attach one asset, found under `r.Provider.Connection.Asset`. To ensure this works for all discovered assets, we need to make sure there's a proper runtime for each of those assets. This is where I am not sure if this is the right fix as we're now creating a runtime for each asset.

### DIFF
--- a/explorer/scan/local_scanner.go
+++ b/explorer/scan/local_scanner.go
@@ -13,12 +13,14 @@ import (
 	"time"
 
 	"go.mondoo.com/cnquery/providers-sdk/v1/inventory/manager"
+	"go.mondoo.com/cnquery/providers-sdk/v1/plugin"
 
 	"github.com/mattn/go-isatty"
 	"github.com/rs/zerolog/log"
 	"github.com/segmentio/ksuid"
 	"github.com/spf13/viper"
 	"go.mondoo.com/cnquery"
+	"go.mondoo.com/cnquery/cli/config"
 	"go.mondoo.com/cnquery/cli/progress"
 	"go.mondoo.com/cnquery/explorer"
 	"go.mondoo.com/cnquery/explorer/executor"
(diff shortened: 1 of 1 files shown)


explorer/scan/local_scanner.go
```diff
@@ -184,9 +190,34 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 		// we grab the asset from the connection, because it contains all the
 		// detected metadata (and IDs)
 		// assets = append(assets, runtime.Provider.Connection.Asset)
-		runtimes = append(runtimes, runtime)
 	}
 
+	// for each asset candidate, we initialize a new runtime and connect to it.
+	for _, asset := range assetCandidates {
```
vjeffrey, line 202: what if we keep a map of runtimes as we go through these, so that we can reuse an already created runtime if its been created?

preslavgerchev: I started doing something similar, but then I read this comment above the runtime struct:
```
// Runtimes are associated with one asset and carry all providers
// and open connections for that asset.
```

I am unsure whether we should be reusing the runtime now or just create 1 per asset.

explorer/scan/local_scanner.go
```diff
@@ -175,7 +181,7 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 			return nil, false, err
 		}
 		log.Debug().Msgf("adding %d asset(s)", len(runtime.Provider.Connection.Inventory.Spec.Assets))
-		assets = append(assets, runtime.Provider.Connection.Inventory.Spec.Assets...)
+		assetCandidates = append(assetCandidates, runtime.Provider.Connection.Inventory.Spec.Assets...)
```
vjeffrey: i think we also need to add a guard here, https://github.com/mondoohq/cnquery/compare/vj/gitlab?expand=1

vjeffrey:
i'm gonna vote for getting this merged in and creating an issue to revisit. i sent the pr to dom but he didn't have time to look and he's driving all day today, so let's just move forward and loop back around as needed

vjeffrey:
ill re-test, but last time i tested this things seemed to be working well

vjeffrey (approved):
approving -- but please open an issue with this pr linked and assign it to dom, so we remember to loop back around to this

preslavgerchev:
> approving -- but please open an issue with this pr linked and assign it to dom, so we remember to loop back around to this

will do, thanks @vjeffrey 

i will re-test as i have a wip commit and then merge

vjeffrey:
@preslavgerchev i think your wip commit is just the guard for inventory.spec.assets, so looks good 👍

vjeffrey:
alright, tested again, getting this merged in, we can followup as needed - thank you!

PR #1650, opened 2023-09-06
✨ gcp gcr repository scanning v9

### LINKED ISSUE #1648: `gcp gcr` scan with project does not work as described
**Describe the bug**
```
cnquery run gcp gcr mondoo-dev-XYZ --repository ubuntu -c "container.image" --verbose 
DBG parsing asset asset-type=14 provider=CONTAINER_REGISTRY
DBG run resolver resolver="Container Registry Discover" resolver-id=cr
→ fetch meta information from container registry registry=gcr.iomondoo-dev-XYZ/ubuntu
x could not fetch container images error="Get \"https://gcr.iomondoo-dev-XYZ/v2/\": dial tcp: lookup gcr.iomondoo-dev-XYZ on 127.0.0.53:53: no such host"
x could not resolve asset error="Get \"https://gcr.iomondoo-dev-XYZ/v2/\": dial tcp: lookup gcr.iomondoo-dev-XYZ on 127.0.0.53:53: no such host" asset=gcr.iomondoo-dev-XYZ/ubuntu
FTL failed to run query error="could not find an asset that we can connect to"
```

There is a slash missing in the URL: `https://gcr.iomondoo-dev-XYZ/v2/`

When you add the slash to the project id, it works.

**To Reproduce**

1. Run the above command
2. Note the error

**Expected behavior**
The command should scan the gcr repository.

- cnquery v8.25.0

czunker:
Fixes #1648 and ports the v8 connection to v9


providers/gcp/resources/discovery/gcr/gcr.go
```diff
@@ -52,21 +49,22 @@ func (a *GcrImages) ListRepository(repository string, recursive bool) ([]*asset.
 			repoURL := repo.String()
 			imageUrl := repoURL + "@" + digest
 
-			asset := &asset.Asset{
+			asset := &inventory.Asset{
 				PlatformIds: []string{MondooContainerImageID(digest)},
 				Name:        containerid.ShortContainerImageID(digest),
```
imilchev: this looks like the k8s container-image discovery. I think we should match the approach and just return an asset with a connection config and nothing else. We should let the os provider fill in the platform IDs, name, etc.

providers/gcp/resources/discovery/gcr/gcr.go
imilchev, line 1: do we need to put this in a separate folder? the rest of the discovery is all in the `discovery.go`

providers/gcp/resources/discovery/gcr/gcr.go
```diff
@@ -112,13 +110,8 @@ func (a *GcrImages) ListRepository(repository string, recursive bool) ([]*asset.
...
-	assets := []*asset.Asset{}
-	// repoAssets, err := a.ListRepository("index.docker.io/mondoo/client", false)
-	// if err == nil && repoAssets != nil {
-	// 	assets = append(assets, repoAssets...)
-	// }
-	// return assets, nil
+func (a *GcrImages) List() ([]*inventory.Asset, error) {
+	assets := []*inventory.Asset{}
 
 	resSrv, err := cloudresourcemanager.NewService(context.Background())
```
imilchev: Can we do this via MQL?

providers/gcp/resources/discovery/gcr/resolver_gcr.go
```diff
@@ -20,18 +18,19 @@ func (r *GcrResolver) Name() string {
...
+	return []string{"auto", "all"}
 }
 
-func (r *GcrResolver) Resolve(ctx context.Context, root *asset.Asset, t *providers.Config, credsResolver vault.Resolver, sfn common.QuerySecretFn, userIdDetectors ...providers.PlatformIdDetector) ([]*asset.Asset, error) {
-	resolved := []*asset.Asset{}
-	repository := t.Host
+// func (r *GcrResolver) Resolve(ctx context.Context, root *inventory.Asset, t *inventory.Config, credsResolver vault.Resolver, sfn common.QuerySecretFn, userIdDetectors ...providers.PlatformIdDetector) ([]*inventory.Asset, error) {
+func (r *GcrResolver) Resolve(ctx context.Context, root *inventory.Asset, conf *inventory.Config, credsResolver vault.Resolver) ([]*inventory.Asset, error) {
+	resolved := []*inventory.Asset{}
+	repository := conf.Host
```
imilchev: i guerss this code can go in `discovery.go`

providers/gcp/resources/discovery/gcr/gcr.go
czunker: :thinking: Strange. I used your PR as an example and removed this. Not sure why it is back in.

Thanks for catching this.

providers/gcp/resources/discovery/gcr/gcr.go
czunker: Done

providers/gcp/resources/discovery/gcr/resolver_gcr.go
czunker: Done

providers/gcp/resources/discovery/gcr/gcr.go
czunker: Perhaps...

But this would require adding multiple MQL resources. And I'd prefer to actually get rid of the repository resource we currently have: https://github.com/mondoohq/cnquery/issues/1651

This command is only about the images. We do not care about the rest at this point.

providers/gcp/resources/discovery.go
```diff
@@ -362,3 +384,127 @@ func cloneConfig(invConf *inventory.Config) *inventory.Config {
...
+	if err != nil {
+		return nil, err
+	}
+	return imgs, nil
+}
+
+// List uses your GCP credentials to iterate over all your projects to identify potential repos
+func (a *GcrImages) List() ([]*inventory.Asset, error) {
+	assets := []*inventory.Asset{}
+
```
vjeffrey, line 479: why not use the resource directly here instead of calling the api?

czunker: My understanding is, we do not have these resources.

At the end, `gcp gcr` and `container registry` commands, both create `container.image` resources, but how we get there is different.

`gcp gcr` does not use a container registry API, but the GCP API. We do not have the intermediate resources `gcp.gcr` and `gcp.gcr.repository`. As I understand this provider, we do not care about these.

So, we start by iterating over all repositories which are present in a GCP project. We take these repositories and iterate over all the images which in these repositories. These images end up as resources, but nothing in between.

providers/gcp/resources/discovery/gcr/gcr.go
czunker: As @vjeffrey had the same question, I dug a bit deeper. Now, I think it isn't possible: https://github.com/mondoohq/cnquery/pull/1650#discussion_r1319335451

PR #1655, opened 2023-09-07
🐛 fix github discovery for v9

vjeffrey:
fixes https://github.com/mondoohq/cnquery/issues/1644

### DIFF
--- a/providers/github/provider/provider.go
+++ b/providers/github/provider/provider.go
@@ -60,16 +60,23 @@ func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error)
 	}
 	conf.Credentials = append(conf.Credentials, vault.NewPasswordCredential("", token))
 
+	// discovery flags
+	discoverTargets := []string{}
+	if x, ok := flags["discover"]; ok && len(x.Array) != 0 {
+		for i := range x.Array {
+			entry := string(x.Array[i].Value)
+			discoverTargets = append(discoverTargets, entry)
+		}
+	}
+	conf.Discover = &inventory.Discovery{Targets: discoverTargets}
+
 	// Do custom flag parsing here
 	switch req.Args[0] {
 	case "org":
-		conf.Discover.Targets = []string{connection.DiscoveryOrganization}
 		conf.Options["organization"] = req.Args[1]
 	case "user":
-		conf.Discover.Targets = []string{connection.DiscoveryUser}
 		conf.Options["user"] = req.Args[1]
 	case "repo":
-		conf.Discover.Targets = []string{connection.DiscoveryRepository}
 		conf.Options["repository"] = req.Args[1]
 	default:
 		return nil, errors.New("invalid GitHub sub-command")
@@ -98,15 +105,16 @@ func (s *Service) Connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 			return nil, err
 		}
 	}
-	inventory, err := s.discover(conn)
+	inv, err := s.discover(conn)
 	if err != nil {
 		return nil, err
 	}
+
 	return &plugin.ConnectRes{
 		Id:        conn.ID(),
 		Name:      conn.Name(),
 		Asset:     req.Asset,
-		Inventory: inventory,
+		Inventory: inv,
 	}, nil
 }
 
--- a/providers/github/provider/provider.go
+++ b/providers/github/provider/provider.go
@@ -38,9 +38,14 @@ func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error)
 		flags = map[string]*llx.Primitive{}
 	}
 
+	if len(req.Args) == 0 {
+		return nil, errors.New("invalid. must specify org, repo, or user")
+	}
+
 	conf := &inventory.Config{
-		Type:    req.Connector,
-		Options: map[string]string{},
+		Type:     req.Connector,
+		Options:  map[string]string{},
+		Discover: &inventory.Discovery{},
 	}
 
 	token := ""
@@ -58,10 +63,13 @@ func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error)
 	// Do custom flag parsing here
 	switch req.Args[0] {
 	case "org":
+		conf.Discover.Targets = []string{connection.DiscoveryOrganization}
 		conf.Options["organization"] = req.Args[1]
 	case "user":
+		conf.Discover.Targets = []string{connection.DiscoveryUser}
 		conf.Options["user"] = req.Args[1]
 	case "repo":
+		conf.Discover.Targets = []string{connection.DiscoveryRepository}
 		conf.Options["repository"] = req.Args[1]
 	default:
 		return nil, errors.New("invalid GitHub sub-command")
@@ -90,12 +98,15 @@ func (s *Service) Connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 			return nil, err
 		}
 	}
-
+	inventory, err := s.discover(conn)
+	if err != nil {
+		return nil, err
+	}
 	return &plugin.ConnectRes{
 		Id:        conn.ID(),
 		Name:      conn.Name(),
 		Asset:     req.Asset,
-		Inventory: nil,
+		Inventory: inventory,
 	}, nil
 }
 
@@ -168,7 +179,6 @@ func (s *Service) detect(asset *inventory.Asset, conn *connection.GithubConnecti
 	if err != nil {
 		return err
 	}
-	// TODO: Add platform IDs
 	asset.PlatformIds = []string{id}
 	return nil
 }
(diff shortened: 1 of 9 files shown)


providers/github/resources/discovery.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func repo(runtime *plugin.Runtime, repoName string, conn *connection.GithubConnection) ([]*inventory.Asset, error) {
+	assetList := []*inventory.Asset{}
+
+	repo, err := getMqlGithubRepo(runtime, repoName)
+	if err != nil {
+		return nil, err
+	}
+	// repo.Owner.Data
```
czunker: nit: This can be removed.

providers/github/resources/github_org.go
```diff
@@ -161,7 +157,7 @@ func (g *mqlGithubOrganization) owners() ([]interface{}, error) {
 			id = *member.ID
 		}
 
-		r, err := CreateResource(g.MqlRuntime, "github.user", map[string]*llx.RawData{
+		r, err := g.MqlRuntime.CreateResource(g.MqlRuntime, "github.user", map[string]*llx.RawData{
```
czunker: This looks strange to me.

I'm only aware of `CreateResource` usage, not `g.MqlRuntime.CreateResource`. The runtime is now passed in as a parameter.

czunker (commented):
Thanks for fixing this @vjeffrey

czunker:
I gave it a try:
```
cnquery run github org mondoohq --discover repos -c "github.repository.license.spdxId" --verbose                           ✔ │ 06:20:34 
DBG using provider github with connector github
DBG no need to update provider last-refresh=58.448175428s provider=github
DBG running provider plugin path=/home/christian/.config/mondoo/providers/github/github
DBG resolved 27 assets
DBG starting query execution qrid=Xiag1Adi3l0=
DBG finished query execution qrid=Xiag1Adi3l0=
DBG rmN/PU9BjTuMDKP3oZQ+j/kU/Cfvkp43OHrhOb+XrZl8zyhYpzNgNYLj9DnlWxhtCnFZPFGJU0gBGky8CvE4lw== finished
DBG graph has received all datapoints
rpc error: code = Unknown desc = no user provided
github.repository.license.spdxId: no data available
DBG starting query execution qrid=Xiag1Adi3l0=
DBG rmN/PU9BjTuMDKP3oZQ+j/kU/Cfvkp43OHrhOb+XrZl8zyhYpzNgNYLj9DnlWxhtCnFZPFGJU0gBGky8CvE4lw== finished
DBG graph has received all datapoints
DBG finished query execution qrid=Xiag1Adi3l0=
rpc error: code = Unknown desc = no user provided
....
```

To me, it looks like the discovery of the repos is working. But then some data is missing to actually create the repo resource.

vjeffrey:
if i try to run the command you tried in v8, i also error --- i think it needs to be:
```
go run apps/cnquery/cnquery.go run github org mondoohq  -c "github.organization.repositories{license.spdxId }" --verbose
```

czunker:
> if i try to run the command you tried in v8, i also error --- i think it needs to be:
> 
> ```
> go run apps/cnquery/cnquery.go run github org mondoohq  -c "github.organization.repositories{license.spdxId }" --verbose
> ```

I did a bit more testing. My first command was wrong, but what I tried to achieve, works in v8 in case you use the correct parameters:
```
cnquery run github org mondoohq --discover repository -c "github.repository.license.spdxId"                                                                       
→ resolved assets resolved-assets=26
Query encountered errors:
cannot cast resource to resource type: <nil>
github.repository.license.spdxId: no data available
...
github.repository.license.spdxId: "Apache-2.0"
github.repository.license.spdxId: "MPL-2.0"
github.repository.license.spdxId: "MPL-2.0"

...
```

In my first try, I used the wrong discovery option: `--discover repo` instead of `--discover repository`

But using the correct discovery in v9, does not show the same result as in v8.

providers/github/provider/provider.go
```diff
@@ -58,10 +63,13 @@ func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error)
 	// Do custom flag parsing here
 	switch req.Args[0] {
 	case "org":
+		conf.Discover.Targets = []string{connection.DiscoveryOrganization}
```
czunker: I think, this is not correct. It uses the first arg as discovery target and not what is specified by the discovery flag:

```
cnquery run github org mondoohq --discover repository -c "github.repository.license.spdxId"
```
In this example, `req.Args[0]` is `org`, but I want to discover the repositories.

czunker:
Also the error message in my first try was misleading. I opened a PR to fix this: https://github.com/mondoohq/cnquery/pull/1657

providers/github/provider/provider.go
vjeffrey: ah gotcha. i didn't realize we actually had discovery targets in v8 for GitHub, it was discovering repos within the org automatically for me without the discovery options -- ill get this all sorted, thanks for testing!

vjeffrey:
hmm, actually i have the platform ids, they look a little off

czunker:
> alright, this is fixed up now, `cnquery shell github org mondoohq --discover all` `cnquery shell github org mondoohq --discover repos,users`

The discovery options are different for v8:
```
cnquery run github org mondoohq --discover repository -c "asset{ ids }" 
→ resolved assets resolved-assets=26
asset: {
  ids: [
    0: "//platformid.api.mondoo.app/runtime/github/owner/mondoohq/repository/installer"
  ]
}
...
```
v9:
```
cnquery run github org mondoohq --discover repos -c "asset{ ids }" 
asset: {
  ids: [
    0: "//platformid.api.mondoo.app/runtime/github/organization/Mondoo Inc"
  ]
}
asset: {
  ids: [
    0: "//platformid.api.mondoo.app/runtime/github/owner/Mondoo Inc/repository/installer"
  ]
}

```
And as already mentioned the platformIDs: "Mondoo Inc" vs. "mondoohq"

providers/github/resources/discovery.go
```diff
@@ -0,0 +1,209 @@
...
+			assetList = append(assetList, &inventory.Asset{
+				PlatformIds: []string{connection.NewGitHubRepoIdentifier(org.Name.Data, repo.Name.Data)},
+				Name:        org.Name.Data + "/" + repo.Name.Data,
+				Platform:    connection.GithubRepoPlatform,
+				Labels:      make(map[string]string),
+				Connections: []*inventory.Config{cloneInventoryConf(conn.Conf)},
+			})
+		}
+	}
+	if stringx.Contains(targets, connection.DiscoveryUsers) {
```
czunker: Nice! That's actually a new feature. We don't have this in v8.

providers/github/resources/github_repo.go
```diff
@@ -470,8 +470,26 @@ func (g *mqlGithubBranch) protectionRules() (*mqlGithubBranchprotection, error)
...
+				"requiredSignatures":             llx.BoolData(false),
+			})
+			if err != nil {
+				return nil, err
+			}
+			return res.(*mqlGithubBranchprotection), nil
+		}
 		// TODO: figure out if the client has the permission to fetch the protection rules
-		return nil, nil
+		return nil, err
```
czunker, line 479: @vjeffrey I'd like your opinion on this one.

When I keep it as is, this panics:
```
cnquery run github repo czunker/go-vex -c "github.repository{ name branches[0]{*} commits[0]{*} }" 
! using builtin provider for github
! using builtin provider for github
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x1180360]

goroutine 14 [running]:
go.mondoo.com/cnquery/providers/github/resources.(*mqlGithubBranchprotection).MqlID(0x2f0000c000cd04ea?)
	/home/christian/workspace/mondoo/github.com/cnquery/providers/github/resources/github.lr.go:3937
go.mondoo.com/cnquery/llx.resource2result({0x22861a0, 0x0}, {0xc000fea348, 0x18})
	/home/christian/workspace/mondoo/github.com/cnquery/llx/data_conversions.go:313 +0x52
```

When I only create a resource with an ID, it prevents the panic, but I get "cannot convert" error.

In v8, this simply displayed as "null":
```
cnquery run github repo czunker/go-vex -c "github.repository{ name branches[0]{ name protectionRules {*} } commits[0]{ sha commit {*}} }"
...
github.repository: {
...
  name: "go-vex"
  branches[0]: {
    protectionRules: null
    name: "christian/extend_csaf_advisories"
  }
}
```

This change works, but is quite different from v8. Also, this might apply to other resources.

vjeffrey: i'll check that out, i think i ran into that before on aws

vjeffrey: this looks like it's an issue not limited to this resource -- returning nil in these types of situations isn't behaving the way we expect it to - so we can probably move this pr forward and fix that in a followup

czunker: I rebased this on main, as https://github.com/mondoohq/cnquery/pull/1726 was merged.

I changed the field accordingly, and now it looks good:
```
nquery run github repo czunker/go-vex -c "github.repository{ name branches[0]{ name protectionRules {*} }}" --verbose 
DBG using provider github with connector github
DBG no need to update provider last-refresh=29.146104804s provider=github
DBG running provider plugin path=/home/christian/.config/mondoo/providers/github/github
DBG resolved 1 assets
...
github.repository: {
  name: "go-vex"
  branches[0]: {
    protectionRules: null
    name: "christian/extend_csaf_advisories"
  }
}
```

czunker:
I gave some other fields a try where we also return `nil, nil`, but those seem fine, e.g.:
```
cnquery run github repo czunker/backend -c "github.repository.openMergeRequests{*}"
...
github.repository.openMergeRequests: []
```

@vjeffrey I think this PR is done for now, or do you see any open topics?

czunker (approved):
LGTM. @vjeffrey should have another look as I also did part of the commits.

providers/github/resources/github_repo.go
vjeffrey: awesome! thank you @arlimus for that fix!

PR #1656, opened 2023-09-07
🐛 clarify resource extensions in LR files

arlimus:
We saw a race condition where the load order of providers would decide which provider became the initializer and owner of a resource. This especially led to issues with the `asset` resource, which would randomly get assigned to `core` and `os`, depending on who came first. When it was assigned to `core` that was fine and intended. But whenever it got assigned to `os` that provider would get a `connect` call with a backend that it just did not support (unless you used `os` already). For example: `cnquery run host https://... -c asset` would try to connect with the `os` provider to fill in contents for the `asset` resource, just because `os` extended it.

Now we have an explicit keyword that is used to extend resources: `extend` (doh). You can see an example in the `os` provider:

```
extend asset {
  vulnerabilityReport() dict
}
```


imilchev (approved):
Just tested this and it seems to fix the issues we had with `asset` and `mondoo` resources. Thanks @arlimus :shipit:

PR #1663, opened 2023-09-07
🐛 make sure providers are not killed after an asset is scanned

imilchev:
This makes sure that scans for a list of assets work. The current implementation would kill the provider after the first asset is scanned. We do a simple fix here that will only kill open providers after the whole scan is complete. As a long term fix we should probably look into keeping track of assets that need a specific provider. Only when all assets for the provider have been scanned, it can be killed.


vjeffrey (approved):
thank you, lets get this merged in and followup as needed since this seems to be unblocking issues

PR #1665, opened 2023-09-08
🐛 fix provider shutdown

imilchev:
- [x] implement `Shutdown` for the provider server
- [x] Do not kill the provider after an asset is scanned
- [x] do we need to call shutdown when the runtime is closed or when the provider is closed?

### DIFF
--- a/providers-sdk/v1/plugin/grpc.go
+++ b/providers-sdk/v1/plugin/grpc.go
@@ -87,6 +87,10 @@ func (m *GRPCServer) Connect(ctx context.Context, req *ConnectReq) (*ConnectRes,
 	return m.Impl.Connect(req, a)
 }
 
+func (m *GRPCServer) Shutdown(ctx context.Context, req *ShutdownReq) (*ShutdownRes, error) {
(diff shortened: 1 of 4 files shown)


preslavgerchev:
I jus ran a full azure scan and things are looking good 👌

providers-sdk/v1/plugin/grpc.go
```diff
@@ -87,6 +87,10 @@ func (m *GRPCServer) Connect(ctx context.Context, req *ConnectReq) (*ConnectRes,
 	return m.Impl.Connect(req, a)
 }
 
+func (m *GRPCServer) Shutdown(ctx context.Context, req *ShutdownReq) (*ShutdownRes, error) {
```
preslavgerchev, line 90: Does that get called anywhere? Seems it's unused in this PR

imilchev: yes, the calling part was implemented in #1632 but wasn't test with non-builtin providers. This code was missing so for all installed provider a not implemented error was returned

preslavgerchev (approved):

PR #1668, opened 2023-09-10
🐛 fix detection of provider via connector

arlimus:
We have conflated connection types (which are coming from asset detection) and connection names (which are coming from CLI). One of the last changes has broken the support of CLI-based connector-detection (at least when the connection type and name differed, like in the case of OS).

This PR clarifies the difference between these types, by allowing the `EnsureProvider` call to either take the type or the name for the connection. In the CLI, we get names like `docker`, which then translate to the `os` provider. In contrast, the connection type is something like `docker-image` which translates to the `os` provider. We could have mixed them all together, but it may lead to confusing results (it should not, but it can and then it really falls apart).

This is also important because `DefaultProviders` uses connector names, fully aligned with the way the CLI config of providers works, which requires a search by name, not by type.

@imilchev fyi ^^ had to clarify the difference to fix regular execution of CLI commands like `run` and `shell`, which had broken

### DIFF
--- a/cli/providers/providers.go
+++ b/cli/providers/providers.go
@@ -34,12 +34,12 @@ func AttachCLIs(rootCmd *cobra.Command, commands ...*Command) error {
 		return err
 	}
 
-	connectorName, autoUpdate := detectConnector(os.Args, rootCmd, commands)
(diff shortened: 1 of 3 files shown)


chris-rock (approved):
Thank you @arlimus

Issue #1662, opened 2023-09-07
mondoo.capabilities is not working


imilchev:
The current implementation of `mondoo.capabilities` does not work. 

https://github.com/mondoohq/cnquery/blob/main/providers/core/resources/mondoo.go#L37-L41

The `core` provider does not have any connections, therefore the code panics since `m.MqlRuntime.Connection` is `nil`

PR #1677, opened 2023-09-11
🐛 fix mondoo.capabilities

arlimus:
1. the detection was broken and crashed and
2. it pulled content from the os provider (sacrilege! luckily only a small file but still)

There were a few options to approach this, but trying to detect this dynamically or shoving it into untyped metadata was not a good approach. Instead we now have a field attached to the connection config which can transmit capabilities per connection.

We still only expose this via the `mondoo.capabilities` field which should be deprecated and replaced with a cleaner approach.


imilchev (approved):
Nice! This fixes #1662

PR #1678, opened 2023-09-11
🐛 support bridge-resources

arlimus:
These are resources without providers, e.g. `my.name.is.bob` as a resource creates the bridge resources `my`, `my.name` and `my.name.is`. These cannot be created in the provider that created `my.name.is.bob` because they are only bridging in nature. That said, users should still be able to type `my.name` and not be greeted by crimson letters desperately proclaiming an error. Not helpful.

This PR not only adds support for this mechanism but also adds more testing to LR generation.

### DIFF
--- a/providers-sdk/v1/lr/lr_test.go
+++ b/providers-sdk/v1/lr/lr_test.go
@@ -8,234 +8,217 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
-func parse(t *testing.T, cmd string, f func(*LR)) {
+func parse(t *testing.T, cmd string) *LR {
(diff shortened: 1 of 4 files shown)


providers-sdk/v1/lr/lr_test.go
```diff
@@ -8,234 +8,217 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
-func parse(t *testing.T, cmd string, f func(*LR)) {
```
arlimus, line 13: embarrassed about these tests, i created those before learning about `require` :see_no_evil: 
all of this is just a rewrite eliminating the function call, 
no functionality was changed in creating this PR

PR #1679, opened 2023-09-11
🐛 os.Connect should not return the asset in the inventory

arlimus:
The behavior of the connect call is defined in the proto here: https://github.com/mondoohq/cnquery/blob/main/providers-sdk/v1/plugin/plugin.proto\#L35-L38 It clarifies that the inventory only returns other discovered assets, while the asset field contains the one asset that the connection call was requesting.

Remove the code that added the connection asset to the inventory, since it does not count as a discovered asset.

### DIFF
--- a/providers/os/provider/provider.go
+++ b/providers/os/provider/provider.go
@@ -206,14 +206,6 @@ func (s *Service) Connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 		}
 	}
 
-	if inv == nil {
-		inv = &inventory.Inventory{
-			Spec: &inventory.InventorySpec{
-				Assets: []*inventory.Asset{req.Asset},
-			},
-		}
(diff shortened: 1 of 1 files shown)


providers/os/provider/provider.go
```diff
@@ -206,14 +206,6 @@ func (s *Service) Connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 		}
 	}
 
-	if inv == nil {
```
imilchev, line 209: I remember I added this code here. I also added a comment why it ended up being like this https://github.com/mondoohq/cnquery/pull/1617/files#r1311634791

imilchev: just tested this and it seems it's no longer an issue

imilchev (approved):

PR #1682, opened 2023-09-11
🐛 fix running on windows

imilchev:
- [x] skip permissions checks for cnquery config on windows
- [x] allow building providers for windows on non-Windows machine
- [x] make sure provider binaries end up on `.exe` for Windows
- [x] if the runtime is Windows look for `.exe` binaries of providers

```
PS C:\Users\ivan\Desktop> .\cnquery.exe run local -c os.name
os.name: "IVAN-WIN"
```


chris-rock (approved):
Thank you @imilchev

PR #1685, opened 2023-09-11
Christian/v9 container image

czunker:


imilchev (approved):
Tested it and it is working correctly! I found another bug though and opened an issue for it #1699

PR #1692, opened 2023-09-12
🐛 fix aws config

vjeffrey:
thanks to @czunker for the catch on this, https://github.com/mondoohq/cnquery/pull/1642#discussion_r1321636162


czunker (approved):
Thanks @vjeffrey

PR #1693, opened 2023-09-12
🐛 fix aliased resources

arlimus:
We could no longer call aliased resource properly. This was partly because the way the schema generation had changed (where I missed to migrate the `IsImplicitResource` field) and partly because the resource lookup had also changed (since we want to call the base resource now with the provider, not the alias).


imilchev (approved):
I tested this and it seems to be working correctly. Thanks!

PR #1696, opened 2023-09-12
🐛 Fix aliases for connectors

czunker:


imilchev (changes requested):
This change only fixes the issue if the provider is builtin. It doesn't work if we are using an installed provider.

czunker:
> This change only fixes the issue if the provider is builtin. It doesn't work if we are using an installed provider.

Nice catch, thanks. I fixed it with another commit:
```
cnquery run fs providers/os/connection/fs/testdata/centos8 -c "file('/etc/hostname').content" --verbose                                                                                                                    
DBG using provider os with connector filesystem
DBG no need to update provider last-refresh=38m38.131337891s provider=os
DBG running provider plugin path=/home/christian/.config/mondoo/providers/os/os
...
file.content: "testhost"
```

chris-rock:
I think you also need to update https://github.com/mondoohq/cnquery/blob/main/providers/defaults.go#L22

czunker:
> I think you also need to update https://github.com/mondoohq/cnquery/blob/main/providers/defaults.go#L22

I gave that a try, and it would also work. (commit is not pushed)

But with this approach, I must add this code at two different places. And every plugin would be required to maintain its aliases in core cnquery. Is this something we would fix in a later iteration of v9?

imilchev (approved):
LGTM now! :shipit:

PR #1697, opened 2023-09-12
🐛 fix GKE cluster resource

imilchev:
Fixes the following error for GKE clusters:
```
  masterAuthorizedNetworksConfig: failed to convert dict to primitive, unsupported child type: []map[string]interface {}
```

### DIFF
--- a/providers/gcp/resources/gke.go
+++ b/providers/gcp/resources/gke.go
@@ -6,12 +6,13 @@ package resources
 import (
 	"context"
 	"fmt"
+	"strings"
+
 	"go.mondoo.com/cnquery/llx"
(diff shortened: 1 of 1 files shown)


czunker:
Looking good:
```
cnquery run gcp project mondoo-edge -c 'gcp.project.gke.clusters{masterAuthorizedNetworksConfig}'                                                                                                                                      
gcp.project.gke.clusters: [
  0: {
    masterAuthorizedNetworksConfig: {
      cidrBlocks: []
      enabled: false
    }
  }
]
```

czunker (approved):
Thanks @imilchev

PR #1698, opened 2023-09-12
🧹 Remove deprecated ioutil

czunker:
These were all deprecated since Go 1.16: https://pkg.go.dev/io/ioutil


chris-rock (approved):
Thank you @czunker

PR #1701, opened 2023-09-12
✨ Add container discovery to v9 os provider

czunker:


providers/os/config/config.go
```diff
@@ -34,7 +34,7 @@ var Config = plugin.Provider{
 			MinArgs: 0,
 			MaxArgs: 0,
 			Discovery: []string{
-				"containers",
+				"container",
```
arlimus, line 37: nice catch, this would have been a breaking change!

arlimus (approved):
works as advertised, thank you @czunker !!

PR #1714, opened 2023-09-13
✨ implement related assets for k8s provider

imilchev:
- Added platform ownership for k8s
- Set platform runtime correctly
- Make sure we don't execute discovery multiple times


preslavgerchev (approved):

arlimus (approved):
Thank you for porting this!

PR #1716, opened 2023-09-13
Skip publish on alpha/beta tags

benr:
To help with testing of v9 we want to create releases for test but not distribute them via the normal channels, they will only be available on the Github release page.


vjeffrey (approved):

vjeffrey:
thank you!

PR #1731, opened 2023-09-14
🐛 store+load recordings with errors

arlimus:
Up until this point, errors were marshalled into `error:{}` into JSON files. This caused problems when trying to load the structure (because the error type is not familiar with empty object).

This fixes the issue in the RawData to JSON marshal/unmarshal calls and additionally handles type problems for int (which is by default read as a float) and time (which is read as a string). It also fixes a problem with reconnecting resources when the field in question is an error.


chris-rock (approved):
Thank you @arlimus

PR #1738, opened 2023-09-14
🔧 add configuration for providers in dev env

arlimus:
This has now happened a few too many times: You get a PR with a change to a provider, but forgot to update your builtin.go file. After 30min you realize you have been testing the providers on disk, instead of the ones in the repo. Also configuring providers is a bit annoying (with all the things you need to change in `builtin.go`).

This is now streamlined in `make providers/config`. You can configure it in `providers.yaml` in the root folder.


vjeffrey:
yay! 🎉  could you also update https://github.com/mondoohq/cnquery/blob/main/docs/development.md#debug-providers ?

arlimus:
> yay! 🎉 could you also update https://github.com/mondoohq/cnquery/blob/main/docs/development.md#debug-providers ?

Great catch!! I ended up seeing more problems with the modular providers (eg `aws` which has a `go.mod`) and extending the automation to configure everything in the dev environment automatically. 

The guide has also been updated.

As a follow-up, it would be a good idea to run a step that checks that no providers are configured when the code is pushed and re-running `make providers/config` to ensure we don't accidentally pull dependencies (which we can now verify).

docs/development.md
```diff
@@ -60,34 +60,16 @@ In v9 we introduced providers, which split up the providers into individual go m
 
 To debug a provider locally with cnquery:
 
-1. Copy the resources json file `providers/TYPE/resources/TYPE.resources.json` to the `providers` top-level dir
-2. Add the provider to `providers/builtin.go`
-
+1. Modify the `providers.yaml` in the root folder and add providers you want to test to the `builtin` field. Example:
```
vjeffrey, line 63: nice! this is so much easier! i'll give it a test locally in a bit, thank you!

chris-rock (approved):
Nice 👍

PR #1739, opened 2023-09-14
🐛 fix missing slack team name

vjeffrey:
local and ssh are also missing asset name --- i think it's because we didn't port over all the id detection stuff we had in v8...looking into it


czunker (approved):
Thanks @vjeffrey

PR #1740, opened 2023-09-15
🐛 fix asset name for local, ssh conn

vjeffrey:
this fixes the asset name detection for local and ssh - it's a portover from v8.

~this still needs some cleanup, ill get back to it this weekend. ran outta steam.~

### DIFF
--- a/providers/os/id/ids/ids.go
+++ b/providers/os/id/ids/ids.go
@@ -10,10 +10,5 @@ const (
 	IdDetector_SshHostkey  = "ssh-host-key"
 	IdDetector_AwsEcs      = "aws-ecs"
 
-	// FIXME: DEPRECATED, remove in v9.0 vv
-	// this is now cloud-detect
-	IdDetector_AwsEc2 = "aws-ec2"
-	// ^^
-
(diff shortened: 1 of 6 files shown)


providers/os/id/ids/ids.go
```diff
@@ -0,0 +1,19 @@
...
+const (
+	IdDetector_Hostname    = "hostname"
+	IdDetector_MachineID   = "machine-id"
+	IdDetector_CloudDetect = "cloud-detect"
+	IdDetector_SshHostkey  = "ssh-host-key"
+	IdDetector_AwsEcs      = "aws-ecs"
+
+	// FIXME: DEPRECATED, remove in v9.0 vv
+	// this is now cloud-detect
+	IdDetector_AwsEc2 = "aws-ec2"
```
czunker: As this PR is for v9 and detection, perhaps we can directly remove it here.

czunker:
Gave it a try with local and ssh to a vagrant box and it works. :+1: 


Directly using the vagrant provider does not work. I'll create an issue for that.

czunker (approved):

czunker:
Thanks @vjeffrey

PR #1741, opened 2023-09-15
🐛 fix for terraform <cannot get hcl block> err

vjeffrey:
fixes https://github.com/mondoohq/cnquery/issues/1737

### DIFF
--- a/providers/terraform/resources/hcl.go
+++ b/providers/terraform/resources/hcl.go
@@ -172,25 +172,19 @@ func newMqlHclBlock(runtime *plugin.Runtime, block *hcl.Block, file *hcl.File) (
 
 	snippet := extractHclCodeSnippet(file, block.TypeRange)
 
-	r, err := CreateResource(runtime, "terraform.block", map[string]*llx.RawData{
+	res, err := CreateResource(runtime, "terraform.block", map[string]*llx.RawData{
 		"type":    llx.StringData(block.Type),
 		"labels":  llx.ArrayData(llx.TArr2Raw(block.Labels), types.String),
 		"start":   llx.ResourceData(start, "terraform.fileposition"),
 		"end":     llx.ResourceData(end, "terraform.fileposition"),
 		"snippet": llx.StringData(snippet),
 	})
-
-	/*
-		if err == nil {
-			r.MqlResource().Cache.Store("_hclblock", &resources.CacheEntry{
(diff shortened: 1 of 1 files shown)


providers/terraform/resources/hcl.go
```diff
@@ -172,24 +172,16 @@ func newMqlHclBlock(runtime *plugin.Runtime, block *hcl.Block, file *hcl.File) (
...
-			r.MqlResource().Cache.Store("_hclblock", &resources.CacheEntry{
-				Data: block,
-			})
-			r.MqlResource().Cache.Store("_hclfile", &resources.CacheEntry{
-				Data: file,
-			})
-		}
-	*/
+	r := res.(*mqlTerraformBlock)
+	r.block = plugin.TValue[*hcl.Block]{State: plugin.StateIsSet, Data: block}
```
preslavgerchev, line 186: can't we pass this in the args as an mql resource, i.e. `llx.ResourceData`?

providers/terraform/resources/hcl.go
```diff
@@ -172,24 +172,16 @@ func newMqlHclBlock(runtime *plugin.Runtime, block *hcl.Block, file *hcl.File) (
...
 
-	r, err := CreateResource(runtime, "terraform.block", map[string]*llx.RawData{
+	res, err := CreateResource(runtime, "terraform.block", map[string]*llx.RawData{
 		"type":    llx.StringData(block.Type),
 		"labels":  llx.ArrayData(llx.TArr2Raw(block.Labels), types.String),
 		"start":   llx.ResourceData(start, "terraform.fileposition"),
 		"end":     llx.ResourceData(end, "terraform.fileposition"),
 		"snippet": llx.StringData(snippet),
 	})
-
```
chris-rock, line 182: we need to err out when err != nil, otherwise we get a panic

chris-rock (changes requested):
We need to add error handling, then LGTM

providers/terraform/resources/hcl.go
vjeffrey: this is for the internal implementation, 

```

type mqlTerraformBlockInternal struct {
	block      plugin.TValue[*hcl.Block]
	cachedFile plugin.TValue[*hcl.File]
}
```

jaym (approved):

PR #1744, opened 2023-09-15
🐛 fix scan with no upstream config err

vjeffrey:


apps/provider-scaffold/template/provider/provider.go.template
```diff
@@ -97,7 +97,7 @@ func (s *Service) connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 	}
 
 	var upstream *upstream.UpstreamClient
-	if req.Upstream != nil {
+	if req.Upstream != nil && req.Upstream.ApiEndpoint != "" {
```
preslavgerchev: I think we should look into how are we getting an upstream config without an API endpoint, thats weird

vjeffrey: that's not really the issue -

this logic assumes that req.Upstream is nil if you're running in incognito. right now, upstream is _not nil_ when in incognito. passing in nil for upstream resulted in a different error, so i went with this fix when i was looking at things. we should prob just make it so that passing in nil for upstream doesn't err out

vjeffrey: actually, i just realized we have a field for this case on the upstream config. ill switch things up.

arlimus (approved):
Sweet, thank you @vjeffrey !

PR #1752, opened 2023-09-17
⭐ add initial terraform git repo support

arlimus:
This is one final addition to v8, before we migrate it all to v9. It adds the ability to use terraform with git repositories like so:

```
cnquery shell terraform git+https://gitlab.com/mondoolabs/example-gitlab
```

Works with GitHub and GitLab, as well as any git.


chris-rock (approved):
Thank you @arlimus

PR #1756, opened 2023-09-17
🧹 let `providers/config` use `providers/builtin_dev.go`

arlimus:
After https://github.com/mondoohq/cnquery/pull/1753

This way we avoid generating the same core+mock code all the time and are more resiliant to changes of these static builtins. At the same time it creates one clean file that is always auto-generated.


chris-rock (approved):
Yeah!

PR #1757, opened 2023-09-17
🐛 support Never in json read/write

arlimus:
Serializing to string tries to follow timestamp norms. However, the two infinity times are set as beyond regular date ranges and cause it to crash.

We decided to use integers to represent the unix time (which is intuitive from reading it) and is much faster to read/write.

However, due to limitations in JSON precision for integers (RFC 7159) we had to limit the range and eagerly interpret things to be either Never-times (future and past).

This needs to be merged before the v9 release to avoid breaking changes in recordings.


chris-rock (approved):
Thank you @arlimus

PR #1762, opened 2023-09-18
🐛 revert error on non-existent package, service, and kernel.module

arlimus:
This is in line with v8 and would otherwise break a number of existing queries. It is inconsistent with a few other resources (like `user`) and needs further alignment (as descriped in https://github.com/mondoohq/cnquery/issues/1622).

That said, some fields are now null instead of empty strings:

```coffee
cnquery> package("blah"){*}
package: {
  name: "blah"
  version: null
  epoch: null
  available: null
  description: null
  status: null
  outdated: false
  arch: null
  origin: null
  installed: false
  format: null
}

cnquery> service("blah"){*}
service: {
  name: "blah"
  installed: false
  description: null
  type: null
  enabled: false
  running: false
  masked: false
}

cnquery> kernel.module("blah"){*}
kernel.module: {
  size: null
  name: "blah"
  loaded: false
}
```

### DIFF
--- a/providers/os/resources/kernel.go
+++ b/providers/os/resources/kernel.go
@@ -322,7 +322,12 @@ func initKernelModule(runtime *plugin.Runtime, args map[string]*llx.RawData) (ma
 	if res, ok := kernel.moduleByName[name]; ok {
 		return nil, res, nil
 	}
-	return args, nil, nil
+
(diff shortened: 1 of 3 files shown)


chris-rock (approved):
Thank you @arlimus

PR #1774, opened 2023-09-18
🐛 fix -h/--help for all attached subcommands

arlimus:
This fixes a regression in v9

Before:

image

After:

image

Works for `run`, `shell`, and `scan` now

Here is the same thing when the provider is not yet installed:

image


vjeffrey (approved):
nice! works great!

PR #1776, opened 2023-09-18
Fix missing space in invalid path error

tas50:
Add the missing space to errors like this:

```
x could not connect to asset error="path 'https://gitlab.com/mondoolabs/example-gitlab.git'is not a valid file or directory" asset=
```


vjeffrey (approved):

tas50:
BTW: This is not present in V9 so no need to fix it there.

PR #1779, opened 2023-09-19
Improvements to provider building

benr:
This change fixes several oddities in the provider build action, namely using the version variable for the providers list.  It also removes manual invocation inputs because it currently builds the latest.json file based on what its built in dist/


vjeffrey (approved):
thank you!

PR #1780, opened 2023-09-19
🐛 add builtin flags for cli pre-processing

arlimus:
This avoids an error-message which makes users think that recordings were not loaded.

Before 

```bash
..go/src/go.mondoo.com/cnquery ±> go run apps/cnquery/cnquery.go run -c sshd.config.params --use-recording a.json
! pre-processing of cli had an error error="unknown flag: --use-recording"
```

After:

```bash
..go/src/go.mondoo.com/cnquery ±> go run apps/cnquery/cnquery.go run -c sshd.config.params --use-recording a.json
```

### DIFF
--- a/providers-sdk/v1/plugin/flag.go
+++ b/providers-sdk/v1/plugin/flag.go
@@ -1,0 +1,38 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package plugin
+
+type FlagType byte
+
+const (
+	FlagType_Bool FlagType = 1 + iota
+	FlagType_Int
+	FlagType_String
+	FlagType_List
+	FlagType_KeyValue
+)
+
+type FlagOption byte
+
+const (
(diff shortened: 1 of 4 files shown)


providers-sdk/v1/plugin/flag.go
arlimus, line 1: Extracted flags to a separate file for easier discovery. However: We don't want to have any cobra/pflags dependencies in this part of the plugin sdk, so it is restricted to just the flags

providers/mock.go
```diff
@@ -11,8 +11,9 @@ import (
 
 var mockProvider = Provider{
 	Provider: &plugin.Provider{
-		Name: "mock",
-		ID:   "go.mondoo.com/cnquery/providers/mock",
+		Name:    "mock",
+		ID:      "go.mondoo.com/cnquery/providers/mock",
+		Version: "9.0.0",
```
arlimus, line 16: without this it shows without version in `cnquery providers`

chris-rock (approved):
Thank you @arlimus

PR #1786, opened 2023-09-19
🐛 Fix networkAcl portRange and egress

arlimus:
Ported https://github.com/mondoohq/cnquery/pull/1508

Fixes https://github.com/mondoohq/cnquery/issues/1510


chris-rock (approved):
Thank you @arlimus

PR #1790, opened 2023-09-19
🐛 Fix tree nodes ordering, makes query checksums deterministic.

preslavgerchev:


arlimus (approved):
thanks for fixing the flip flop issue!!

PR #1793, opened 2023-09-19
🐛 ensure we use the credential token for gitlab scan

vjeffrey:
the error i encountered is reproducible by running `unset SSH_AUTH_SOCK` and then running `DEBUG=1 cnquery shell gitlab --group mondoolabs --token TOKEN`

you should see it fail out when cloning the repos 

(with this change it should just work. in my test i have one repo i can't clone due to perms, so it get 36 assets discovered)

------
i'm a little uncertain about this change, i would have expected some process before resolve, like the enrich assets one,  to take care of doing this

but i did test this on two diff machines and got the same result on both


arlimus (approved):
Thank you @vjeffrey !

PR #1795, opened 2023-09-20
Statically build providers

benr:


benr:
I need to update the Makefile as well.

arlimus:
I ran into a few problems with these build flags in local testing with an updated makefile: https://github.com/mondoohq/cnquery/pull/1796

PR #1809, opened 2023-09-20
🐛 use system path on default when root user is used

chris-rock:
This change handle the special case where we run cnquery inside of a container. When root (uid-0) is used we want to use the system path as default. While we handled the case for home path detection we have not incorporated that when we install a plugin. Now we determine the default path, this allows us to run well in container land:

```
docker run -it cnspec-alpine /bin/sh
/ # cnquery shell local
! can't find any paths for providers, none are configured system-path=/opt/mondoo/providers
→ installing provider 'os' version=9.0.0
→ successfully installed os provider path=/opt/mondoo/providers/os version=9.0.0
→ connected to Alpine Linux v3.15

cnquery> 
```

In addition, I added code that DEBUG=1 is also respected for global logging before the commands are injected

### DIFF
--- a/apps/cnquery/cmd/root.go
+++ b/apps/cnquery/cmd/root.go
@@ -81,6 +81,13 @@ func init() {
 	logger.CliCompactLogger(logger.LogOutputWriter)
 	zerolog.SetGlobalLevel(zerolog.InfoLevel)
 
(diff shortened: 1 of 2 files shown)


arlimus (approved):
Awesome, so this fixes that home is detected as `""`.

PR #1818, opened 2023-09-21
🐛 detect asset kind

arlimus:
Fixes https://github.com/mondoohq/cnquery/issues/1691


tas50:
Is it possible to add a test for this fallback?

chris-rock (approved):
Thank you @arlimus

chris-rock:
> Is it possible to add a test for this fallback?

We are going to add more integration test to cover this case in the future.

PR #1819, opened 2023-09-21
🐛 fix for users { * }: check for nil file data before returning path

vjeffrey:
fixes https://github.com/mondoohq/cnquery/issues/1724

### DIFF
--- a/providers/os/resources/privatekey.go
+++ b/providers/os/resources/privatekey.go
@@ -16,7 +16,6 @@ func initPrivatekey(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[
 		if !ok {
 			return nil, nil, errors.New("wrong type for 'path' in privatekey initialization, it must be a string")
 		}
-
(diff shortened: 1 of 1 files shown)


providers/os/resources/privatekey.go
```diff
@@ -33,7 +32,7 @@ func (r *mqlPrivatekey) id() (string, error) {
 	// TODO: use path or hash depending on initialization
 
 	file := r.GetFile()
-	if file.Error != nil {
+	if file.Error != nil || file.Data == nil {
```
preslavgerchev: I am a bit confused by this, it implies that the Error and Data might be nil at the same time

if that is the case, we should probably split those two checks, else `return "", file.Error` might evaluate to `"", nil`

arlimus: +1, we can return a custom error if `file.Data == nil` and just let users know that no file is provided

preslavgerchev (approved):

Issue #1794, opened 2023-09-19
Add AWS VPC subnet resource


scottford-io:
**Is your feature request related to a problem? Please describe.**
MQL does not contain a resource for assessing AWS VPC subnets. This is a requirement for AWS security recommendations such as:

https://docs.aws.amazon.com/securityhub/latest/userguide/ec2-controls.html#ec2-15

scottford-io:
See https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-subnets.html

PR #1827, opened 2023-09-21
✨ Add AWS VPC subnet resource

mariuskimmina:
#1794

first time making changes to providers, so please review carefully. 
Example usage

image

I wasn't sure what kind of fields to add here. Just went with `id` and `cidrs` for now - there's probably more that could make sense.

### DIFF
--- a/providers/aws/resources/aws.lr.manifest.yaml
+++ b/providers/aws/resources/aws.lr.manifest.yaml
@@ -1423,6 +1423,8 @@ resources:
       region: {}
       routeTables: {}
       state: {}
+      subnets:
+        min_mondoo_version: 9.0.0
       tags: {}
     is_private: true
     min_mondoo_version: latest
@@ -1447,6 +1449,17 @@ resources:
       routes: {}
     is_private: true
     min_mondoo_version: latest
+    platform:
+      name:
+      - aws
+  aws.vpc.subnet:
+    fields:
+      arn: {}
+      cidrs: {}
+      id: {}
+      mapPublicIpOnLaunch: {}
+    is_private: true
+    min_mondoo_version: 9.0.0
     platform:
       name:
       - aws
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -50,6 +50,8 @@ private aws.vpc @defaults("arn isDefault") {
   flowLogs() []aws.vpc.flowlog
   // List of route tables for the VPC
   routeTables() []aws.vpc.routetable
+  // List of subnets for the VPC
+  subnets() []aws.vpc.subnet
   // Tags on the VPC
   tags map[string]string
 }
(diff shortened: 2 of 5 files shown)


mariuskimmina:
Added the filed `mapPublicIpOnLaunch` which is the actual requirement for AWS security recommendations discussed in #1794

```
cnquery> aws.vpcs.where( arn == "arn:aws:vpc:eu-central-1:455954587013:id/vpc-091edd43468ca60e5" )[0].subnets[0].mapPublicIpOnLaunch
[ok] value: true
```

providers/aws/resources/aws.lr.manifest.yaml
```diff
@@ -1423,6 +1423,7 @@ resources:
       region: {}
       routeTables: {}
       state: {}
+      subnets: {}
```
```suggestion
      subnets:
        min_mondoo_version: 9.0.0
```

providers/aws/resources/aws.lr.manifest.yaml
```diff
@@ -1450,3 +1451,13 @@ resources:
     platform:
       name:
       - aws
+  aws.vpc.subnet:
+    fields:
+      cidrs: {}
+      id: {}
+      mapPublicIpOnLaunch: {}
+    is_private: true
+    min_mondoo_version: latest
```
```suggestion
    min_mondoo_version: 9.0.0
```

vjeffrey (changes requested):
can you please add in the subnet arn, and add an id function for the subnet `func (s *mqlAwsVpcSubnet) id() (string, error) {` that returns the arn?

this will help ensure the objects are uniquely identifiable

mariuskimmina:
I have added the arn filed, I noticed a bug tho.
Currently I have 3 subnets in AWS, the output in cnquery looks like this

aws.vpcs.where[0].subnets: [
  0: aws.vpc.subnet id="subnet-021eb62bc781a5ed5"
  1: aws.vpc.subnet id="subnet-021eb62bc781a5ed5"
  2: aws.vpc.subnet id="subnet-021eb62bc781a5ed5"
]

So the number of subnets is correct but they all have the same id.
I copied this part from the exisiting routeTables function and found that the same bug exists there as well

I have 2 route tables in my aws account

cnquery> aws.vpcs.where( arn == "arn:aws:vpc:eu-central-1:455954587013:id/vpc-091edd43468ca60e5" )[0].routeTables
aws.vpcs.where[0].routeTables: [
  0: aws.vpc.routetable id="rtb-0d28b6f55dfe767c2"
  1: aws.vpc.routetable id="rtb-0d28b6f55dfe767c2"
]

I see two route tables here but they both have the same id

vjeffrey:
did you add in the id function?
```
func (s *mqlAwsVpcSubnet) id() (string, error) {
  return s.Arn.Data, nil
}
```

route table might also need the id fix

mariuskimmina:
> did you add in the id function?
> 
> ```
> func (s *mqlAwsVpcSubnet) id() (string, error) {
>   return s.Arn.Data, nil
> }
> ```
> 
> route table might also need the id fix

Ah, adding the `id` function does seem to fix it. 

I also saw now this comment
```
to override the _id implement id()
```

But, doesn't that mean that there is probably a bug in the default _id implementation because it should have been able to use the correct `id` field for each subnet and not the same one three times?

vjeffrey:
using the arn as the id guarantees we get the unique one since it includes the region and account id - whenever dealing with aws we always want to use the arn where possible because of that.

good question about the _id thing tho, i haven't looked at it much yet in our new v9 world...i'll dig around for an answer for u on that

mariuskimmina:
Anything still needs to be done here from my side? @vjeffrey

vjeffrey:
sorry for the delay! this is great, thank you!

vjeffrey (approved):

PR #1834, opened 2023-09-21
WIP 🐛 provider install: create system path if home path does not exist

vjeffrey:
if im reading the code right, this should fix https://github.com/mondoohq/cnquery/issues/1833


arlimus:
I'm holding off on it for a sec to figure out why the path is not getting set in the first place.

vjeffrey:
i think we're good to close this 👍

PR #1839, opened 2023-09-21
update GitHub from v44 -> v55

tas50:
- 18 months of updates
- Better handling of secondary API limits


arlimus (approved):
kudos for the update!

PR #1841, opened 2023-09-22
🐛 move default provider path on macos to /Library/Mondoo

arlimus:
Add special handling for MacOS / Darwin and make sure all configs + providers are in the same shared directory `/Library/Mondoo`

Fixes https://github.com/mondoohq/cnquery/issues/1832


atomic111:
@arlimus thank you for the fix

atomic111 (approved):

PR #1842, opened 2023-09-22
🐛 avoid printing home-path for root

arlimus:
Root just uses the system path, so trying to print a home path that is an empty string is confusing and wrong.

Before:

image

After:

image


atomic111 (approved):
@arlimus thank you for the improvement

Issue #1775, opened 2023-09-18
Add AWS VPC endpoint service resource


scottford-io:
**Is your feature request related to a problem? Please describe.**
Amazon Security Hub suggests that VPC endpoint services are created for various services like EC2. cnquery should provide a resource to test that VPC endpoints are configured.

https://docs.aws.amazon.com/securityhub/latest/userguide/ec2-controls.html#ec2-10

https://docs.aws.amazon.com/whitepapers/latest/aws-privatelink/what-are-vpc-endpoints.html

**Describe the solution you'd like**
```
aws.vpc:                            Amazon Virtual Private Cloud (VPC)
  arn string:                       ARN of the VPC
  endpoints                        A list of VPC endpoints configured
  flowLogs []aws.vpc.flowlog:       A list of flowlogs for the VPC
  id string:                        ID of the VPC
  isDefault bool:                   Whether the VPC is the default one
  region string:                    Region the VPC exists in
  routeTables []aws.vpc.routetable: List of route tables for the VPC
  state string:                     State of the VPC (pending or available)
  tags map[string]string:           Tags on the VPC
```

**Describe alternatives you've considered**

**Additional context**

PR #1845, opened 2023-09-22
✨ Add AWS VPC Endpoints

mariuskimmina:
#1775

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -50,6 +50,8 @@ private aws.vpc @defaults("arn isDefault cidrBlock") {
   instanceTenancy string
   // Region the VPC exists in
   region string
+  // A list of endpoints for the VPC
+  endpoints() []aws.vpc.endpoint
   // A list of flowlogs for the VPC
   flowLogs() []aws.vpc.flowlog
   // List of route tables for the VPC
@@ -78,6 +80,24 @@ private aws.vpc.subnet @defaults("arn") {
   cidrs string
   // Indicates whether instances launched in this subnet receive a public IPv4 address
   mapPublicIpOnLaunch bool
+}
+
+// Amazon Virtual Private Cloud (VPC) Endpoint
+private aws.vpc.endpoint @defaults("id") {
+  // Unique ID of the endpoint
+  id string
+  // Type of the endpoint
+  type string
+  // VPC the endpoint exists in
+  vpc string
+  // Region the VPC exists in
+  region string
+  // The name of the endpoint service
+  serviceName string
+  // The policy document associated with the endpoint, if applicable
+  policyDocument string
+  // The subnets for the (interface) endpoint
+  subnets []string
 }
 
 // Amazon Virtual Private Cloud (VPC) Flow Log
--- a/providers/aws/resources/aws_vpc.go
+++ b/providers/aws/resources/aws_vpc.go
@@ -103,6 +103,55 @@ func (a *mqlAws) getVpcs(conn *connection.AwsConnection) []*jobpool.Job {
 	return tasks
 }
 
+func (a *mqlAwsVpcEndpoint) id() (string, error) {
+	return a.Id.Data, nil
+}
+
+func (a *mqlAwsVpc) endpoints() ([]interface{}, error) {
+	conn := a.MqlRuntime.Connection.(*connection.AwsConnection)
+	vpc := a.Id.Data
+
+	svc := conn.Ec2(a.Region.Data)
+	ctx := context.Background()
+	endpoints := []interface{}{}
+	filterKeyVal := "vpc-id"
+	nextToken := aws.String("no_token_to_start_with")
+	params := &ec2.DescribeVpcEndpointsInput{Filters: []vpctypes.Filter{{Name: &filterKeyVal, Values: []string{vpc}}}}
+	for nextToken != nil {
+		endpointsRes, err := svc.DescribeVpcEndpoints(ctx, params)
+		if err != nil {
+			return nil, err
+		}
+		nextToken = endpointsRes.NextToken
+		if endpointsRes.NextToken != nil {
+			params.NextToken = nextToken
+		}
+
+		for _, endpoint := range endpointsRes.VpcEndpoints {
+			var subnetIds []interface{}
+			for _, subnet := range endpoint.SubnetIds {
+				subnetIds = append(subnetIds, subnet)
+			}
+			mqlEndpoint, err := CreateResource(a.MqlRuntime, "aws.vpc.endpoint",
+				map[string]*llx.RawData{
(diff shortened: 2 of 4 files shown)


providers/aws/resources/aws_vpc.go
```diff
@@ -101,6 +101,50 @@ func (a *mqlAws) getVpcs(conn *connection.AwsConnection) []*jobpool.Job {
...
+		}
+
+		for _, endpoint := range endpointsRes.VpcEndpoints {
+			var subnetIds []interface{}
+			for _, subnet := range endpoint.SubnetIds {
+				subnetIds = append(subnetIds, subnet)
+			}
+			mqlEndpoint, err := CreateResource(a.MqlRuntime, "aws.vpc.endpoint",
+				map[string]*llx.RawData{
+					"id":             llx.StringData(*endpoint.VpcEndpointId),
```
vjeffrey: i haven't tested this one yet, but same comment as the other pr applies - if you can please add in the arn and have that be returned by the id function. if the arn isn't returned on the response, you can assemble it (there are a bunch of examples of this in the aws_ec2 resources  file

mariuskimmina:
I have added the arn filed, I noticed a bug tho. 
Currently I have 3 subnets in AWS, the output in cnquery looks like this

```
aws.vpcs.where[0].subnets: [
  0: aws.vpc.subnet id="subnet-021eb62bc781a5ed5"
  1: aws.vpc.subnet id="subnet-021eb62bc781a5ed5"
  2: aws.vpc.subnet id="subnet-021eb62bc781a5ed5"
]
```

So the number of subnets is correct but they all have the same id. 
I copied this part from the exisiting routeTables function and found that the same bug exists there as well

I have 2 route tables in my aws account

```
cnquery> aws.vpcs.where( arn == "arn:aws:vpc:eu-central-1:455954587013:id/vpc-091edd43468ca60e5" )[0].routeTables
aws.vpcs.where[0].routeTables: [
  0: aws.vpc.routetable id="rtb-0d28b6f55dfe767c2"
  1: aws.vpc.routetable id="rtb-0d28b6f55dfe767c2"
]
```

I see two route tables here but they both have the same id

mariuskimmina:
Do endpoints have their own arn? I can't find any example of one online.

mariuskimmina:
Added the id function with `Id.Data` instead since there doesn't seem to be an arn. Also I saw `Id.Data` also being used in other places across the code base such as 

```
func (a *mqlAwsS3BucketGrant) id() (string, error) {
	return a.Id.Data, nil
}
```

vjeffrey:
this is working well!
if there's no arn for endpoints, could you prefix it with the region, so we return region/id ? 
the point of the id function here is to give the object a unique identifier. since we query across regions by default, adding region can help us ensure we don't confuse one object for another for multi-region resources

mariuskimmina:
There seem to be a lot of resources that just return `a.Id.Data` and I couldn't find a case where we add the region to those yet. I agree that it would make sense to add the region here and also to all the other cases where we return `Id.Data` as I don't think the uniqueness of ids is guaranteed across regions (at least I couldn't find any such guarantee).

mariuskimmina:
I have added the region. I did not test this part yet as I am now on my work notebook and don't have an AWS account setup. 
To get the region of the endpoint I just passed the region of the VPC through 

```
"region":         llx.StringData(a.Region.Data),
```

I suppose that's fine?

vjeffrey (approved):

providers/aws/resources/aws.lr
```diff
@@ -62,6 +64,24 @@ private aws.vpc.routetable @defaults("id") {
   routes []dict
 }
 
+// Amazon Virtual Private Cloud (VPC) Endpoint
+private aws.vpc.endpoint @defaults("id") {
+  // Unique ID of the endpoint
+  id string
+  // type of the endpoint
```
```suggestion
  // Type of the endpoint
```

providers/aws/resources/aws.lr
```diff
@@ -62,6 +64,24 @@ private aws.vpc.routetable @defaults("id") {
...
+  type string
+  // VPC the endpoint exists in
+  vpc string
+  // Region the VPC exists in
+  region string
+  // The name of the endpoint service
+  serviceName string
+  // The policy document associated with the endpoint, if applicable.
```
```suggestion
  // The policy document associated with the endpoint, if applicable
```

providers/aws/resources/aws.lr
```diff
@@ -62,6 +64,24 @@ private aws.vpc.routetable @defaults("id") {
...
+  policyDocument string
+  // the subnets for the (interface) endpoint
```
```suggestion
  // The subnets for the (interface) endpoint
```

tas50:
@mariuskimmina can you attach the output this resource provides so I can use it in the release notes?

mariuskimmina:
@tas50 

```
cnquery> aws.vpcs.where( arn == "arn:aws:vpc:eu-central-1:455954587013:id/vpc-091edd43468ca60e5" )[0].endpoints.first.policyDocument
aws.vpcs.where[0].endpoints.first.policyDocument: "{
  \"Statement\": [
    {
      \"Action\": \"*\",
      \"Effect\": \"Allow\",
      \"Principal\": \"*\",
      \"Resource\": \"*\"
    }
  ]
}"
```


```
cnquery> aws.vpcs.where( arn == "arn:aws:vpc:eu-central-1:455954587013:id/vpc-091edd43468ca60e5" )[0].endpoints.first.type
aws.vpcs.where[0].endpoints.first.type: "Interface"
```


```
cnquery> aws.vpcs.where( arn == "arn:aws:vpc:eu-central-1:455954587013:id/vpc-091edd43468ca60e5" )[0].endpoints.first.serviceName
aws.vpcs.where[0].endpoints.first.serviceName: "com.amazonaws.eu-central-1.ec2"
```

Hmm, adding the region to the id didn't seem to have worked tho :thinking: 
```
aws.vpcs.where( arn == "arn:aws:vpc:eu-central-1:455954587013:id/vpc-091edd43468ca60e5" )[0].endpoints.first.id
aws.vpcs.where[0].endpoints.first.id: "vpce-0211b655551cf8b01"
```

mariuskimmina:
Now I got the `region/id` thing to work 

```
cnquery> aws.vpcs[12].endpoints.first
! using builtin provider for aws
aws.vpcs[12].endpoints.first: aws.vpc.endpoint id="eu-central-1/vpce-05052690cb8ca48ac"
cnquery> aws.vpcs[12].endpoints.first.id
aws.vpcs[12].endpoints.first.id: "eu-central-1/vpce-05052690cb8ca48ac"
```

mariuskimmina:
I believe this one should be ready now @vjeffrey

vjeffrey:
👍  this is looking good, thank you! can you rebase to resolve the conflicts? after that we can merge

mariuskimmina:
> 👍 this is looking good, thank you! can you rebase to resolve the conflicts? after that we can merge

Done

PR #1855, opened 2023-09-22
🐛 aws resource bugfixes

vjeffrey:
this gets matching reports for aws-foundations, aws-operational-bp, and mondoo-aws-sec


arlimus (approved):
thank you for catching all these!

PR #1860, opened 2023-09-23
🔧 add utility to help version providers

arlimus:
The last few beta releases have shown how hard it is to keep all provider versions updated. To ease the pain, we are starting a small utility.

It is still in its infancy, but it is already really helpful:

```bash
alias version="go run providers-sdk/v1/util/version/version.go"
```

Check what needs updating (fast)

```bash
version check providers/*/ --fast
```

image

Update one provider, automatically bump the patch version when applicable, and add a commit:

```bash
version update providers/o* --fast --increment=patch --commit
```

image

Note 1: We use Golang's `go-git` for git processing, which is cleaner than using git in go, but also a lot slower, especially with larger files (beware: it sometimes may look like it's blocking on things, but it just takes long to process files).

Note 2: We are not looking into commits right now to find the last version bump, but that will be added next.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -102,12 +102,25 @@ require (
 )
 
 require (
+	dario.cat/mergo v1.0.0 // indirect
 	github.com/GaijinEntertainment/go-exhaustruct/v3 v3.1.0 // indirect
+	github.com/acomagu/bufpipe v1.0.4 // indirect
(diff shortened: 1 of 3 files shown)


chris-rock (approved):
So cool

PR #1863, opened 2023-09-23
🐛 fix time parsing in github

arlimus:
The child object might be non-pointers, but their parents are all pointers now. Correctly handle nil-values.


tas50:
Works great for me locally. Thanks @arlimus

tas50 (approved):

PR #1864, opened 2023-09-23
Add missing versions in the GH LR file

tas50:
Slap a few more release versions in here


arlimus (approved):
appreciate you caught these!

PR #1868, opened 2023-09-23
🐛 pre-process known provider flags

arlimus:
We were getting errors about pre-processing failing when used in conjunction with full-fledged CLI commands.

Example: If you didn't have `cnquery` installed and copy-pasted:

```bash
cnquery run github org mondoohq --token github_.. -c ".."
```

... you'd get an error about not being able to parse `--token`.

This is a bit inefficient, but it at least results in a clean run for now.


chris-rock (approved):
Thank you, this makes it more clean and removes the warning for the user.

PR #1869, opened 2023-09-24
🐛 support sudo over ssh on alpine

arlimus:
stat does not support all the flags on Alpine, particularly printf. We can skip newline escapes, since we only grab simple fields. The only limitation here is the SElinux context, which we do not return yet (so we can revisit that then). Even so, it is the last value returned, so we can stick with `-c`


chris-rock (approved):
Thank you @arlimus

PR #1870, opened 2023-09-24
🐛 report on crashed provider + dedupe shutdown + fix multierr nil

arlimus:
1. Crashed providers print their plugin panic and then an error message about being unable to shutdown the provider. This is wrong and misleading.
2. Since we have potentialy multiple runtimes, the shutdown call can come down multiple paths to the same provider, leading to duplication of errors and reporting. Fix it by giving providers one clear shutdown.
3. Multierr did not handle nil errors correctly and would then end up in panic-land. Fixed.

With these updates we get:

image

Future ideas: hide away the panic behind a DEBUG flag

### DIFF
--- a/providers/runtime.go
+++ b/providers/runtime.go
@@ -16,6 +16,7 @@ import (
 	"go.mondoo.com/cnquery/providers-sdk/v1/upstream"
 	"go.mondoo.com/cnquery/types"
 	"go.mondoo.com/cnquery/utils/multierr"
+	"google.golang.org/grpc/status"
 )
 
 const defaultShutdownTimeout = time.Duration(time.Second * 120)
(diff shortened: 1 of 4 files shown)


providers/runtime.go
```diff
@@ -347,6 +357,23 @@ func (r *Runtime) watchAndUpdate(resource string, resourceID string, field strin
 	return raw, nil
 }
 
+func (r *Runtime) handlePluginError(err error, provider *ConnectedProvider) (bool, error) {
+	st, ok := status.FromError(err)
+	if !ok {
+		return false, err
+	}
+
+	switch st.Code() {
```
chris-rock, line 366: nice

chris-rock (approved):
Thank you for the better error handling

PR #1872, opened 2023-09-24
🐛 make filter queries optional

arlimus:
Analogous to https://github.com/mondoohq/cnspec/pull/741

Also adjust naming of the filter function in the executor. This needs more deduplication of course, but a small step ... 🤏


chris-rock (approved):
big plus one

PR #1878, opened 2023-09-24
🧹 remove vmware guest connection

chris-rock:
The VMware guest connection is not performant enough to act as a proper connection. Therefore we remove the support from cnquery v9.


arlimus (approved):
Oh awesome cleanup :grin: :broom:

PR #1879, opened 2023-09-24
🧹 remove old mock discovery

chris-rock:
We have an all-new mock system in cnquery v9. The old implementation is not needed anymore.


arlimus (approved):
:broom: :broom: :broom:

PR #1880, opened 2023-09-24
🧹 new logic to customise inventory config clone

chris-rock:
We use a similar logic in multiple providers to clone a config but remove the discovery options. This is useful since we most likely do not want the discovery options when we clone. This change makes the the `config.Clone()` more smart by providing more options that can be provided. It makes the code of the providers more readable.


arlimus (commented):
this is going to help across discovery

arlimus (approved):

PR #1881, opened 2023-09-24
⭐️ discovery for vsphere

chris-rock:
This change implements discovery and platform detection for discovered assets. It also makes sure that the platform ids are set properly so that calling `esxi.host` on a discovered ESXi host works as expected.


arlimus (approved):
Glad to have the discovery back, thank you

PR #1884, opened 2023-09-25
⭐ introduce empty

arlimus:
Avoids having to double-check values, instead you can simply say:

```coffee
[] == empty
null == empty
'' == empty
{} == empty
```

If merged, this feature will be supported in v9, but remain experimental at first. 

Helps with https://github.com/mondoohq/cnquery/issues/1874

Follow-up: This needs support for list resources, which don't work right now. You have to explicitly request the list:

```
users.list == empty
```

Follow-up: Nicer printing on failure/success


chris-rock (approved):
Nice addition, and makes writing checks much easier.

PR #1912, opened 2023-09-25
🐛 Add incognito flag to run command

### LINKED ISSUE #1889: upstream config not handled correctly
**Describe the bug**
This command does not recognize the upstream config in v9:
```
cnquery run local -c "platform.cves{*}" 
To use this resource, you must authenticate with Mondoo Platform.
To learn how, read:
platform.cves.list: no data available
```

Using the same config in v8 works:
```
cnquery run local -c "platform.cves{*}" 
platform.cves.list: [
  0: {
    mrn: "//vadvisor.api.mondoo.app/cves/CVE-2023-29824"
    worstScore: audit.cvss id = cvss/9.80/vector/9.8/CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H
    summary: "** DISPUTED ** A use-after-free issue was discovered in Py_FindObjects() function in SciPy versions prior to 1.8.0. NOTE: the vendor and discoverer indicate that this is not a security issue."
    state: "PUBLIC"
    published: 2023-07-17 18:51:00 +0000 UTC
    modified: 2023-07-17 18:51:00 +0000 UTC
    unscored: false
    id: "CVE-2023-29824"
  }
....
```

**To Reproduce**

1. Use latest main
2. Run above query with a valid upstream config
3. Note the error

**Expected behavior**
cnquery should use the config and fetch data from upstream.

czunker:
Fixes #1889


preslavgerchev:
Why do we need to report anything for a `run` command, I would imagine that just executes the query and exits?

czunker:
> Why do we need to report anything for a `run` command, I would imagine that just executes the query and exits?

For most queries, that is true. And I do not need upstream to report something but to fetch the data for advisories, CVEs, ... We need to authenticate to the platform to get this information.

preslavgerchev:
> > Why do we need to report anything for a `run` command, I would imagine that just executes the query and exits?
> 
> For most queries, that is true. And I do not need upstream to report something but to fetch the data for advisories, CVEs, ... We need to authenticate to the platform to get this information.

makes sense, i didnt think of that

PR #1918, opened 2023-09-25
✨ Install a specific provider version

preslavgerchev:
Allows users to install providers by just name or ver
```
➜ ~/go/bin/cnquery providers  install okta@9.0.0                             
→ installing provider 'okta' version=9.0.0
→ successfully installed okta provider path=/Users/preslavgerchev/.config/mondoo/providers/okta version=9.0.0
```

```
➜ ~/go/bin/cnquery providers  install okta@9.0.1
→ installing provider 'okta' version=9.0.1
FTL failed to install error="cannot find provider okta-9.0.1 under url https://releases.mondoo.com/providers/okta/9.0.1/okta_9.0.1_darwin_arm64.tar.xz"
```
There's an issue with the CLI pre-processing, I will look into it, seems unrelated to the changes in this PR

### DIFF
--- a/apps/cnquery/cmd/providers.go
+++ b/apps/cnquery/cmd/providers.go
@@ -82,10 +82,11 @@ var installProviderCmd = &cobra.Command{
 }
 
 func installProviderByVersion(name string, version string) {
-	_, err := providers.Install(name, version)
+	installed, err := providers.Install(name, version)
 	if err != nil {
 		log.Fatal().Err(err).Msg("failed to install")
 	}
+	providers.PrintInstallResults([]*providers.Provider{installed})
 }
 
 func installProviderUrl(u string) {
--- a/apps/cnquery/cmd/providers.go
+++ b/apps/cnquery/cmd/providers.go
@@ -52,7 +52,7 @@ var listProvidersCmd = &cobra.Command{
 var installProviderCmd = &cobra.Command{
 	Use:    "install <NAME>",
 	Short:  "Install or update a provider.",
-	Args:   cobra.MinimumNArgs(1),
+	Args:   cobra.ExactArgs(1),
 	Long:   "",
 	PreRun: func(cmd *cobra.Command, args []string) {},
 	Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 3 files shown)


apps/cnquery/cmd/providers.go
```diff
@@ -51,9 +52,17 @@ var listProvidersCmd = &cobra.Command{
 var installProviderCmd = &cobra.Command{
 	Use:    "install <NAME>",
 	Short:  "Install or update a provider.",
+	Args:   cobra.MinimumNArgs(1),
```
chris-rock: that does not work with --file and --url

apps/cnquery/cmd/providers.go
```diff
@@ -51,9 +52,17 @@ var listProvidersCmd = &cobra.Command{
 var installProviderCmd = &cobra.Command{
 	Use:    "install <NAME>",
```
chris-rock: I think we should use the formal [provider-name]@[version-number]. This has the advantage that we do not need to validate if --version and --file or --url are conflicting

apps/cnquery/cmd/providers.go
preslavgerchev: I changed that to `ExactArgs(1)`, so we can only pass in the provider's name, i don't think the --file and --url flags count towards that number, those are flags and not args

apps/cnquery/cmd/providers.go
preslavgerchev: makes sense, i changed the code

apps/cnquery/cmd/providers.go
```diff
@@ -26,6 +26,7 @@ func init() {
 
 	installProviderCmd.Flags().StringP("file", "f", "", "install a provider via a file")
 	installProviderCmd.Flags().String("url", "", "install a provider via URL")
+	installProviderCmd.Flags().String("version", "", "install a specific version of a provider")
```
chris-rock: we do not need this anymore, right?

apps/cnquery/cmd/providers.go
```diff
@@ -51,6 +52,7 @@ var listProvidersCmd = &cobra.Command{
 var installProviderCmd = &cobra.Command{
 	Use:    "install <NAME>",
 	Short:  "Install or update a provider.",
+	Args:   cobra.ExactArgs(1),
```
chris-rock: This will not work when url or file is provided

chris-rock (approved):
Woot, cool stuff.

PR #1922, opened 2023-09-26
🧹 more gitlab group and project fetching cleanup

vjeffrey:
we're still having trouble finding projects in _some_ groups.  

i haven't been able to repro, but this is more cleanup that gets us one less api call


arlimus (approved):
:+1: thank you!

PR #1923, opened 2023-09-26
🐛 fix error when cloning resources

arlimus:
Mandatory arguments were never submitted, because the new provider concept meant that we had to request all the fields individually. We were missing a good pattern to execute this.

Turns out the CreateResourceWithID was only used in one place and its primary purpose was to clone the resource. This method has now been adjusted to better reflect its purpose and its arguments can now do the heavy lifting or getting all the necessary fields for cloning the resource. This method can be further refined to grab mandatory fields automatically.

Before:

image

After:

image

### DIFF
--- a/llx/builtin_resource.go
+++ b/llx/builtin_resource.go
@@ -8,7 +8,6 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/rs/zerolog/log"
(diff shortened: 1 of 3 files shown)


vjeffrey (approved):
🎉

PR #1924, opened 2023-09-26
⭐️ Provider Builder v2

benr:
This change introduces several improvements to building of providers:

- Only providers which are changed are built.  This is determined by examining the version of each provider in the repo and comparing that version number against that of the latest.json on releases.mondoo.com.  If the number is the same the provider is skipped in the build process.  If it is different the provider is build and deployed.
- Providers are built any time a push to main includes a change in the providers/ directory
- The manual invocation of the workflow includes the ability to forcably build and deploy all providers and also the ability to skip publishing
- Providers that are built but not deployed (such as by forcing a skip such as above) will retain the built artifacts with the job for inspection or testing purposes
- Providers are automatically discovered.  A list (env var) is provided to explicitly exclude providers, such as 'core'

### DIFF
--- a/.github/workflows/providers.yaml
+++ b/.github/workflows/providers.yaml
@@ -21,6 +21,7 @@ env:
 
 jobs:
   scoping:
+    name: "Scoping"
     runs-on: self-hosted
     timeout-minutes: 10
     outputs:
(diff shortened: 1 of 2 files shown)


chris-rock (approved):
Nice nice nice

PR #1927, opened 2023-09-26
🧹 fix slack asset name and discovery/req args

vjeffrey:
no args needed, its just `shell/scan slack` with a token, just like v8


chris-rock (approved):
Thank you @vjeffrey

PR #1928, opened 2023-09-26
🐛 Fix handling of socketMap

### LINKED ISSUE #1434: Unify the way `_socketsMap` is populated
**What is not working as you expected it?**
We currently have two different ways, to fill the `_socketsMap`:

https://github.com/mondoohq/cnquery/blob/05e6754a2c260d1c35050f51e38fef1be45a1cc1/_resources/packs/core/processes.go#L214

and

https://github.com/mondoohq/cnquery/blob/05e6754a2c260d1c35050f51e38fef1be45a1cc1/_resources/packs/core/port.go#L285

The implementation in port was introduced with https://github.com/mondoohq/cnquery/pull/531.

The current assumption is, we could use the way it is done in `port.go` also in `processes.go` to only have one way of getting the information.

**Where on the platform does it happen?**
`cnquery`

czunker:
Fixes #1434


czunker:
This PR basically moves code from the port resource to the process resource to have the logic for the socketMap in one place.

czunker:
I tried this with a local, ssh, and running container connection. OS were Ubuntu 22.04 and Alpine 3.something.

czunker:
@arlimus Is this how you thought it should be?

arlimus:
Sorry I'm late to respond @czunker !

I think it's good direction to move this to the process resource, as you are approaching it.

A few things:
1. The port resource has changed and a rebase is necessary. 
2. The trickier part: The way the updated port resource works is it doesn't bug down the engine until the user requests anything related to processes. I.e. it is very fast simple checks that look for exposure and ports, without going into processes. I'd like to keep this around, even if we move resources, to keep things running fast. An alternative approach could be to find a more optimized solution for the the sockets (since we know most native tools do this in a split second, whereas the search takes a lot of effort).

No need to press this for v9, I have updated the tags.

arlimus:
@czunker Given that the primary issue for this PR is closed right now, I'd recommend we get to this after the other priority items are resolved first.

czunker:
This PR now also fixes an issue, where I didn't see data with a debian 11.

arlimus (approved):
Awesome improvements! This should be green once rebased, nice job :metal:

PR #1937, opened 2023-09-26
🐛 Fix k8s.ingress certificates

### LINKED ISSUE #1867: k8s scan compare results
```
FTL failed to resolve policies error="failed to compile bundle: failed to compile //local.cnspec.io/run/local-execution/queries/mondoo-kubernetes-best-practices-ingress-cert-expiration: failed to compile query 'k8s.ingress.tls.all(\n  certificates.all(\n    expiresIn.days > 15\n  )\n)\n': cannot find resource that is called by '>' of type core.certificate\nfailed to compile //local.cnspec.io/run/local-execution/queries/mondoo-kubernetes-best-practices-ingress-cert-expiration: failed to compile query 'k8s.ingress.tls.all(\n  certificates.all(\n    expiresIn.days > 15\n  )\n)\n': cannot find resource that is called by '>' of type core.certificate\n"
exit status 1
```

czunker:
Fixes #1867

### DIFF
--- a/providers/k8s/resources/ingress.go
+++ b/providers/k8s/resources/ingress.go
@@ -20,10 +20,9 @@ import (
 )
 
 type mqlK8sIngressInternal struct {
-	lock   sync.Mutex
-	obj    *networkingv1.Ingress
-	objId  string
-	mqlK8s *mqlK8s
+	lock  sync.Mutex
+	obj   *networkingv1.Ingress
+	objId string
 }
 
 func (k *mqlK8s) ingresses() ([]interface{}, error) {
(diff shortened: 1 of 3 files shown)


czunker:
This works now with Ingress objects which have and don't have TLS certificates configured:
```
cnquery run k8s --discover ingresses -c "k8s.ingress.tls.all( certificates.all( expiresIn.days > 15 ) )"
! using builtin provider for k8s
! using builtin provider for k8s
[ok] value: true
```

providers/k8s/resources/ingress.go
```diff
@@ -20,8 +20,10 @@ import (
 )
 
 type mqlK8sIngressInternal struct {
-	lock sync.Mutex
-	obj  *networkingv1.Ingress
+	lock   sync.Mutex
+	obj    *networkingv1.Ingress
+	objId  string
+	mqlK8s *mqlK8s
```
czunker: I only need this to later access the `GetSecrets` func. Perhaps there is a nicer way of doing this. Ideas are very welcome.

arlimus: You can request resources on the fly, especially useful if they are static like in `k8s` case:

```go
	o, err := CreateResource(runtime, "k8s", map[string]*llx.RawData{})
	if err != nil {
		return nil, err
	}
	k8s := o.(*mqlK8s)
```

czunker: Thanks for this. Works like a charm.

providers/k8s/resources/ingress.go
```diff
@@ -60,16 +56,33 @@ func (k *mqlK8s) ingresses() ([]interface{}, error) {
...
 
+func (k *mqlK8sIngress) tls() ([]interface{}, error) {
+	o, err := CreateResource(k.MqlRuntime, "k8s", map[string]*llx.RawData{})
+	if err != nil {
+		return nil, err
+	}
+	k8s := o.(*mqlK8s)
+
+	ingress := k.obj
+	objId := k.objId
```
arlimus: nit: since these variables are only used once, you could just apply them

arlimus (approved):
Looks great, thank you for the update!

providers/k8s/resources/ingress.go
czunker: Good point. Thanks.

PR #1944, opened 2023-09-27
🐛 avoid crashing terraform on unsupported plan calls

arlimus:
Happens when we call plan resources without operating on a Terraform plan. This needs better handling for unsupported or invalid data. I strongly recommend we instead inform users that there are no plan files via an error in the future. For now, create compatibility with v8 to get it release-ready.

Fixes https://github.com/mondoohq/cnquery/issues/1866


vjeffrey (approved):
working better for sure! ⭐️

i am still seeing other errs
```
cnspec> terraform.settings
terraform.settings: terraform.settings id = terraform.settings
cnspec> terraform.settings { *}
1 error occurred:
	* cannot convert primitive with NO type information
terraform.settings: {
  requiredProviders: cannot convert primitive with NO type information
  block: cannot convert primitive with NO type information
  ```
but we can address those in followup

PR #1946, opened 2023-09-27
🧹 fallback to directly fetching a project if not found

vjeffrey:


resources/packs/gitlab/gitlab.go
```diff
@@ -151,8 +151,23 @@ func (g *mqlGitlabProject) init(args *resources.Args) (*resources.Args, GitlabPr
...
 		}
 	}
+	// try to fetch it directly
+	gt, err := gitlabProvider(g.MotorRuntime.Motor.Provider)
+	if err != nil {
+		return nil, nil, err
+	}
+	prj, err := gt.GetProjectById(projectId)
+	if err != nil {
+		return nil, nil, errors.New("project not found")
```
imilchev: can we wrap the original error too so the root cause is not hidden

imilchev (approved):

PR #1966, opened 2023-09-28
⭐️ full IMDSv2 support for Amazon instances

vjeffrey:
fixes https://github.com/mondoohq/cnquery/issues/1813

This change updates the platform detection to work well with IMDSv2 for both Linux and Windows. In addition it slickly tweaks the default platform ids:

- machine-id is disable by default since it is not reliable
- cloud-detect is enabled as default now


.vscode/launch.json
```diff
@@ -177,6 +177,15 @@
...
         "status",
       ],
+    },
+    {
+      "name": "cnquery-shell",
+      "type": "go",
+      "request": "launch",
+      "program": "${workspaceRoot}/apps/cnquery/cnquery.go",
+      "args": [
+        "shell", "ssh", "administrator@18.215.249.49", "--password", "8!dVWy&WRh)kgaTHB=7A9Chfw9v9I.94",
```
preslavgerchev: we should remove that

vjeffrey (commented):
GitHub won't let me +1 this because i opened it,  but +1 , thank you for the updates! 🎉

.vscode/launch.json
chris-rock: I 100% agree. Machine is destroyed, but anyway

arlimus (approved):
noice!

PR #1967, opened 2023-09-28
🐛 pre-process all CLI flags

arlimus:
We were getting errors for commands that had not been pre-processed, like `login --token`.

This PR changes the CLI pre-processor and makes it traverse the root command to find all possible flags across any command and add them to the pre-processing flags. This avoid the warning/error.

Addresses parts of: https://github.com/mondoohq/cnquery/pull/1965
namely: `! CLI pre-processing encountered an issue error="unknown shorthand flag: 't' in -t"`


chris-rock (approved):
Nice, thank you!

PR #1974, opened 2023-09-28
⭐ gitlab provider with discovery

arlimus:
Add discovery options to GitLab provider.

Example:

```bash
cnquery run gitlab --group mondoolabs --discover projects
```

We also support `--discover groups` if you want to add the affected group to the set as well.


vjeffrey:
this appears to be changing the platform id compared to v8 - when i sent the results into my space in edge i suddenly had 64 projects instead of 32

vjeffrey:
(admittedly, i do prefer the v9 version of the platform id, since its using ids instead of the path)

vjeffrey (approved):
thank you! this is working great, ill pick up the platform id thing as a followup

PR #1975, opened 2023-09-29
✨ discover terraform in gitlab + overhaul discovery

arlimus:
Massive overhaul to the discovery code to make it clearer, reduce duplicate calls to the API, and make it ready for groups discovery.

It also now discovers Terraform assets, the next PR will have the terraform+git implementation to use these.

### DIFF
--- a/providers/gitlab/provider/discovery.go
+++ b/providers/gitlab/provider/discovery.go
@@ -1,0 +1,219 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package provider
+
+import (
+	"strconv"
+	"strings"
+
+	"github.com/rs/zerolog/log"
+	"github.com/xanzy/go-gitlab"
+	"go.mondoo.com/cnquery/providers-sdk/v1/inventory"
+	"go.mondoo.com/cnquery/providers/gitlab/connection"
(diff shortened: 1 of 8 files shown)


providers/gitlab/provider/discovery.go
```diff
@@ -0,0 +1,216 @@
...
+
+	groupAssets, groups, err := s.discoverGroups(root, conn)
+	if err != nil {
+		return nil, err
+	}
+	if slices.Contains(targets, DiscoveryGroup) {
+		assets = append(assets, groupAssets...)
+	}
+
+	projectAssets, projects, err := s.discoverProjects(root, conn, groups)
```
vjeffrey, line 44: is this not guarded by an if because its needed for the terraform?

arlimus: yes, and for any other mechanisms we may support in the future. The new code for both `discoverGroups` and `discoverProjects` behaves in a way where it will end up with a list of groups and projects either way!

This means: 
1. if you specify one group and one project, and set discover to groups,projects => you get 1 group and 1 project
2. if you specify one group and no project, and set discovery to projects,terraform => you get N projects and N terraforms
3. if you specify one group and no projects, and set discovery to ONLY terraform => you get N terraforms (only!)

chris-rock (approved):

PR #1978, opened 2023-09-29
🐛 Do not try and parse github repo if owner is set.

### LINKED ISSUE #1971: Github provider doesn't work with an inventory file
**Describe the bug**
Trying to scan a github asset by using an inventory-file only will fail, the provider expects that there's always an arg passed in her:
https://github.com/mondoohq/cnquery/blob/main/providers/github/provider/provider.go#L82

We should consider that it's possible that the scan is happening via an inventory-file, i.e. `cnquery scan github --inventory-file test.yaml` and not rely on the args being there

preslavgerchev:
Fixes #1971


chris-rock (approved):
Thank you @preslavgerchev

PR #1979, opened 2023-09-29
🧹 Apply root-level asset labels to inventory assets.

preslavgerchev:
I am not sure if this is the place to fix this. Maybe indeed it's to have each provider decide that for themselves (whether they want to apply root-level asset labels to discovered assets)

### DIFF
--- a/explorer/scan/local_scanner.go
+++ b/explorer/scan/local_scanner.go
@@ -207,6 +207,11 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 			log.Error().Err(err).Msg("unable to connect to asset")
 			continue
 		}
+
+		// for all discovered assets, we apply mondoo-specific labels that come from the root asset
+		for _, a := range runtime.Provider.Connection.GetInventory().GetSpec().GetAssets() {
+			a.AddMondooLabels(asset)
+		}
 		processedAssets, err := providers.ProcessAssetCandidates(runtime, runtime.Provider.Connection, upstream, "")
 		if err != nil {
 			return nil, false, err
--- a/providers-sdk/v1/inventory/asset.go
+++ b/providers-sdk/v1/inventory/asset.go
@@ -138,3 +138,15 @@ func (s *AssetCategory) UnmarshalJSON(data []byte) error {
 	}
 	return nil
 }
+
+// Merges the mondoo-specific labels from the provided root into the provided asset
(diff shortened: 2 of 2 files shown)


providers/assets.go
```diff
@@ -19,7 +19,17 @@ func ProcessAssetCandidates(runtime *Runtime, connectRes *pp.ConnectRes, upstrea
 		return []*inventory.Asset{connectRes.Asset}, nil
 	} else {
 		logger.DebugDumpJSON("inventory-resolved", connectRes.Inventory)
-		assetCandidates = connectRes.Inventory.Spec.Assets
+		// We want to apply all root asset labels to the inventory assets. We could also do this
+		// per provider, but we'd have to apply the same logic in all providers.
+		for _, a := range connectRes.Inventory.Spec.Assets {
```
chris-rock: I think we should leave it to the provider to decide that.

providers/azure/provider/provider.go
```diff
@@ -134,6 +134,10 @@ func (s *Service) Connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 		return nil, err
 	}
 
+	for _, a := range inventory.GetSpec().GetAssets() {
```
chris-rock: I am not 100 sure I understand this, wouldn't this be required to be pulled from the resource itself?

preslavgerchev: The use case  I am trying to solve is passing in custom labels on the root that are not necessarily computable on each resource level.

An example is `mondoo.com/integration-mrn` which can only be passed it, it cannot be determined by each resource

vjeffrey: ohh, bc when we connect the labels are added to the req.Asset and then all the discovered ones don't have the label. 

this made me think about annotations, e.g. `cnspec scan local --annotation test=vj` -- looks like we're not passing the annotations through either. ill open an issue

chris-rock: I think there are two valid use cases:
1. Mondoo specific labels that should be propagated. I agree that we should do that outside of the providers itself, but then the copy needs to be restricted to specific labels that start with mondoo.com
2. I think it makes sense to support annotations and they can be copied over since they are user provided. We can do that also outside of the providers.

preslavgerchev: I have changed to only take `mondoo.com/` labels into account and have moved the logic into the scanning flow itself

chris-rock (approved):
Thank you @preslavgerchev

PR #1983, opened 2023-09-29
⭐️ unique platform id for terraform when used with git

chris-rock:
This is a followup to https://github.com/mondoohq/cnquery/pull/1975

```
cnquery shell gitlab --group example --discover terraform 
→ loaded configuration from /Users/chris/.config/mondoo/mondoo.yml using source default

    Available assets                                                               
                                                                                   
  > 1. Terraform Static Analysis example/example-gitlab (terraform-hcl)         
    2. Terraform Static Analysis example/example-playground (terraform-hcl)     
    3. Terraform Static Analysis example/example-code (terraform-hcl) 
    4. Terraform Static Analysis example/example-demo-environment (terraform-hcl)            
```

### DIFF
--- a/providers/terraform/provider/detector.go
+++ b/providers/terraform/provider/detector.go
@@ -6,6 +6,8 @@ package provider
 import (
 	"crypto/sha256"
 	"encoding/hex"
+	"errors"
+	"fmt"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


providers/terraform/provider/detector.go
```diff
@@ -48,26 +50,37 @@ func (s *Service) detect(asset *inventory.Asset, conn *connection.Connection) er
...
+	// we always prefer the git url since it is more reliable
+	url, ok := asset.Connections[0].Options["ssh-url"]
+	if ok {
+		domain, org, repo, err := parseSSHURL(url)
+		if err != nil {
+			return err
+		}
+		platformID := "//platformid.api.mondoo.app/runtime/terraform/domain/" + domain + "/org/" + org + "/repo/" + repo
+		asset.Connections[0].PlatformId = platformID
+		asset.PlatformIds = []string{platformID}
```
vjeffrey, line 62: could we add both platform ids to the platform ids array? so that if someone scans terraform directly and via gitlab they would come up with the same asset

vjeffrey (approved):
thank you! this is working well, i can make the platform ids change in a followup since i'm working on the list all groups stuff right now

providers/terraform/provider/detector.go
chris-rock: Yes we should, lets do this in a followup PR.

PR #1995, opened 2023-09-30
✨ terraform.settings.backend

arlimus:
Easy accessor for the terraform backend settings.

Note: The Terraform provider needs get its testing extended considerably, which is when we will expand testing for these fields as well.

Note2: The backend settings are pulled in, but backend settings blocks are a different story.


chris-rock (approved):
Nice 👍🏻

PR #1997, opened 2023-09-30
Add back more LR content that didn't make it to v9

tas50:
Diffing back some content from the v8 LR files


chris-rock (approved):
Thank you @tas50

PR #1998, opened 2023-09-30
🐛 download os provider for login+status

arlimus:
Both commands need to get system infos at the start. Without the os provider this is not possible (at the time of writing).

This PR also clarifies how the method `EnsureProviders` works and makes the list of existing providers optional.


chris-rock (approved):
Thank you

PR #1999, opened 2023-09-30
🐛 resource == null fixed

arlimus:
The binding is non-null, we need to compare the value, not the binding struct.


chris-rock (approved):
Thank you @arlimus

PR #2000, opened 2023-09-30
🐛 only scan local queries that are supported

arlimus:
We may not always have all providers installed. When we run against an upstream server we only deal with the resolved bundle, which usually is sufficient to filter out anything we cannot run. However, with local bundles this is different. Especially when we pull bundles and compile them locally, they may contain a lot of things that the system has no provider for.

This PR filters out anything from the local execution that is not covered by the currently supported providers. These are tied to the active connection.

Before:

image

After:

image


chris-rock (approved):
Great addition @arlimus

PR #2004, opened 2023-09-30
🐛 fix panic on uninitialized tf state

vjeffrey:
i mentioned the err here: https://github.com/mondoohq/cnquery/pull/1990


chris-rock:
It will turn green after #2001 is merged

chris-rock (approved):

PR #2005, opened 2023-10-01
Add creationDate to aws.s3.bucket resource

tas50:
This is a handy this to know for the asset overview


chris-rock (approved):
Thank you @tas50

PR #2015, opened 2023-10-02
⚡ re-use runtimes across providers

arlimus:
We minimize runtime creation as much as we can in this iteration.

1. Whenever an existing runtime connects to an asset (and collects identifying information about the asset like platform identifiers or MRN), it can be re-used in subsequent runs with the same asset. Discovery runs provide assets with discovery information, so they can be re-used. 
2. Runtimes that never connect to assets (like the initial runtime on a scan) will be discarded. There are very few of these.
3. On re-use, the provider-discovery is not necessary anymore, saving more calls.


vjeffrey:
🎉 ✨ 🥳  im excited for this one!

chris-rock (approved):
This is an great improvement

PR #2016, opened 2023-10-02
Adds EC2 metadata httpEndpoint field

scottford-io:
This PR adds a new field for EC2 instances to determine the status of the IMDS service, and allows users to determine whether IMDS is `enabled` or `disabled`. 

image


arlimus (approved):
thank you!

PR #2019, opened 2023-10-02
🐛 get value of incognito for shell (instead of always true)

vjeffrey:
fixes https://github.com/mondoohq/cnspec/issues/746


chris-rock (approved):
Thank you @vjeffrey

chris-rock:
This is okay to change but it is still not the correct implementation for #746. Setting incognito or not should not have any effect on vulnerability fetching. Incognito does not mean you cannot fetch stateless information, it just means nothing is being stored. Therefore even when incognito is set it should be able to pull vulnerability information.

PR #2020, opened 2023-10-02
🧹 bring back v8 ProviderType

imilchev:
The enum is still needed in order to support v8 inventory files and scan API clients. We should delete it once we drop v8 support


chris-rock (approved):
Thank you @imilchev

PR #2021, opened 2023-10-02
🐛 update provider defaults + missing connection types + v8 compat

arlimus:
Connection types were missing and thus caused any request that would look for them to fail. In most cases this is not noticable at all (because you eg connect to github to continue to do things with github as you discover more stuff).

But there is a very nasty case where this hurt: When we are in serve mode and don't have a primary connection that triggers provider installations. In this case we receive asset types (or rather v8 Backend fields!), which then didn't find any way to look up their provider.

So:
1. Add a script to help with defaults generation (it does 80% of the work, you only have to copy it over): ```bash version defaults providers/*/ ```
2. Update all providers with it
3. Support v8 `Backend` field (needs in-depth testing since both fields are not fully compatible; add translations where necessary)

### DIFF
--- a/providers/runtime.go
+++ b/providers/runtime.go
@@ -160,11 +160,8 @@ func (r *Runtime) DetectProvider(asset *inventory.Asset) error {
 	for i := range asset.Connections {
 		conn := asset.Connections[i]
 		if conn.Type == "" {
-			if conn.Backend != "" {
-				conn.Type = conn.Backend
-			} else {
-				continue
-			}
+			log.Warn().Msg("no connection `type` provided in inventory, falling back to deprecated `backend` field")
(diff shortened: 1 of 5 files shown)


providers/runtime.go
```diff
@@ -160,7 +160,8 @@ func (r *Runtime) DetectProvider(asset *inventory.Asset) error {
 	for i := range asset.Connections {
 		conn := asset.Connections[i]
 		if conn.Type == "" {
-			continue
+			log.Warn().Msg("no connection type provided, falling back to deprecated option backend")
```
chris-rock:
```suggestion
			log.Warn().Msg("no connection 'type' provided in inventory, falling back to deprecated 'backend' field")
```

chris-rock (approved):
Thank you @arlimus and @imilchev

PR #2039, opened 2023-10-02
✨ provide annotations upon registration

preslavgerchev:
Add support for annotations in the config, that gets generated during the login/registration.

`~/go/bin/cnquery login --token <token> --annotation a=b --annotation c=d`

Every asset, that gets scanned, using this config will get those annotations applied.

This PR also fixes the manually passed in annotations during the `scan` cmd, i.e. `~/go/bin/cnquery scan --annotation x=y`.

User-provided annotations have precedence over the config ones


vjeffrey (approved):
working beautifully!

Issue #2037, opened 2023-10-02
ms365: pkcs12: error reading P12 data: asn1: syntax error: sequence truncated


chris-rock:
**Describe the bug**

v9 ms365 does not load the pfx files correctly:

FTL could not load asset information error="rpc error: code = Unknown desc = cannot fetch credentials for microsoft provider: could not parse provided certificate at ./certificate.pfx: pkcs12: error reading P12 data: asn1: syntax error: sequence truncated"

The same credentials work without any issues in v8

**To Reproduce**

try to connect to ms365

**Expected behavior**

PR #2041, opened 2023-10-02
🐛 use inventory in shell to resolve credentials properly.

preslavgerchev:
Use an inventory to resolve assets before opening a shell. The inventory will ensure that all secrets are processed correctly and that any secrets, specified by a local path (certificates, private keys) are loaded in memory correctly.

This PR also adjusts the ms365 provider slightly by dropping the azure CLI support

### DIFF
--- a/apps/cnquery/cmd/shell.go
+++ b/apps/cnquery/cmd/shell.go
@@ -4,6 +4,7 @@
 package cmd
 
 import (
+	"errors"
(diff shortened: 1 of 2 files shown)


chris-rock (approved):
Thank you @preslavgerchev

chris-rock:
fixes #2037

PR #2050, opened 2023-10-03
⭐ support ephemeral vs normal runtimes

arlimus:
Ephemeral runtimes are used as throw-away runtimes, that come with their own plugin runner (and execution space). These runtimes can be thrown away after they are used, while other runtimes and plugins remain shared.

This is particularly useful when scanning targets with the discover mode, where potentially many runtimes/plugins need to be instantiated on the fly and are only required for the lifetime of the scan.

image

Ephemeral providers:

1. Have their own plug instance that is not share with any other provider.
2. They can be safely shut down, without impacting other providers.
3. Ephemeral runtimes may spawn additional providers, which are shared in nature. On shutdown of ephemeral providers these shared providers are not closed.
4. Non-ephemeral plugin instances are now only destroyed when the coordinator shuts down.

Fixes https://github.com/mondoohq/cnquery/issues/2049

### DIFF
--- a/explorer/scan/local_scanner.go
+++ b/explorer/scan/local_scanner.go
@@ -235,10 +235,11 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 	for i := range assetCandidates {
 		candidate := assetCandidates[i]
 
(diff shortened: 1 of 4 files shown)


imilchev (approved):
this fixes the errors! thanks 🎉

PR #2051, opened 2023-10-03
🐛 aws and terraform resource fixes

vjeffrey:
fixes https://github.com/mondoohq/cnquery/issues/2044

run cnquery scan aws and cnquery scan terraform against the server with the asset overview policy is now error free
fixes https://github.com/mondoohq/cnquery/issues/2003

### DIFF
--- a/providers/aws/resources/aws.lr.go
+++ b/providers/aws/resources/aws.lr.go
@@ -467,7 +467,7 @@ func init() {
 			Create: createAwsEc2InstanceDevice,
 		},
 		"aws.ec2.securitygroup": {
-			// to override args, implement: initAwsEc2Securitygroup(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error)
+			Init: initAwsEc2Securitygroup,
 			Create: createAwsEc2Securitygroup,
 		},
 		"aws.ec2.securitygroup.ippermission": {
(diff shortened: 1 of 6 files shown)


providers/aws/resources/aws_s3.go
```diff
@@ -198,7 +204,7 @@ func (a *mqlAwsS3Bucket) policy() (*mqlAwsS3BucketPolicy, error) {
 	}
 
 	// no bucket policy found, return nil for the policy
-	return &mqlAwsS3BucketPolicy{}, nil
+	return nil, errors.New("no policy found")
```
chris-rock: why are we returning an error and not just nil?

chris-rock (approved):

providers/aws/resources/aws_s3.go
vjeffrey: hmm, good point. ill check that

chris-rock: I think a NilData would be great because no policy is defined.

arlimus: fyi: we might have to set the field in this case (ie: `TValue[T]{State: StateIsNull | StateIsSet, Error: err}`). The hard thing here is that I had a hard time detecting with go if something is actually nil, because the nil gets wrapped in other types. I'll help solve this in the future to make it clear, but for now just set the TValue manually and return nil, nil if you need this :)

vjeffrey: 👍  pushed up the fix - er, actually, nevermind - pushing up fix in a few

vjeffrey: ugh, this resource is giving me issues when nil
```
cnquery> aws.s3.buckets { policy.id}
1 error occurred:
	* cannot cast resource to resource type: <nil>
	* 
	````

PR #2053, opened 2023-10-03
🐛 add compatibility layer for v8 inventory

arlimus:
The `kind` field was missing to be compatible with the API. Added a layer to translate v9 kind data to the v8 kind enum (which was removed because we no longer have a centralized authority on asset kinds.


providers-sdk/v1/inventory/deprecated_v8.go
```diff
@@ -0,0 +1,71 @@
...
+
+	for i := range asset.Connections {
+		conn := asset.Connections[i]
+		conn.Kind = Kind2DeprecatedV8Kind(asset.KindString)
+	}
+
+	// FIXME: Remove this and solve it at its core
+	var ids []string
+	for _, id := range asset.PlatformIds {
+		if strings.HasPrefix(id, "//") {
```
arlimus, line 31: I noticed a bunch of platform IDs that shouldn't be sent up:

image

arlimus: we need to fix these at their core, this is a sanity-check for v8, but it needs to be removed long-term

chris-rock (approved):
Thank you @arlimus

PR #2064, opened 2023-10-03
🐛 fix panic in azure discovery

vjeffrey:
fixes https://github.com/mondoohq/cnquery/issues/2054


chris-rock (approved):
Thank you @vjeffrey

PR #2069, opened 2023-10-03
⭐ terraform.resources( filter )

arlimus:
Support both string and regex filters for easily accessing all the resources that you need via their label.

image


chris-rock (approved):
Thank you @arlimus

PR #2075, opened 2023-10-04
🐛 avoid warnings on unknown CLI flags

arlimus:
We were still getting some leftover warnings on CLI flags. To work around this issue, this PR changes the way the detection works:

1. All unknown flags are now blindly accepted. These flags will continue to show up in the CLI and for the vast majority of use-cases they don't matter. At this early stage we don't have to get it right.
2. Only add root and attaching command flags now. This means we no longer recurse deep into all commands that exist under the root command. This saves a ton of time. Only works because of (1).


chris-rock (approved):
Thank you @arlimus

PR #2076, opened 2023-10-04
✨ add aws mock connection

arlimus:
Follows the pattern of the slack mock connection. Similar to slack, this pattern is not fully streamlined yet. I'll add that in a follow-up shortly, for now this works great :)

Now you can:

```
cnquery shell mock --use-recording recording.json 
```


chris-rock (approved):
Thank you @arlimus

PR #2077, opened 2023-10-04
🐛 fix empty aws.s3.bucket.policy + 🐛 fix its ID

arlimus:
The ID needs to be assigned after parsing at the moment. This is probably better stored in the generate file, but for now the manual assignment fixes things.

### DIFF
--- a/providers/aws/resources/aws_s3.go
+++ b/providers/aws/resources/aws_s3.go
@@ -164,13 +164,25 @@ func (a *mqlAwsS3Bucket) id() (string, error) {
 	return a.Arn.Data, nil
 }
 
+func emptyAwsS3BucketPolicy(runtime *plugin.Runtime) (*mqlAwsS3BucketPolicy, error) {
+	res, err := CreateResource(runtime, "aws.s3.bucket.policy", map[string]*llx.RawData{
+		"name":       llx.StringData(""),
+		"document":   llx.StringData("{}"),
+		"version":    llx.StringData(""),
+		"id":         llx.StringData(""),
+		"statements": llx.ArrayData([]interface{}{}, types.Dict),
+	})
+	if err != nil {
+		return nil, err
+	}
+	return res.(*mqlAwsS3BucketPolicy), nil
+}
+
 func (a *mqlAwsS3Bucket) policy() (*mqlAwsS3BucketPolicy, error) {
-	bucketname := a.Name.Data
-
-	location := a.Location.Data
-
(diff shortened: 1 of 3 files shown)


providers/aws/resources/aws_s3.go
```diff
@@ -164,13 +164,25 @@ func (a *mqlAwsS3Bucket) id() (string, error) {
...
 }
 
-func (a *mqlAwsS3Bucket) policy() (*mqlAwsS3BucketPolicy, error) {
-	bucketname := a.Name.Data
-
-	location := a.Location.Data
+func emptyAwsS3BucketPolicy(runtime *plugin.Runtime) (*mqlAwsS3BucketPolicy, error) {
+	res, err := CreateResource(runtime, "aws.s3.bucket.policy", map[string]*llx.RawData{
+		"name":       llx.StringData(""),
+		"document":   llx.StringData("{}"),
```
arlimus, line 170: without this, it's impossible to parse the document, since json-unmarshal expect it to at least read the structure. Without the document, we cannot determine an ID for the object, which is mandatory.

providers/aws/resources/aws_s3.go
```diff
@@ -164,13 +164,25 @@ func (a *mqlAwsS3Bucket) id() (string, error) {
...
+		"version":    llx.StringData(""),
+		"id":         llx.StringData(""),
+		"statements": llx.ArrayData([]interface{}{}, types.Dict),
```
arlimus, line 173: adding these as empty pre-cached fields so the runtime doesn't attempt to compute them

providers/aws/resources/aws_s3.go
```diff
@@ -179,9 +191,9 @@ func (a *mqlAwsS3Bucket) policy() (*mqlAwsS3BucketPolicy, error) {
 	})
 	if err != nil {
 		if isNotFoundForS3(err) {
-			return &mqlAwsS3BucketPolicy{}, nil
+			return emptyAwsS3BucketPolicy(a.MqlRuntime)
 		}
-		return &mqlAwsS3BucketPolicy{}, err
+		return nil, err
```
arlimus, line 196: if we have an error, we don't need to return the object

chris-rock (approved):
Thank you @arlimus

providers/aws/resources/aws_s3.go
vjeffrey: oooh! thank you!

PR #2082, opened 2023-10-04
✨ re-use provider instance for all assets when scanning k8s

imilchev:
Same as mondoohq/cnspec#802


czunker (approved):
LGTM. Thanks @imilchev

PR #2087, opened 2023-10-04
Restore docs generation + lr updates

tas50:
- Get make working for docs generation
- Backport some missing fixes in LR files that went into v8


arlimus (approved):
Thank you for updating this!

PR #2089, opened 2023-10-04
🐛 fix v8 compat issues

arlimus:
Fixes: https://github.com/mondoohq/cnquery/issues/2088


providers/vcd/resources/vcd.lr
```diff
@@ -120,6 +120,9 @@ vcd.networkPool {
 
 // VMware Cloud Director External Network
 private vcd.externalNetwork @defaults("name") {
+  // URN of the network.
+  // Deprecated: Please use the urn field instead.
```
tas50: Can we shove this on one line. Otherwise the docs get generated with a funky comma after the period

vjeffrey (approved):

PR #2095, opened 2023-10-05
✨ introduce dev-only llx run command (v9 edition)

arlimus:
v8 to v9 compatibility has bitten me once too often... actually it goes beyond that with past versions.

To ease these problems, this command is introduced to make the llx bundles shareable and testable across versions:

```bash
> cnquery run -c "command('echo hi 🌎').stdout.trim" --llx bundle.llx

> cnquery8 run --use-llx bundle.llx
command.stdout.trim: "hi 🌎"
```


chris-rock (approved):
Thank you @arlimus

PR #2100, opened 2023-10-05
✨ implement k8s resource filters

imilchev:
Something that we forgot to implement in v9 but it is needed by the mondoo-operator. To test this use the following inventory file:
```yaml
apiVersion: v1
kind: Inventory
metadata:
  name: inventory
spec:
  assets:
    - connections:
        - type: k8s
          options:
            k8s-resources: pod:default:nginx,deployment:mondoo-operator:mondoo-operator-controller-manager
          discover:
            targets:
              - auto
```

run it with `cnquery scan k8s --inventory-file inv.yaml`. Only the pod `nginx` in the `default` namespace and the pod `mondoo-controller-manager` in the `mondoo-operator` namespace should be scanned

### DIFF
--- a/providers/k8s/resources/discovery.go
+++ b/providers/k8s/resources/discovery.go
@@ -109,6 +109,11 @@ func Discover(runtime *plugin.Runtime) (*inventory.Inventory, error) {
 		nsFilter.exclude = strings.Split(exclude, ",")
 	}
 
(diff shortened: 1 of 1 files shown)


chris-rock (approved):
Thank you @imilchev

PR #2109, opened 2023-10-06
Update cnquery description from fleet -> infra

tas50:
We're not using the fleet word anymore. Update to infrastructure to match the same change we made in v8


chris-rock (approved):
Thank you @tas50

PR #2123, opened 2023-10-08
🧹 allow url option for custom gitlab urls

vjeffrey:


providers/gitlab/config/config.go
```diff
@@ -48,6 +48,12 @@ var Config = plugin.Provider{
 					Default: "",
 					Desc:    "a GitLab project to scan",
 				},
+				{
+					Long:    "url",
+					Type:    plugin.FlagType_String,
+					Default: "",
+					Desc:    "custom url of the gitlab org",
```
tas50:
```suggestion
					Desc:    "custom url of the GitLab org",
```

imilchev: I think this isn't valid. That's the gitlab base url, so I will change the desc

PR #2125, opened 2023-10-09
🧹 Cleanup workaround for GCP shutdown

### LINKED ISSUE #1823: 🧹 remove gcp-snapshot cleanup workaround
**Describe the bug**
I think, this can now be removed: https://github.com/mondoohq/cnquery/blob/ebe8c80f79d1faf596b52668f860f6bae7d4f590/apps/cnquery/cmd/plugin.go#L105

This also happens here: https://github.com/mondoohq/cnquery/blob/main/providers/gcp/provider/provider.go#L184

**To Reproduce**

1. Remove the workaround
2. Run a gcp snapshot scan on a GCP VM
3. Check that the scan works and the mount & disk is cleaned up afterwards

**Expected behavior**
The cleanup should happen without the workaround.

czunker:
Fixes #1823

### DIFF
--- a/apps/cnquery/cmd/plugin.go
+++ b/apps/cnquery/cmd/plugin.go
@@ -122,14 +122,6 @@ func (c *cnqueryPlugin) RunQuery(conf *run.RunQueryConfig, runtime *providers.Ru
 		out.WriteString("[")
 	}
 
-	// FIXME: workaround for gcp-snapshot
-	// For a gcp-snapshot asset, we start with a GCP connection.
-	// This get's overridden by a filesystem connection. The filesystem connection is what we need for the scan
-	// But later, we need the GCP runtime to cleanup the snapshot disk
-	if runtime.Provider.Instance.Name == "gcp" && runtime.Provider.Connection.Name == "filesystem" {
-		defer runtime.Close()
-	}
-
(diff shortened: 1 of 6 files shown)


czunker:
This only works, when https://github.com/mondoohq/cnquery/issues/2126 is solved.

czunker:
> This only works, when #2126 is solved.

According to https://github.com/mondoohq/cnquery/pull/2187#issuecomment-1769986452, this should no longer be needed.

apps/cnquery/cmd/plugin.go
```diff
@@ -122,14 +122,6 @@ func (c *cnqueryPlugin) RunQuery(conf *run.RunQueryConfig, runtime *providers.Ru
 		out.WriteString("[")
 	}
 
-	// FIXME: workaround for gcp-snapshot
-	// For a gcp-snapshot asset, we start with a GCP connection.
-	// This get's overridden by a filesystem connection. The filesystem connection is what we need for the scan
-	// But later, we need the GCP runtime to cleanup the snapshot disk
-	if runtime.Provider.Instance.Name == "gcp" && runtime.Provider.Connection.Name == "filesystem" {
```
chris-rock, line 129: we can clean this because it happens here https://github.com/mondoohq/cnquery/blob/main/providers/gcp/provider/provider.go#L184 now

chris-rock (approved):
Thank you @czunker

PR #2127, opened 2023-10-09
🐛 Use correct provider for current connection

### LINKED ISSUE #2126: GCP snapshot scan fails
**Describe the bug**
```
sudo ./cnquery run gcp snapshot --project-id mondoo-dev-262313 suse15 -c "asset{ name kind platform ids }" --verbose
! can't find any paths for providers, none are configured system-path=/opt/mondoo/providers
DBG using provider gcp with connector gcp
DBG Started a new runtime (1 total)
! using builtin provider for gcp
...
DBG created tmp scan dir dir=/tmp/cnspec-scan1952179752
DBG search for target volume
...
DBG load filesystem path=/tmp/cnspec-scan1952179752
DBG could not execute os release detection command error="provider does not implement RunCommand"
DBG platform> cannot parse lsb config on this linux system error="open /tmp/cnspec-scan1952179752/etc/lsb-release: no such file or directory"
DBG could not execute os release detection command error="provider does not implement RunCommand"
DBG platform> detected os family=["suse","linux","unix","os"] platform=sles
FTL failed to run query error="provider type does not match"
```

At the point, where the execution fails, cnquery tries to use a `filesystem` connection with the gcp provider. At this point it should use the `filesystem` connection with the os provider.

We first need the gcp provider to get the needed metadata and attach the disk, the we use the os provider to scan the filesystem.

**To Reproduce**

1. Spin up a GCP VM
2. Run the above cnquery with an existing snapshot
3. Note the error

**Expected behavior**
The command should get executed.

- cnspec v9.0.1

**Additional context**
Perhaps related to https://github.com/mondoohq/cnquery/issues/2120

czunker:
Even, when a runtime already exists, it might have the wrong provider associated.

In cases where we swtich the provider, this has the wrong provider.

Fixes #2126


czunker:
As discussed with @arlimus, I'll try to find a different spot to solve this, when possible.

czunker:
#2168 fixed a similar issue for the switch from, e.g., was instance-connect to ssh.
But this isn't working in this case.

For instance-connect -> ssh the connection type is directly set to ssh in ParseCli.
For the snapshot case, we still need the cloud provider to handle some tasks via the cloud API and later change the connection type to "filesystem" in the Connect func.

czunker:
This would also fix ebs snapshot scanning: https://github.com/mondoohq/cnquery/issues/2158

czunker:
According to https://github.com/mondoohq/cnquery/pull/2187#issuecomment-1769986452 this should no longer be needed.

PR #2162, opened 2023-10-10
✨ Add new azure network resources

preslavgerchev:
The new resources:
   - virtual networks
   - virtual network gateways
   - load balancers
   - subnets
   - nat gateways
   - firewalls
   - firewall policies
   - application security groups
  
Still need to do:
 - [x] firewalls
 - [x] firewall policies
 - [x] application security groups

cc @HRouhani, feel free to test it out. I have added most of the stuff that was required, but ping me if there's something missing

### DIFF
--- a/providers/azure/resources/azure.lr
+++ b/providers/azure/resources/azure.lr
@@ -188,6 +188,10 @@ private azure.subscription.networkService {
   virtualNetworks() []azure.subscription.networkService.virtualNetwork
   // List of virtual network gateways
   virtualNetworkGateways() []azure.subscription.networkService.virtualNetworkGateway
+  // List of network firewalls
+  firewalls() []azure.subscription.networkService.firewall
+  // List of firewall policies
+  firewallPolicies() []azure.subscription.networkService.firewallPolicy
 }
 
 // Azure Network Virtual Network Gateway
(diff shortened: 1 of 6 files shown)


providers/azure/resources/azure.lr.manifest.yaml
```diff
@@ -953,16 +953,35 @@ resources:
...
       - azure
+  azure.subscription.networkService.backendAddressPool:
+    fields:
+      etag: {}
+      id: {}
+      name: {}
+      properties: {}
+      type: {}
+    is_private: true
+    min_mondoo_version: latest
```
tas50: Can we get the intended version on all these?

preslavgerchev: I updated it, but we should revisit this whole approach. This should switch to provider version and not cnquery version

imilchev (approved):

providers/azure/resources/azure.lr
```diff
@@ -180,6 +180,464 @@ private azure.subscription.networkService {
...
+  // Outbound Rule id
+  id string
+  // Outbound Rule name
+  name string
+  // Outbound Rule type
+  type string
+  // Outbound Rule etag
+  etag string
+  // Outbound Rule properties
+  properties dict
```
vjeffrey, line 640: oh dang! so many new resources!

PR #2164, opened 2023-10-10
🐛 Prevent error when no providers are build

czunker:


czunker:
This should prevent this error: https://github.com/mondoohq/cnquery/actions/runs/6465525126

image

imilchev (approved):

PR #2171, opened 2023-10-11
🐛 Fix AWS EBS missing region

### LINKED ISSUE #2158: ebs scan failing with describe instances missing region error
```
FTL could not load asset information error="rpc error: code = Unknown desc = unable to validate: operation error EC2: DescribeInstances, failed to resolve service endpoint, endpoint rule error, Invalid Configuration: Missing Region"
```

this is new, when i last tested a couple weeks ago things worked.

i'm also super unclear on how we're getting here...none of the logging i would expect is showing up
```
x unable to connect to asset error="rpc error: code = Unknown desc = unable to validate: operation error EC2: DescribeInstances, failed to resolve service endpoint, endpoint rule error, Invalid Configuration: Missing Region"
```

ssm and instanceconnect on the cli are also misbehaving. these all worked before, im not sure what's going on yet

czunker:
Fixes #2158

### DIFF
--- a/providers/aws/connection/awsec2ebsconn/provider.go
+++ b/providers/aws/connection/awsec2ebsconn/provider.go
@@ -62,6 +62,7 @@ func NewAwsEbsConnection(id uint32, conf *inventory.Config, asset *inventory.Ass
 	if err != nil {
 		return nil, errors.Wrap(err, "could not load instance info: aws-ec2-ebs provider only valid on ec2 instances")
 	}
+	log.Debug().Interface("info", i).Msg("received instance info")
 
 	// ec2 client for the scanner region
 	cfg.Region = i.Region
@@ -69,7 +70,7 @@ func NewAwsEbsConnection(id uint32, conf *inventory.Config, asset *inventory.Ass
 
 	targetRegion := conf.Options["region"]
 	if targetRegion == "" {
-		log.Info().Msg("flag --region not specified, using scanner instance region")
+		log.Info().Msg("flag --region not specified, using instance region")
 		targetRegion = i.Region
 	}
 
@@ -86,6 +87,7 @@ func NewAwsEbsConnection(id uint32, conf *inventory.Config, asset *inventory.Ass
 			PlatformId: conf.PlatformId,
 			Region:     targetRegion,
 			Id:         conf.Options["id"],
+			AccountId:  i.AccountID,
 		},
 		targetType: conf.Options["type"],
 		scannerInstance: &awsec2ebstypes.InstanceId{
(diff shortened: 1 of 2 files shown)


providers/aws/connection/awsec2ebsconn/provider.go
```diff
@@ -62,14 +62,20 @@ func NewAwsEbsConnection(id uint32, conf *inventory.Config, asset *inventory.Ass
...
 		return nil, errors.Wrap(err, "could not load instance info: aws-ec2-ebs provider only valid on ec2 instances")
 	}
+	log.Debug().Interface("info", i).Msg("received instance info")
 
 	// ec2 client for the scanner region
 	cfg.Region = i.Region
 	scannerSvc := ec2.NewFromConfig(cfg)
 
+	targetRegion := conf.Options["region"]
+	if targetRegion == "" {
```
vjeffrey, line 72: if conf.Options["region"] is empty we actually need to fail out, we can't properly scan unless we know the target's region

apps/cnquery/cmd/plugin.go
```diff
@@ -120,7 +120,7 @@ func (c *cnqueryPlugin) RunQuery(conf *run.RunQueryConfig, runtime *providers.Ru
 
 	for i := range assets {
 		connectAsset := assets[i]
-		connectAssetRuntime, err := providers.Coordinator.RuntimeFor(connectAsset, runtime)
+		connectAssetRuntime, err := providers.Coordinator.EphemeralRuntimeFor(connectAsset)
```
vjeffrey: ooh, this must be the provider switch fix?

czunker: No, that's for the cleanup.

I think, I need to test this with multiple assets and a `run` command. Not sure how this would behave.

I didn't need a provider switch fix for ebs. But I don't understand why.

providers/aws/connection/awsec2ebsconn/provider.go
czunker: Oh, I thought I could asume the same region as the scanner.

apps/cnquery/cmd/plugin.go
vjeffrey: ah, i just tested this, i think we're still getting the wrong provider, we shouldn't be hitting describe instances before we hit some other logs, looks like we're redirecting this to the aws provider

vjeffrey: i tried testing with your other pr and it somehow didn't fix the connection here...ill dig in today

providers/aws/connection/awsec2ebsconn/provider.go
czunker: Before, we relied on the `--region` flag. When this wasn't present, the describes error in the issue happened. Now, we still use this flag when present, but fallback to the scanner region.

czunker:
Superseded by https://github.com/mondoohq/cnquery/pull/2293

PR #2172, opened 2023-10-11
🧹 Fix deprecated goreleaser flags

czunker:
https://goreleaser.com/deprecations/\#-skip


czunker:
When approved, we can port this change to cnspec and server.

imilchev (approved):

PR #2182, opened 2023-10-11
✨ add filters to aws discovery

vjeffrey:
i don't love the strict, hidden syntax im using to parse this all from the cli...but it does work. 
open to ideas on how to improve! 

i need filtering for the lambda (we had ec2 filtering in v8)

```
go run apps/cnquery/cnquery.go shell aws --discover instances  --filters ec2:region=us-east-2

go run apps/cnquery/cnquery.go shell aws --discover lambda-functions  --filters all:region=us-east-2
```

### DIFF
--- a/providers/aws/config/config.go
+++ b/providers/aws/config/config.go
@@ -84,6 +84,12 @@ var Config = plugin.Provider{
 					Default: "",
 					Desc:    "Override option for EBS scanning that tells it to not create the snapshot or volume",
 				},
+				{
+					Long:    "filters",
+					Type:    plugin.FlagType_KeyValue,
+					Default: "",
+					Desc:    "Filter options",
+				},
 			},
 		},
 	},
--- a/providers/aws/connection/connection.go
+++ b/providers/aws/connection/connection.go
@@ -6,6 +6,7 @@ package connection
 import (
 	"context"
 	"errors"
+	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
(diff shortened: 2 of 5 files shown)


czunker:
This would fix https://github.com/mondoohq/cnquery/issues/913

czunker:
I like these filters. For the region, perhaps removing the prefix makes it more intuitive. Combining the above example would result in:
```
aws --discover instances,lambda-functions  --filters ec2:region=us-east-2,all:region=us-east-2
```
I'd expect to use something, like:
```
aws --discover instances,lambda-functions  --filters region=us-east-2
```

mariuskimmina:
I think it should be possible to just have a filter like `region=us-east-2` without specifying the service.
So this would be valid

```
go run apps/cnquery/cnquery.go shell aws --discover instances  --filters region=us-east-2 
```

And this would also be valid (just wouldn't do anything in this case)

```
go run apps/cnquery/cnquery.go shell aws --discover lambda-functions  --filters instance-id=xxxxx
```

Maybe we could just have one struct `DiscoveryFilters` have the functions `instanceMatchesFilters` figure out what it needs to use?

vjeffrey:
👍  i like the idea of not requiring the prefix on the region

i would need to keep the prefix as an option for backwards compatibility - restricting everything to a region vs just ec2 scans to a region are two very diff things, but could make it so cli can use region without prefix and we just translate it to both

vjeffrey:
i pushed the changes to support region alone

vjeffrey:
> Maybe we could just have one struct DiscoveryFilters have the functions instanceMatchesFilters figure out what it needs to use?

I'm not entirely certain i understand this suggestion, can you expand?

arlimus (approved):
A ton of power added to these filters!! :star:

PR #2184, opened 2023-10-12
🧹 Remove _motor directory.

preslavgerchev:
Get rid off the v8 `_motor` directory, we shouldn't need that anymore


chris-rock:
platform detection tests are migrated here https://github.com/mondoohq/cnquery/pull/2225.

Here is the list of things we need to check before we remove

- [x] we can delete discovery since that is not required anymore
- [ ] we need to validate if there is something left from ssh
- [ ] we need to validate if something is missing from winrm

arlimus:
Thank you for catching these old dependencies!! We finally included this work in the v10 bump, given the amount of testing we've had in the last 3 months. Kudos!

See: https://github.com/mondoohq/cnquery/pull/2966

PR #2187, opened 2023-10-12
✨🐛 support switching providers post-connect

arlimus:
In the original design, the connect call would always return the connected asset. However, we have some legacy code that behaved differently in v8. When it was translated to v9, we had to return the fully discovered asset, which is the asset that we actually want to connect.

For now, the connect call will continue to fulfill two functions:
1. discover the actual asset and
2. establish the connection

During this time, we will check once if the provider changed after the connect call and switch providers if that happened.

We want a distinct discover and connect call in the future.

Note:
1. The connect call on the shell behaves differently, happy to adjust it if we need it there too.
2. If additional providers get connected via resources, they do not go through this procedure. This is intentional. Example: The os provider calls out to the network provider. There is no way for that process to all of a sudden end up with a completely different provider (that would also be a nightmare to debug)

### DIFF
--- a/providers/runtime.go
+++ b/providers/runtime.go
@@ -242,6 +242,7 @@ func (r *Runtime) Connect(req *plugin.ConnectReq) error {
 		return err
 	}
 	if postProvider.ID != r.Provider.Instance.ID {
+		req.Asset = r.Provider.Connection.Asset
 		r.UseProvider(postProvider.ID)
 		r.Provider.Connection, err = r.Provider.Instance.Plugin.Connect(req, &callbacks)
 		if err != nil {
--- a/providers/runtime.go
+++ b/providers/runtime.go
@@ -164,11 +164,19 @@ func (r *Runtime) addProvider(id string, isEphemeral bool) (*ConnectedProvider, 
 // but haven't initialized any provider. It will also try to install providers
 // if necessary (and enabled)
 func (r *Runtime) DetectProvider(asset *inventory.Asset) error {
+	provider, err := r.providerForAsset(asset)
+	if err != nil {
+		return err
+	}
+	return r.UseProvider(provider.ID)
+}
+
+func (r *Runtime) providerForAsset(asset *inventory.Asset) (*Provider, error) {
 	if asset == nil {
-		return errors.New("please provide an asset to detect the provider")
+		return nil, errors.New("please provide an asset to detect the provider")
 	}
 	if len(asset.Connections) == 0 {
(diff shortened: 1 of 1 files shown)


arlimus:
Feel free to hold off on it if it doesn't solve the problem

czunker:
I tried it with the ebs snapshot scan (#2171). It was only working with builtin providers. https://github.com/mondoohq/cnquery/pull/2127/files was working in both cases (builtin and plugin). Perhaps we can combine these two somehow. I've currently no clue, what's the problem, sorry.

imilchev:
If this works for builtin providers but not for installed ones, then it's probably an issue with pointers that get set in the original asset when using builtin because we are changing the pointer in the same process. For installed providers that doesn't work since we adjust the asset over gRPC. That's just a guess based on stuff I have seen so far

vjeffrey:
sorry it took me so long to get back to this, i kept trying to test in between other things and not paying enough attention (its working well against most machines, but the machine i was testing against seems to be having troubles, but that's unrelated to this)

with this change + one other small change in the ebs provider, things work! thank you!

vjeffrey:
hmmm, i think something else changed since this first came up -- things seem to work without this change now

arlimus:
Holding off on it for now.

czunker:
Still needed for #2126 .

providers/runtime.go
```diff
@@ -216,6 +224,32 @@ func (r *Runtime) Connect(req *plugin.ConnectReq) error {
...
+	//
+	// Additionally, we do not loop this connect+recheck approach indefinitely.
+	// We only run it once and only accept one asset switch. This will be
+	// changed once we have an explicit discover call in plugins.
+	postProvider, err := r.providerForAsset(r.Provider.Connection.Asset)
+	if err != nil {
+		return err
+	}
+	if postProvider.ID != r.Provider.Instance.ID {
+		req.Asset = r.Provider.Connection.Asset
```
czunker, line 245: @arlimus I needed to add this line to make it also work with plugins. With this change, the results also looks like the one in v8.

I tried to set all this inside the GCP provider, but it does not get transferred to the os provider.

czunker:
@vjeffrey Any objects to merge this? You mentioned EBS scanning is now working without this change.

@arlimus Do you agree with the additional change?

providers/runtime.go
arlimus: Makes sense, thank you for the catch!

czunker (approved):
Thanks @arlimus

PR #2189, opened 2023-10-12
✨ Add more fields to microsoft.serviceprincipal, retrieve enterprise applications

preslavgerchev:
Add support for `microsoft.enterpriseApplications`, which is a list of service principals with a special tag

### DIFF
--- a/providers/ms365/resources/ms365.lr
+++ b/providers/ms365/resources/ms365.lr
@@ -166,7 +166,7 @@ private microsoft.application @defaults("id displayName") {
 }
 
 // Microsoft Service Principal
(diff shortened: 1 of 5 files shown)


providers/ms365/resources/ms365.lr
```diff
@@ -167,6 +169,38 @@ private microsoft.application @defaults("id displayName") {
 private microsoft.serviceprincipal @defaults("id") {
```
chris-rock: id is not a useful default, switching it to name

chris-rock: updated the definition

chris-rock (approved):
Thank you @preslavgerchev

PR #2194, opened 2023-10-12
Use MQL functions to retrieve aws resources, making use of the cache.

preslavgerchev:
By using the MQL implementation of `Get`, we hit the cache once it's computed, this saves a lot of time when iterating over resources

There are probably other places where we need to change this, I only made sure that `aws.ec2.instances` queries are fixed. Will follow through with others later

Before:
```
➜ ~/go/bin/cnquery run aws -c 'aws.ec2.instances.length'
aws.ec2.instances.length: 10

cnquery on  preslav/aws-fix [$!?] via 🐹 v1.21.1 on 🐳 v20.10.16 using ☁️  default/mondoo-prod-243414 took 3m 7s
```

After:
``` 
➜ ~/go/bin/cnquery run aws -c 'aws.ec2.instances.length'
! using builtin provider for aws
! using builtin provider for aws
! using builtin provider for aws
aws.ec2.instances.length: 10

cnquery on  preslav/aws-fix [$!?] via 🐹 v1.21.1 on 🐳 v20.10.16 using ☁️  default/mondoo-prod-243414 took 12s 
```

### DIFF
--- a/providers/aws/resources/aws_ec2.go
+++ b/providers/aws/resources/aws_ec2.go
@@ -614,7 +614,7 @@ func (a *mqlAwsEc2) getInstances(conn *connection.AwsConnection) []*jobpool.Job 
 	for _, region := range regions {
 		regionVal := region
 		f := func() (jobpool.JobResult, error) {
-			log.Debug().Msgf("calling aws with region %s", regionVal)
+			log.Debug().Msgf("calling aws func with region %s", regionVal)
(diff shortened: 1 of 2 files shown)


vjeffrey:
niiice, thank you! sorry about all these 🙈 not sure why i ended up using the other stuff, mustve just fallen into a bad pattern

PR #2195, opened 2023-10-12
🧹 Improve debug logging for AWS calls, do not fetch ec2's instance VPC if not needed

preslavgerchev:
* Improve logging for all AWS client calls. This is useful when debugging to see what API calls are going out to see where the problem might be
 * Make sure we only load an instance's VPC if actually requested, i.e. compute it once needed. To do so, I've added a `vpcArn` field that holds the vpc's ARN and can be used to request it

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -1766,6 +1766,8 @@ private aws.ec2.instance @defaults("arn state") {
   keypair() aws.ec2.keypair
   // Time when the last state transition occurred
   stateTransitionTime time
+  // The ARN of the VPC, associated with the instance
+  vpcArn string
 }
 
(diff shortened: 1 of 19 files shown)


providers/aws/resources/aws.lr
```diff
@@ -1766,6 +1766,8 @@ private aws.ec2.instance @defaults("arn state") {
   keypair() aws.ec2.keypair
   // Time when the last state transition occurred
   stateTransitionTime time
+  // The ARN of the VPC, associated with the instance
```
tas50:
```suggestion
  // The ARN of the VPC associated with the instance
```

chris-rock (approved):
Thank you @preslavgerchev

PR #2196, opened 2023-10-12
🐛 hot fix for too long upstream scan time

vjeffrey:
i haven't been able to find a good policy with examples of the os + network resources so i didn't do the cross provider types thing yet, im just skipping over the check if its network or core. gonna go find a test example for the network stuff now

```
with this branch, reporting upstream, and all providers installed:
`DEBUG=1 go run apps/cnquery/cnquery.go scan local  2.71s user 1.89s system 39% cpu 11.757 total`
`DEBUG=1 go run apps/cnspec/cnspec.go scan local  4.12s user 3.00s system 42% cpu 16.767 total`
```

im not patient enough to time out the version of this scan on main. i can say its more than 5 minutes for sure.

### DIFF
--- a/providers-sdk/v1/plugin/start.go
+++ b/providers-sdk/v1/plugin/start.go
@@ -17,7 +17,18 @@ type Provider struct {
 	ID              string
 	Version         string
 	ConnectionTypes []string
-	Connectors      []Connector
+	// CrossProviderTypes are asset providers that already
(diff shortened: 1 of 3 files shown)


arlimus (approved):
Nice stopgap, let's overhaul cross provider calls later :)

PR #2198, opened 2023-10-13
🐛 Fix aws dynamodb table init function, make all AWS init funcs use the mql cache

preslavgerchev:
Followup to #2194

### DIFF
--- a/providers/aws/resources/aws_dynamodb.go
+++ b/providers/aws/resources/aws_dynamodb.go
@@ -79,7 +79,7 @@ func (a *mqlAwsDynamodb) getBackups(conn *connection.AwsConnection) []*jobpool.J
 	return tasks
 }
 
-func initDynamoDbTable(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error) {
+func initAwsDynamodbTable(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error) {
 	if len(args) > 2 {
 		return args, nil, nil
 	}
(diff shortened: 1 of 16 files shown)


providers/aws/resources/aws_dynamodb.go
```diff
@@ -102,14 +102,17 @@ func initDynamoDbTable(runtime *plugin.Runtime, args map[string]*llx.RawData) (m
 	}
 	dynamodb := obj.(*mqlAwsDynamodb)
 
-	rawResources, err := dynamodb.tables()
+	rawResources := dynamodb.GetTables()
 	if err != nil {
 		return nil, nil, err
 	}
```
mariuskimmina: This is now checking an error that has been checked before and can be removed I think

providers/aws/resources/aws_dynamodb.go
```diff
@@ -360,14 +363,17 @@ func initAwsDynamodbGlobaltable(runtime *plugin.Runtime, args map[string]*llx.Ra
...
 	dynamodb := obj.(*mqlAwsDynamodb)
 
-	rawResources, err := dynamodb.globalTables()
+	rawResources := dynamodb.GetGlobalTables()
 	if err != nil {
 		return nil, nil, err
 	}
```
mariuskimmina: This is also now checking an error that has been checked before and can be removed I think

mariuskimmina (commented):
Nicely done, just some small suggestions

providers/aws/resources/aws_dynamodb.go
preslavgerchev: great catch, resolved

mariuskimmina (approved):

PR #2201, opened 2023-10-13
🧹 Allow running microsoft queries from an azure asset

preslavgerchev:
Assume we can get no credentials if running just `shell azure`. This also enables `shell ms365 --tenant-id <value>` which makes querying a bit easier

### DIFF
--- a/providers/ms365/connection/auth.go
+++ b/providers/ms365/connection/auth.go
@@ -18,7 +18,10 @@ func getTokenCredential(credential *vault.Credential, tenantId, clientId string)
 	var azCred azcore.TokenCredential
 	var err error
 	if credential == nil {
-		return nil, errors.New("no credentials provided")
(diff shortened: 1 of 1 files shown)


preslavgerchev:
Note: this currently doesn't work cause of #2196, I tested by dropping this commit and then re-building cnquery

vjeffrey:
we need to do the cross provider connection types stuff so we can include this

vjeffrey:
we need to add the cross provider types and then read those where we're doing that provider check 
https://github.com/mondoohq/cnquery/blob/main/providers/network/config/config.go#L16
https://github.com/mondoohq/cnquery/blob/main/providers/runtime.go#L578

arlimus:
I'll get the cross-provider mechanism rolling to tackle this.

preslavgerchev:
Closing this as the auth files have been moved, the fix in this PR is no longer applicable

PR #2211, opened 2023-10-14
⭐ http.get + url + STS support

arlimus:
We are starting with one specific use-case, but it's a fun one:

```coffee
http.get('https://www.gov.uk') { version statusCode header.sts }
```

which gets us:

```coffee
http.get: {
  statusCode: 200
  header.sts: http.header.sts maxAge=365 days  includeSubDomains=false preload=true
  version: "2.0"
}
```

### DIFF
--- a/providers/network/resources/network.lr
+++ b/providers/network/resources/network.lr
@@ -14,6 +14,67 @@ socket @defaults("protocol port address") {
   address string
 }
 
+// HTTP endpoint
+http {}
+
+// HTTP GET requests
+http.get @defaults("url statusCode") {
+  init(rawUrl string)
+  // URL for this request
+  url url
+  // Returned header for this request
+  header() http.header
+  // Status returned by this request
+  statusCode() int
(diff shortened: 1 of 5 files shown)


providers/network/resources/network.lr
```diff
@@ -14,6 +14,67 @@ socket @defaults("protocol port address") {
...
+  password string
+  // Host, either registered name or IP; e.g. mondoo.com
+  host string
+  // Port, optional decimal number; e.g. 80
+  port int
+  // Path, consisting of segments separated by '/'
+  path string
+  // Query, optional, attached to path via '?', parsed via '&' and ';' delimiters
+  query map[string]string
+  // Raw Query, optional raw string attched to path after '?'
```
tas50:
```suggestion
  // Raw Query, optional raw string attached to path after '?'
```

arlimus: thank you!

chris-rock (approved):

PR #2213, opened 2023-10-14
✨ more HTTP security headers

arlimus:
After: https://github.com/mondoohq/cnquery/pull/2211

adds:
- X-Frame-Options
- X-XSS-Protection
- X-Content-Type-Options
- Referrer-Policy

```coffee
http.get('https://console.mondoo.com') { version header{ xFrameOptions xXssProtection xContentTypeOptions referrerPolicy }}
```

```coffee
http.get: {
  header: {
    xContentTypeOptions: "nosniff"
    referrerPolicy: "same-origin"
    xFrameOptions: "SAMEORIGIN"
    xXssProtection: http.header.xssProtection enabled=true mode="block" report=null
  }
  version: "2.0"
}
```

... getting closer to CSP

kudos @benr

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -34,6 +34,7 @@ mfs
 mgroup
 Mpim
 nodepool
+nosniff
 nullgroup
 opcplc
 Pids
@@ -42,6 +43,7 @@ pushconfig
 querypack
 resourcegroup
 Sas
+SAMEORIGIN
 serviceprincipals
 Snat
 spdx
--- a/providers/network/resources/http.go
+++ b/providers/network/resources/http.go
@@ -90,16 +90,29 @@ func (x *mqlHttpGet) header() (*mqlHttpHeader, error) {
 		for j := range vals {
 			ivals[j] = vals[j]
 		}
-		params[mkey] = ivals
+		params[normalizeHeaderKey(mkey)] = ivals
 	}
 
 	o, err := CreateResource(x.MqlRuntime, "http.header", map[string]*llx.RawData{
+		"__id":   llx.StringData(x.__id),
 		"params": llx.MapData(params, types.Array(types.String)),
 	})
 	if err != nil {
 		return nil, err
 	}
 	return o.(*mqlHttpHeader), nil
+}
+
+var normHeaderKeys = map[string]string{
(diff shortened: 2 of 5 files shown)


scottford-io:
@arlimus This is looking really great. Is there a way to attach `http.get` to the `host` provider? Maybe I am not understanding this correctly, but my hope is that you can simply run this with 

```
cnspec scan host https://mondoo.com
```

Then be able to run checks such as:

```
http.header.sts.maxAge <= 365
```

arlimus:
@scottford-io Awesome questions!!

**1: host targets** 

Fixed in https://github.com/mondoohq/cnquery/pull/2217 see that PR, it has a great example. That will also allow use to just `cnspec scan host ...`

**2: STS maxAge**

This is a fun one: by default the STS maxAge is transmitted as a number of seconds:

```
Strict-Transport-Security: max-age=31536000; includeSubDomains
```

However, as you can see the result in cnquery looks like this:

```coffee
header.sts: http.header.sts maxAge=365 days  includeSubDomains=true preload=false
```

The fun part is: we already convert it to a proper duration internally. This means you can now run queries against it, that map to durations, like this:

```coffee
cnquery> http.get.header.sts.maxAge > 10 * time.day
[ok] value: 365 days 

cnquery> http.get.header.sts.maxAge <= 365 * time.day
[ok] value: 365 days 

cnquery> http.get.header.sts.maxAge <= 355 * time.day
[failed] http.get.header.sts.maxAge < <ref>
  expected: < 355 days 
  actual:   365 days 
```

So you can use any of the `time` values we support (which is day, hour, minute, and second). 

(Tangent: We don't really support `time.year` yet, curious to see if we have a use-case. I'm very hesitant to add `time.month` because that may be 30 or 31 days.)

chris-rock:
We also have the fqdn exposed via asset.fqdn which allows you to write tests for scenarios where the url is based on the domain.

```bash
cnquery shell host https://mondoo.com

cnquery> asset.fqdn
asset.fqdn: "mondoo.com"
```

This allows you to write:

```
http.get(asset.fqdn + "/my/path") { version statusCode header.sts }
```

Right now, 9.1.0 has the issue that fqdn is not exposed which is tracked in https://github.com/mondoohq/cnquery/issues/2222

chris-rock (approved):
Thank you @arlimus

PR #2226, opened 2023-10-15
🐛 fix platform detection override for lr docs

chris-rock:
This is fixing an oversight that was introduced in https://github.com/mondoohq/cnquery/pull/2102. Maps do not have a consistent order in Go. Therefore the override was not stable. This adjust the code that platform detection works fully deterministic. This was not a problem for any other than terraform resources. Terraform uses `terraform` for hcl and `terraform.plan` as well as `terraform.state`. Therefore we cannot simply check for a prefix. Terraform hcl should better be called `terraform.hcl` but that is for another PR.


arlimus (approved):
Ohhhhh that's how that happened!

PR #2042, opened 2023-10-02
[WIP] ⭐️ atlassian provider

mariuskimmina:
Work in progress

PR #2253, opened 2023-10-16
⭐️ atlassian provider

mariuskimmina:
Reopening #2042 from a branch instead of a fork


Makefile
```diff
@@ -313,6 +314,9 @@ providers/build/aws: providers/lr
 providers/install/aws:
 	@$(call installProvider, providers/aws)
 
+providers/build/atlassian: providers/lr
```
imilchev, line 318: can you add the `providers/install/atlassian` target too

providers/atlassian/connection/admin/connection.go
```diff
@@ -0,0 +1,85 @@
...
+type AdminConnection struct {
+	id     uint32
+	Conf   *inventory.Config
+	asset  *inventory.Asset
+	client *admin.Client
+	host   string
+}
+
+func NewConnection(id uint32, asset *inventory.Asset, conf *inventory.Config) (*AdminConnection, error) {
+	adminToken := conf.Options["admintoken"]
```
imilchev: this won't be set. Same applies to the other options you assume in the other connection types. You should define the flags that are needed for these and handle them

imilchev (commented):
Looks good! Almost there. Added a few small things, also need to check the failing Github actions 🙌

providers/atlassian/connection/admin/connection.go
mariuskimmina: They should be working now

providers/atlassian/connection/admin/connection.go
```diff
@@ -0,0 +1,86 @@
...
+	if adminToken == "" {
+		adminToken = os.Getenv("ATLASSIAN_ADMIN_TOKEN")
+	}
+	if adminToken == "" {
+		return nil, errors.New("you need to provide atlassian admin token via ATLASSIAN_ADMIN_TOKEN env")
```
imilchev: this error is no longer correct. Either the flag or the var needs to be set. Make sure that's updated for the other flags too

providers/atlassian/config/config.go
```diff
@@ -0,0 +1,63 @@
...
+	Connectors: []plugin.Connector{
+		{
+			Name:      "atlassian",
+			Use:       "atlassian",
+			Short:     "atlassian",
+			MaxArgs:   2,
+			Discovery: []string{},
+			Flags: []plugin.Flag{
+				{
+					Long:    "admintoken",
```
imilchev: to follow the convention set by the other providers this should be `admin-token`

providers/atlassian/config/config.go
```diff
@@ -0,0 +1,63 @@
...
+					Desc:    "Provide atlassian hostname (e.g. https://example.atlassian.net).",
+				},
+				{
+					Long:    "user",
+					Type:    plugin.FlagType_String,
+					Default: "",
+					Desc:    "Provide atlassian user name (e.g. example@example.com).",
+				},
+				{
+					Long:    "usertoken",
```
imilchev: `user-token`

providers/atlassian/config/config.go
```diff
@@ -0,0 +1,63 @@
...
+					Type:    plugin.FlagType_String,
+					Default: "",
+					Desc:    "Provide atlassian user api token (used for jira / confluence).",
+				},
+				{
+					Long:    "scimtoken",
```
imilchev: `scim-token`

Makefile
mariuskimmina: It's there

providers/atlassian/provider/provider.go
```diff
@@ -0,0 +1,242 @@
...
+		conf.Options["user"] = string(x.Value)
+	}
+	if x, ok := flags["host"]; ok && len(x.Value) != 0 {
+		conf.Options["host"] = string(x.Value)
+	}
+	if x, ok := flags["admintoken"]; ok && len(x.Value) != 0 {
+		conf.Options["admintoken"] = string(x.Value)
+	}
+	if x, ok := flags["scimtoken"]; ok && len(x.Value) != 0 {
+		conf.Options["scimtoken"] = string(x.Value)
```
imilchev: you can also call the option the same way as the flag, so `scim-token`

providers/defaults.go
```diff
@@ -46,6 +46,18 @@ var DefaultProviders Providers = map[string]*Provider{
 			},
 		},
 	},
+	"atlassian": {
+		Provider: &plugin.Provider{
+			Name:            "atlassian",
+			ConnectionTypes: []string{"atlassian"},
```
imilchev: the types should be set here too

PR #2280, opened 2023-10-17
⭐ add run info

arlimus:
Add a new info functionality to the run-command. This collects stats about aquery you are interested in. For example: it collects information about the resources and fields that the query is going to execute:

```bash
> cnquery run -c "sshd.config.params[Version] == mondoo.version" --info
```

which returns:

```bash
Resources and Fields used:
- sshd.config
  - params
- mondoo
  - version
```

We can extend it to more information in the future.


chris-rock (approved):
Very nice

PR #2285, opened 2023-10-18
🐛 ensure that atlassian connection authenticated successfully

### LINKED ISSUE #2268: atlassian provider do permission checks properly
When using the provider without access to the jira instance it returns empty results instead of saying something like "not authorized"

image

mariuskimmina:
Closes #2268


chris-rock (approved):
Works as advertised! Thank you @mariuskimmina

PR #2288, opened 2023-10-18
✨ ensure that the core provider overrides

arlimus:
The core provider is the only special provider in our current list, treated with higher priority.

We are looking to be more explicit about the order in which providers are added and how things are getting merged together. Until we get there, we can force core to at least not be overwritten by another provider.


chris-rock (approved):
Thank you @arlimus

PR #2289, opened 2023-10-18
🐛 use NewResource to load init functions in core provider

chris-rock:
The init functions have not been called for the core provider because we used CreateResource instead of NewResource


arlimus (approved):
noice catch!

PR #2293, opened 2023-10-19
🐛 ensure ebs connection config has type filesystem

vjeffrey:
picked up this hint from https://github.com/mondoohq/cnquery/pull/2171

just trying to get the small we-definitely-want-this pieces in to try to break down this ebs scan issue a bit

### DIFF
--- a/providers/aws/connection/awsec2ebsconn/provider.go
+++ b/providers/aws/connection/awsec2ebsconn/provider.go
@@ -173,6 +173,7 @@ func NewAwsEbsConnection(id uint32, conf *inventory.Config, asset *inventory.Ass
 
 	// Create and initialize fs provider
 	fsConn, err := connection.NewFileSystemConnection(id, &inventory.Config{
+		Type:       "filesystem",
 		Path:       volumeMounter.ScanDir,
 		PlatformId: conf.PlatformId,
 		Options:    conf.Options,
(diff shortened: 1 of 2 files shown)


providers/os/connection/snapshot/blockdevices.go
```diff
@@ -112,7 +112,7 @@ func (blockEntries blockDevices) GetUnmountedBlockEntry() (*fsInfo, error) {
 }
 
 func (entry blockDevice) IsNoBootVolume() bool {
-	return entry.Uuid != "" && entry.FsType != "" && entry.FsType != "vfat" && entry.Label != "EFI"
+	return entry.Uuid != "" && entry.FsType != "" && entry.FsType != "vfat" && entry.Label != "EFI" && entry.Label != "boot"
```
vjeffrey, line 115: this is the reason my tests kept failing so much - i was testing against a rhel machine and i was picking up the boot volume, which we would then of course fail to mount and everything would go a little haywire

providers/aws/connection/awsec2ebsconn/provider.go
```diff
@@ -173,6 +173,7 @@ func NewAwsEbsConnection(id uint32, conf *inventory.Config, asset *inventory.Ass
 
 	// Create and initialize fs provider
 	fsConn, err := connection.NewFileSystemConnection(id, &inventory.Config{
+		Type:       "filesystem",
```
vjeffrey, line 186: thanks to @czunker for the hint on this one

PR #2205, opened 2023-10-13
🐛 Fix vSphere platform vulnerabilities and eol

czunker:

PR #2301, opened 2023-10-19
🐛 Fix vSphere platform vulnerabilities & support eol for non-os platforms

chris-rock:
superseeds #2205

This change implements vulnerabilities for vsphere by implementing:

```
extend asset {
  vulnerabilityReport() dict
}
```

In the vsphere provider.

Previously `asset.eol` was part of the os provider but is actually independent. This change moves asset.eol to core.

It required some changes to work:

- support init functions in core provider https://github.com/mondoohq/cnquery/pull/2289
- ensure resources from core provider cannot be overwritten https://github.com/mondoohq/cnquery/pull/2288

TODO:

- [ ] do some extra testing with cnspec
- [ ] ensure the migration is smooth since there is the edge case where older cnspec which do not have assset.eol in core pull an updated os provider. Upgrading to latest cnspec will fix it then but we need to think about a migration. We could test a scenario where both os and core provider have asset.eol. It may be easier to bite it and ask users to upgrade cnspec then


providers/vsphere/resources/vsphere.lr.manifest.yaml
```diff
@@ -105,6 +109,16 @@ resources:
       name:
       - vmware-esxi
       - vmware-vsphere
+  platform:
```
czunker, line 112: I do not understand why these are showing up here. They aren't part of the `vsphere.lr` file.

czunker:
I gave `asset.eol` a try and do not get data:
```
cnquery run vsphere christian@vsphere.local@192.168.5.24 -p XXX --config ~/workspace/mondoo/sa-edge-default.json -c "asset.eol"
→ loaded configuration from /home/christian/workspace/mondoo/sa-edge-default.json using source --config
To use this resource, you must authenticate with Mondoo Platform.
To learn how, read:
asset.eol: no data available
To use this resource, you must authenticate with Mondoo Platform.
To learn how, read:
asset.eol: no data available
To use this resource, you must authenticate with Mondoo Platform.
To learn how, read:
asset.eol: no data available
To use this resource, you must authenticate with Mondoo Platform.
To learn how, read:
asset.eol: no data available
```
I rebuild cnquery, os, and vsphere. Core as builtin should have been rebuild with cnquery.

`asset.vulnerabilityReport` is working.

providers/vsphere/resources/vsphere.lr.manifest.yaml
chris-rock: They need to be removed here

chris-rock:
More work has been extracted from here since we need to tackle a few more detailed issues

- EOL handling is here https://github.com/mondoohq/cnquery/pull/2362
- Vuln handling is here https://github.com/mondoohq/cnquery/pull/2363

Since all work has been extracted into individual PRs, we close this one here

PR #2305, opened 2023-10-20
🐛 provider installation should refresh schema

arlimus:
Extensible schemas are used to go across multiple providers and aggregate a schema. We use them extensively, in order to load all available resources and fields. However, if a new provider is installed after a schema has been fully loaded, it won't show up, because the schema doesn't know it has to refresh.

This is now fixed. We keep track of the last time a provider was installed to the system. Whenever that happens, a timestamp is updated. Each extensible schema keeps track of this timestamp. If it changes, the schema gets refreshed.

### DIFF
--- a/providers/extensible_schema.go
+++ b/providers/extensible_schema.go
@@ -13,11 +13,11 @@ import (
 type extensibleSchema struct {
 	resources.Schema
 
-	loaded    map[string]struct{}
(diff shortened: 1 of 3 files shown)


chris-rock (approved):
Nice improvement @arlimus

PR #2311, opened 2023-10-20
🐛 Fix k8s manifests platformIDs and asset name

### LINKED ISSUE #2302: `scan k8s <manifest-file>` does not work for some files
**Describe the bug**
```
cnquery scan k8s .github/test_files/k8s-manifest.yaml --verbose                                                                                                                                                                                                                                                            
DBG using provider k8s with connector k8s
DBG Started a new runtime (1 total)
DBG running provider plugin path=/home/christian/.config/mondoo/providers/k8s/k8s
DBG Started a new runtime (2 total)
DBG Started a new runtime (3 total)
DBG resolved 2 assets
DBG Started a new runtime (4 total)
DBG Started a new runtime (5 total)
→ synchronize assets
DBG initialize client authentication issuer=mondoo/ams kid=//agents.api.mondoo.app/spaces/dazzling-golick-767384/serviceaccounts/2AyMRoi8pfJa6gOqq16QWstmKPK
DBG non-ok http request body="could not sync assets" status=500
FTL failed to run scan error="rpc error: code = Internal desc = could not sync assets"
```
The same happens for cnspec.

This is the used file: https://github.com/mondoohq/actions/blob/main/.github/test_files/k8s-manifest.yaml

Bug was observed here: https://github.com/mondoohq/actions/actions/runs/6571602900/job/17851138836?pr=73

But also happens when running locally.

**To Reproduce**

1. Run above command with the **mentioned** file
2. Note the error

**Expected behavior**
The manifest file should get scanned.

- cnquery/cnspec v9.2.2
- k8s 9.0.13 with connectors: k8s

**Additional context**
I gave it a try with another file and that worked. Will add it to a comment.

An `incognito` scan of the actions test file works:
```
cnspec scan k8s .github/test_files/k8s-manifest.yaml --incognito                                                                                                                                                                                                                                                            → loaded configuration from /etc/opt/mondoo/mondoo.yml using source default

...

Asset: mondoo-operator/mondoo-operator-controller-manager
---------------------------------------------------------

Checks:
✓ Pass:  A 100  Container should have a memory limit
✓ Pass:  A 100  Container should request CPU
✓ Pass:  A 100  Container should configure a livenessProbe
✓ Pass:  A 100  Pods should not run Kubernetes dashboard
✓ Pass:  A 100  Container should not mount the Docker socket
✓ Pass:  A 100  Deployments should not run in the default namespace
✓ Pass:  A 100  Deployments should not run Tiller (Helm v2)
✓ Pass:  A 100  Container should not mount the containerd socket
✓ Pass:  A 100  Container should configure a readinessProbe
✓ Pass:  A 100  Container should not run as root
✓ Pass:  A 100  Container should not allow privilege escalation
✓ Pass:  A 100  Deployments should not bind to a host port
✕ Fail:  C  40  Container image pull should be consistent
✓ Pass:  A 100  Container should not mount the CRI-O socket
✓ Pass:  A 100  Pod should not run with the default service account
✓ Pass:  A 100  Pod should not define hostAliases
✓ Pass:  A 100  Deployments should not run with SYS_ADMIN capability
✓ Pass:  A 100  Deployments should mount any host path volumes as read-only
✓ Pass:  A 100  Container should use an immutable root filesystem
✓ Pass:  A 100  Pod should not run with hostPID
✕ Fail:  D  20  Deployments should not run with NET_RAW capability
✓ Pass:  A 100  Container should request memory
✓ Pass:  A 100  Container should not run as a privileged container
✓ Pass:  A 100  Pod should not run with hostNetwork
✓ Pass:  A 100  Container should have a CPU limit
✓ Pass:  A 100  Pod should not run with hostIPC

Vulnerabilities:
error: Could not find the vulnerability report.

Scanned 2 assets

Kubernetes Deployment
    B mondoo-operator/mondoo-operator-controller-manager

Kubernetes Manifest
    X 

Summary
=======

Score Distribution		Asset Distribution
------------------		------------------
A   0 assets      		Kubernetes Manifest     1
B   1 assets      		Kubernetes Deployment   1
C   0 assets
D   0 assets
F   0 assets
X   1 assets
```

czunker:
Fixes #2302


providers/k8s/connection/shared/connection.go
```diff
@@ -63,56 +64,6 @@ type ResourceResult struct {
 	AllNs     bool
 }
 
-func getPlatformInfo(objectKind string, runtime string) *inventory.Platform {
```
czunker, line 66: This wasn't referenced anywhere, so I removed it.

providers/k8s/connection/manifest/connection.go
```diff
@@ -123,11 +130,11 @@ func (c *Connection) AssetId() (string, error) {
 	// the same resource multiple times but it will result in different assets because of the random
 	// file name.
 
-	if len(c.Objects) == 1 {
+	if len(c.Objects) == 1 && c.asset.Platform.Runtime == "k8s-admission" {
```
czunker, line 139: @imilchev Please double-check.

This was also broken in v8. When we have a manifest file with only one object in it, we would use the uid. But as I understand the comment, this was only meant to be used with Admission scans.

v8:
```
  ids: [
    0: "//platformid.api.mondoo.app/runtime/k8s/uid/3eae16c2-2b55-4048-a84b-03a849d12e40"
  ]
```

v9 with this PR:
```
  ids: [
    0: "//platformid.api.mondoo.app/runtime/k8s/uid/044dbf3db4656a3863d97901ab262e11d184641014d4cb10bf6d57c78da8f437"
  ]
```

chris-rock:
Can we add test cases with sample manifests to determine the platform id?

czunker:
> Can we add test cases with sample manifests to determine the platform id?

Would like to. We need to figure out how to add back the tests we had with v8: https://github.com/mondoohq/cnquery/issues/2157

providers/k8s/connection/manifest/connection.go
imilchev: your change is the correct behaviour here! nice catch!

imilchev (approved):
lgtm

chris-rock:
I cannot reproduce this behavior since I am missing sample data, can you add sample manifests into a testdata directory so that this PR can be verified?

chris-rock:
I added a manifest and a test case and fixed the assignment of the platform id to the asset object that is returned during sync.

PR #2312, opened 2023-10-20
✨ Rename recording to AssetRecording, allow overwriting of asset.

preslavgerchev:
We already have predefined recordings, which are great, but they cannot be modified in any way. 

For example, assume there's a test that has a policy with a filter that looks like `asset.name == 'my-test'`. We could load a recording and all we'd need to do is change the asset's name to match that policy and start asserting against it.

This PR exposes the internal `recording` by renaming it to `AssetRecording`. It also exposes a few more functions in the test utils to return just the recording, not the runtime. This allows for something like
```
myAsset := &inventory.Asset{Name: "my-test"}
recording := testutils.LinuxAssetRecording()
recording.SetAsset(myAsset)
runtime := testutils.MockFromRecording(recording)
```

### DIFF
--- a/providers-sdk/v1/testutils/testutils.go
+++ b/providers-sdk/v1/testutils/testutils.go
@@ -233,9 +233,36 @@ func mockRuntime(testdata string) llx.Runtime {
 	return mockRuntimeAbs(filepath.Join(TestutilsDir, testdata))
 }
 
-func mockRuntimeAbs(testdata string) llx.Runtime {
+func MockFromRecording(recording providers.Recording) llx.Runtime {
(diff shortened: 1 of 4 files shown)


arlimus:
I promise I'll get it done in the morning, so sorry for being slow :see_no_evil:

preslavgerchev:
I also opened https://github.com/mondoohq/cnquery/pull/2323 which I think is a bit simpler approach, does not change that many things, sorry for pushing this that late!

preslavgerchev:
Closing in favor of #2323

PR #2320, opened 2023-10-21
🐛 fix aws platform id

chris-rock:
Under certain circumstances the discovery has not overwritten the platform id.


arlimus (approved):
uh, nice catch!

PR #2321, opened 2023-10-21
🐛 Change data conversion to allow for nil resources. Fix azure monitor as an example

preslavgerchev:
Some of the resources we try to fetch may not be present, consider this as an example:
```

// Azure Monitor Diagnostic Setting
private azure.subscription.monitorService.diagnosticsetting @defaults("id name") {
  .....
  // Diagnostic setting storage account
  storageAccount() azure.subscription.storageService.account
}
```

When computing the storage account it is totally possible that this resource is not present. Until now, trying to return a nil resource resulted in a panic which was fixed by returning an error, something like
```
return nil, errors.New("storage account not present for the diagnostic setting resource")
```
which led to
```
cnquery> azure.subscription.monitor.diagnosticSettings{storageAccount}
1 error occurred:
        * diagnostic settings has no storage account
azure.subscription.monitor.diagnosticSettings: [
  0: {
    storageAccount: diagnostic settings has no storage account
  }
]
```

The problem with this is that there's not really an error, it's just that the storage account isn't present.

To fix this, I saw that we have a `plugin.NotReady` error which seemed like a good indicator of this use case. This now results in
```
cnquery> azure.subscription.monitor.diagnosticSettings{storageAccount}
azure.subscription.monitor.diagnosticSettings: [
  0: {
    storageAccount: null
  }
]
```

which will allow for better assertions

Chaining on null resources remains functional:
```
cnquery> azure.subscription.monitor.diagnosticSettings{storageAccount.id}
azure.subscription.monitor.diagnosticSettings: [
  0: {
    storageAccount.id: null
  }
]
```

### DIFF
--- a/providers-sdk/v1/plugin/runtime.go
+++ b/providers-sdk/v1/plugin/runtime.go
@@ -239,9 +239,15 @@ func GetOrCompute[T any](cached *TValue[T], compute func() (T, error)) *TValue[T
 
 	x, err := compute()
 	if err != nil {
-		res := &TValue[T]{Data: x, Error: err}
-		if err != NotReady {
-			res.State = StateIsSet | StateIsNull
+		res := &TValue[T]{State: StateIsSet}
+		// Note: if we're getting back a NotReady error, it means the computed
+		// resource is not present and we should return nil.
+		// To avoid panics down the stack, we only set the data and err if it's anything other than a NotReady error
+		if err == NotReady {
+			res.State |= StateIsNull
+		} else {
+			res.Data = x
+			res.Error = err
 		}
 		return res
 	}
--- a/providers-sdk/v1/plugin/runtime.go
+++ b/providers-sdk/v1/plugin/runtime.go
@@ -239,13 +239,12 @@ func GetOrCompute[T any](cached *TValue[T], compute func() (T, error)) *TValue[T
 
 	x, err := compute()
 	if err != nil {
-		res := &TValue[T]{State: StateIsSet}
+		// why do we always set to IsNull here?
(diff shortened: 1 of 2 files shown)


chris-rock:
I am not sure I understand why we are not just return nil if the resource does not exist. We use the return nil behavior in other places.

preslavgerchev:
> I am not sure I understand why we are not just returning nil if the resource does not exist. We use the return nil behavior in other places.

`return nil, nil` does not work, it panics. It's a long stack but summarized: the code in `GetOrCompute` (plugin/runtime.go) will assume there's no error and return a `TValue` object where both `Data` and `Error` are nil. The conversion to `DataRes` then goes wrong as it's thinking there's something inside the struct

arlimus:
> > I am not sure I understand why we are not just returning nil if the resource does not exist. We use the return nil behavior in other places.
> 
> `return nil, nil` does not work, it panics. It's a long stack but summarized: the code in `GetOrCompute` (plugin/runtime.go) will assume there's no error and return a `TValue` object where both `Data` and `Error` are nil. The conversion to `DataRes` then goes wrong as it's thinking there's something inside the struct

I have an example in a few other places: you set the field locally and then you can return nil. 

Here is an example from GCP storage buckets. It both sets the value and returns `nil, nil`:

```go
	g.StorageBucket.State = plugin.StateIsSet | plugin.StateIsNull
	return nil, nil
```

See https://github.com/mondoohq/cnquery/blob/main/providers/gcp/resources/logging.go#L324-L325

I'm closing this PR for now just to make sure we don't accidentally merge it, I think it has some far-reaching consequences.

PR #2322, opened 2023-10-22
🐛 fromrecording: check for 0 length typ before calling containsresource

vjeffrey:
saw this when working on some tests that use  a recording

### DIFF
--- a/providers-sdk/v1/plugin/runtime.go
+++ b/providers-sdk/v1/plugin/runtime.go
@@ -44,6 +44,9 @@ func (r *Runtime) ResourceFromRecording(name string, id string) (map[string]*llx
 	// recording and initialize them.
 	// TODO: we could use the provided information for a later request.
 	for k, v := range data.Fields {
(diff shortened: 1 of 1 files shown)


providers-sdk/v1/plugin/runtime.go
```diff
@@ -44,6 +44,9 @@ func (r *Runtime) ResourceFromRecording(name string, id string) (map[string]*llx
 	// recording and initialize them.
 	// TODO: we could use the provided information for a later request.
 	for k, v := range data.Fields {
+		if len(v.Data.Type) == 0 {
```
vjeffrey, line 47: without this we would land in IsArray, which would read typ[0], which would in turn panic

arlimus:
Do you by any chance remember what produced those recordings?
I think we'll want to produce some form of warnings (makes searching easier if something got filtered) and probably want to fix the underlying problem

chris-rock:
Closing this as its not the right fix for the problem.

PR #2323, opened 2023-10-23
✨ Allow creating of a recording from just a single asset.

preslavgerchev:
A bit simpler approach, compared to #2312 

We already have predefined recordings, which are great, but they cannot be modified in any way. 

This PR lets us create a recording from a simple asset, containing just that asset. It holds no resources/connections on the side, although we can extend

```
myAsset := &inventory.Asset{Name: "my-test"}
recording := testutils.RecordingFromAsset(myAsset)
runtime := testutils.MockFromRecording(recording)
```


arlimus:
I like this approach a lot more, love the simplified interface so we can separate testing from regular interactions.

arlimus (approved):
Great job on these improvements and yay for exposing this capability in the tests!!

PR #2324, opened 2023-10-23
✨ be more lenient with errors on registrykey

arlimus:
Adjust the behavior of registry key to be more in line with resources like package, service, and user. If a key is not found, return the empty resource instead of an error, with fields that are null.

Additionally:
- align registrykey getEntries with the exists call
- exit out of registrykey props init earlier
- initialize properties with empty values if requested


chris-rock (approved):
@arlimus tested this and it works great with larger benchmarks

PR #2325, opened 2023-10-23
🐛 Support protobuf string values for VaultType.

preslavgerchev:
In v8, the inventory proto had its own `VaultConfiguration`
```
message VaultConfiguration {
  string name = 1;
  string type = 2;
  map<string, string> options = 3;
}
```

Note the type here is a string. This meant that we'd usually pass in the string representation of the enum, which is using the auto-generated map, e.g.
```
VaultType_name = map[int32]string{
		0: "None",
		1: "KeyRing",
		2: "LinuxKernelKeyring",
		3: "EncryptedFile",
		4: "HashiCorp",
		5: "GCPSecretsManager",
		6: "AWSSecretsManager",
		7: "AWSParameterStore",
		8: "GCPBerglas",
		9: "Memory",
	}
```

In v9, we've gotten rid of the the duplicate `VaultConfiguration` in the inventory.proto and it now directly uses the proper config, which is defined (in `vault.proto`) as:
```

message VaultConfiguration {
  string name = 1;
  VaultType type = 2;
  map <string, string> options = 3;
}
```

Note the type here is correct, it references to the enum type instead of a string. To support conversion from v8, extend the 
`NewVaultType` func to also consider the auto-generated protobuf string map

### DIFF
--- a/providers-sdk/v1/vault/config.go
+++ b/providers-sdk/v1/vault/config.go
@@ -9,19 +9,32 @@ import (
 	"strings"
 )
 
+// supports both the protobuf string and the custom defined string
+// representations of a vault type
 func NewVaultType(name string) (VaultType, error) {
 	entry := strings.TrimSpace(strings.ToLower(name))
 	var code VaultType
-	ok := false
+	marshalMapOk := false
 	for k := range vaultMarshalNameMap {
 		if vaultMarshalNameMap[k] == entry {
-			ok = true
+			marshalMapOk = true
 			code = k
 			break
 		}
 	}
-	if !ok {
-		return VaultType_None, errors.New("unknown type value: " + string(name))
+	if !marshalMapOk {
+		// also support the auto-generated protobuf string values for all the enum values
+		protoMapOk := false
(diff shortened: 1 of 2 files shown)


providers-sdk/v1/vault/config.go
```diff
@@ -9,19 +9,32 @@ import (
...
 		if vaultMarshalNameMap[k] == entry {
-			ok = true
+			marshalMapOk = true
 			code = k
 			break
 		}
 	}
-	if !ok {
-		return VaultType_None, errors.New("unknown type value: " + string(name))
+	if !marshalMapOk {
```
chris-rock, line 25: Do we then still need vaultMarshalNameMap? Or can we simplify the code?

preslavgerchev: Looks like we're still using this in the `vault`cmds, e.g.  `vault list`

I think we should leave it to support inventory files where the type was defined with this custom map

chris-rock (approved):
Thank you @preslavgerchev

PR #2332, opened 2023-10-23
🧹 Prevent nil panic when recording is not set

czunker:
In case recording isn't specififed during scanner initialisation, set the NullRecording


chris-rock:
followup for https://github.com/mondoohq/packer-plugin-cnspec/pull/132

chris-rock (approved):
Thank you @czunker

PR #2304, opened 2023-10-20
🐛 Use configured api-proxy

czunker:
Fixes #2277

PR #2343, opened 2023-10-23
🐛 Use configured api-proxy for eol and vulnReport

### LINKED ISSUE #2341: cnquery and cnspec not fully working behind proxy
i have tested it with the following env https://github.com/mondoohq/samples/tree/main/aws/ec2-instance-proxy

I have compiled the cnspec gateway from main:

```bash
./cnspec version
cnspec v9.2.3 (f21f22c, unknown)
```

cnspec login command

```bash
./cnspec login --token $MONDOO_REGISTRATION_TOKEN --api-proxy http://10.0.101.144:3128 --config /etc/opt/mondoo/mondoo.yml
```

I get the following error:

```bash
groups.where.list: []
error: rpc error: code = Unknown desc = failed to do request: Post "https://us.api.mondoo.com/AdvisoryScanner/IsEol": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
platform.eol.date: null



asset.platform: "debian"
error: failed to do request: Post "https://us.api.mondoo.com/AdvisoryScanner/AnalyseAsset": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
platform.vulnerabilityReport: null
```

czunker:
Fixes #2341

### DIFF
--- a/providers/os/resources/asset_eol.go
+++ b/providers/os/resources/asset_eol.go
@@ -66,8 +66,20 @@ func initAssetEol(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[st
 
 	// get new advisory report
 	// start scanner client
+	upstreamClient := mcc
+	if mcc.UpstreamConfig.ApiProxy != "" {
+		// The proxy config is part of the http client, but only as pointer
+		// This get's not transfered into the plugin.
+		// Because of that, we have to re-create the http client here, when we use an API proxy
+		log.Info().Str("api_proxy", mcc.UpstreamConfig.ApiProxy).Msg("use api proxy")
+		var err error
+		upstreamClient, err = mcc.UpstreamConfig.InitClient()
+		if err != nil {
+			return nil, nil, err
+		}
+	}
 
-	scannerClient, err := newAdvisoryScannerHttpClient(mcc.ApiEndpoint, mcc.Plugins, mcc.HttpClient)
+	scannerClient, err := newAdvisoryScannerHttpClient(mcc.ApiEndpoint, mcc.Plugins, upstreamClient.HttpClient)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/providers/os/resources/platform_advisories.go
+++ b/providers/os/resources/platform_advisories.go
@@ -61,7 +61,20 @@ func fetchVulnReport(runtime *plugin.Runtime) (interface{}, error) {
 
 	// get new advisory report
 	// start scanner client
-	scannerClient, err := newAdvisoryScannerHttpClient(mcc.ApiEndpoint, mcc.Plugins, mcc.HttpClient)
+	upstreamClient := mcc
+	if mcc.UpstreamConfig.ApiProxy != "" {
+		// The proxy config is part of the http client, but only as pointer
+		// This get's not transfered into the plugin.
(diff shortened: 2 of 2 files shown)


czunker:
This is a follow-up to #2304 

This code snippet was removed from https://github.com/mondoohq/cnquery/pull/2304#pullrequestreview-1689690298 

It looks like, the previous fix only worked for builtin but not for plugin.

We create the HTTP client with proxy config inside the cnquery binary. The proxy is set inside the Transport, which is a pointer. This pointer does not get transferred across gRPC into the plugin, where we use an HTTP client again, to talk to the API to get EOL and vulnerability data.

https://github.com/mondoohq/cnquery/pull/2301 might make this unnecessary, because there we move these resources into core, which is a builtin.

providers/os/resources/asset_eol.go
```diff
@@ -66,8 +66,20 @@ func initAssetEol(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[st
 
 	// get new advisory report
 	// start scanner client
+	upstreamClient := mcc
+	if mcc.UpstreamConfig.ApiProxy != "" {
+		// The proxy config is part of the http client, but only as pointer
+		// This get's not transfered into the plugin.
+		// Because of that, we have to re-create the http client here, when we use an API proxy
+		log.Info().Str("api_proxy", mcc.UpstreamConfig.ApiProxy).Msg("use api proxy")
```
chris-rock, line 74: I think this is to late. We do not want to handle this here? Can we not do this when we init upstream?

providers/os/resources/platform_advisories.go
```diff
@@ -61,7 +61,20 @@ func fetchVulnReport(runtime *plugin.Runtime) (interface{}, error) {
...
 	// start scanner client
-	scannerClient, err := newAdvisoryScannerHttpClient(mcc.ApiEndpoint, mcc.Plugins, mcc.HttpClient)
+	upstreamClient := mcc
+	if mcc.UpstreamConfig.ApiProxy != "" {
+		// The proxy config is part of the http client, but only as pointer
+		// This get's not transfered into the plugin.
+		// Because of that, we have to re-create the http client here, when we use an API proxy
+		log.Info().Str("api_proxy", mcc.UpstreamConfig.ApiProxy).Msg("use api proxy")
+		var err error
+		upstreamClient, err = mcc.UpstreamConfig.InitClient()
```
vjeffrey, line 71: i was trying to think of a way to unit test this but i can't think of anything...we should add some integration test for proxy support in the near future - let's get that added to our testing card

providers/os/resources/asset_eol.go
czunker: Looks like my assumption was wrong. We already do an `InitClient` inside the provider plugin: https://github.com/mondoohq/cnquery/blob/main/providers/os/provider/provider.go#L442

That should also set the proxy, but I double-checked, and when I don't have this fix in the provider, I get a timeout.

providers/os/resources/platform_advisories.go
czunker: Already done.

czunker:
I can no longer reproduce the error.

See https://github.com/mondoohq/cnquery/issues/2341#issuecomment-1776821129 for details.

PR #2345, opened 2023-10-24
🧹 update go dependencies

chris-rock:
This updates the go dependencies in cnquery itself


providers/github/go.mod
```diff
@@ -42,20 +42,20 @@ require (
 	github.com/alingse/asasalint v0.0.11 // indirect
 	github.com/ashanbrown/forbidigo v1.6.0 // indirect
 	github.com/ashanbrown/makezero v1.1.1 // indirect
-	github.com/aws/aws-sdk-go v1.45.26 // indirect
+	github.com/aws/aws-sdk-go v1.46.2 // indirect
```
czunker, line 45: Not specific to this PR, but do we know why we have aws and k8s in here?

czunker (approved):
LGTM. Thanks @chris-rock

providers/github/go.mod
chris-rock: That is caused by os provider requiring those sdks for platform detection, eg. see https://github.com/mondoohq/cnquery/blob/main/providers/os/id/awsec2/awsec2.go

PR #2348, opened 2023-10-24
🧹 simplify code that communicates with mvd

chris-rock:
Extracted from #2301 to focus 2301 on the resource optimization. This is purely refactoring and simplifying the code. No functional change.


czunker:
eol looks the same:
```
cnquery run -c "asset.eol platform.eol mondoo.eol.date"                                                                                                                                                                                                   
platform.eol: platform.eol date=2027-04-01 02:00:00 +0200 CEST
asset.eol: asset.eol date=2027-04-01 02:00:00 +0200 CEST
mondoo.eol.date: Never
```

`asset.vulnerabilityReport.advisories.length` and `platform.vulnerabilityReport.advisories.length` also return the same as before.

providers/os/resources/asset_eol.go
```diff
@@ -54,51 +83,51 @@ func initPlatformEol(runtime *plugin.Runtime, args map[string]*llx.RawData) (map
 
 // ^^
 
-func initAssetEol(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error) {
-	conn := runtime.Connection.(shared.Connection)
-	platform := conn.Asset().Platform
-	eolPlatform := convertPlatform2VulnPlatform(platform)
+// FIXME: DEPRECATED, update in v10.0 vv
```
czunker, line 86: Do we need to add a note in the lr file? For `platform.eol`, we have:
```
// Information about the platform end-of-life. Deprecated in favor of `asset.eol`. This resource will be removed in version 10.0.
// Please use asset.eol instead.
```

chris-rock: We have that already added to platform eol resource.

czunker (approved):
Thanks @chris-rock

PR #2357, opened 2023-10-25
🐛 avoid TLS hangups and errors

arlimus:
TLS checks have a tendency to timeout and hang for a long time, especially when there is nothing listening on a port at all.

This PR introduces a 2sec timeout on all initial OSI L4 connections to make sure the endpoint actually exists.

Additionally it now avoid printing a lot of very cryptic errors for TLS. Instead, this mechanism now returns `null` data if no TLS was detected for any of its fields.


chris-rock (approved):
Adding a timeout and improving the output is amazing

PR #2359, opened 2023-10-25
🐛 fix TLS on open endpoint without TLS

arlimus:
After https://github.com/mondoohq/cnquery/pull/2358

A valid endpoint with TLS returns very cryptic error messages.

We already handled the case of having no open endpoint in https://github.com/mondoohq/cnquery/pull/2357 . This extends the approach to also work for open endpoints that are not TLS.

OLD:

```coffee
> tls { socket versions }
1 error occurred:
        * 1 error occurred:
        * failed to get a TLS response
tls: {
  versions: 1 error occurred:
        * failed to get a TLS response

  socket: socket protocol="tcp" port=80 address="mondoo.com"
}
```

NEW

```coffee
> tls { socket versions }
tls: {
  versions: null
  socket: socket protocol="tcp" port=80 address="mondoo.com"
}
```


chris-rock (approved):
Thank you

PR #2360, opened 2023-10-25
🐛 compile variant queries in bundles earlier

arlimus:
Technically adding the filters should happen after the variant queries have been compiled, not before. This aligns with cnspec.

Partly related to the issue below vv
Fixes https://github.com/mondoohq/cnquery/issues/2279


chris-rock:
I added variant bundle test case to load the bundle. We can extend the compile testing in future iterations.

chris-rock (approved):

PR #2361, opened 2023-10-25
🧹 introduce structured mqlc errors

arlimus:
These help us communicate when resources or fields have not been found. They are indications that can be used to infer that providers may be missing on the system.


chris-rock (approved):
Love it. More structured errors

PR #2362, opened 2023-10-25
🧹 move asset.eol to core provider

chris-rock:
This change is an extraction from https://github.com/mondoohq/cnquery/pull/2301. The goal of this change is to make sure the `asset.eol` resource is available for all providers. To achieve this we need to move the resource into the core provider instead of the provider.

To ensure we have a smooth migration for older cnquery v9 versions we cannot simply remove the `asset.eol` resource from the os provider. With https://github.com/mondoohq/cnquery/pull/2288 we make sure that resources from core always have priority and cannot be overwritten by other providers. Therefore we have the implementation for `asset.eol` in both core and os provider for a while.

TODO:

- [x] Seems like the priority is not properly handled yet 

```
cnquery> asset.eol
incorrect provider for asset, not adding
```


chris-rock:
The current issue left with the PR is that the upstream configuration does not work for the core provider in the init function.

PR #2366, opened 2023-10-25
🐛 fix reading never time in raw data json

vjeffrey:


llx/rawdata_json.go
```diff
@@ -343,9 +343,10 @@ func rawDataJSON(typ types.Type, data interface{}, codeID string, bundle *CodeBu
 			return nil
 		}
 
-		// if *time == NeverPastTime || *time == NeverFutureTime {
-		// 	TODO: ... unclear
-		// }
+		if *time == NeverPastTime || *time == NeverFutureTime {
+			buf.WriteString("Never")
```
jaym: doesn't this need to be a valid json value?

vjeffrey: bleeeehhhh to json! 
updated :) 
thank you!

jaym (approved):

PR #2367, opened 2023-10-25
📄 Add equinix docs, update azure and core docs

misterpantz:


providers/azure/resources/azure.lr
```diff
@@ -48,7 +48,7 @@ azure.subscription @defaults ("name") {
   mariaDb() azure.subscription.mariaDbService
   // Cosmos DB resources in the subscription
   cosmosDb() azure.subscription.cosmosDbService
-  // Key vault resources in the subscription
+  // Vault resources in the subscription
```
tas50: Isn't this a reference to the product name not the objects

providers/azure/resources/azure.lr
```diff
@@ -938,19 +938,19 @@ private azure.subscription.storageService.account.service.properties.retentionPo
   id string
   // Number of days that metrics or logging data is retained
   retentionDays int
-  // Whether a retention policy is enabled for the Azure Storage
```
tas50: That's a brand name

providers/azure/resources/azure.lr
misterpantz:
```suggestion
  // Azure Key Vault resources in the subscription
```

providers/azure/resources/azure.lr
```diff
@@ -938,19 +938,19 @@ private azure.subscription.storageService.account.service.properties.retentionPo
...
+  // Whether a retention policy is enabled for the Azure storage
```
misterpantz:
```suggestion
  // Whether a retention policy is enabled for the Azure Storage
```

tas50 (approved):

PR #2370, opened 2023-10-26
✨ prioritize connected provider schema

arlimus:
Currently schemas are loaded in almost random order (except for the core provider). This is problematic if 2 providers provide the same resource, like `asset.eol` which is provided by both `os` and `vsphere`. Since we now have this use-case, it became more important to control the load-order of schemas.

Notes:
1. At the very start we still `loadAllSchemas`. This will be removed in the future because we really don't need all schemas at all times. We also initialize early runtimes that we tend to throw away - so all this work is wasted.
   - In this early stage we only prioritize the core provider
2. Once users connect to a provider via `runtime.Connect(..)` it sets the default provider for the runtime. It is also the signal for us to set a priority schema, based on whatever the users connected to. For example: If we connect to e.g. `vsphere` this provider gets prioritized.
   - Once prioritized, the schema is refreshed
   - During the refresh the core provider comes first, then the connecting provider.
3. Extending the schema behaved incorrectly when two non-extended resources were added. In this case the base resource survived, even though the new resource should have overwritten it (to be consistent with the remaining code).
4. We were passing pointers around for schemas, which is nice and fast, but also led to object changes that were unintended. Most notably: When re-prioritizing schemas it became evident that the original resource had changed (and adopted all the fields of merged resources). This approach is slower but safer from side-effects.

### DIFF
--- a/providers-sdk/v1/resources/schema.go
+++ b/providers-sdk/v1/resources/schema.go
@@ -3,6 +3,9 @@
 
 package resources
 
+// Add another schema and return yourself. other may be nil.
+// The other schema overrides specifications in this schema, unless
+// a resource is flagged as an extension.
 func (s *Schema) Add(other *Schema) *Schema {
 	if other == nil {
 		return s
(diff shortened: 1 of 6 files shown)


chris-rock (approved):
Great addition. Works well!

PR #2371, opened 2023-10-26
Add kind to default fields in azure cosmodb

tas50:
This gets us the DB type which was MongoDB in my testing


czunker (approved):
LGTM Thanks @tas50

PR #2372, opened 2023-10-26
⭐️ ability to fetch group roles and members

chris-rock:
This change allows users to fetch the okta groups with its roles and members:

```
cnspec> okta.groups.where(roles.one(type =="SUPER_ADMIN")) { name roles { * } members members.length < 2 }
okta.groups.where: [
  0: {
    roles: [
      0: {
        created: 2023-04-08 22:11:00 +0200 CEST
        lastUpdated: 2023-04-08 22:11:00 +0200 CEST
        assignmentType: "GROUP"
        id: "gra91lrbpr7vabz4l5d7"
        type: "SUPER_ADMIN"
        status: "ACTIVE"
        label: "Super Administrator"
      }
    ]
    name: "Super Admins"
    members.length < 2: true
    members: [
      0: okta.user profile.email="ben@example.com"
    ]
  }
]
```

You can also check which permissions custom roles have:

```
cnspec> okta.customRoles { * }
okta.customRoles: [
  0: {
    label: "Custom Role"
    id: "cr0cyddwhxapVOPrB5d7"
    description: "Custom Role"
    permissions: []
  }
]
```

This allows us to make dramatic performance improvement for the okta health sight policy https://github.com/mondoohq/cnspec-policies/blob/main/core/mondoo-okta-security.mql.yaml#L183. 

```
okta.users.where( roles.map(type).contains("SUPER_ADMIN") ).length < props.maxOktaSuperAdmins
```

Instead of iterating over all users (which is slow for large accounts), we just have to look for the group and the role.

```
okta.groups.where(roles.one(type =="SUPER_ADMIN")).all(members.length < 2 )
```

### DIFF
--- a/providers/okta/resources/okta.lr
+++ b/providers/okta/resources/okta.lr
@@ -144,7 +144,7 @@ private okta.role @defaults("label status") {
 private okta.group @defaults("name") {
   // Unique key for Group
   id string
-	// Group Name
+	// Group name
   name string
 	// Group description
 	description string
@@ -158,9 +158,9 @@ private okta.group @defaults("name") {
   lastUpdated time
   // The Group's Profile properties
   profile dict
-  // Group Members
+  // Group members
   members() []okta.user
-  // Group Roles
+  // Group roles
   roles() []okta.role
 }
 
@@ -338,7 +338,7 @@ private okta.threatsConfiguration @defaults("action") {
 
 // Okta Custom Role
 private okta.customRole @defaults("label") {
-  // Identifier of the Policy
+  // Identifier for the policy
   id string
   // Name of the Custom Role
   label string
--- a/providers/okta/resources/okta.lr
+++ b/providers/okta/resources/okta.lr
@@ -10,6 +10,8 @@ okta {
   users() []okta.user
   // Okta groups
   groups() []okta.group
+  // Okta group rules
+  groupRules() []okta.groupRule
   // Okta domains
   domains() []okta.domain
   // Okta applications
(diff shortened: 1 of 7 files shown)


providers/okta/resources/okta.lr
```diff
@@ -137,9 +141,13 @@ private okta.role @defaults("id status") {
 }
 
 // Okta Group
-private okta.group @defaults("id") {
+private okta.group @defaults("name") {
   // Unique key for Group
   id string
+	// Group Name
```
tas50:
```suggestion
	// Group name
```

providers/okta/resources/okta.lr
```diff
@@ -150,6 +158,21 @@ private okta.group @defaults("id") {
   lastUpdated time
   // The Group's Profile properties
   profile dict
+  // Group Members
```
tas50:
```suggestion
  // Group members
```

providers/okta/resources/okta.lr
```diff
@@ -150,6 +158,21 @@ private okta.group @defaults("id") {
...
+  members() []okta.user
+  // Group Roles
```
tas50:
```suggestion
  // Group roles
```

providers/okta/resources/okta.lr
```diff
@@ -310,4 +333,17 @@ private okta.threatsConfiguration @defaults("action") {
   created time
   // Timestamp when the Network Zone was last updated
   lastUpdated time
+
+}
+
+// Okta Custom Role
+private okta.customRole @defaults("label") {
+  // Identifier of the Policy
```
tas50:
```suggestion
  // Identifier for the policy
```

arlimus (approved):
Very cool!

PR #2376, opened 2023-10-26
🐛 shutdown providers

mariuskimmina:
I think in these cases we would also want to shutdown before log.fatal - we can't return an error here I think


arlimus (approved):
great catch, thank you!! :+1:  

really need to find a different pattern down the line, i didn't consider the coordinator shutdown well enough

PR #2381, opened 2023-10-26
⭐ introduce provider plugin heartbeats

arlimus:
Introduce a heartbeat to all provider plugins.

- Older versions of cnquery/cnspec that install newer versions of plugins will not feel a difference. The heartbeat-kill routine is only initialized with the first heartbeat request
- Newer versions of cnquery/cnspec will require the new release of all provider plugins - mandatory. They now require the heartbeat routine to be present. This is due to the grave effects of having old providers in memory and is thus a mandatory ugrade for all users. 
- Heartbeats are done every 2 seconds. There is a grace-period of 3 additional seconds before the plugin kills itself. Thus: worst-case the plugin takes 5 seconds to shutdown, best-case it terminates after 3 seconds without heartbeat.


providers/coordinator.go
```diff
@@ -62,6 +63,33 @@ type RunningProvider struct {
 	lock sync.Mutex
 }
 
+// initialize the heartbeat with the provider
+func (p *RunningProvider) heartbeat() {
+	interval := 2 * time.Second
+	gracePeriod := 3 * time.Second
+	go func() {
+		for !p.isClosed && !p.isShutdown {
```
jaym: this looks like a data race

providers/coordinator.go
```diff
@@ -62,6 +63,33 @@ type RunningProvider struct {
...
+			})
+			if status, ok := status.FromError(err); ok {
+				if status.Code() == 12 {
+					log.Fatal().
+						Str("plugin-ID", p.ID).
+						Msg("please update the provider plugin for " + p.Name)
+				}
+			}
+			if err != nil {
+				log.Fatal().
```
jaym: log.Fatal prevents defers from running, meaning we do not get a chance to properly clean up things like the providers

providers/coordinator.go
```diff
@@ -62,6 +63,33 @@ type RunningProvider struct {
 	lock sync.Mutex
 }
 
+// initialize the heartbeat with the provider
+func (p *RunningProvider) heartbeat() {
+	interval := 2 * time.Second
+	gracePeriod := 3 * time.Second
+	go func() {
```
jaym, line 70: I think this needs to take some sort of context or channel to properly shut down.

providers-sdk/v1/plugin/service.go
```diff
@@ -0,0 +1,34 @@
...
+
+func (s *Service) Heartbeat(req *HeartbeatReq) (*HeartbeatRes, error) {
+	if req.Interval == 0 {
+		return nil, errors.New("heartbeat failed, requested interval is 0")
+	}
+
+	now := time.Now().UnixNano()
+	s.lastHeartbeat = now
+
+	go func() {
```
jaym, line 30: We should avoid launching goroutines like this as we risk leaking memory. it would be better to send a message to some sort of heartbeat component, which there is a single instance of that would be responsible for killing the process

providers-sdk/v1/plugin/service.go
```diff
@@ -0,0 +1,34 @@
...
+		time.Sleep(time.Duration(req.Interval))
+		if s.lastHeartbeat == now {
```
jaym: this is a datarace

providers-sdk/v1/plugin/service.go
arlimus: +1, let's improve that long-term

providers-sdk/v1/plugin/service.go
```diff
@@ -21,11 +23,18 @@ func (s *Service) Heartbeat(req *HeartbeatReq) (*HeartbeatRes, error) {
...
 	s.lastHeartbeat = now
+	s.lock.Unlock()
 
 	go func() {
 		time.Sleep(time.Duration(req.Interval))
-		if s.lastHeartbeat == now {
+
+		s.lock.Lock()
+		isDead := s.lastHeartbeat == now
+		defer s.lock.Unlock()
```
jaym: don't need defer here

jaym (approved):

PR #2389, opened 2023-10-27
🐛 filter unsupported queries in bundle

arlimus:
For local bundle execution, we often run into cases where certain providers aren't installed. We could search to get a bundle compiled, but we don't have an exhaustive list of resources and fields, so it's not feasable right now. We also don't explicitly specify required providers inside of bundles just yet, so this won't do.

For now, there is an option to remove unsupported queries from the execution. This option has been around, but still had some shortcomings.

Namely: Until this point it never worked correctly with Variants. Queries from variants would not get removed and old references were still lingering around.

This PR cleans up the code, makes it more re-usable, and adds testing around this use-case.


chris-rock (commented):
Thank you @arlimus

PR #2391, opened 2023-10-27
🐛 support empty bundle execution

arlimus:
It is possible for bundles to end up having no queries at all after filtering. I noticed that the execution would hang until timeout is reached in this case.

@jaym You know this code much better than I do, could you add a test for this use-case? Also feel free to fix this up if you prefer a different approach.

### DIFF
--- a/explorer/executor/executor.go
+++ b/explorer/executor/executor.go
@@ -90,6 +90,14 @@ func ExecuteFilterQueries(runtime llx.Runtime, queries []*explorer.Mquery, timeo
 }
 
 func (e *instance) runCode(queries map[string]*explorer.ExecutionQuery, timeout time.Duration) error {
+	if len(queries) == 0 {
+		e.progressReporter.Completed()
+		go func() {
+			e.done <- struct{}{}
+		}()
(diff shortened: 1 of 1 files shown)


explorer/executor/executor.go
```diff
@@ -90,6 +90,14 @@ func ExecuteFilterQueries(runtime llx.Runtime, queries []*explorer.Mquery, timeo
 }
 
 func (e *instance) runCode(queries map[string]*explorer.ExecutionQuery, timeout time.Duration) error {
+	if len(queries) == 0 {
+		e.progressReporter.Completed()
+		go func() {
+			e.done <- struct{}{}
```
jaym, line 96: we should circle back around to this. `e.done` should just be closed to signal done. there should be no need to launch a goroutine

jaym (approved):
i can look into writing a test later. There is no existing pattern in that repo I see

explorer/executor/executor.go
arlimus: +1, my approach here should be replaced

PR #2398, opened 2023-10-29
⭐ More improvements to the Azure defaults

tas50:
Improve the data we return in a few more places by default


arlimus (approved):
These are great! kudos

PR #2426, opened 2023-10-31
🐛 do not panic on empty list assertions

arlimus:
Fixes https://github.com/mondoohq/cnquery/issues/2223#issuecomment-1780528051


mariuskimmina (approved):
This fixes the panic, thanks. 
The issue should remain open tho, as the filter is still not working.

PR #2427, opened 2023-10-31
🐛 fix concurrent access to extensible schema

arlimus:
1. Reduce the complexity of locks (kudos @jaym for pointing this out)
2. Split all methods into safe and unsafe calls (to reduce complexity)
3. Make sure the aggregate object is never modified, only re-created

... and add lots of instructions on how to use this.

### DIFF
--- a/providers/extensible_schema.go
+++ b/providers/extensible_schema.go
@@ -12,19 +12,22 @@ import (
 )
 
 type extensibleSchema struct {
-	aggregate      resources.Schema
+	// Note: this object is re-created every time we refresh. It is treated
+	// as unsafe and may be returned to concurrent processes for reading.
+	// Thus, its contents are read-only and may only be replaced entirely.
+	roAggregate resources.Schema
+	// These are all individual schemas that have been added (not their aggregate)
+	loaded map[string]*resources.Schema
+	// Optional prioritization order of select schemas in aggregation.
 	prioritization []string
-
-	loaded        map[string]*resources.Schema
-	runtime       *Runtime
-	lastRefreshed int64
-	lockAll       sync.Mutex // only used in getting all schemas
-	lockAdd       sync.Mutex // only used when adding a schema
+	runtime        *Runtime
+	lastRefreshed  int64
+	sync           sync.Mutex
 }
 
 func newExtensibleSchema() extensibleSchema {
 	return extensibleSchema{
-		aggregate: resources.Schema{
+		roAggregate: resources.Schema{
 			Resources: map[string]*resources.ResourceInfo{},
 		},
 		loaded:         map[string]*resources.Schema{},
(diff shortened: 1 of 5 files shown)


providers/runtime.go
```diff
@@ -130,7 +130,7 @@ func (r *Runtime) UseProvider(id string) error {
 func (r *Runtime) AddConnectedProvider(c *ConnectedProvider) {
 	r.providers[c.Instance.ID] = c
 	r.schema.Add(c.Instance.Name, c.Instance.Schema)
-	r.schema.refresh()
+	r.schema.unsafeRefresh()
```
jaym: why is it safe to call `unsafeRefresh` without holding the lock

providers/extensible_schema.go
```diff
@@ -105,13 +117,45 @@ func (x *extensibleSchema) LookupField(resource string, field string) (*resource
...
+	// If another goroutine started to load this before us, it will be locked until
+	// we complete to load everything and then it will be dumped into this
+	// position. At this point, if it has been loaded we can return safely, since
+	// we don't unlock until we are finished loading.
+	if x.lastRefreshed >= LastProviderInstall {
+		return
+	}
+	x.lastRefreshed = LastProviderInstall
+
+	providers, err := ListActive()
```
jaym, line 152: are all calls to `ListActive` guaranteed to be synchronized through a single `extensibleSchema`. I ask because this call ends up modifying global state without taking any locks

providers/runtime.go
```diff
@@ -253,7 +253,7 @@ func (r *Runtime) Connect(req *plugin.ConnectReq) error {
 
 	r.Recording.EnsureAsset(r.Provider.Connection.Asset, r.Provider.Instance.ID, r.Provider.Connection.Id, asset.Connections[0])
 	r.schema.prioritizeIDs(BuiltinCoreID, r.Provider.Instance.ID)
-	r.schema.refresh()
+	r.schema.unsafeRefresh()
```
jaym: same question about why this is safe without a lock

providers/mock.go
```diff
@@ -117,7 +117,7 @@ func (s *mockProviderService) Init(running *RunningProvider) {
 	s.initialized = true
 
 	rt := s.coordinator.NewRuntime()
-	rt.schema.loadAllSchemas()
-	rt.schema.refresh()
+	rt.schema.unsafeLoadAll()
```
jaym: i think these calls may not be needed. `newRuntime` already calls them:

```
func (c *coordinator) newRuntime(isEphemeral bool) *Runtime {
	res := &Runtime{
		coordinator:     c,
		providers:       map[string]*ConnectedProvider{},
		schema:          newExtensibleSchema(),
		Recording:       NullRecording{},
		shutdownTimeout: defaultShutdownTimeout,
		isEphemeral:     isEphemeral,
	}
	res.schema.runtime = res

	// TODO: do this dynamically in the future
	res.schema.unsafeLoadAll()
	// TODO: this step too shouild be optional only, even when loading all.
	// It is executed when the we connect via a provider, so doing it here is
	// overkill.
	res.schema.unsafeRefresh()
```

providers/extensible_schema.go
```diff
@@ -105,13 +117,45 @@ func (x *extensibleSchema) LookupField(resource string, field string) (*resource
...
+	if err != nil {
+		log.Error().Err(err).Msg("failed to list all providers, can't load additional schemas")
+		return
+	}
+
+	for name := range providers {
+		schema, err := x.runtime.coordinator.LoadSchema(name)
```
jaym, line 159: Is it possible for `LoadSchema` to get called concurrently? It looks to me like different runtimes share coordinators. `LoadSchema` has a path where it can end up modifying the data the coordinator holds through `provider.LoadResources()`.

providers/runtime.go
arlimus: it's not, following the description and where this is located, it's an oversight

providers/extensible_schema.go
arlimus: Not in scope of this PR, we can follow up on this. I don't see a problem at this time, but we can secure providers separately too

providers/runtime.go
arlimus: same, this was an oversight; adjusting the prioritizeIDs to be a concurrency-safe method that handles the refresh

providers/mock.go
arlimus: You're right, I'll comment them out and add a comment that once we remove unsafeLoadAll from newRuntime we need to add it here.

providers/extensible_schema.go
arlimus: +1, yes, I think we need to separately make the coordinator thread-safe. Not in scope for this PR.

jaym (approved):

PR #2431, opened 2023-10-31
🐛 Fix GKE node scanning

czunker:
We are missing some data from COS, that's why we supply some infomration via an inventory file. This infomration needs to be preserved, so we can sync the asset.

### DIFF
--- a/providers/os/provider/provider.go
+++ b/providers/os/provider/provider.go
@@ -419,10 +419,17 @@ func (s *Service) connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 		if err != nil {
 			return nil, err
 		}
-		fingerprint, err := IdentifyPlatform(conn, asset.Platform, []string{ids.IdDetector_Hostname})
-		if err == nil {
-			asset.Name = fingerprint.Name
-			asset.PlatformIds = fingerprint.PlatformIDs
+		// This is a workaournd to set Google COS platfomr IDs when scanned from inside k8s
+		pID, err := conn.(*connection.FileSystemConnection).Identifier()
+		if err != nil {
+			fingerprint, err := IdentifyPlatform(conn, asset.Platform, []string{ids.IdDetector_Hostname})
+			if err == nil {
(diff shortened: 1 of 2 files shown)


czunker:
image

And from the Pod logs:
```
! Error:        Ensure that the Kubelet only makes use of Strong Cryptographic Ciphers

error: Could not load the vulnerability report: rpc error: code = InvalidArgument desc = package analysis is not supported for the provided platform


Scanned 1 asset

Container-Optimized OS from Google
    D gke-mondoo-gke-cluster-sandboxed-pool-a09384ab-igc6

```

czunker:
Testing this is a bit complicated. You need:
- a GKE cluster
- a custom cnspec container image including the os provider build from this PR
- an inventory file like we create it with the mondoo-operator

This PR does not really introduce new functionality. It ports the code from v8 to v9.

providers/os/provider/provider.go
```diff
@@ -419,10 +419,17 @@ func (s *Service) connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 		if err != nil {
 			return nil, err
 		}
-		fingerprint, err := IdentifyPlatform(conn, asset.Platform, []string{ids.IdDetector_Hostname})
-		if err == nil {
-			asset.Name = fingerprint.Name
-			asset.PlatformIds = fingerprint.PlatformIDs
+		// This is a workaournd to set Google COS platfomr IDs when scanned from inside k8s
```
imilchev, line 422:
```suggestion
		// This is a workaournd to set Google COS platform IDs when scanned from inside k8s
```

imilchev (approved):

providers/os/provider/provider.go
czunker: Oh, there are even more typos in this sentence... I'll get a PR up for this.

czunker: https://github.com/mondoohq/cnquery/pull/2464/files

PR #2436, opened 2023-11-01
🧹 fix aws ec2 instance name

vjeffrey:


mariuskimmina (approved):
Thanks for the clean up @vjeffrey

PR #2439, opened 2023-11-01
✨ cnquery status show latest version available

mariuskimmina:
This adds a new output to `cnquery status` which shows the version number of the latest release available on github and shows a hint that a newer version is available if it the currently installed version and the latest version don't match.

```
./cnquery status
→ Platform:             ubuntu
→ Version:              22.04
→ Hostname:             mondoopad
→ IP:                   192.168.178.32
→ Time:                 2023-11-01T13:36:01+01:00
→ Version:              unstable (API Version: unstable)
→ Latest Version:       9.4.0
! A newer version is available
→ API ConnectionConfig: https://us.api.mondoo.com
→ API Status:           SERVING
→ API Time:             2023-11-01T12:36:02Z
→ API Version:          9
```

Related to: https://github.com/mondoohq/cnspec/issues/887

### DIFF
--- a/apps/cnquery/cmd/status.go
+++ b/apps/cnquery/cmd/status.go
@@ -186,7 +186,7 @@ func (s Status) RenderCliStatus() {
 		log.Info().Msg("Features:\t\t" + strings.Join(s.Upstream.Features, ","))
 	}
 
-	if s.Client.ParentMrn == "" {
+	if s.Client.ParentMrn != "" {
 		log.Info().Msg("Owner:\t\t" + s.Client.ParentMrn)
 	}
 
--- a/apps/cnquery/cmd/status.go
+++ b/apps/cnquery/cmd/status.go
@@ -159,7 +159,7 @@ func (s Status) RenderCliStatus() {
 
 	latestVersion, err := cnquery.GetLatestVersion()
 	if err != nil {
-		log.Warn().Msg("failed to get latest version")
+		log.Warn().Err(err).Msg("failed to get latest version")
 	}
 
 	log.Info().Msg("Time:\t\t\t" + s.Client.Timestamp)
--- a/apps/cnquery/cmd/status.go
+++ b/apps/cnquery/cmd/status.go
@@ -149,37 +149,42 @@ type ClientStatus struct {
 func (s Status) RenderCliStatus() {
 	if s.Client.Platform != nil {
 		agent := s.Client
-		log.Info().Msg("Platform:\t" + agent.Platform.Name)
-		log.Info().Msg("Version:\t" + agent.Platform.Version)
-		log.Info().Msg("Hostname:\t" + agent.Hostname)
-		log.Info().Msg("IP:\t\t" + agent.IP)
+		log.Info().Msg("Platform:\t\t" + agent.Platform.Name)
+		log.Info().Msg("Version:\t\t" + agent.Platform.Version)
+		log.Info().Msg("Hostname:\t\t" + agent.Hostname)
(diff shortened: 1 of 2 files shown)


cnquery.go
```diff
@@ -42,6 +47,47 @@ func GetVersion() string {
...
+	}
+
+	return release.Name, nil
+}
+
+// GetLatestVersion returns the latest version available on Github
+func GetLatestVersion() string {
+	releaseName, err := GetLatestReleaseName()
+	if err != nil {
+		panic(err)
```
preslavgerchev: i dont think we want to panic here, failing to get latest version should not be a cause panic on the status call, we should just push the err through

cnquery.go
```diff
@@ -42,6 +47,47 @@ func GetVersion() string {
...
 }
 
+// Release represents a GitHub release
+type Release struct {
+	Name string `json:"name"`
+}
+
+// GetLatestReleaseName fetches the name of the latest release from the specified GitHub repository
+func GetLatestReleaseName() (string, error) {
+	url := "https://api.github.com/repos/mondoohq/cnquery/releases/latest"
```
preslavgerchev: nit: can we make this a param and rename to `defaultUrl`. we can reuse this func for cnspec as well, if needed

apps/cnquery/cmd/status.go
```diff
@@ -149,37 +149,42 @@ type ClientStatus struct {
...
 	if s.Upstream.API.Version != cnquery.APIVersion() {
 		log.Warn().Msg("API versions do not match, please update the client")
 	}
 
 	if len(s.Upstream.Features) > 0 {
-		log.Info().Msg("Features:\t" + strings.Join(s.Upstream.Features, ","))
+		log.Info().Msg("Features:\t\t" + strings.Join(s.Upstream.Features, ","))
 	}
 
 	if s.Client.ParentMrn == "" {
```
preslavgerchev: shouldn't this check be `!=`?

cnquery.go
mariuskimmina: Right, changed it

apps/cnquery/cmd/status.go
mariuskimmina: Hm, as long as the numbers can't be negative it behaves the same I think

preslavgerchev: I meant the `s.Client.ParentMrn == ""` block, it should probably be `s.Client.ParentMrn != ""`,

there is no point in showing an empty parent mrn

apps/cnquery/cmd/status.go
```diff
@@ -149,37 +149,50 @@ type ClientStatus struct {
 func (s Status) RenderCliStatus() {
 	if s.Client.Platform != nil {
 		agent := s.Client
-		log.Info().Msg("Platform:\t" + agent.Platform.Name)
-		log.Info().Msg("Version:\t" + agent.Platform.Version)
-		log.Info().Msg("Hostname:\t" + agent.Hostname)
-		log.Info().Msg("IP:\t\t" + agent.IP)
+		log.Info().Msg("Platform:\t\t" + agent.Platform.Name)
```
preslavgerchev, line 152: why is this changed to a double tab?

mariuskimmina: They looked more properly lined up this way for me.

apps/cnquery/cmd/status.go
mariuskimmina: Oh, you are right, changed it

preslavgerchev (approved):

PR #2442, opened 2023-11-01
🐛 fix systemd service listing

### LINKED ISSUE #2432: :bug: Linux/Unix systemctl status `masked` is not recognized by cnquery
**Describe the bug**
The Linux service status `masked` is not recognized by cnquery:

**To Reproduce**

1. Mask a service
```
systemctl mask avahi-daemon.service
```
2. Check status via systemctl
```
# systemctl status avahi-daemon.service
○ avahi-daemon.service
     Loaded: masked (Reason: Unit avahi-daemon.service is masked.)
     Active: inactive (dead)
```
3. Check status via cnquery
```
cnquery run -c 'service("avahi-daemon.service") {*}'
service: {
  running: false
  masked: false
  enabled: false
  installed: false
  name: "avahi-daemon.service"
  type: null
  description: null
```

**Expected behavior**
The service status `masked` needs to be displayed correctly.

**Screenshots or CLI Output**
image

**Additional context**
```
cnquery version
cnquery 9.4.0 (a9102f6d, 2023-10-27T00:17:51Z
```

mariuskimmina:
Fixes #2432
Extends #2437

The command we have been using this far `systemctl --all list-units --type service`  does not return services that are disabled.
We have to use `systemctl list-unit-files --type=service --all` instead. Which also required the parsing to be changed.

```
systemctl status rsync.service
○ rsync.service
     Loaded: masked (Reason: Unit rsync.service is masked.)
     Active: inactive (dead)
```

Before:

```
service: {
  installed: false
  enabled: false
  description: null
  type: null
  running: false
  masked: false
  name: "rsync.service"
}
```

After:
```
service: {
  installed: false
  enabled: false
  description: ""
  type: "systemd"
  running: false
  masked: true
  name: "rsync.service"
}
```

TODO: 

- [x] add new unit tests


mariuskimmina:
I fixed the `TestParseServiceSystemDUnitFiles` by creating a new testdata file for ubuntu2204 using the `systemctl list-unit-files --type service --all` command. There were more test in `manager_test.go` that relied on the old implementation of `ParseServiceSystemDUnitFiles` that have not been recreated here. I currently left them commented out.

mariuskimmina:
Noticed another problem, using `systemctl list-unit-files` requires follow up commands to get all the information about a service. I am doing this with `exec.Command` on my system and it works fine but it seems this doesn't work in all cases, it shoud use `s.conn.RunComand` - but I can't use this inside `ParseServiceSystemDUnitFiles` ....

mariuskimmina:
Now it seems to be working

```
cnquery> service("avahi-daemon") {*}
service: {
  installed: true
  running: true
  name: "avahi-daemon"
  enabled: true
  description: "Avahi mDNS/DNS-SD Stack"
  masked: false
  type: "systemd"
}
cnquery> service("rsync") {*}
service: {
  enabled: false
  running: false
  type: "systemd"
  description: ""
  name: "rsync"
  masked: true
  installed: false
}
```

mariuskimmina:
On a site note, `Installed` is always the same and `Enabled` - it has been that way before this change as well, as far as I can tell.

providers/os/resources/services/systemd.go
```diff
@@ -32,49 +32,104 @@ func ResolveSystemdServiceManager(conn shared.Connection) OSServiceManager {
...
 			Name:      name,
-			Installed: m[i][2] == "loaded",
-			Running:   m[i][3] == "active",
-			// TODO: we may need to revist the enabled state
-			Enabled:     m[i][2] == "loaded",
-			Masked:      m[i][2] == "masked",
-			Description: m[i][5],
-			Type:        "systemd",
+			Enabled:   fields[1] == "enabled",
+			Installed: fields[1] == "enabled",
```
mick-lue: I'd like to propose a differentiation between _Installed_ and _Enabled_:
If you found the service through a call to systemd, it is installed. So you could probably go with `Installed: true`.
Enabled can stay the same, as it will tell the user if the service will start at boot.

czunker:
Gave it a try with a debian 11 Vagrant box. LGTM:
```
cnquery run vagrant default -c "services{*}"                                                                                                                                                                                                         ✔ │ 26s │ 10:27:16 
services.list: [
  0: {
    enabled: true
    type: "systemd"
    running: true
    installed: true
    name: "apparmor"
    masked: false
    description: "Load AppArmor profiles"
  }
  1: {
    enabled: false
    type: "systemd"
    running: false
    installed: false
    name: "apport-autoreport"
    masked: false
    description: "Process error reports when automatic reporting is enabled"
  }
...
  108: {
    enabled: false
    type: "systemd"
    running: false
    installed: false
    name: "screen-cleanup"
    masked: true
    description: ""
  }
...
  113: {
    enabled: false
    type: "systemd"
    running: false
    installed: false
    name: "snap.lxd.daemon"
...
```

providers/os/resources/services/manager_test.go
```diff
@@ -191,3 +96,41 @@ func TestManagerWindows(t *testing.T) {
 
 	assert.Equal(t, 1, len(serviceList))
 }
+
+func TestManagerUbuntu2204(t *testing.T) {
+	// tests fallback to upstart service
```
czunker: nit: These distros no longer have upstart

czunker (approved):
Thanks @mariuskimmina

PR #2449, opened 2023-11-02
🧹 fixups for aws ebs scanning and ec2 image fetch

vjeffrey:
the first thing is a guard for https://github.com/mondoohq/cnquery/issues/2344

also added some labels we depend on for ebs scans on pre-discovered assets


providers/aws/resources/discovery_conversion.go
```diff
@@ -710,6 +722,7 @@ func EbsConnectAsset(args []string, opts map[string]string) *inventory.Asset {
 	asset.IdDetector = []string{ids.IdDetector_Hostname} // do not use cloud detect or host key here
 	asset.Connections = []*inventory.Config{{
 		Type:     string(awsec2ebsconn.EBSConnectionType),
+		Backend:  inventory.ProviderType_AWS_EC2_EBS,
```
czunker: nit: Isn't this deprecated?

czunker (approved):
Thanks @vjeffrey

providers/aws/resources/discovery_conversion.go
vjeffrey: oh ya i can remove that, i think i placed it there while testing

PR #2458, opened 2023-11-03
🐛 ensure we use the connection asset so we do not lose info from connect call

vjeffrey:
i found that we had the name in the provider runtime code, but when we popped back out to the local scanner code the asset did not have a name. this sets the asset to the one set by the connection, which should have the info we want

DO NOT use builtin provider to test this - things work without this change with the builtin provider. 

easiest way to test is ssh scan to linux machine - you should see a valid instance id, e.g. `i-9038439043`

### DIFF
--- a/explorer/scan/local_scanner.go
+++ b/explorer/scan/local_scanner.go
@@ -212,6 +212,7 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 			log.Error().Err(err).Msg("unable to connect to asset")
 			continue
 		}
+		asset = runtime.Provider.Connection.Asset // to ensure we get all the information the connect call gave us
 
 		// for all discovered assets, we apply mondoo-specific labels and annotations that come from the root asset
 		for _, a := range runtime.Provider.Connection.GetInventory().GetSpec().GetAssets() {
@@ -272,6 +273,7 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 			Asset:    candidate.asset,
 			Upstream: upstream,
 		})
+		candidate.asset = runtime.Provider.Connection.Asset // to ensure we get all the information the connect call gave us
 
 		if err != nil {
 			log.Error().Err(err).Str("asset", candidate.asset.Name).Msg("unable to connect to asset")


explorer/scan/local_scanner.go
```diff
@@ -302,6 +302,7 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 
 	justAssets := []*inventory.Asset{}
 	for _, asset := range assets {
+		asset.asset = asset.runtime.Provider.Connection.Asset // to ensure we get all the information the connect call gave us
```
preslavgerchev: i think we should set this right after the `Connect` for that asset returns, such that any guards/checks that follow are executed against just one asset

explorer/scan/local_scanner.go
```diff
@@ -212,6 +212,7 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 			log.Error().Err(err).Msg("unable to connect to asset")
 			continue
 		}
+		asset = runtime.Provider.Connection.Asset // to ensure we get all the information the connect call gave us
```
vjeffrey, line 215: if you look at line https://github.com/mondoohq/cnquery/blob/f04541179ee30d2d3e88a96cdce811346455fc80/explorer/scan/local_scanner.go#L236-L238 there's some commented out code there that was doing this same thing

explorer/scan/local_scanner.go
```diff
@@ -272,6 +273,7 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 			Asset:    candidate.asset,
 			Upstream: upstream,
 		})
+		candidate.asset = runtime.Provider.Connection.Asset // to ensure we get all the information the connect call gave us
```
preslavgerchev, line 276: do we still need this with the changes on line 215?

vjeffrey: we did when i tested - because one is for that main clires asset and the other is for the candidate assets

preslavgerchev: ooh, i see

vjeffrey:
tested with `go run apps/cnquery/cnquery.go scan ssh ec2-user@ec2-75-101-226-172.compute-1.amazonaws.com` and `go run apps/cnquery/cnquery.go scan aws --discover lambda-functions,vpcs --profile home`

preslavgerchev (approved):

PR #2462, opened 2023-11-03
🐛 tls extension value parsing

mariuskimmina:
Fixes https://github.com/mondoohq/cnquery/issues/2445

The problem is that there are many potential extensions, some which are quite niche. 
We should at least implement parsing for the most common ones.

This PR covers:

- [x] Subject Key Identifier (SKID)
- [x] Subject Alternative Names (SAN)


imilchev (approved):
LGTM

PR #2465, opened 2023-11-03
🐛 ensure we do not return ecr image assets with no connection info from discovery call

vjeffrey:
this can happen if an image has no tag or we were unable to fetch the metadata for the image

### DIFF
--- a/providers/aws/resources/discovery.go
+++ b/providers/aws/resources/discovery.go
@@ -208,7 +208,6 @@ func Discover(runtime *plugin.Runtime, filters connection.DiscoveryFilters) (*in
 		}
 		in.Spec.Assets = append(in.Spec.Assets, list...)
 	}
-
 	return in, nil
 }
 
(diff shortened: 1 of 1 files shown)


providers/aws/resources/discovery.go
```diff
@@ -295,7 +294,10 @@ func discover(runtime *plugin.Runtime, awsAccount *mqlAwsAccount, target string,
 			if !imageMatchesFilters(a, filters) {
 				continue
 			}
-			assetList = append(assetList, addConnectionInfoToEcrAsset(a, conn))
+			ecrAsset := addConnectionInfoToEcrAsset(a, conn)
+			if len(ecrAsset.Connections) > 0 {
```
czunker, line 298: Do we have meaningful data for an error message, so that the user gets a clue which image we cannot scan?

czunker (approved):
Thanks for finding this edge case @vjeffrey

PR #2468, opened 2023-11-06
🧹 Use dependabot groups for update PRs

czunker:
This should help with such PRs https://github.com/mondoohq/cnquery/pull/2418

The grouping should create one PR, which includes the current available minor and patch updates.

https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,10 +1,6 @@
 ---
 version: 2
 updates:
-  - package-ecosystem: gomod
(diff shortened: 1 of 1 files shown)


vjeffrey (approved):

czunker:
> oh having just one pr per provider sounds wonderful!

When it works here, we can port it to the other repos.

PR #2508, opened 2023-11-07
🐛 aws: do not fail on missing discover, use correct id detectors

vjeffrey:


providers/aws/resources/discovery_conversion.go
```diff
@@ -289,73 +289,6 @@ func addConnectionInfoToEc2Asset(instance *mqlAwsEc2Instance, accountId string,
 	return asset
 }
 
-func addSSMConnectionInfoToEc2Asset(instance *mqlAwsEc2Instance, accountId string, conn *connection.AwsConnection) *inventory.Asset {
```
vjeffrey, line 292: this was unused

vjeffrey:
i pushed a second commit that fixes a thing i noticed:
`cnquery shell aws --discover instances` would run the ssh call but not actually fill in the platform info, and we would be left with empty platform name on the asset. i added a thing that makes sure that if the platform name is empty we still run detect

preslavgerchev (approved):

PR #2510, opened 2023-11-07
🧹 add CompilerConfig on all compile-related calls

arlimus:
Remaining todos to move this out of draft:
- [ ] update cnspec: https://github.com/mondoohq/cnspec/pull/937
- [ ] update other components

-----

Until now, the only thing developers can configure when it comes to compile-related calls is the schema.

However, there are use-cases where we may want to configure other things in the compiler, like:
- which feature-flags are enabled/disabled
- use of stats

I ran into the last use-case myself, which would have caused large copy-paste rewrites of the Bundle-compiler, just in order to get access to all the stats.

At this point it is a more comprehensive experience to not move the schema around, but instead wire the compiler config directly through all these calls. This means that more future-facing features will be configurable and that a chain of calls (like `Bundle.Compile`) uses a consistent set of compiler configs.

@ Reviewers: I recommend merging this post-release, since it will require changes in all upstream projects that depend on it. I have already updated cnspec, but there are more.

### DIFF
--- a/apps/cnquery/cmd/scan.go
+++ b/apps/cnquery/cmd/scan.go
@@ -21,6 +21,7 @@ import (
 	"go.mondoo.com/cnquery/v9/cli/theme"
 	"go.mondoo.com/cnquery/v9/explorer"
 	"go.mondoo.com/cnquery/v9/explorer/scan"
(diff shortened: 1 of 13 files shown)


chris-rock (approved):
Thank you @arlimus

Issue #1606, opened 2023-08-29
:bug: ? Sub service `.socket` not picked up


mm-weber:
**Describe the bug**
The sub-service `systemd-journal-remote.socket` isn't picked up by cnspec:

```
cnspec> services.where(name == /remote/)
services.where.list: [
  0: service name="systemd-journal-remote" running=false enabled=true type="systemd"
]
```

**To Reproduce**
Follow https://github.com/mondoo-community/workspace-lecos/issues/6

**Expected behavior**
The sub-service `.remote` should be shown in the `services`:

```
cnspec> services.where(name == /remote/)
services.where.list: [
  0: service name="systemd-journal-remote" running=false enabled=true type="systemd"
  1: service name="systemd-journal-remote.socket" running=false enabled=true type="systemd" 
]

```

**Desktop (please complete the following information):**
```
[rocky@ip-10-0-101-136 ~]$ cnspec version
cnspec 8.25.0 (b999413, 2023-08-28T20:50:59Z)
```

mariuskimmina:
We are filtering out the .sockets in our command `systemctl list-unit-files --type=service --all` by saying `type=service` other types are

* socket
* path
* slice
* target
* mount

There might be more but that's all I have on my system. 

We need to decide on how we want to handle these. Treating them all as services and removing the filter would be easy (#2531) but doesn't really seem right. We might need to create these as different entities such that we could say `socket.where()`  or `service.where()` depending on what we are looking for

chris-rock:
The service can be checked with MQL right now with:

```
parse.ini(content: command("systemctl show --no-pager  systemd-journal-remote").stdout.trim).params
```

chris-rock:
I changed this from bug to feature request since cnquery works as designed and the systemctl information can be checked with native MQL mechanisms. We still want to make it easier to access the data going forward.

tas50:
Reviewed as part of a full backlog triage — **confirmed still open** in the current codebase.

providers/os/resources/services/systemd.go:248 `SystemDServiceManager.List()` runs `systemctl list-unit-files --type service --all` and `list-units --type service --all` (line 259), restricting to service-type units; `.socket` units are excluded from the `services` list. A separate `systemd.socket` resource exists (providers/os/resources/services/systemd_socket.go, PR #7097) but it is not merged into the `services` collection the issue references.

[...]

PR #2531, opened 2023-11-09
🐛 fix service socket being filtered

mariuskimmina:
#1606 

```
cnquery> services.where(name == /remote/)
services.where.list: [
  0: service name="systemd-journal-remote" running=false enabled=false type="systemd"
  1: service name="systemd-journal-remote.socket" running=false enabled=false type="systemd"
  2: service name="remote-cryptsetup.target" running=false enabled=false type="systemd"
  3: service name="remote-fs-pre.target" running=false enabled=false type="systemd"
  4: service name="remote-fs.target" running=true enabled=true type="systemd"
  5: service name="remote-veritysetup.target" running=false enabled=false type="systemd"
]
```

- [ ] adjust tests

### DIFF
--- a/providers/os/resources/services/systemd.go
+++ b/providers/os/resources/services/systemd.go
@@ -95,7 +95,7 @@ func SystemDExtractDescription(systemctlOutput string) string {
 // List returns a slice of Service structs representing the state of all services
 func (s *SystemDServiceManager) List() ([]*Service, error) {
 	var services []*Service
-	cmdList, err := s.conn.RunCommand("systemctl list-unit-files --type service --all")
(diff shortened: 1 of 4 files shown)


tas50:
I think we really need to think through if this is the behavior we want. If we merge this change we're going to be treating timers and mount unit files as services, which doesn't seem correct.

mariuskimmina:
> I think we really need to think through if this is the behavior we want. If we merge this change we're going to be treating timers and mount unit files as services, which doesn't seem correct.

Right, I'll convert this to draft for now

chris-rock:
We do not think this is the right solution

PR #2544, opened 2023-11-09
🐛 fix comments and add tests for latest release

mariuskimmina:
Follow up on #2543

### DIFF
--- a/cnquery.go
+++ b/cnquery.go
@@ -46,7 +46,7 @@ func GetVersion() string {
 	return Version
 }
 
-// Release represents a GitHub release
+// Release represents a release
 type Release struct {
 	Name    string `json:"name"`
 	Version string `json:"version"`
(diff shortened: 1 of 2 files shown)


cnquery.go
```diff
@@ -46,15 +46,15 @@ func GetVersion() string {
...
 
-// Release represents a GitHub release
+// Release represents a release
 type Release struct {
 	Name    string `json:"name"`
 	Version string `json:"version"`
 }
 
-var cnqueryGithubReleaseUrl = "https://releases.mondoo.com/cnquery/latest.json?ignoreCache=1"
+var cnqueryReleaseUrl = "https://releases.mondoo.com/cnquery/latest.json?ignoreCache=1"
```
imilchev:
```suggestion
var cnqueryLatestReleaseUrl = "https://releases.mondoo.com/cnquery/latest.json?ignoreCache=1"
```

cnquery_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetLatestVersion(t *testing.T) {
+	version, err := GetLatestVersion()
+
+	assert.NoError(t, err)
+	assert.NotNil(t, version)
+	assert.NotContains(t, cnqueryReleaseUrl, "github.com")
```
imilchev: check that it is `releases.mondoo.com`. We don't want this to be different than github but we explicitly want it to be `releases.mondoo.com`

mariuskimmina:
Added your suggestions @imilchev

imilchev (approved):

PR #2556, opened 2023-11-10
🐛 fix different behavior between scan container and scan docker

### LINKED ISSUE #2518: different behavior between `cnspec scan container` and `cnspec scan docker`
**Describe the bug**

There is a different behavior between the target `container` and `docker`

i scan a container image with the following command and it fails:

```bash
cnspec scan container docker.io/pmuench/dvwa-container-escape
→ loaded configuration from /Users/atomic111/.config/mondoo/mondoo.yml using source default
x unable to connect to asset error="rpc error: code = Unknown desc = Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?"
```

if I scan the same container image with the following command it scans the upstream container image:

```bash
cnspec scan docker docker.io/pmuench/dvwa-container-escape
→ loaded configuration from /Users/atomic111/.config/mondoo/mondoo.yml using source default
→ synchronize assets

 docker.io/pmuench/dvwa-container-escape ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% score


Asset: docker.io/pmuench/dvwa-container-escape
----------------------------------------------
```

My expectation is that it is the same behavior.

mariuskimmina:
Fixes #2518


providers/os/provider/provider.go
```diff
@@ -89,7 +89,7 @@ func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error)
 		port = 5985
 	case "vagrant":
 		conf.Type = "vagrant"
-	case "docker":
+	case "docker", "container":
```
preslavgerchev: i think we should keep those separate. there should be a difference between  `container image scan` which should attempt to pull from the registry you've specified and `docker image scan` which should try to pull from the local docker that is running

mariuskimmina: But on this switch statement, both do exactly the same thing on `image` as far as I can tell (and also for all other options)

chris-rock: I agree with Preslav. This is how it should work

mariuskimmina: Okay, we can close the issue as intended behavior then?

chris-rock: No because `cnspec scan container docker.io/pmuench/dvwa-container-escape` needs to work without docker

chris-rock:
The intended behavior:

*container*

```
cnspec scan container docker.io/pmuench/dvwa-container-escape
-> does not use docker
-> pulls the container image directly from the registry
```

*docker*

```
cnspec scan docker docker.io/pmuench/dvwa-container-escape
-> tries to pull the image from local docker host
-> if image is not there use same logic as container to pull the image from registry
```

mariuskimmina:
Okay, so currently the `docker` version uses `FetchConnectionType` to determine if it's 

1. a container
2. a locally present image
3. a remote image

and it uses docker to do that.

On the other hand `container` currently simply assumes that it's a container. 
So we need to allow `container` to also auto-detect if it's a remote image instead. But without the use of docker.
I have now added a simple function that does this differentiation assuming that if it contains a colon `:` it's an image, if it contains a slash `/` it's also an image but otherwise it's a container. 

I believe this lines up with what is described here: https://mondoo.com/docs/cnspec/cli/cnspec_scan_container/

PR #2563, opened 2023-11-13
🐛 Keep idDetectors across connects

czunker:
With local connections, we saw the case in cnspec, that on first connect, assets had platformIDs as expected. With local connections this should only be the hostname and perhaps cloud data.

But on later connects, the set id detectors were emtpy and another default kicked in: https://github.com/mondoohq/cnquery/blob/main/providers/os/provider/detector.go\#L39-L41

This added the ssh hostkey detector to the asset which is not expected.

### DIFF
--- a/providers/os/provider/platform.go
+++ b/providers/os/provider/platform.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/rs/zerolog/log"
 	"go.mondoo.com/cnquery/v9/providers-sdk/v1/inventory"
+	"go.mondoo.com/cnquery/v9/providers/os/connection"
 	"go.mondoo.com/cnquery/v9/providers/os/connection/shared"
 	"go.mondoo.com/cnquery/v9/providers/os/detector"
 	"go.mondoo.com/cnquery/v9/providers/os/id/awsec2"
@@ -26,6 +27,7 @@ type PlatformFingerprint struct {
 	Runtime       string
 	Kind          string
 	RelatedAssets []PlatformFingerprint
+	idDetectors   []string
 }
 
 type PlatformInfo struct {
@@ -44,8 +46,21 @@ func IdentifyPlatform(conn shared.Connection, p *inventory.Platform, idDetectors
 	}
 
 	var fingerprint PlatformFingerprint
-	var ids []string
+	var platformIds []string
 	var relatedIds []string
+
+	if len(idDetectors) == 0 {
+		// fallback to default id detectors
+		switch conn.Type() {
+		case connection.Local:
+			idDetectors = []string{ids.IdDetector_Hostname, ids.IdDetector_CloudDetect}
+		case connection.SSH:
+			idDetectors = []string{ids.IdDetector_Hostname, ids.IdDetector_CloudDetect, ids.IdDetector_SshHostkey}
+		case connection.Tar, connection.FileSystem, connection.DockerSnapshot:
+			idDetectors = []string{ids.IdDetector_Hostname}
+		}
+	}
+	fingerprint.idDetectors = idDetectors
 
 	for i := range idDetectors {
 		idDetector := idDetectors[i]
@@ -56,7 +71,7 @@ func IdentifyPlatform(conn shared.Connection, p *inventory.Platform, idDetectors
 			continue
 		}
 		if len(platformInfo.IDs) > 0 {
-			ids = append(ids, platformInfo.IDs...)
+			platformIds = append(platformIds, platformInfo.IDs...)
 		}
 		if len(platformInfo.RelatedPlatformIDs) > 0 {
 			relatedIds = append(relatedIds, platformInfo.RelatedPlatformIDs...)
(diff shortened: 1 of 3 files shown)


czunker:
A `cnspec scan` looked like this before:
```
DBG update asset asset=clone platform-ids=["//platformid.api.mondoo.app/hostname/clone","//platformid.api.mondoo.app/runtime/ssh/hostkey/SHA256-2TTaTQRSyf8OKVDYf3mH4yV+V4+/4qG8gZTWWI4ijwk","//platformid.api.mondoo.app/runtime/ssh/hostkey/SHA256-qyiuEyGAHMdvzlQ+vcIXH3z9IE2FgzzQQquustvikhg","//platformid.api.mondoo.app/runtime/ssh/hostkey/SHA256-4+jQYT8PFCVPL1rKSjb0fFX2mf+qK2xdR/CrI9df9Vo"]
DBG using API endpoint https://api.edge.mondoo.com
DBG run scan asset=clone
DBG client> request policies bundle for asset asset=//assets.api.mondoo.app/spaces/elastic-payne-456551/assets/2Xz6Fazj0N291974pKKzNo84eUQ
```

Now, it only has the hostname:
```
DBG update asset asset=debian-11-clone platform-ids=["//platformid.api.mondoo.app/hostname/debian-11-clone"]
DBG using API endpoint http://192.168.1.87:8989
DBG run scan asset=debian-11-clone

```

providers/os/provider/provider.go
```diff
@@ -319,6 +319,7 @@ func (s *Service) connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 		if len(idDetectors) == 0 {
 			// fallback to default id detectors
 			idDetectors = []string{ids.IdDetector_Hostname, ids.IdDetector_CloudDetect}
+			asset.IdDetector = idDetectors
```
chris-rock: why do we need to set this here to the asset proto? can we build a test case for that behavior? if we need to set this, I would prefer to do this as part of the fingerprint setting.

czunker: Testing this isn't that easy. We would need some sort of high level tests for this. The problem only happens when used with plugins. As soon as we use builtins, the problem disappears.

I'm not sure what you mean by "fingerprint setting". The detectors are what is used to determine the platform ID, so doing the fingerprinting. In my understanding they are the setting.
And we also need to set it inside the asset, because we later need them again. My current assumption is, this is needed because connect and discover are currently not handled separatly.

chris-rock: This does not depend on MQL, therefore we can just test it within the provider. We just need to connect to an asset, run detection and disconnect and reconnect. See https://github.com/mondoohq/cnquery/blob/main/providers/vsphere/provider/provider_test.go#L19. 

We set multiple asset fields like name and platform ids here https://github.com/mondoohq/cnquery/pull/2563/files#diff-2ca4387e72bfea051603beafed7cf18d56fd372d0b3ada6287eb91de2ddf5b8aR385, why not set the id detectors there as well?

czunker: Thanks for the pointers.

I refactored the fingerprinting and added a test for the idDetectors. But this test only tests the current code. It wouldn't have caught the issue this PR solves because the gRPC part is missing.

providers/os/provider/platform.go
```diff
@@ -44,9 +46,22 @@ func IdentifyPlatform(conn shared.Connection, p *inventory.Platform, idDetectors
 	}
 
 	var fingerprint PlatformFingerprint
-	var ids []string
+	var platformIds []string
 	var relatedIds []string
 
+	if len(idDetectors) == 0 {
```
chris-rock, line 52: I like the handing here

providers/os/provider/platform.go
```diff
@@ -26,6 +27,7 @@ type PlatformFingerprint struct {
 	Runtime       string
 	Kind          string
 	RelatedAssets []PlatformFingerprint
+	idDetectors   []string
```
chris-rock: to avoid confusion, we may want to rename this to `activeIdDetectors`

providers/os/provider/provider_test.go
```diff
@@ -0,0 +1,65 @@
...
+	require.NoError(t, err)
+	require.NotNil(t, connectResp)
+
+	require.Len(t, connectResp.Asset.IdDetector, 2)
+	require.Contains(t, connectResp.Asset.IdDetector, ids.IdDetector_Hostname)
+	require.Contains(t, connectResp.Asset.IdDetector, ids.IdDetector_CloudDetect)
+	require.NotContains(t, connectResp.Asset.IdDetector, ids.IdDetector_SshHostkey)
+	// here we have the hostname twice, as platformid and stand alone
+	// This get's cleaned up later in the code
+	require.Len(t, connectResp.Asset.PlatformIds, 2)
```
chris-rock, line 40: We should really not return hostnames that do not conform to be a platform id. Therefore we should only have one result here.

czunker: As discussed, this will be part of https://github.com/mondoohq/cnquery/issues/1811

czunker: I'll add a comment in the code to make also clear.

providers/os/provider/platform.go
czunker: Thanks

providers/os/provider/platform.go
czunker: Changed

chris-rock (approved):
Thank you @czunker

PR #2566, opened 2023-11-13
Update yet to be released field to match others

tas50:
We use CreatedAt more than we use CreatedTime. Eventually we should unify this, but let's not make it worse.

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -1235,7 +1235,7 @@ private aws.dynamodb.table @defaults("name region") {
   // Tags for the table
   tags() map[string]string
   // The creation time of the table
-  createdTime time
+  createdAt time
   // Is deletion protection enabled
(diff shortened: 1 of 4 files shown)


providers/aws/resources/aws.lr
```diff
@@ -1235,7 +1235,7 @@ private aws.dynamodb.table @defaults("name region") {
   // Tags for the table
   tags() map[string]string
   // The creation time of the table
-  createdTime time
+  createdAt time
```
chris-rock, line 1238: This is a breaking change of the resource. I think we need consensus first on how we structure those create and update timestamps. I am in full support to make this a consistent approach. We follow the API that we are exposing and they are not consistent either.

tas50:
At the time of the PR this wasn't a breaking change since the resource had yet to be released. It is now, so closing this out.

PR #2570, opened 2023-11-14
🐛 fix json escaping

imilchev:
Fixes https://github.com/mondoohq/server/issues/6811

### DIFF
--- a/llx/rawdata_json.go
+++ b/llx/rawdata_json.go
@@ -52,8 +52,12 @@ func float2json(f float64) string {
 }
 
 // Takes care of escaping the given string
-func string2json(s string) string {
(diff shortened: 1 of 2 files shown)


llx/rawdata_json.go
```diff
@@ -52,8 +52,12 @@ func float2json(f float64) string {
 }
 
 // Takes care of escaping the given string
-func string2json(s string) string {
-	return fmt.Sprintf("%#v", s)
```
imilchev, line 56: i am not sure why we used this weirdness instead of proper json marshaller. It seems like this are working as expected with this change. Unless there is some hidden reason why we didn't use that in the first place...

PR #2592, opened 2023-11-16
🐛 Try to make mquery checksums stable

jaym:
If we change a value inside the query, we need to recalculate the checksum. If checksums aren't stable, this ends up breaking parts of the resolver code in cnspec

### DIFF
--- a/explorer/mquery.go
+++ b/explorer/mquery.go
@@ -280,6 +280,10 @@ func (m *Mquery) refreshChecksumAndType(queries map[string]*Mquery, props map[st
 
 // RefreshAsFilter filters treats this query as an asset filter and sets its Mrn, Title, and Checksum
 func (m *Mquery) RefreshAsFilter(mrn string, schema llx.Schema) (*llx.CodeBundle, error) {
+	if m.Title == "" {
+		m.Title = m.Query
+	}
+
 	bundle, err := m.refreshChecksumAndType(nil, nil, schema)
 	if err != nil {
 		return bundle, err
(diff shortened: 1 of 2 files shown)


explorer/mquery.go
```diff
@@ -280,6 +280,10 @@ func (m *Mquery) refreshChecksumAndType(queries map[string]*Mquery, props map[st
 
 // RefreshAsFilter filters treats this query as an asset filter and sets its Mrn, Title, and Checksum
 func (m *Mquery) RefreshAsFilter(mrn string, schema llx.Schema) (*llx.CodeBundle, error) {
+	if m.Title == "" {
```
preslavgerchev: is there any reason why we fall back to the query if title is empty? is title being empty bad?

chris-rock: I also do not understand why this is required

imilchev: the title is part of the fields used to calculate the checksum of the query

jaym: I also do not understand. I can remove it if everyone agrees

jaym: @imilchev the checksumming doesn't care if the title is empty. What would be important is its consistently empty.

The one way it may break is our checksum is something like:
```
c = c.Add(A)
c = c.Add(B)
c = c.Add(C)
```
and i think it does it a byte at a time, so the checksums for `(A, B, C) = ("foo", "bar", "")` would be the same as `(A, B, C) = ("foo", "", "bar")`. But I didn't look deep into it and I may be misremembering how it actually works

preslavgerchev: I suspect the query-as-title fallback has been used to make printing on CLI pretty, so we should probably move that logic there

jaym: do we print asset filters?

preslavgerchev: nope, i keep forgetting this change is only for asset filters

preslavgerchev (approved):

PR #2607, opened 2023-11-19
Avoid mdx compilation errors in docs markdown

tas50:
Bare values like <> are no longer allowed in markdown so this causes errors when the parser is trying to parse it. This avoids the parsing.


chris-rock (approved):
Ok for now, but this needs to be fixed in the markdown converter itself.

PR #2610, opened 2023-11-20
🐛 adjust type of empty+errored array maps

arlimus:
Couple of problems we saw:
1. Reported in https://github.com/mondoohq/cnquery/issues/2604 we see that the child type of the list is never set correctly, if the list is empty. A bit lazy on the return, glad to have this fixed
2. I also noticed that on errors the type is retained. Technically, this shouldn't matter (because it's an error), but just to be sure, we should set it correctly when we return the results.

Fixes https://github.com/mondoohq/cnquery/issues/2604


chris-rock (approved):
Thank you @arlimus

PR #2633, opened 2023-11-23
🐛 Fix asset.name for remote images

### LINKED ISSUE #2297: asset names for container images are different between v8 and v9
**Describe the bug**
Screenshot from 2023-10-19 11-29-20

The older image is scanned with v9.2.1 the newer one with v8.1.1.

The command was:
```
cnspec scan container image registry.access.redhat.com/ubi8/ubi:8.6
```

**To Reproduce**

1. Run above command with a v8 and a v9 client
2. Compare results
3. Note the difference

**Expected behavior**
Names should be the same.

- os provider for v9: 9.0.14

czunker:
Fixes #2297

### DIFF
--- a/.github/workflows/pr-test-lint.yml
+++ b/.github/workflows/pr-test-lint.yml
@@ -109,6 +109,9 @@ jobs:
           git push
         shell: bash
 
+      - name: 'Set up gcloud CLI'
+        uses: 'google-github-actions/setup-gcloud@v1'
+
       - name: Test cnquery
         run: make test/go/plain-ci
         
--- a/providers/os/connection/docker_container.go
+++ b/providers/os/connection/docker_container.go
@@ -225,11 +225,17 @@ func NewContainerRegistryImage(id uint32, conf *inventory.Config, asset *invento
 		conn.PlatformIdentifier = identifier
 		conn.Metadata.Name = containerid.ShortContainerImageID(hash.String())
 
-		asset.PlatformIds = []string{identifier}
 		repoName := ref.Context().Name()
 		imgDigest := hash.String()
 		name := repoName + "@" + containerid.ShortContainerImageID(imgDigest)
-		asset.Name = name
+		if asset.Name == "" {
+			asset.Name = name
+		}
+		if len(asset.PlatformIds) == 0 {
+			asset.PlatformIds = []string{identifier}
+		} else {
+			asset.PlatformIds = append(asset.PlatformIds, identifier)
+		}
 
 		// set the platform architecture using the image configuration
 		imgConfig, err := img.ConfigFile()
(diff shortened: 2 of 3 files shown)


czunker:
This is how it looked with v8:
```
cnquery-8.28.4 run container image redis:7.2.3-bookworm -c "asset.name" 
asset.name: "index.docker.io/library/redis@dd5380318220"
```

v9:
```
cnquery run container image redis:7.2.3-bookworm -c "asset.name" 
asset.name: "index.docker.io/library/redis@aa853dac4144"
```

v9 + updated os provider:
```
cnquery run container image redis:7.2.3-bookworm -c "asset.name" 
asset.name: "index.docker.io/library/redis@dd5380318220"
```

.github/workflows/pr-test-lint.yml
```diff
@@ -78,6 +78,9 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-go-
 
+      - name: 'Set up gcloud CLI'
```
czunker, line 112: This is needed to prevent the error:
```
error getting credentials - err: exec: "docker-credential-gcloud": executable file not found in $PATH, out: ``
```

providers/os/connection/docker_container.go
```diff
@@ -225,11 +225,15 @@ func NewContainerRegistryImage(id uint32, conf *inventory.Config, asset *invento
...
 
-		asset.PlatformIds = []string{identifier}
 		repoName := ref.Context().Name()
 		imgDigest := hash.String()
 		name := repoName + "@" + containerid.ShortContainerImageID(imgDigest)
-		asset.Name = name
+		if asset.Name == "" {
+			asset.Name = name
+		}
+		if len(asset.PlatformIds) == 0 {
```
chris-rock, line 234: Why do we need that check?

czunker: In the case I was testing, the platformID was already set to the correct sha, the same as for the asset name.

As I'm not sure whether we end up here with an empty name or platformID, I added the `if`s instead of removing this code part.

chris-rock: 1. If is a different platform id, then we need to figure out why, but this can happen if users define their own
2. We should only add the platform id that we determined, when its different to the list so that we ensure uniqness (i think we validate this later, but we need to double-check)
3. We should definitely always add the platform id as long as it is different, not just when `len(asset.PlatformIds) == 0` is matched

Would love to see a test case for that since this seems to be an error case we can run more often in

czunker: As discussed, I now append the platformID in case there is already one present.

I also extended the test case to check for the platformID, I'd expect for this remote image.

chris-rock (approved):
Thank you @czunker

PR #2644, opened 2023-11-26
🧹 update license header in lr files

chris-rock:


czunker (approved):
Nice catch. Thanks @chris-rock 

Latest copywrite now also knows about lr files: https://github.com/hashicorp/copywrite/releases/tag/v0.17.0

PR #2647, opened 2023-11-27
🐛 dependabot: also update providers go.mod

czunker:
From time to time we have dependabot PRs which update the cnquery go.mod and would also require an update of the providers go.mod files. But dependabot can only update one go.mod. This PR should update the providers go.mod files in these cases.

### DIFF
--- a/.github/workflows/pr-test-lint.yml
+++ b/.github/workflows/pr-test-lint.yml
@@ -81,6 +81,27 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-go-
 
+      - name: Updates to cnquery core go.mod?
+        id: cnquerygomod
+        if: github.event_name == 'pull_request' && github.actor == 'dependabot[bot]'
+        run: |
+          echo "COUNT_GOMOD=$(git diff-tree --name-only -r HEAD~1..HEAD | grep -c -E "^go.mod$")" >> $GITHUB_OUTPUT
+
+      - name: Fix providers go.mod for dependabot PRs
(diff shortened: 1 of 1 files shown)


czunker:
This is an example of such a failed PR:
https://github.com/mondoohq/cnquery/actions/runs/6968070246

dependabot only updated the cnquery go.mod, but later during tests go complains, that ipmi would also require these updates.

.github/workflows/pr-test-lint.yml
```diff
@@ -81,6 +81,27 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-go-
 
+      - name: Updates to cnquery core go.mod?
+        id: cnquerygomod
+        if: github.event_name == 'pull_request' && github.actor == 'dependabot[bot]'
+        run: |
+          echo "COUNT_GOMOD=$(git diff-tree --name-only -r HEAD~1..HEAD | grep -c -E "^go.mod$")" >> $GITHUB_OUTPUT
+
+      - name: Fix providers go.mod for dependabot PRs
```
chris-rock, line 90: Just to fully understand this approach: this is pushing a commit to the dependabot PR that is open?

czunker: Yes, that's what it should do from what I could read about it.

From my understanding, this is possible with newer versions of the `checkout` action. It checks out the branch of the current PR. In case it is a dependabot PR that changes `/go.mod`, this adds a commit that updates the providers `go.mod` files.

This should then allow the following tests to pass. We get the tests across the complete PR and later have one large PR, including all the updates.

chris-rock: Lets try it.

PR #2654, opened 2023-11-27
✨ New vulnmgmt resources

czunker:
This moves the vuln scanning server-side and adds new resources.

E.g.:
```
cnspec> vulnmgmt.advisories
vulnmgmt.advisories: [
  0: vuln.advisory id="USN-5181-1"
  1: vuln.advisory id="USN-5181-1"
  2: vuln.advisory id="USN-5181-1"
  3: vuln.advisory id="USN-5181-1"
]
```


czunker:
This requires https://github.com/mondoohq/mondoo-go/pull/10

providers-sdk/v1/upstream/gql/vulnmgmt_gql.go
```diff
@@ -0,0 +1,208 @@
+package gql
```
czunker, line 4: The GQL queries are in core cnquery, because I assume we need to use them in different providers. Currently it is only os, but vsphere will follow and perhaps more.

providers/os/resources/vulnmgmt.go
```diff
@@ -0,0 +1,235 @@
...
+)
+
+func (v *mqlVulnmgmt) lastAssessment() (*time.Time, error) {
+	mcc := v.MqlRuntime.Upstream
+	if mcc == nil || mcc.ApiEndpoint == "" {
+		return nil, resources.MissingUpstreamError{}
+	}
+
+	// get new gql client
+	mondooClient, err := gql.NewClient(mcc.UpstreamConfig, mcc.HttpClient)
```
imilchev: you are creating a new GQL client in multiple functions. Isn't it better to make the GQL client part of the connection or part of the `mqlVulnmgmt` instead?

providers/os/resources/vulnmgmt.go
```diff
@@ -0,0 +1,235 @@
...
+func (v *mqlVulnmgmt) advisories() ([]interface{}, error) {
+	return nil, v.populateData()
+}
+
+func (v *mqlVulnmgmt) packages() ([]interface{}, error) {
+	return nil, v.populateData()
+}
+
+func (v *mqlVulnmgmt) stats() (*mqlAuditCvss, error) {
+	return nil, v.populateData()
```
imilchev, line 93: shouldn't that return also the actual value?

providers/os/resources/vulnmgmt.go
```diff
@@ -0,0 +1,235 @@
...
+
+	pkgsRes, err := CreateResource(v.MqlRuntime, "packages", nil)
+	if err != nil {
+		return nil, err
+	}
+	pkgs := pkgsRes.(*mqlPackages)
+	pkgsList := pkgs.GetList().Data
+
+	gqlPackages := make([]mondoogql.PackageInput, len(pkgsList))
+	for i, p := range pkgs.GetList().Data {
```
imilchev:
```suggestion
	for i, p := range pkgsList {
```

providers/vsphere/resources/vsphere.lr
```diff
@@ -18,6 +20,74 @@ platform {
...
+  // Package version
+  version string
+  // Available package version
+  available string
+  // Architecture of this package
+  arch string
+}
+
+// Common Vulnerability Scoring System (CVSS) score
+private audit.cvss {
```
imilchev: should we add a default?

providers/os/resources/vulnmgmt.go
czunker: Nice catch. Thanks.

providers/os/resources/vulnmgmt.go
czunker: Nice idea.
I added it to a cache, so we can re-use it, but do not need to create it upfront.

providers/os/resources/vulnmgmt.go
czunker: Not needed. That's some MQL black magic: https://github.com/mondoohq/cnquery/blob/main/providers/os/resources/command.go#L58

But a comment might also be helpful here. I added it.

providers/vsphere/resources/vsphere.lr
czunker: Good idea. Added.

imilchev (approved):

PR #2682, opened 2023-11-29
✨ New AWS WAF resource

mariuskimmina:
- [x] ACLs
- [x] Rules
- [x] RuleGroups
- [x] IP Sets


mariuskimmina:
IP Sets 

```
cnquery> aws.waf.ipSets.first{ * }
aws.waf.ipSets.first: {
  addressType: "IPV4"
  addresses: [
    0: "10.0.0.0/32"
  ]
  name: "marius-test"
  id: "b78a84f9-5fbe-4cbd-8e1f-e5ccf331a148"
  description: ""
  arn: "arn:aws:wafv2:us-east-1:xxxxx:regional/ipset/marius-test/b78a84f9-5fbe-4cbd-8e1f-e5ccf331a148"
}
```

mariuskimmina:
This works now as a starting point, the rules are still missing a lot important information

mariuskimmina:
Parsed the statement into a dict now

```
  1: {
    name: "marius-test"
    statement: {
      AndStatement: null
      ByteMatchStatement: null
      GeoMatchStatement: null
      IPSetReferenceStatement: null
      LabelMatchStatement: null
      ManagedRuleGroupStatement: null
      NotStatement: null
      OrStatement: null
      RateBasedStatement: null
      RegexMatchStatement: null
      RegexPatternSetReferenceStatement: null
      RuleGroupReferenceStatement: null
      SizeConstraintStatement: null
      SqliMatchStatement: {
        FieldToMatch: {
          AllQueryArguments: null
          Body: null
          Cookies: null
          HeaderOrder: null
          Headers: null
          JA3Fingerprint: null
          JsonBody: null
          Method: null
          QueryString: null
          SingleHeader: {
            Name: "xxxxxxx"
          }
          SingleQueryArgument: null
          UriPath: null
        }
        SensitivityLevel: "LOW"
        TextTransformations: [
          0: {
            Priority: 0.000000
            Type: "MD5"
          }
        ]
      }
      XssMatchStatement: null
    }
    priority: 1
  }
  ```

File: providers/aws/resources/aws_waf.go [hunk 1]
```diff
@@ -0,0 +1,290 @@
...
+		Id:    &a.Id.Data,
+		Name:  &a.Name.Data,
+		Scope: scope,
+	}
+	aclDetails, err := svc.GetWebACL(ctx, params)
+	if err != nil {
+		return nil, err
+	}
+	for _, rule := range aclDetails.WebACL.Rules {
+		ruleStatement, err := convert.JsonToDict(rule.Statement)
```
vjeffrey: marius and i chatted about this - the top level items under statement should be objects instead of dict to make querying easier

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,74 @@ private aws.vpc.flowlog @defaults("id region status") {
   tags map[string]string
 }
 
+// Amazon WAF v2
+private aws.waf {
```
scottford-io: A couple of questions here:

1. Does this resource need to be `private`? 
2. Should we define `@defaults()`?

mariuskimmina:
setting to draft again while I work on turning the statement into more fine grained resources

providers/aws/resources/aws.lr
mariuskimmina: What to use as a default here? the `aws.waf` has no fields, it's just a holder for functions

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  // List of waf acls
```
```suggestion
  // List of WAF ACLs
```

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  acls() []aws.waf.acl
+  // List of waf rules
```
```suggestion
  // List of WAF rules
```

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  ruleGroups() []aws.waf.rulegroup
+  // List of waf ip sets
```
```suggestion
  // List of WAF IP sets
```

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+private aws.waf.acl @defaults("name") {
+  // ARN of the ACL
+  arn string
+  // ID of the ACL
+  id string
+  // Name of the ACL
+  name string
+  // Description of the ACL
+  description string
+  // If the ACL is managed by FirewallManager
```
```suggestion
  // Whether the ACL is managed by Firewall Manager
```

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  managedByFirewallManager bool
+  // List of waf rules
```
misterpantz, line 164:
```suggestion
  // List of WAF rules
```

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  rules() []aws.waf.rule
+}
+
+// Amazon WAF rule
+private aws.waf.rule @defaults("name") {
+  // Name of the rule
+  name string
+  // Priority
```
misterpantz: What's the range?

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  priority int
+  // Statement
```
```suggestion
  // Part of the rule that tells WAF how to inspect a web request
```

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  statement aws.waf.rule.statement
+  // Action
```
```suggestion
  // Part of the rule that tells WAF what to do with a web request when it matches the criteria defined in the rule
```

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+}
+
+private aws.waf.rule.statement.sqlimatchstatement.fieldtomatch.singlequeryargument {
+  name string
+}
+
+private aws.waf.rule.statement.sqlimatchstatement.fieldtomatch.singleheader {
+  name string
+}
+
```
misterpantz: missing docs?
If you add basic help text, I'll flesh it out

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+// Amazon WAF v2 ACL
+private aws.waf.rulegroup @defaults("name") {
+  // ARN of the rulegroup
```
```suggestion
  // ARN of the rule group
```

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  arn string
+  // ID of the rulegroup
```
```suggestion
  // ID of the rule group
```

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  id string
+  // Name of the rulegroup
```
```suggestion
  // Name of the rule group
```

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  name string
+  // Description of the rulegroup
```
```suggestion
  // Description of the rule group
```

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  description string
+  // List of waf rules
```
```suggestion
  // List of WAF rules
```

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  rules() []aws.waf.rule
+}
+
+// Amazon WAF v2 ACL
```
misterpantz: This looks like copy/paste :)

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+
+private aws.waf.rule.statement.sqlimatchstatement.fieldtomatch.singlequeryargument {
+  name string
+}
+
+private aws.waf.rule.statement.sqlimatchstatement.fieldtomatch.singleheader {
+  name string
+}
+
+// Amazon WAF v2 ACL
```
misterpantz: This looks like copy/paste :)

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  name string
+  // Description of the rulegroup
+  description string
+  // List of waf rules
+  rules() []aws.waf.rule
+}
+
+// Amazon WAF v2 ACL
+private aws.waf.ipset @defaults("name") {
+  // ARN of the ipset
```
misterpantz, line 424:
```suggestion
  // ARN of the IP set
```

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  arn string
+  // ID of the ipset
```
misterpantz, line 426:
```suggestion
  // ID of the IP set
```

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  id string
+  // Name of the ipset
```
misterpantz, line 430:
```suggestion
  // Name of the IP set
```

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  name string
+  // Description of the ipset
```
misterpantz, line 432:
```suggestion
  // Description of the IP set
```

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,159 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  description string
+  // AddressType either ipv4 or ipv6
```
misterpantz, line 434:
```suggestion
  // Address type: ipv4 or ipv6
```

misterpantz (commented):
Hi @mariuskimmina! @tas50 asked me to take a look at the docs. I might have jumped in before you're ready. 😄

mariuskimmina:
Tracking progress on the more fine grained structs

- [x] SizeConstraintStatement
- [x] GeoMatchStatement
- [ ] XssMatchStatement
- [ ] SqliMatchStatement
- [ ] ByteMatchStatement
- [ ] LabelMatchStatement
- [ ] IPSetReferenceStatement
- [ ] ManagedRuleGroupStatement
- [ ] RatebasedStatement
- [ ] RegexMatchStatement
- [ ] RegexPatternSetReferenceStatement
- [ ] AndStatement
- [ ] OrStatement
- [ ] NotStatement
- [ ] Actions

mariuskimmina:
Also still a lot of docs missing, this was a lot too add ...

providers/aws/resources/aws.lr
chris-rock: That cannot be private, otherwise the resource does not show up properly in auto-completion. Since the user need to call `aws.waf.acls()` it is a good idea to do it.

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,269 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+}
+
+private aws.waf.rule.action {
+  allow aws.waf.rule.action.allow
+  block aws.waf.rule.action.block
+  count aws.waf.rule.action.count
+  captcha aws.waf.rule.action.captcha
+}
+
+private aws.waf.rule.action.allow { }
```
chris-rock: the benefit of this approach is unclear to me at this point

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,269 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  // Description of the ACL
+  description string
+  // Whether the ACL is managed by Firewall Manager
+  managedByFirewallManager bool
+  // List of WAF rules
+  rules() []aws.waf.rule
+}
+
+// Amazon WAF rule
+private aws.waf.rule @defaults("name") {
```
chris-rock, line 171: aws.waf.rulegroup should move above rule to make it easier to read

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,269 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  count aws.waf.rule.action.count
+  captcha aws.waf.rule.action.captcha
+}
+
+private aws.waf.rule.action.allow { }
+private aws.waf.rule.action.block { }
+private aws.waf.rule.action.count { }
+private aws.waf.rule.action.captcha { }
+
+private aws.waf.rule.statement {
```
chris-rock: all those resources need better defaults and best match what we see in aws dashboard

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,269 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  // Priority
+  priority int
+  // Part of the rule that tells WAF how to inspect a web request
+  statement aws.waf.rule.statement
+  // Part of the rule that tells WAF what to do with a web request when it matches the criteria defined in the rule
+  action aws.waf.rule.action
+}
+
+private aws.waf.rule.action {
+  allow aws.waf.rule.action.allow
```
chris-rock: I think we can dramatically simplify this. In the UI I see three different things attached to a action. Maybe something more like this:

action {
  action string
  response string
  labels []string
}

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,269 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+
+private aws.waf.rule.action.allow { }
+private aws.waf.rule.action.block { }
+private aws.waf.rule.action.count { }
+private aws.waf.rule.action.captcha { }
+
+private aws.waf.rule.statement {
+  // random uuid, to make the statement unique, these don't exist in AWS
+  id string
+  sqliMatchStatement aws.waf.rule.statement.sqlimatchstatement
```
chris-rock, line 202: This makes it difficult to check, since you as a user do not really know which check type is used. I do not think this current approach is good to iterate through a tree of statements

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,269 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+}
+
+private aws.waf.rule.statement.sqlimatchstatement {
+  ruleName string
+  fieldToMatch aws.waf.rule.fieldtomatch
+  sensitivityLevel string
+}
+
+// Amazon WAF v2 ACL
+private aws.waf.rulegroup @defaults("name") {
```
chris-rock: should probably be `aws.waf.ruleGroup`

providers/aws/resources/aws_waf.go
```diff
@@ -0,0 +1,725 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/wafv2"
+	waftypes "github.com/aws/aws-sdk-go-v2/service/wafv2/types"
+	"github.com/google/uuid"
+
```
chris-rock, line 14: nit: not properly formatted, use gofumpt

providers/aws/resources/aws_waf.go
```diff
@@ -0,0 +1,725 @@
...
+	"go.mondoo.com/cnquery/v9/providers/aws/connection"
+	"go.mondoo.com/cnquery/v9/types"
+)
+
+func (a *mqlAwsWaf) id() (string, error) {
+	return "aws.waf", nil
+}
+
+func (a *mqlAwsWafAcl) id() (string, error) {
+	return a.Id.Data, nil
```
chris-rock: If possible, those ids should be the arn to make sure they are globally unique

providers/aws/resources/aws.lr
mariuskimmina: I have been mimicking what the aws api is giving me, how would you do it? I could add an additional field `kind` that contains the kind of statement we are dealing with.

providers/aws/resources/aws.lr
mariuskimmina: There is more under allow that we could add here (custom response for exmaple)

providers/aws/resources/aws.lr
mariuskimmina: So I could try to flatten the whole thing so that we have something like

```
rule.statement { 
  kind: sqliMatchStatement
  sensitivityLevel: low
  ... whatever other fields we need for a sqliMatchStatement
}

Comment (mariuskimmina) [reply providers/aws/resources/aws.lr:202]: What do you think about the statement like this @chris-rock 

```
aws.waf.acls.first.rules[1].statement: {
  size: null
  searchString: "zzzzz"
  statements: null
  key: null
  fallbackBehavior: null
  VendorName: null
  ruleName: "marius-test-two"
  Name: null
  kind: "ByteMatchStatement"
  headerName: null
  statement: null
  fieldToMatch: aws.waf.rule.fieldtomatch id = marius-test-two
  countryCodes: null
  regexString: null
  comparisonOperator: null
  position: null
  scope: null
  arn: null
```

This puts all the fields directly on the `statement`, I added the `kind` field to tell what kind of statement it is - all fields that are not part of a `ByteMatchStatement` in this case are null - would that be better?

File: providers/aws/resources/aws.lr (hunk 29 above)
Comment (mariuskimmina) [reply providers/aws/resources/aws.lr]: I thought we never capitalize in these?
We also have for example 

```
private aws.iam.usercredentialreportentry @defaults("arn") {
```

Without any capitalization

File: providers/aws/resources/aws.lr (hunk 27 above)
Comment (mariuskimmina) [reply providers/aws/resources/aws.lr]: Adopted this

File: providers/aws/resources/aws_waf.go (hunk 30 above)
Comment (mariuskimmina) [reply providers/aws/resources/aws_waf.go:14]: Running `gofumpt -w providers/aws/resources/aws_waf.go` did not change this

File: providers/aws/resources/aws.lr (hunk 28 above)
Comment (mariuskimmina) [reply providers/aws/resources/aws.lr:202]: For the record, I ended up not doing the above, we still have all the subtypes of statements. Even if we add all fields to the statement, the users still has to first check what kind of statement he is dealing with (which he can now do with the field `kind` ).

> This makes it difficult to check, since you as a user do not really know which check type is used

I don't thing that there is any way around this, you have to first know which check type is used. If you don't know which check type is used you can't know what fields you would be looking for.

File: providers/aws/config/config.go [hunk 32]
```diff
@@ -84,6 +84,12 @@ var Config = plugin.Provider{
 					Default: "",
 					Desc:    "Override option for EBS scanning that tells it to not create the snapshot or volume",
 				},
+				{
+					Long:    "scope",
```
vjeffrey, line 88: i would expect to be able to specify this filter at the resource level, not the cli level

providers/aws/resources/aws.lr
```diff
@@ -121,6 +121,312 @@ private aws.vpc.flowlog @defaults("id region status") {
...
+  byteMatchStatement aws.waf.rule.statement.bytematchstatement
+  regexMatchStatement aws.waf.rule.statement.regexmatchstatement
+  geoMatchStatement aws.waf.rule.statement.geomatchstatement
+  ipSetReferenceStatement aws.waf.rule.statement.ipsetreferencestatement
+  labelMatchStatement aws.waf.rule.statement.labelmatchstatement
+  managedRuleGroupStatement aws.waf.rule.statement.managedrulegroupstatement
+  notStatement aws.waf.rule.statement.notstatement
+  orStatement aws.waf.rule.statement.orstatement
+  andStatement aws.waf.rule.statement.andstatement
+  rateBasedStatement aws.waf.rule.statement.ratebasedstatement
```
vjeffrey, line 213: this really is a perfect example of where we could use union types. 
but lacking union types, what about adding a "resolvedStatementJson" field that just puts the statement into a json dict, so folks can see the thing easily without first knowing the kind?

providers/aws/config/config.go
mariuskimmina: I was thinking about that, but how? When inside the shell providing arguments to resources doesn't seem possible at the moment
```
failed to compile: cannot call resource field with arguments yet
```

providers/aws/resources/aws.lr
mariuskimmina: In addition to the current types you mean? Yea could definetly do that.

vjeffrey: ya in addition, just so there's a way to see it without drilling in all the way

providers/aws/config/config.go
vjeffrey: i was thinking similar to how we init objects sometimes with a value, like what u did here for example: https://github.com/mondoohq/cnquery/pull/2753/files

mariuskimmina: ah, on the waf object we could do it, so that we end up with something like

```
aws.waf(scope: "CLOUDFRONT").rules
```

vjeffrey: yaaa!

mariuskimmina: This doesn't seem to work because changing it does not invalidate the cache, leading to situations like this

```
cnquery> aws.waf(scope: "REGIONAL").acls
aws.waf.acls: [
  0: aws.waf.acl name="marius-test"
]
cnquery> aws.waf(scope: "CLOUDFRONT").acls
aws.waf.acls: [
  0: aws.waf.acl name="marius-test"
]
```

In the example above, I should not get the same result with a changed scope, but the previous cached result is returned.

providers/aws/resources/aws.lr
mariuskimmina:
```
aws.waf.acls.first.rules.first.statement.json: {
  AndStatement: null
  ByteMatchStatement: null
  GeoMatchStatement: null
  IPSetReferenceStatement: null
  LabelMatchStatement: null
  ManagedRuleGroupStatement: null
  NotStatement: null
  OrStatement: null
  RateBasedStatement: null
  RegexMatchStatement: null
  RegexPatternSetReferenceStatement: null
  RuleGroupReferenceStatement: null
  SizeConstraintStatement: null
  SqliMatchStatement: {
    FieldToMatch: {
      AllQueryArguments: null
      Body: {
        OversizeHandling: "MATCH"
      }
      Cookies: null
      HeaderOrder: null
      Headers: null
      JA3Fingerprint: null
      JsonBody: null
      Method: null
      QueryString: null
      SingleHeader: null
      SingleQueryArgument: null
      UriPath: null
    }
    SensitivityLevel: "LOW"
    TextTransformations: [
      0: {
        Priority: 0.000000
        Type: "NONE"
      }
    ]
  }
  XssMatchStatement: null
}
```

providers/aws/config/config.go
vjeffrey: yaa. i think the only alternative i can think of is to do aws.waf.regional.acls and aws.waf.cloudfront.acls kinda thing, so that we have two different aws waf prefixes.

i think this solution works for now, we'll see how users respond

PR #2696, opened 2023-11-29
🐛 Allow mondoo-tools to push

czunker:
We can only assign permissions on a job level.

But a dependabot PR runs on the workflow files from main. It should not include updated workflows


czunker:
This should prevent
```
fatal: unable to access 'https://github.com/mondoohq/cnquery/': The requested URL returned error: 403
```
https://github.com/mondoohq/cnquery/actions/runs/7006338996/job/19057980829

czunker:
We will implement this differently.

PR #2699, opened 2023-11-30
🧹 dump ms365 output in debug files, stop scripts on error.

preslavgerchev:
Use `logger.DebugDump` to get the script output out for an improved debugging experience. Also stop the scripts if there's an error, avoiding issues where we get back an empty JSON object since one of the cmdlets has failed


arlimus (approved):
Helpful for debugging!

PR #2713, opened 2023-11-30
🧹 Align atlassian provider ID

czunker:
All other providers include v9.

### DIFF
--- a/providers/atlassian/config/config.go
+++ b/providers/atlassian/config/config.go
@@ -10,7 +10,7 @@ import (
 
 var Config = plugin.Provider{
 	Name:    "atlassian",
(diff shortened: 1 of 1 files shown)


providers/atlassian/config/config.go
```diff
@@ -10,8 +10,8 @@ import (
 
 var Config = plugin.Provider{
 	Name:    "atlassian",
-	ID:      "go.mondoo.com/cnquery/providers/atlassian",
-	Version: "9.1.6",
+	ID:      "go.mondoo.com/cnquery/v9/providers/atlassian",
+	Version: "9.1.7",
```
czunker: Not sure wether the bump is also needed.

vjeffrey (approved):
thank you!

PR #2717, opened 2023-11-30
Adds additional AWS IAM policy elements

scottford-io:
Adds additional AWS IAM policy elements:

`Principal`
`NotPrincipal`
`NotResource`
`Id`

@vjeffrey We will also need to add `Condition` but I could use your support on that. 

`Condition` - https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition.html

### DIFF
--- a/providers/aws/resources/awspolicy/testdata/iam_policy4.json
+++ b/providers/aws/resources/awspolicy/testdata/iam_policy4.json
@@ -1,0 +1,17 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [{
+      "Effect": "Allow",
+      "Principal" : { 
+        "AWS": [ 
+          "123456789012",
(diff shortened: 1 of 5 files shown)


providers/aws/resources/awspolicy/testdata/iam_policy4.json
```diff
@@ -0,0 +1,17 @@
...
+  "Statement": [{
+      "Effect": "Allow",
+      "Principal" : { 
+        "AWS": [ 
+          "123456789012",
+          "555555555555" 
+          ]
+        },
+      "Action": "s3:*",
+      "Resource": [
```
vjeffrey, line 12: could we modify this test to use "NotResource" ?

vjeffrey (approved):

PR #2719, opened 2023-12-01
🐛 Fix image names for k8s container discovery

czunker:
This fixes the asset name when images are discovered inside a kubernetes cluster.

Now the name is the same as with v8


czunker:
Follow-up to https://github.com/mondoohq/cnquery/pull/2633

imilchev (approved):
LGTM

PR #2726, opened 2023-12-01
🧹 Automatically create PR to update all deps

czunker:
This creates a PR to update cnquery and provider deps. Dpendabot PRs on the cnquery go.mod do not work, because they mostly also require updates on the provider go.mods.


chris-rock (approved):
@czunker That is great, lets try it.

PR #2729, opened 2023-12-03
Put Devuan into Debian family

tas50:
It's a non-systemd forked of Debian. With this change the package resource works.

```
cnquery> asset.family
asset.family: [
  0: "debian"
  1: "linux"
  2: "unix"
  3: "os"
]
cnquery> packages.list.first
packages.list.first: package name="libpam-runtime" version="1.1.8-3.6"
```

### DIFF
--- a/providers/os/detector/detector_all.go
+++ b/providers/os/detector/detector_all.go
@@ -214,6 +214,17 @@ var popos = &PlatformResolver{
 	},
 }
 
+var devuan = &PlatformResolver{
+	Name:     "devuan",
+	IsFamily: false,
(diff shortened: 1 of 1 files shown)


chris-rock:
We need to be careful with putting Devuan into the debian family. Some resources expect specific behavior with specific families. We need to double-check that the resources handle that well. I am not sure if putting it into the same family makes really sense. Other rhel forks like amazon linux are also not in rhel family. Is there any specific reason why this is required?

chris-rock:
I close this for now. Please re-open if you feel there is a strong reason to add devuan to the debian family

PR #2738, opened 2023-12-04
🐛 Fix missing platform for container image discovery

czunker:
This fixes the error:
'unable to detect platform for asset ...'

### DIFF
--- a/providers/runtime.go
+++ b/providers/runtime.go
@@ -250,6 +250,14 @@ func (r *Runtime) Connect(req *plugin.ConnectReq) error {
 		}
 	}
 
+	// We have a special case here, where we loose the platform information
+	// when doing container-image discovery for k8s.
+	// This way we make sure to preserve the platform information across plugins.
+	pf := r.Provider.Connection.Asset.Platform
+	if req.Asset.Platform == nil {
+		req.Asset.Platform = pf
+	}
+
 	r.Recording.EnsureAsset(r.Provider.Connection.Asset, r.Provider.Instance.ID, r.Provider.Connection.Id, asset.Connections[0])
 	r.schema.prioritizeIDs(BuiltinCoreID, r.Provider.Instance.ID)
 	return nil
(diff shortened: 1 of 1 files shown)


czunker:
Related to https://github.com/mondoohq/cnspec/pull/906

I'm not sure why this worked before. I assume the platform got populated later, because we didn't see the container discovery panic.

providers/runtime.go
```diff
@@ -250,6 +250,14 @@ func (r *Runtime) Connect(req *plugin.ConnectReq) error {
 		}
 	}
 
+	// We have a special case here, where we loose the platform information
```
imilchev: nit: 
```suggestion
	// We have a special case here, where we lose the platform information
```

providers/runtime.go
```diff
@@ -250,6 +250,14 @@ func (r *Runtime) Connect(req *plugin.ConnectReq) error {
...
+	// when doing container-image discovery for k8s.
+	// This way we make sure to preserve the platform information across plugins.
+	pf := r.Provider.Connection.Asset.Platform
+	if req.Asset.Platform == nil {
```
imilchev, line 257: shouldn't this be something that the provider needs to fill in? I would expect that when you connect to the provider for a given asset, it will fill in the platform info

czunker: The provider does fill it, but because of the plugins, it's gone at this point.

It's again one of the cases where builtins work and plugins do not. At some point we loose a pointer.

imilchev: I don't want to over-complicate this PR but isn't it better to figure out where we need to explicitly store the response of `Connect`? Otherwise in the future some other field may be missing and we need to add another condition here

czunker: Another solution: https://github.com/mondoohq/cnspec/pull/981

czunker: This is where we get the response from the plugin:
https://github.com/mondoohq/cnquery/blob/main/providers/runtime.go#L223

```
r.Provider.Connection, r.Provider.ConnectionError = r.Provider.Instance.Plugin.Connect(req, &callbacks)
```

But we do not return it from there, so we have to use it explicitly. At least this is my current understanding.

czunker:
Close in favor of https://github.com/mondoohq/cnspec/pull/981

PR #2741, opened 2023-12-05
✨ Get network provider to work with inventory file

mariuskimmina:


providers/network/connection/connection.go
```diff
@@ -14,11 +14,12 @@ type HostConnection struct {
 }
 
 func NewHostConnection(id uint32, asset *inventory.Asset, conf *inventory.Config) *HostConnection {
-	return &HostConnection{
+	conn := &HostConnection{
```
imilchev: i guess this can be reverted

providers/network/resources/tls.go
```diff
@@ -28,19 +28,21 @@ var rexUrlDomain = regexp.MustCompile(regex.UrlDomain)
...
-	if conn.Conf.Port != 0 {
-		return int64(conn.Conf.Port)
+	port := conn.Conf.Options["port"]
+	portNumber, _ := strconv.ParseInt(port, 10, 64)
+	if portNumber != 0 {
+		return int64(portNumber)
 	}
 
-	if conn.Conf.Runtime == "" {
+	if conn.Conf.Options["runtime"] == "" {
```
imilchev: same here

mariuskimmina: The first one I agree, this second one I don't see what you mean

providers/network/provider/provider.go
```diff
@@ -179,15 +188,16 @@ func (s *Service) connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 }
 
 func (s *Service) detect(asset *inventory.Asset, conn *connection.HostConnection) error {
-	asset.Name = conn.Conf.Host
+	asset.Name = conn.Conf.Options["host"]
```
imilchev: the host from the options should be set in the conf already at this point

providers/network/resources/http.go
```diff
@@ -55,21 +55,26 @@ func initHttpGet(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[str
...
 		}
 
-		scheme := conn.Conf.Runtime
+		scheme := conn.Conf.Options["runtime"]
 		if scheme == "" {
 			// At this point we are in best effort territory. Which means we will
 			// go HTTP unless port 443 is specified. Users can always provide a
 			// scheme to be prescriptive.
-			if conn.Conf.Port == 443 {
+			if conn.Conf.Options["port"] == "443" {
```
imilchev: this should be mapped already to the right property. You shouldn't need to use `Options` here

providers/network/provider/provider.go
mariuskimmina: Not the case - Where should this mapping happen?

providers/network/provider/provider.go
```diff
@@ -78,15 +87,17 @@ func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error)
...
+	conf.Options["port"] = fmt.Sprint(port)
+	conf.Options["type"] = "host"
+	conf.Options["path"] = url.Path
+	conf.Options["runtime"] = scheme
+	conf.Options["insecure"] = fmt.Sprintf("%t", insecure)
+	conf.Host = host
+	conf.Port = int32(port)
+
 	asset := inventory.Asset{
-		Connections: []*inventory.Config{{
```
imilchev: this looks like the reverse of parsing the cli. From parsed parameters you are converting them back to strings

providers/network/resources/http.go
```diff
@@ -67,9 +67,14 @@ func initHttpGet(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[str
 			}
 		}
 
+		port := conn.Conf.Options["port"]
```
imilchev: at this point `conn.Conf.Port` should be set. This is what I meant by the previous comment I put here.

providers/network/resources/http.go
```diff
@@ -55,7 +55,7 @@ func initHttpGet(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[str
 			return nil, nil, errors.New("missing URL for http.get")
 		}
 
-		scheme := conn.Conf.Runtime
```
imilchev: runtime should be set. Why do you parse it from the options?

providers/network/resources/tls.go
imilchev: runtime should be set, you shouldn't need to use the options for properties that are part of the config already. you should be reading values from the options only if they don't have a dedicated property in the config itself

providers/network/resources/tls.go
```diff
@@ -32,11 +32,11 @@ func connTlsPort(conn *connection.HostConnection) int64 {
 		return int64(conn.Conf.Port)
 	}
 
-	if conn.Conf.Runtime == "" {
+	if conn.Conf.Options["runtime"] == "" {
 		return 443
 	}
 
-	port := CommonPorts[conn.Conf.Runtime]
```
imilchev: same here

imilchev (approved):

PR #2746, opened 2023-12-05
⭐️ allow users to ignore .terraform directory

chris-rock:
fixes  #605

### DIFF
--- a/providers/terraform/connection/hcl_manifest.go
+++ b/providers/terraform/connection/hcl_manifest.go
@@ -56,9 +56,9 @@ func newHclConnection(path string, asset *inventory.Asset) (*Connection, error) 
 	}
 
 	confOptions := asset.Connections[0].Options
-	includeTerraform := true
+	includeDotTerraform := true
 	if confOptions["ignore-dot-terraform"] == "true" {
-		includeTerraform = false
+		includeDotTerraform = false
 	}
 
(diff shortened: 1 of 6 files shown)


providers/terraform/connection/hcl_manifest.go
```diff
@@ -51,6 +51,15 @@ func NewHclConnection(id uint32, asset *inventory.Asset) (*Connection, error) {
 
 func newHclConnection(path string, asset *inventory.Asset) (*Connection, error) {
 	// NOTE: right now we are only supporting to load either state, plan or hcl files but not at the same time
+	if len(asset.Connections) != 1 {
+		return nil, errors.New("only one connection is supported")
+	}
+
+	confOptions := asset.Connections[0].Options
+	includeTerraform := true
```
czunker: nit: Should we rename this to `includeDotTerraform`? As it is now, it somehow reads strange to not include terraform in a terraform connection.

czunker (approved):
Thanks @chris-rock Gave this a try with some other terraform manifest and it works as expected.

providers/terraform/connection/hcl_manifest.go
chris-rock: Great idea

czunker (approved):
Thanks @chris-rock

Issue #2739, opened 2023-12-04
github.mergeRequest results in `cannot convert primitive with NO type information`


JosiahOne:
**Describe the bug**
`github.mergeRequest` results in `cannot convert primitive with NO type information` instead of returning valid data as part of a GitHub repository.

**To Reproduce**

1. Get yourself a PAT and export: `GITHUB_TOKEN="<stuff>"`
2. Connect to a repo: `cnquery shell github repo <ORG>/<REPO>`
3. Run the following query: `github.mergeRequest`

We get:

```
cnquery> github.mergeRequest
1 error occurred:
	* cannot convert primitive with NO type information
github.mergeRequest: github.mergeRequest id=null state=null
```

Further, you can crash the provider by running:

```
cnquery> github.mergeRequest { * }
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0x1b50740]

goroutine 136 [running]:
go.mondoo.com/cnquery/v9/providers/github/resources.(*mqlGithubMergeRequest).reviews(0x1e604c0?)
	/home/runner/_work/cnquery/cnquery/providers/github/resources/github_repo.go:686 +0x60
go.mondoo.com/cnquery/v9/providers/github/resources.(*mqlGithubMergeRequest).GetReviews.func1()
	/home/runner/_work/cnquery/cnquery/providers/github/resources/github.lr.go:4215 +0x8b
go.mondoo.com/cnquery/v9/providers-sdk/v1/plugin.GetOrCompute[...](0xc000282ad8?, 0xc0002ed7c4)
	/home/runner/_work/cnquery/cnquery/providers-sdk/v1/plugin/runtime.go:240 +0x3a
go.mondoo.com/cnquery/v9/providers/github/resources.(*mqlGithubMergeRequest).GetReviews(0x10076d8?)
	/home/runner/_work/cnquery/cnquery/providers/github/resources/github.lr.go:4204 +0x3f
go.mondoo.com/cnquery/v9/providers/github/resources.glob..func208({0x206f420?, 0xc000282960?})
	/home/runner/_work/cnquery/cnquery/providers/github/resources/github.lr.go:803 +0x30
go.mondoo.com/cnquery/v9/providers/github/resources.GetData({0x206f420, 0xc000282960}, {0xc0004c8488, 0x7}, 0x23?)
[...]

PR #2753, opened 2023-12-06
🐛 Github mergeRequest initialization 

mariuskimmina:
This creates an init function for `github.mergeRequest`

Related to: #2739 

With this we can get a single merge request by it's pr number

```
cnquery> github.mergeRequest(number: 2183){ * }
github.mergeRequest: {
  repoName: "cnquery"
  assignees: [
    0: github.user id=38843153 name="Marius Kimmina"
  ]
  state: "closed"
  number: 2183
  labels: []
  reviews: []
  owner: github.user id=10341541 name="vjeffrey"
  commits: []
  title: "🐛 fix missing asset name during local scan reporting upstream"
  id: 1553252490
  createdAt: 2023-10-12 03:06:43 +0200 CEST
}
```

```
cnquery> github.mergeRequest
number must be set for github.mergeRequest initialization
github.mergeRequest: no data available
```

Prevents the previous error on `gitub.mergeRequest` as well as the provider panic on `github.mergeRequest{ * }`


providers/github/resources/github_repo.go
```diff
@@ -124,6 +124,90 @@ func (g *mqlGithubRepository) id() (string, error) {
...
+	if x, ok := args["number"]; ok {
+		number, ok = x.Value.(int64)
+		if !ok {
+			return nil, nil, errors.New("wrong type for 'number' in github.mergeRequest initialization, it must be a number")
+		}
+	}
+	if number == 0 {
+		return nil, nil, errors.New("number must be set for github.mergeRequest initialization")
+	}
+	if len(args) > 2 {
```
imilchev: i think that should be 1 since you only need 1 argument to be provided in order to load a MR

providers/github/resources/github_repo.go
```diff
@@ -124,6 +124,90 @@ func (g *mqlGithubRepository) id() (string, error) {
...
+		}
+		var assignees []interface{}
+		assignees = append(assignees, assignee)
+		prNumber := int64(*pr.Number)
+		args["id"] = llx.IntDataPtr(pr.ID)
+		args["number"] = llx.IntDataPtr(&prNumber)
+		args["state"] = llx.StringDataPtr(pr.State)
+		args["createdAt"] = llx.TimeData(pr.CreatedAt.Time)
+		args["title"] = llx.StringDataPtr(pr.Title)
+		args["owner"] = llx.ResourceData(owner, owner.MqlID())
```
imilchev: can you double-check what is the right second argument for this function. I can see in here that it's using `owner.MqlName()` https://github.com/mondoohq/cnquery/blob/0150f03fbe01efb086418ecf96ee093f2be4ef8e/providers/github/resources/github_repo.go#L424
I think that should be the right one

imilchev (approved):

PR #2756, opened 2023-12-07
🐛 fix for domain scanning integration

mariuskimmina:
Changing the asset name from just the host to the scheme + host so that if we trigger a scan for both http and https they end up with different names.

Also get the port form the options and then set the scheme based on the port.
This together with https://github.com/mondoohq/server/pull/7032

### DIFF
--- a/providers/network/provider/provider.go
+++ b/providers/network/provider/provider.go
@@ -155,6 +155,20 @@ func (s *Service) connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 		conn.Conf.Host = conn.Conf.Options["host"]
 	}
 
+	if conn.Conf.Options != nil && conn.Conf.Options["port"] != "" {
+		portNumber, err := strconv.ParseInt(conn.Conf.Options["port"], 10, 32)
+		if err != nil {
+			return nil, err
+		}
+		conn.Conf.Port = int32(portNumber)
+		if conn.Conf.Port == 80 {
+			conn.Conf.Runtime = "http://"
+		}
+		if conn.Conf.Port == 443 {
+			conn.Conf.Runtime = "https://"
+		}
+	}
+
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


providers/network/provider/provider.go
```diff
@@ -38,6 +38,31 @@ func Init() *Service {
...
+		}},
+	}
+
+	return &plugin.ParseCLIRes{Asset: &asset}, nil
+}
+
+func parseTarget(target string) (string, int, string, string, error) {
 	// Note on noSchema handling:
 	// A user may type in a target like: `google.com`. Technically, this is not
 	// avalid scheme. We need to make it into a valid url scheme for parsing
```
czunker:
```suggestion
	// a valid scheme. We need to make it into a valid url scheme for parsing
```

providers/network/provider/provider.go
```diff
@@ -183,7 +203,8 @@ func (s *Service) connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 }
 
 func (s *Service) detect(asset *inventory.Asset, conn *connection.HostConnection) error {
-	asset.Name = conn.Conf.Host
+	hostWithSchmeme := conn.Conf.Runtime + conn.Conf.Host
```
czunker:
```suggestion
	hostWithScheme := conn.Conf.Runtime + conn.Conf.Host
```

providers/network/provider/provider.go
```diff
@@ -192,7 +213,7 @@ func (s *Service) detect(asset *inventory.Asset, conn *connection.HostConnection
 	}
 
 	asset.Fqdn = conn.FQDN()
-	asset.PlatformIds = []string{"//platformid.api.mondoo.app/runtime/network/host/" + conn.Conf.Host}
+	asset.PlatformIds = []string{"//platformid.api.mondoo.app/runtime/network/host/" + hostWithSchmeme}
```
czunker: Is the scheme `https://`? I'd suggest to trim off `://` from the schema for the platformID.

czunker (approved):
Thanks @mariuskimmina

PR #2774, opened 2023-12-08
🐛 Fix k8s.kubelet readOnlyPort property

czunker:
When this value was set to '0', we missed it.

Related-to https://github.com/mondoohq/cnquery/issues/1831


vjeffrey (approved):
thank you!

PR #2796, opened 2023-12-11
🐛 update to latest cyclone-dx behaviour

chris-rock:
bug introduced in https://github.com/mondoohq/cnquery/pull/2791


imilchev (approved):

czunker:
Fixes https://github.com/mondoohq/cnquery/issues/2795

PR #2797, opened 2023-12-11
Debug GHA PRs

czunker:
This should help to debug, why we didn't have tests for the last PR.

### DIFF
--- a/.github/pr-body.md
+++ b/.github/pr-body.md
@@ -1,3 +1,5 @@
+!!! ONLY A TEST !!!
+
 Created by Mondoo Tools via GitHub Actions
 
 Workflow:
--- a/.github/workflows/pr-test-lint.yml
+++ b/.github/workflows/pr-test-lint.yml
@@ -58,12 +58,18 @@ jobs:
   go-test:
     runs-on: self-hosted
     timeout-minutes: 120
-    steps:         
+    steps:
       - name: Checkout code
         uses: actions/checkout@v4
 
       - name: Import environment variables from file
         run: cat ".github/env" >> $GITHUB_ENV
+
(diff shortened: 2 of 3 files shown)


.github/pr-body.md
```diff
@@ -1,3 +1,5 @@
+!!! ONLY A TEST !!!
+
 Created by Mondoo Tools via GitHub Actions
```
chris-rock, line 3: I do not understand why we need to merge the test. Can we not trigger this without merging to main?

preslavgerchev:
> This should help to debug, why we didn't have tests for the last PR.

Can you link the mentioned PR? We only trigger tests if the source code has changed:
```
## Only trigger tests if source is changing
on:
  push:
    paths:
      - '**.go'
      - '**.mod'
      - 'go.sum'
```

czunker:
> > This should help to debug, why we didn't have tests for the last PR.
> 
> Can you link the mentioned PR? We only trigger tests if the source code has changed:
> 
> ```
> ## Only trigger tests if source is changing
> on:
>   push:
>     paths:
>       - '**.go'
>       - '**.mod'
>       - 'go.sum'
> ```

https://github.com/mondoohq/cnquery/pull/2791

This had changes to `go.mod`s and `go.sum`s, so I would have expected tests to run.

.github/pr-body.md
czunker: I don't think so.

I need additional output in the `pr-test-lint.yaml` which is triggered by the GHA PR. And I also need the trigger from GHA, otherwise actor and some other things would be different, which I suspect to be part of the problem.

chris-rock (approved):

PR #2798, opened 2023-12-11
🧹 allow more configuration for azure client

chris-rock:


arlimus (approved):
Thank you for adding all the config options!

PR #2829, opened 2023-12-13
🐛 do not include architecture as default field for aws ec2 instance yet

vjeffrey:
its only supported on 9.11+, so putting it in defaults breaks things

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -2247,7 +2247,7 @@ private aws.ec2.volume @defaults("id region volumeType size encrypted state") {
 }
 
 // Amazon EC2 instance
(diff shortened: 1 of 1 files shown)


scottford-io (approved):
👍

vjeffrey:
the other thing here, with @tas50 just brought up, is that the min mondoo version doesn't fully apply in our v9 world. so really that architecture field should be marked as 9.0.0 min mondoo version.

that kinda fixes things. but we will still break v8 clients if we have the architecture field in the defaults

tas50:
I updated everything from 9.0 onwards to be 9.0.0 here: https://github.com/mondoohq/cnquery/pull/2830

vjeffrey:
k, i'm gonna close this pr and merge the other one, bc my brain was not being smart yesterday when i said this would break v8 bc obvs it wouldn't bc client versions!

PR #2833, opened 2023-12-14
🧹 re-enable ec2 & ssm instances api discovery

vjeffrey:
https://github.com/mondoohq/cnquery/issues/2623


scottford-io (approved):
Thanks @vjeffrey I just tested locally and it is working as expected! 🙌

PR #2848, opened 2023-12-15
✨ add out of scope action

imilchev:
add an out of scope action that will be used for compliance controls and assets


explorer/cnquery_explorer.proto
```diff
@@ -136,6 +136,7 @@ enum Action {
   DEACTIVATE = 2;
   ACTIVATE = 3;
   IGNORE = 4;
+  OUT_OF_SCOPE = 5;
```
preslavgerchev, line 139: what is the difference between this and ignore?

imilchev: Out of scope is closer to `DEACTIVATE` rather than `IGNORE`. Essentially, out of scope will create a different type of exception that doesn't show up in reporting or anywhere else but it will still disable execution

chris-rock (approved):
Thank you @imilchev

PR #2849, opened 2023-12-15
✨ Multiple ms365 resource updates

preslavgerchev:
* Add `visibility` to `microsoft.group`
* Add `AllowedToCreateTenants` to the auth policy in `microsoft.policies`
* Ensure PnP version is set to latest so we get back all the required props. Use `-EnumsAsStrings` to get a better representation for the enums in there


HRouhani:
@chris-rock I tested it and it works as expected.

imilchev (approved):

PR #2855, opened 2023-12-16
⭐️  Add ms365.sharepointonline.site resource

HRouhani:


HRouhani:
Screenshot from 2023-12-16 15-22-26

HRouhani:
Screenshot from 2023-12-16 15-31-39

HRouhani:
@preslavgerchev    my guess is that ConvertTo-Json   does not remove [] !

providers/ms365/resources/ms365.lr
```diff
@@ -398,6 +398,16 @@ ms365.sharepointonline {
   spoTenant dict
   // SharePoint Online tenant sync client restriction
   spoTenantSyncClientRestriction dict
+  // SharePoint Online tenant sites
+  spoSites []ms365.sharepointonline.site
```
preslavgerchev, line 402: @HRouhani i have made this to be a list, any reason why it cannot be?

providers/ms365/connection/sharepoint_report.go
```diff
@@ -17,20 +17,22 @@ var sharepointReport = `
...
 Connect-PnPOnline -AccessToken $token -Url $url
 
 $SPOTenant = (Get-PnPTenant)
 $SPOTenantSyncClientRestriction = (Get-PnPTenantSyncClientRestriction)
+$SPOSite = (Get-PnPTenantSite)
 
 $sharepoint = New-Object PSObject
 Add-Member -InputObject $sharepoint -MemberType NoteProperty -Name SPOTenant -Value $SPOTenant
 Add-Member -InputObject $sharepoint -MemberType NoteProperty -Name SPOTenantSyncClientRestriction -Value $SPOTenantSyncClientRestriction
+Add-Member -InputObject $sharepoint -MemberType NoteProperty -Name SPOSite -Value $SPOSite
```
preslavgerchev, line 31: lets return all sites from the script. even it we dont end up using them its better to keep this script as simple as possible and do any other logic in the provider itself

preslavgerchev:
@HRouhani i have made some changes so that we keep adding resources-only instead of dicts (specially for top-level stuff)

providers/ms365/connection/sharepoint_report.go
HRouhani: awesome idea, Thanks for taking care of it

providers/ms365/resources/ms365.lr
HRouhani: Thanks, looks really good.

HRouhani:
@preslavgerchev  The one which is need for now is ms365.sharepointonline.spoSites, which works as expected and therefor this PR can be merged. The one related to ms365.sharepointonline.site {*} is not working and from my perspective is not needed now. So you can improve and fix it whenever you got time. I just put it here for your info. Thanks for your Great work

preslavgerchev:
> @preslavgerchev The one which is needed for now is ms365.sharepointonline.spoSites, which works as expected and therefor this PR can be merged. The one related to ms365.sharepointonline.site {*} is not working and from my perspective is not needed now. So you can improve and fix it whenever you got time. I just put it here for your info. Thanks for your Great work
> 
> Screenshot from 2023-12-18 10-59-07

`ms365.sharepointonline.site` is just the resource itself, we don't support this currently as it has no init function, only `ms365.sharepointonline.spoSites` is supported

imilchev (approved):

PR #2858, opened 2023-12-16
🐛 set default sudo executable

chris-rock:
This fixes an issue reported by @username-is-already-taken2. Packer plugin properly enables the sudo config via the inventory https://github.com/mondoohq/packer-plugin-cnspec/blob/a4c6be26de7ac51f49259fd9ea4501663d43ea87/provisioner/provisioner.go#L335-L340. 

Problem was that no sudo executable was set. The cli flag always set the executable as well https://github.com/mondoohq/cnquery/blob/main/providers/os/connection/shared/shared.go#L197-L212. Therefore the inventory for the packer scan used the following sudo setting:

```json
"sudo": {
  "active": true
}
```

This PR changes the default and always falls back to `sudo` cli if no alternative executable was set.


mariuskimmina (approved):
LGTM

arlimus (approved):
Kudos!

PR #2859, opened 2023-12-17
🧹 improve slack conversations handling

chris-rock:
- allows users to filter by top level categories without where queries which leads to performance improvements since we only fetch the relevant data
- implement handling for rate-limiting since slack go sdk is not doing that for the conversations api

### DIFF
--- a/providers/slack/resources/conversations.go
+++ b/providers/slack/resources/conversations.go
@@ -19,7 +19,7 @@ func (o *mqlSlackConversations) id() (string, error) {
 	return "slack.conversations", nil
 }
 
-func (s *mqlSlackConversations) listChannels(types []string) ([]interface{}, error) {
+func (s *mqlSlackConversations) listChannels(types ...string) ([]interface{}, error) {
 	conn := s.MqlRuntime.Connection.(*connection.SlackConnection)
 	client := conn.Client()
 	if client == nil {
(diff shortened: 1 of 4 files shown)


providers/slack/resources/conversations.go
```diff
@@ -7,14 +7,19 @@ import (
...
 	"go.mondoo.com/cnquery/v9/providers-sdk/v1/util/convert"
 	"go.mondoo.com/cnquery/v9/providers/slack/connection"
 )
 
-func (s *mqlSlack) conversations() ([]interface{}, error) {
+func (o *mqlSlackConversations) id() (string, error) {
+	return "slack.conversations", nil
+}
+
+func (s *mqlSlackConversations) listChannels(types []string) ([]interface{}, error) {
```
arlimus: nit: if we did `listChannels(types ...string)` here we could call it with `listChannels("private_channel")` etc below; but really just a nit, no real impact...

arlimus (approved):
Nice builtin filters!

providers/slack/resources/conversations.go
chris-rock: I like it and adjust the code

chris-rock: fixed

PR #2860, opened 2023-12-18
🧹 add back in ecr-api and ecs-api discovery

vjeffrey:
also adds instance type to ec2 instances asset labels so we can filter on that value later 

https://github.com/mondoohq/cnquery/issues/2623


chris-rock (approved):
Thank you @vjeffrey

PR #2870, opened 2023-12-18
🧹 introduce retry handling for slack api

chris-rock:
- global retry logic for slack api
- exclude archived channels for `slack.conversations.publicChannels` and `slack.conversations.privateChannels`


arlimus (approved):
kudos!

PR #2873, opened 2023-12-19
Update the go-github package from 55 -> 57

tas50:
This includes a few nice new capabilities we can utilize in the future + the usual bumps / fixes.


chris-rock (approved):
Thank you @tas50 Great update.

PR #2882, opened 2023-12-19
🧹 Update deps 20231219

czunker:


czunker:
We need to bump k8s separatly: https://github.com/mondoohq/cnquery/issues/2883

chris-rock (approved):

PR #2892, opened 2023-12-20
⭐️ includes the platform ids and ec2 instance arn in sbom exports

chris-rock:
This PR ensures that the platform ids are also added to the sbom


arlimus (approved):
Thank you!

PR #2904, opened 2023-12-21
🐛 Use ssh key instead of GH token

czunker:
When the PR is created with a GH token, the tests aren't executed. This is becasue of limitation applied to the token.

These limiations do not apply to ssh keys.


czunker:
And older PR without tests:
https://github.com/mondoohq/cnquery/pull/2791

A new one with tests:
https://github.com/mondoohq/cnquery/pull/2902

chris-rock (approved):
Thank you @czunker

PR #2921, opened 2023-12-27
Minor changes to resource manifest docs

misterpantz:
Is it okay to change the platform values like I did here?

### DIFF
--- a/providers/arista/resources/arista.lr.manifest.yaml
+++ b/providers/arista/resources/arista.lr.manifest.yaml
@@ -17,7 +17,7 @@ resources:
     min_mondoo_version: 5.15.0
     platform:
       name:
-      - arista-eos
+      - Arista EOS
     snippets:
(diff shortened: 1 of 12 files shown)


providers/arista/resources/arista.lr.manifest.yaml
```diff
@@ -17,7 +17,7 @@ resources:
     min_mondoo_version: 5.15.0
     platform:
       name:
-      - arista-eos
+      - Arista EOS
```
chris-rock, line 20: Those names match the platform name identifier. They should stay the way they in that document. If we need better names in the docs, we need to adjust the docs command to make this happen.

misterpantz: Ah, thank you. Well is there any reason why we should reveal the platform name identifier to the readers of the docs?

PR #2939, opened 2024-01-04
✨ add gql query for searchPolicy

mariuskimmina:
This will be used by the new `cnspec policy` command which is being worked on here https://github.com/mondoohq/cnspec/pull/1001

This is required to determine the active policies on the connected space (`policy list`)


providers-sdk/v1/upstream/gql/space_report_gql.go
```diff
@@ -0,0 +1,77 @@
...
+			SpaceReport `graphql:"... on SpaceReport"`
+		} `graphql:"spaceReport(input: $input)"`
+	}
+	orderBy := PolicyReportSummaryOrder{
+		Direction: "DESC",
+		Field:     "TITLE",
+	}
+	err := c.Query(context.Background(), &m, map[string]interface{}{
+		"input":   mondoogql.SpaceReportInput{SpaceMrn: mondoogql.String(mrn)},
+		"first":   graphql.Int(10),
```
mariuskimmina: Not sure what to use here

providers-sdk/v1/upstream/gql/policy_gql.go
```diff
@@ -0,0 +1,128 @@
...
+
+type PolicyNode struct {
+	// The Policy struct is embedded to match the inline fragment ... on Policy.
+	// We assume that all these fields are part of the Policy type.
+	*Policy `graphql:"... on Policy"`
+}
+
+// Policy contains fields that are expected to be part of the Policy type
+// within the GraphQL schema.
+type Policy struct {
```
preslavgerchev, line 30: why do we have this in cnquery instead of cnspec? can we not move this file over there? feels weird that cnquery defines something policy-related, specially when it's unused

mariuskimmina: Hmmm yea I guess you are right, I just thought since all the previous gql queries where defined in cnquery I should add it here. I'll close this PR and add it to the cnspec one. Good Point

PR #2943, opened 2024-01-04
⭐️ Azure snapshot/instance scanning

preslavgerchev:
This only support snapshot/instance scanning in the same resource group. We can extend this to work across resource groups in the future by specifying the fully qualified resource ids.

To test, run
```
sudo DEBUG=1 ./cnquery shell azure compute snapshot <snapshot-name> --client-id <id> --tenant-id <id> --client-secret <value>
``` 

in a VM, located in the same RG as the snapshot

Or if you want to scan an instance:

```
sudo DEBUG=1 ./cnquery shell azure compute instance <instance-name> --client-id <id> --tenant-id <id> --client-secret <value>
```

Tasks:
 - [x] tests (whatever can be tested here)


providers/azure/connection/azureinstancesnapshot/provider.go
```diff
@@ -0,0 +1,399 @@
...
+			return nil, err
+		}
+		return nil, fmt.Errorf("failed to list logical unit numbers: %s", outErr)
+	}
+	// output looks like this:
+	// [0:0:0:0]    /dev/sda
+	// [1:0:0:0]    /dev/sdb
+	data, err := io.ReadAll(cmd.Stdout)
+	if err != nil {
+		return nil, err
```
vjeffrey: would be great to have a test around this parsing

providers/azure/connection/azureinstancesnapshot/provider.go
```diff
@@ -0,0 +1,399 @@
...
+
+	c := &AzureSnapshotConnection{
+		opts:            conf.Options,
+		snapshotCreator: sc,
+		scanner:         *scanner,
+		identifier:      conf.PlatformId,
+	}
+
+	// setup disk image so and attach it to the instance
+	mi := mountInfo{}
```
vjeffrey, line 128: when i was testing this for aws a few months back, i ran into an issue where the setup was being called a couple times  due to how we were connecting, did u run into that at all? (just curious)

providers/azure/provider/provider.go
```diff
@@ -143,15 +180,28 @@ func (s *Service) Connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
...
 	asset := req.Asset
 	conf := asset.Connections[0]
 	s.lastConnectionID++
-	conn, err := connection.NewAzureConnection(s.lastConnectionID, asset, conf)
+	var conn shared.AzureConnection
+	var err error
+
+	switch conf.Type {
+	case string(azureinstancesnapshot.SnapshotConnectionType):
+		// A GcpSnapshotConnection is a wrapper around a FilesystemConnection
```
vjeffrey: nit: comment needs updating

vjeffrey (approved):
i didn't give it a real-life test, but code looks good! left a couple small questions

providers/azure/connection/azureinstancesnapshot/provider.go
preslavgerchev: +1, i will add tests

providers/azure/connection/azureinstancesnapshot/provider.go
preslavgerchev: havent noticed that, no

PR #2950, opened 2024-01-05
✨ Support scanning instances/snapshots by a fully qualified azure resource id

preslavgerchev:
We can now scan instances/snapshots across resource groups, they still need to be in the same location though.

```
sudo DEBUG=1 ./cnquery shell azure compute snapshot "/subscriptions/subId/resourceGroups/my-rg/providers/Microsoft.Compute/snapshots/test-debian-snap" --client-id <id> --tenant-id <id>--client-secret <secret>
```


chris-rock (approved):
Thank you @preslavgerchev

PR #2952, opened 2024-01-06
⭐ number.inRange support

arlimus:
Support a simpler methods to see if a number is in range.

So far users have to:

```coffee
num >= 3 && num <= 5
```

Now you can:

```coffee
num.inRange(min,  max)
```

Example:

```coffee
3.inRange(3, 5)
```

This supports both integers and floats. Dict support is coming next.


scottford-io (approved):
Looks good @arlimus ! 🙌

PR #2956, opened 2024-01-06
✨ parse.duration

arlimus:
This now supports parsing durations in seconds (s), minutes (m), hours (m), days (d), and years (y). This works with both the shorthand scalar as well as the long version, e.g.:

```coffee
parse.duration("3d")
parse.duration("3days")
```

The parser is very lenient, but we recommend using the above scalars as so:

```
1s = 1 second
2m = 2 minutes
3h = 3 hours
4d = 4 days
5y = 5 years
```


scottford-io (approved):
Looks good @arlimus !

PR #2958, opened 2024-01-06
✨ string.in(list of strings)

arlimus:
Support the inverse of `list.contains(string)` by adding support for `string.in(list)`. This makes queries much more flexible. Example:

```coffee
"hi".in(["a","hi","b"])
```


scottford-io (approved):
Looks good @arlimus !

Thanks

PR #2959, opened 2024-01-06
✨ time.inRange(min, max)

arlimus:
Add support for the inRange operator for the time type. Makes it easier to check if a time is between two timestamps.


scottford-io (approved):
Looks good @arlimus !

PR #2964, opened 2024-01-07
MERGE AFTER v9.14 -- ⭐ support named arguments in functions

arlimus:
This is an exciting new v10 feature that allows users to set named arguments in functions.

This means that you can do things like this:

```coffee
users.all(user:
  groups.contains(group:
    user.uid == group.gid
  )
)
```

This handles all cases where you felt you had to assign `_` to a variable. This makes it easier in all cases where you could only use one expression (like with `all`, `one`, etc).

Most exciting part about this feature is: It is backwards-compatible. It works with older runtimes of MQL since it only changes the compiler. (tested via binary LLX)

You can still rely on using `_` if no variable name is defined. This remains the default.

### DIFF
--- a/mql/mql_test.go
+++ b/mql/mql_test.go
@@ -220,6 +220,20 @@ func TestNullResources(t *testing.T) {
 	})
 }
 
+func TestNamedFunctions(t *testing.T) {
+	x := testutils.InitTester(testutils.LinuxMock())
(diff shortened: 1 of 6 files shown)


scottford-io (approved):
Great stuff @arlimus !!! ⭐ ⭐

arlimus:
Reverted! This was not to be merged until v9.14 is merged

PR #2965, opened 2024-01-07
🧹 ensure that all providers are installed for bundle lint

chris-rock:
This will make the manual installation of the providers in github actions obsolete, eg. https://github.com/mondoohq/cnquery-packs/blob/main/.github/workflows/lint.yaml#L21-L35


scottford-io (approved):
⭐  thanks @chris-rock

PR #2969, opened 2024-01-07
🧹 support sysinfo without providers

arlimus:
Normally we require to download and install the OS provider when we want to gather the OS info. This is now no longer necessary. Instead we grab a few very isolated libraries from the OS provider itself and embed them into the runtime. This yielded no significant increase in the runtime (500kb). While we want to further decrease this size over time, it is important to support scanning other provider targets without requiring the OS provider to be pulled.

This needs a follow-up where we decouple the connection dependency from the resolver, since it pulls in a chain of dependencies for Tar + Docker right now (because the resolver type-casts to `TarConnection` and `DockerContainerConnection`)


chris-rock (approved):
Thank you @arlimus

PR #2973, opened 2024-01-08
🐛 update `for` field in props

arlimus:
We were never fully implementing referential properties until now. This PR fixes the fact that the `for` field is referencing other properties in the same bundle using their `uid`. It now correctly generates their `mrn` and sets the `uid` to empty string like in the property itself.

### DIFF
--- a/explorer/property.go
+++ b/explorer/property.go
@@ -4,7 +4,6 @@
 package explorer
 
 import (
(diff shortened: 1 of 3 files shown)


chris-rock (approved):
Thank you @arlimus

PR #2974, opened 2024-01-08
✨ Add support for scanning azure disks.

preslavgerchev:
Adds support for scanning azure disks: `cnquery scan azure compute disk <id> ....`.

This means that the azure snapshot scanner now supports disks, instances and snapshot targets. This PR also adds some cleanup by removing unused properties and making things private where possible.

### DIFF
--- a/providers/azure/connection/azureinstancesnapshot/snapshot.go
+++ b/providers/azure/connection/azureinstancesnapshot/snapshot.go
@@ -20,9 +20,39 @@ const (
 	createdValue   = "cnspec"
 )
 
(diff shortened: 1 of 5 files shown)


providers/azure/connection/azureinstancesnapshot/snapshot.go
```diff
@@ -20,6 +20,35 @@ const (
 	createdValue   = "cnspec"
 )
 
+type SnapshotCreator struct {
```
preslavgerchev: these are all non-functional changes, moving things around and marking fields as private

imilchev (approved):

PR #2979, opened 2024-01-08
🐛 only use native registry key implementation if we are scanning local machine from windows

chris-rock:


arlimus (approved):
local pocal, ty!

PR #2981, opened 2024-01-09
🐛  ensure platform id gets set for no-setup awsec2ebs scans

vjeffrey:


providers/aws/connection/awsec2ebsconn/provider.go
```diff
@@ -153,6 +153,9 @@ func NewAwsEbsConnection(id uint32, conf *inventory.Config, asset *inventory.Ass
 			asset.Connections[0].Options["createdBy"] = "Mondoo"
 		}
 	}
+	if conf.Options[snapshot.NoSetup] == "true" {
```
preslavgerchev, line 156: curious: does NoSetup means that the snapshot we're targeting is already mounted at the target instance?

preslavgerchev (approved):

providers/aws/connection/awsec2ebsconn/provider.go
vjeffrey: yam it means we can skip all the find and attach step cause the volume is already on the machine

PR #2988, opened 2024-01-09
🧹 proto changes for the progressmessage and errormessage

vjeffrey:
adding progress and error messages to the SBOM proto


sbom/sbom.proto
```diff
@@ -36,6 +36,30 @@ enum Status {
...
   // status typically requires a thorough review and reprocessing of the BOM.
   STATUS_FAILED = 3;
+
+  // STARTED indicates that the SBOM scan was initialized. It is used for tracking
+  // progress of a scan over time.
+  STATUS_STARTED = 4;
+}
+
+// ProgressMessage 
+message ProgressMessage {
```
chris-rock: I do not think we really need this. We could return the sbom message. timestamp is already part of it and asset has the external id included, status too.

sbom/sbom.proto
```diff
@@ -36,6 +36,30 @@ enum Status {
...
+}
+
+// ErrorReport is an error that occurs when performing an SBOM scan
+message ErrorReport {
+  // The error message from the scan.
+  string msg = 1;
+  // The external id associated with the scan error.
+  ExternalIDType externalID = 2;
+  // Status, e.g. "FAILED" being communicated
+  Status status = 3;
```
chris-rock: Should we simply add the error msg field to the sbom message since the other fields are already available in Sbom message?

vjeffrey: 👍

vjeffrey: updated!

chris-rock (approved):
Great addition @vjeffrey Please update the generated code before merging.

PR #3000, opened 2024-01-11
🧹 Attach zerolog logger to plugin clients.

preslavgerchev:
Attach the zerolog logger to the plugin. This way we gain extra logging around the plugins' lifecycle:

```
cnquery> exit
x plugin process exited error="exit status 1" id=8679 plugin=/opt/mondoo/providers/azure/azure
! failed to shut down provider error="rpc error: code = Unavailable desc = error reading from server: EOF" provider=azure
→ plugin process exited id=9127 plugin=/opt/mondoo/providers/os/os
```


chris-rock (approved):
Thank you @preslavgerchev

PR #3001, opened 2024-01-11
🐛 Allow for provider shutdown to take longer than a heartbeat.

preslavgerchev:
Until now, both `Shutdown` and `Heartbeat` shared a lock. This means that if shutting down ever takes more then the heartbeat interval, the provider plugin gets killed as it cannot keep doing heartbeats since the lock is occupied.

This PR introduces a second lock to only be used for the heartbeat-related stuff: Setting `isClosed` and `isShutdown` on the provider and then using this in the heartbeat logic. The shutdown logic uses a separate lock to ensure that both can run in parallel.


arlimus (approved):
Great catch, love this improvement!

PR #3013, opened 2024-01-12
🧹 Rename region label to location label for azure vms.

preslavgerchev:
In azure-land it's called a location, not a region


chris-rock (approved):
Thank you @preslavgerchev

PR #3016, opened 2024-01-12
✨ aws.iam.loginProfile added for IAM users

arlimus:


providers/aws/resources/aws.lr
```diff
@@ -777,6 +777,13 @@ private aws.iam.user @defaults("arn name") {
   groups() []string
   // List of access keys metadata associated with the user
   accessKeys() []dict
+  // Login profile for the user
+  loginProfile() aws.iam.loginProfile
+}
+
+private aws.iam.loginProfile @defaults("createDate") {
```
chris-rock: comment for resource is missing

chris-rock (approved):
Thank you @arlimus

providers/aws/resources/aws.lr
arlimus: good catch, thank you!

providers/aws/resources/aws.lr
```diff
@@ -777,6 +777,14 @@ private aws.iam.user @defaults("arn name") {
...
   // List of access keys metadata associated with the user
   accessKeys() []dict
+  // Login profile for the user
+  loginProfile() aws.iam.loginProfile
+}
+
+// AWS IAM login profile for a user
+private aws.iam.loginProfile @defaults("createDate") {
+  // Time when the login profile was created
+  createDate time
```
tas50: Can we make this createdAt?  AWS is super inconsistent in their API and we've mostly just copied that historically. Lately I've been trying to make them all createdAt since that was the most common use in the SDK. It'll really help things to be predictable for users and avoid us having breaking changes later to harmonize it all:

Here's the current numbers:

- createdAt: 17
- createdTime: 6
- createdDate: 1

arlimus: awesome suggestion, we'll get it adjusted

providers/aws/resources/aws.lr.manifest.yaml
```diff
@@ -1703,6 +1703,14 @@ resources:
     platform:
       name:
       - aws
+  aws.iam.loginProfile:
+    fields:
+      createdAt: {}
+    is_private: true
+    min_mondoo_version: latest
```
```suggestion
    min_mondoo_version: 10.0
```

providers/aws/resources/aws.lr.manifest.yaml
```diff
@@ -1769,6 +1777,8 @@ resources:
       createDate: {}
       groups: {}
       id: {}
+      loginProfile:
+        min_mondoo_version: latest
```
```suggestion
        min_mondoo_version: 10.0
```

PR #3019, opened 2024-01-12
🧹 Rename the ec2-volume and ec2-snapshot platforms to be EBS

tas50:
The API calls these EC2, which makes sense from an SDK organization standpoint, but doesn't reflect what these platforms are call. The console and documentation calls these EBS volumes/snapshots. This is a breaking change, but the blast radius is very small at this point since discover is not enabled by default. V10 is the perfect time to make this before we enable fine grained scanning by default.


chris-rock:
Makes sense to me

ping @vjeffrey @preslavgerchev to make them aware as well.

chris-rock (approved):
Thank you @tas50

PR #3020, opened 2024-01-13
✨ Automatically create cnquery update PR

czunker:
A new cnquery release will trigger a workflow in the cnspec repo to create a PR which updates the dep.

### DIFF
--- a/.github/workflows/bump-cnspec.yml
+++ b/.github/workflows/bump-cnspec.yml
@@ -1,0 +1,19 @@
+name: Trigger cnquery bump in cnspec
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  trigger-cnspec:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger cnquery bump in cnspec
+        uses: peter-evans/repository-dispatch@v2
+        with:
+          token: ${{ secrets.RELEASR_ACTION_TOKEN }}
+          repository: "mondoohq/cnspec"
+          event-type: update-cnquery
+          client-payload: '{
+              "version": "${{  github.ref_name }}"
+            }'


.github/workflows/bump-cnspec.yml
```diff
@@ -0,0 +1,27 @@
...
+        default: "v9.14.0"
+
+jobs:
+  trigger-cnspec:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger cnquery bump in cnspec
+        uses: peter-evans/repository-dispatch@v2
+        with:
+          token: ${{ secrets.CNSPEC_REPO_API_TOKEN }}
```
benr: Use RELEASR_ACTION_TOKEN for the moment.

.github/workflows/bump-cnspec.yml
```diff
@@ -0,0 +1,27 @@
...
+  push:
+    tags:
+      - '*'
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'cnquery version to use'
+        type: string
+        required: false
+        default: "v9.14.0"
```
chris-rock: Should this not be empty? I also feel that the workflow dispatch should be in cnspec not in cnquery

czunker: Good point.

We could use the `workflow_dispatch` to test the complete workflow, but then I'd need to rewrite this workflow. I'll remove it for now.

.github/workflows/bump-cnspec.yml
```diff
@@ -0,0 +1,20 @@
+name: Trigger cnquery bump in cnspec
+
+on:
+  push:
+    tags:
+      - '*'
```
chris-rock: Ben just discovered that this should only happen on release. This way we would avoid that we have accidental pushes to other repos

czunker: Thanks. I changed this.

I'm not sure what you mean by "pushes to other repos". This workflow does not (git) push anything. It dispatches an event to the other repo and will there trigger a workflow.

benr (approved):

.github/workflows/bump-cnspec.yml
benr: Thanks, looks great.

PR #3025, opened 2024-01-14
🧹 remove versioned providers for crossProviderList

arlimus:
info.Provider is only ever non-versioned and we have updated recordings, so this is no longer necessary


chris-rock (approved):
Thank you @arlimus

arlimus:
I'm closing this for now in favor of https://github.com/mondoohq/cnquery/pull/3037 , I think @chris-rock 's intuition/direction on this is much better than what I was trying to do. Removing it here is too early, considering we have versioned providers out there. The referenced PR also adds deprecation notices.

PR #3027, opened 2024-01-15
⭐ Add discovery for aws-rds-dbcluster

tas50:
RDS clusters are very important and should be scanned as their own assets with checks applied to those assets. This starts that process.

```
cnquery shell aws --discover rds-dbclusters

    Available assets

  > 1. othercluster (aws-rds-dbcluster)
    2. timcluster (aws-rds-dbcluster)








    ↑/k up • ↓/j down • q quit • ? more

```


chris-rock (approved):
Thank you @tas50

PR #3028, opened 2024-01-15
✨ introduce I/O retries for provider installs

arlimus:
Fixes https://github.com/mondoohq/cnspec/issues/998

The implementation is flexible to be used in other places. We can consider packaging it separately down the line, but for now it is in a good place to be very specific and purpose-built for this use-case.

### DIFF
--- a/providers/providers.go
+++ b/providers/providers.go
@@ -541,7 +541,7 @@ func InstallIO(reader io.ReadCloser, conf InstallConf) ([]*Provider, error) {
 		// We don't set a max retry, since we can indefinitely try to remove this
 		err := osRetry(func() error {
 			return os.RemoveAll(tmpdir)
-		}, -1)
+		}, maxInstallConfRetries)
 		if err != nil {
 			log.Error().Err(err).Msg("failed to remove temporary folder for unpacked provider")
 		}
--- a/providers/providers.go
+++ b/providers/providers.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/cockroachdb/errors"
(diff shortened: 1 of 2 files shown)


providers/providers.go
```diff
@@ -482,7 +523,11 @@ func InstallIO(reader io.ReadCloser, conf InstallConf) ([]*Provider, error) {
 	// so we don't spam the system with unnecessary data. Optionally we could
 	// keep them and re-use them, so they don't have to download again.
 	defer func() {
-		if err = os.RemoveAll(tmpdir); err != nil {
+		// We don't set a max retry, since we can indefinitely try to remove this
+		err := osRetry(func() error {
```
chris-rock, line 542: could that end up in an infinite loop?

providers/runtime.go
```diff
@@ -585,6 +585,7 @@ func (r *Runtime) lookupResourceProvider(resource string) (*ConnectedProvider, *
 	}
 
 	if info.Provider != providerConn && !stringx.Contains(crossProviderList, info.Provider) {
+		log.Error().Str("infoProvider", info.Provider).Str("connectionProvider", providerConn).Msg("mismatch between expected and received provider, ignoring provider")
```
chris-rock, line 588: would be helpful to log what is not matching, eg. via debug log

providers/providers.go
arlimus: yes. If you feel better, I could give it a high limit?

providers/runtime.go
arlimus: I found this message super helpful in understanding what is not matching. The corssProviderList is static, so this message only happens when the two are different.

chris-rock: we are on the same page, just missed that providerConn == info.Provider

providers/providers.go
chris-rock: an upper bound would be great.

arlimus: added an upper bound. I don't think it's helpful in this particular case, but it's definitely something we can learn as we go. These boundaries are estimates based on the data that is processed, curious to see how we tweak it over time.

chris-rock (approved):
Thank you @arlimus

PR #3033, opened 2024-01-15
🧹 remove trailing dot for short cmd description

chris-rock:
after #3031 

simplifies the short description of commands since it is unusual to use trailing dots for short descriptions.


misterpantz (approved):
Exquisitely consistent!

PR #3037, opened 2024-01-16
🐛 support legacy+modern provider ID lookups

arlimus:
Legacy providers contain the version in their ID. Modern providers after v10 do not do this anymore.

Support both forms of lookup in this transition time.


chris-rock (approved):
Thank you @arlimus

PR #3073, opened 2024-01-19
✨ Add zones, state, isRunning properties to azure compute VMs.

preslavgerchev:
Adds 3 new properties to the azure compute VMS:
 * `zones`
 * `state` - either `unknown`, `stopped` or `running`
 * `isRunning` -> a shortcut for verifying that `state == 'running'`

### DIFF
--- a/providers/azure/resources/compute.go
+++ b/providers/azure/resources/compute.go
@@ -22,6 +22,22 @@ func (a *mqlAzureSubscriptionComputeService) id() (string, error) {
 	return "azure.subscription.compute/" + a.SubscriptionId.Data, nil
 }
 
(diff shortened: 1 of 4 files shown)


providers/azure/resources/compute.go
```diff
@@ -79,6 +96,48 @@ func (a *mqlAzureSubscriptionComputeService) vms() ([]interface{}, error) {
 	return res, nil
 }
 
+func (a *mqlAzureSubscriptionComputeServiceVm) state() (string, error) {
```
preslavgerchev, line 99: for getting the state, we need to call `InstanceView` for each machine. It does not look like this information is present from the bulk retrieval API

vjeffrey (approved):

PR #3076, opened 2024-01-19
🐛 Fix port resource panic

jaym:
It is possible for the port resource to panic on mac and windows if the process is requested. This can happen because we query for all the processes first, then we query for ports which gets us process ids. A new process showing up between the first and second queries can cause the panic to happen.

Linux has a similar check


chris-rock (approved):
Thank you @jaym

PR #3081, opened 2024-01-21
🧹 ensure example query packs pass our own lint requirements

chris-rock:
Use our lint requirements to validate our examples. To make this work, we temporarily install the required providers and remove them afterwards. This ensures that other tests like cli tests are not affected by the providers we need here.


arlimus (approved):
So true!

PR #3082, opened 2024-01-21
⭐ dict.recurse( .. )

arlimus:
Recurse through an arbitrary dict (JSON/YAML/...) structure and find entries that match a search criteria.

For example, you can set up a data structure like this:

```json
{
  "users": [
    {"name": "bob"}
  ],
  "owners": {
    "admins": [
      {"name": "joy", "isOwner": true}
    ]
  }
}
```

Finding any user in this list is difficult with traditional mechanisms, since you'd have to understand the data structure and where to search.

With `recurse` it is made much easier:

```coffee
jdata.recurse( name != empty )
```

```
[
  0: {
    name: "bob"
  }
  1: {
    isOwner: true
    name: "joy"
  }
]
```

You could then just grab the names and continue:

```coffee
jdata.recurse( name != empty ).map(name)
```

```
[
  0: "bob"
  1: "joy"
]
```


scottford-io (approved):
This worked well for me @arlimus. ⭐

PR #3095, opened 2024-01-23
🐛 make sure the os provider can be cross referenced from another provider

imilchev:
this is needed by the k8s provider since it is referencing the containerImage resource, which is defined in the os provider

we also have the issue where filters that shouldn't be executed for an asset may request a provider that shouldn't be called. For example, trying to run aws filters on a k8s asset. This results in printing too many error messages, so I changed the log level of that error to debug

### DIFF
--- a/providers/runtime.go
+++ b/providers/runtime.go
@@ -580,6 +580,7 @@ func (r *Runtime) lookupResourceProvider(resource string) (*ConnectedProvider, *
 	crossProviderList := []string{
 		"go.mondoo.com/cnquery/providers/core",
 		"go.mondoo.com/cnquery/providers/network",
+		"go.mondoo.com/cnquery/providers/os",
 		// FIXME: DEPRECATED, remove in v12.0 vv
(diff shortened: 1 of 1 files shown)


providers/runtime.go
```diff
@@ -588,11 +589,12 @@ func (r *Runtime) lookupResourceProvider(resource string) (*ConnectedProvider, *
 		// version support, we can safely remove this.
 		"go.mondoo.com/cnquery/v9/providers/core",
 		"go.mondoo.com/cnquery/v9/providers/network",
+		"go.mondoo.com/cnquery/v9/providers/os",
```
chris-rock, line 592: I think we need to review those resources that require it from the k8s provider.

chris-rock (approved):
Thank you @imilchev

providers/runtime.go
imilchev: it's just `containerImage` resource that is referenced

PR #3096, opened 2024-01-23
⭐ implement StoreResults for resources data

arlimus:
This extends the proto to store resources data directly. This is the next step in the data recordings journey that makes this data available to upstream targets. We are starting our journey in v10, so this is still behind a feature-flag for now.

1. The recording interface moves to LLX, which also defines the runtime. This allows anyone to request the recording layer from the runtime and standardizes setters/getters.
2. `StoreResultsReq` is extended to support resources recording data via the new `resources` field. To avoid confusion, the `data` field has been renamed to `query_data`, because it captures MQL query data only. The `StoreQueryData` method was adjusted analogously.
3. The implementation of recordings stays with runtime for now, analogous to runtime itself having a proto/implementation split
4. The feature-flag is `StoreResourcesData`. Only when it is active will scan results be sent upstream. We continue to not send `shell` and `run` data by default. I think we may want to consider sending these but may want to think about a CLI flag to explicitly send data to avoid unwanted side-effects. `scan` semantically already sends data up, so users expect it to behave this way.
5. The new proto message `ResourceRecording` captures all recording data for each resource and was adopted from the previous recording implementation (where it was non-proto).

**Follow-ups:** 

As this is only the start, there are 2 vital things before we can spread this more broadly. These have to be done in the v10 timeframe before the feature becomes default:
1. Implement this analogously for cnspec
2. Dedupe resources and query data. We currently duplicate query and recording data, which is annoying.


explorer/cnquery_explorer.proto
```diff
@@ -355,7 +355,8 @@ message UpdateAssetJobsReq {
 // Store results for a given asset
 message StoreResultsReq {
   string asset_mrn = 1;
-  map<string, cnquery.llx.Result> data = 3;
+  map<string, cnquery.llx.Result> query_data = 3;
```
chris-rock: While there is no technical reason, I'd like to do avoid the rename for now. Can we keep the old name for the field and just extend the comments?

chris-rock (approved):
Very nice improvement @arlimus  Just added one comment.

explorer/cnquery_explorer.proto
arlimus: yes, happy to adjust it

chris-rock (approved):
Thank you @arlimus

PR #3097, opened 2024-01-23
🧹 only log warnings and higher for providers

imilchev:
Make sure we don't show info messages for the providers since that is a bit too much


chris-rock (approved):
Thank you @imilchev

PR #3103, opened 2024-01-23
🐛 fix default providers IDs

imilchev:
makes sure we are not installing the same provider version over and over


chris-rock (approved):
Thank you @imilchev

PR #3134, opened 2024-01-26
🧹 log the error that caused a provider to be unavailable

imilchev:
sometimes we have seen this error in the logs while scanning. However, the actual error message causing the provider to shutdown isn't printed. This should help investigate the issue further


chris-rock (approved):
Thank you @imilchev

PR #3135, opened 2024-01-27
🐛 Fix wrong resource name for sql firewall rules.

### LINKED ISSUE #3128: Azure:  Issue with Fetching Firewall Rules for PostgreSQL Server
Related to:
Ensure 'Allow access to Azure services' for PostgreSQL Database Server is disabled

**Problem**:

In the cnquery resources, there seems to be a discrepancy in how firewall rules for PostgreSQL servers are fetched. The code in **postgresql.go** currently references **azure.subscription.sql.firewallrule** which appears incorrect for PostgreSQL servers. I suspect it should be azure.subscription.postgresql.firewallrule instead.

```
	pager := dbFirewallClient.NewListByServerPager(resourceID.ResourceGroup, server, &postgresql.FirewallRulesClientListByServerOptions{})
	res := []interface{}{}
	for pager.More() {
		page, err := pager.NextPage(ctx)
		if err != nil {
			return nil, err
		}
		for _, entry := range page.Value {
			mqlFireWallRule, err := CreateResource(a.MqlRuntime, "azure.subscription.sql.firewallrule",
				map[string]*llx.RawData{
					"id":             llx.StringData(convert.ToString(entry.ID)),
					"name":           llx.StringData(convert.ToString(entry.Name)),
					"type":           llx.StringData(convert.ToString(entry.Type)),
					"startIpAddress": llx.StringData(convert.ToString(entry.Properties.StartIPAddress)),
					"endIpAddress":   llx.StringData(convert.ToString(entry.Properties.EndIPAddress)),
				}) 
```

**Error encountered:**

"firewallRules: cannot find resource azure.subscription.sql.firewallrule in this provider"

**Issue:**

The error suggests that the code is attempting to fetch SQL firewall rules instead of PostgreSQL firewall rules, leading to a resource not found error.

**Request**:

We need a review and correction of the code to ensure that the firewall rules for PostgreSQL servers are fetched correctly in line with Azure's SDK and API structure. 

Screenshot from 2024-01-26 10-18-23

preslavgerchev:
Fixes #3128

### DIFF
--- a/providers/azure/resources/mysql.go
+++ b/providers/azure/resources/mysql.go
@@ -208,7 +208,7 @@ func (a *mqlAzureSubscriptionMySqlServiceServer) firewallRules() ([]interface{},
 			return nil, err
 		}
 		for _, entry := range page.Value {
-			mqlFireWallRule, err := CreateResource(a.MqlRuntime, "azure.subscription.sql.firewallrule",
+			mqlFireWallRule, err := CreateResource(a.MqlRuntime, "azure.subscription.sqlService.firewallrule",
 				map[string]*llx.RawData{
 					"id":             llx.StringData(convert.ToString(entry.ID)),
 					"name":           llx.StringData(convert.ToString(entry.Name)),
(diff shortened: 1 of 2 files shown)


chris-rock (approved):
Thank you @preslavgerchev

PR #3137, opened 2024-01-27
🧹 make panic handler more flexible to allow ui message logging

chris-rock:
We've seen some cases where panics happened in our packer provider. Therefore we want to introduce the panic handler there. Beside pure reporting, the packer plugin would like to hook into the panic handler and print the panic via the packer UI printer. This allows users to send handler


scottford-io (approved):
👍

PR #3138, opened 2024-01-27
🐛 return proper error when no provider could be detected

chris-rock:
The `DetectProvider` https://github.com/mondoohq/cnquery/blob/main/providers/runtime.go#L168C1-L178C2 relies on the fact that we either have an error or a proper provider object. 

```go
// DetectProvider will try to detect and start the right provider for this
// runtime. Generally recommended when you receive an asset to be scanned,
// but haven't initialized any provider. It will also try to install providers
// if necessary (and enabled)
func (r *Runtime) DetectProvider(asset *inventory.Asset) error {
	provider, err := r.providerForAsset(asset)
	if err != nil {
		return err
	}
	return r.UseProvider(provider.ID)
}
```

Turns out this was not always the case. `EnsureProvider` has a case where it returns `nil, nil`. This changes the logic to make the error explicit https://github.com/mondoohq/cnquery/blob/main/providers/providers.go#L318-L323.

### DIFF
--- a/providers/providers.go
+++ b/providers/providers.go
@@ -285,6 +285,14 @@ func ListAll() ([]*Provider, error) {
 	return res, nil
 }
 
+type ProviderNotFoundError struct {
+	lookup ProviderLookup
+}
+
+func (e *ProviderNotFoundError) Error() string {
+	return "cannot find provider for " + e.lookup.String()
+}
+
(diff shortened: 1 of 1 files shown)


imilchev (approved):
LGTM

PR #3139, opened 2024-01-27
🧹 refactor os connection types

chris-rock:
The os provider had two places where connection types where defined. This was confusing. The values where the same though. This cleans up the different parts to just use one definition.


imilchev (approved):
LGTM

PR #3141, opened 2024-01-27
⭐️ WinRM support

chris-rock:
After #3139

This PR brings back winrm support from v8


arlimus (approved):
Thank you for adding it back!!

PR #3142, opened 2024-01-27
🚤 improve performance for large slack accounts

chris-rock:
If you are a large organisation with 1000s of conversations, the previous implementation fetched information about the creator via the user rest call. The problem is that the slack api heavily rate limits the user rest call.

This change makes the creator information asynchronous and therefore allows much speedier fetching of the conversations. This avoids reaching the rate limit.

### DIFF
--- a/providers/slack/connection/connection_test.go
+++ b/providers/slack/connection/connection_test.go
@@ -1,0 +1,38 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+//go:build debugtest
+// +build debugtest
+
+package connection
+
+import (
+	"context"
+	"testing"
+
+	"github.com/slack-go/slack"
+	"github.com/stretchr/testify/require"
+	"go.mondoo.com/cnquery/v10/providers-sdk/v1/inventory"
+	"go.mondoo.com/cnquery/v10/providers-sdk/v1/vault"
+)
+
+func TestSlackProvider(t *testing.T) {
+
+	cred := &vault.Credential{
+		Type:     vault.CredentialType_password,
+		Password: "<slack-token>",
+	}
+	cred.PreProcess()
+
+	conn, err := NewSlackConnection(1, &inventory.Asset{}, &inventory.Config{
+		Type:        "slack",
+		Credentials: []*vault.Credential{cred},
+	})
+	require.NoError(t, err)
+
+	client := conn.Client()
+	ctx := context.Background()
+	users, err := client.GetUsersContext(ctx, slack.GetUsersOptionLimit(999))
+	require.NoError(t, err)
+	require.NotNil(t, users)
+}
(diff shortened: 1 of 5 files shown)


imilchev (commented):
That's a great improvement. However, I think in certain cases (depending on how you write the MQL query), this approach may still be suboptimal. 

If I understand it correctly, this relies on the init function for `mqlSlackUser` to fetch the relevant user when requested ([here](https://github.com/mondoohq/cnquery/blob/45cc9c336c298d1a9c70c76ebee254a0555db447/providers/slack/resources/users.go#L249)). If we have an MQL query that would be grabbing all creators for all conversations:
```coffee
slack.conversations {
  creator
}
```
That would be calling the `GetUserContext` API for each unique creator. Won't it be better to refactor this too? We can just fetch all users with 1 API call and cache them

chris-rock:
> That's a great improvement. However, I think in certain cases (depending on how you write the MQL query), this approach may still be suboptimal.
> 
> If I understand it correctly, this relies on the init function for `mqlSlackUser` to fetch the relevant user when requested ([here](https://github.com/mondoohq/cnquery/blob/45cc9c336c298d1a9c70c76ebee254a0555db447/providers/slack/resources/users.go#L249)). If we have an MQL query that would be grabbing all creators for all conversations:
> 
> ```coffee
> slack.conversations {
>   creator
> }
> ```
> 
> That would be calling the `GetUserContext` API for each unique creator. Won't it be better to refactor this too? We can just fetch all users with 1 API call and cache them

I am not 100% sure I understand. It is correct that we do the user call as soon as the user calls creator. The init function would then fetch the data for each user. Is your idea to batch the user fetching. Is so we would need to implement caching logic. Totally doable but was not in scope for this PR.

imilchev:
> > That's a great improvement. However, I think in certain cases (depending on how you write the MQL query), this approach may still be suboptimal.
> > If I understand it correctly, this relies on the init function for `mqlSlackUser` to fetch the relevant user when requested ([here](https://github.com/mondoohq/cnquery/blob/45cc9c336c298d1a9c70c76ebee254a0555db447/providers/slack/resources/users.go#L249)). If we have an MQL query that would be grabbing all creators for all conversations:
> > ```coffee
> > slack.conversations {
> >   creator
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That would be calling the `GetUserContext` API for each unique creator. Won't it be better to refactor this too? We can just fetch all users with 1 API call and cache them
> 
> I am not 100% sure I understand. It is correct that we do the user call as soon as the user calls creator. The init function would then fetch the data for each user. Is your idea to batch the user fetching. Is so we would need to implement caching logic. Totally doable but was not in scope for this PR.

yes, something like that. Just thought that in some cases this will still result in many API calls. I think the solution can be simple. If instead of calling `client.GetUserInfo` [here](https://github.com/mondoohq/cnquery/blob/45cc9c336c298d1a9c70c76ebee254a0555db447/providers/slack/resources/users.go#L249), we use `slack.users` MQL resource to get all users, then we can loop over them and find the one that is being requested. IIRC we use this approach for k8s resources too

chris-rock:
Lets chat about your improvement idea!

PR #3146, opened 2024-01-28
⭐️ support scope mrn reading from service account

chris-rock:
This adds support to read scope mrns from service account. Its a preparation to support organization and space service account in cnquery and cnspec.


cli/config/config.go
```diff
@@ -254,22 +258,36 @@ func (c *CommonOpts) GetServiceCredential() *upstream.ServiceAccountCredentials
...
 	}
 
-	return parent
+	return scopeMrn
+}
+
+// GetParentMrn returns the scope mrn that is used for the service account.
+// This is either the organization mrn or the space mrn.
+// Deprecated: Use GetScopeMrn instead
+func (c *CommonOpts) GetParentMrn() string {
```
preslavgerchev, line 289: can we just not remove this function? i dont think this is used by an end user in some way so should be safe to delete it and migrate cnspec to `GetScopeMrn` (if cnspec makes use of that)

preslavgerchev (approved):

cli/config/config.go
chris-rock: Lets leave it deprecated for now.

PR #3151, opened 2024-01-29
🧹 Retry login command when concurrent IAM updates are detected

jaym:
If the server returns aborted for RegisterAgent, that means it aborted the change because of concurrent writes to the IAM policy, likely meaning concurrent registrations. Clients are safe to backoff and retry in this case.


chris-rock (approved):
Great addition @jaym

PR #3159, opened 2024-01-30
✨ go benchmark for scans

imilchev:
Something we already have for cnspec. This will run a couple of scans and benchmark CPU and memory. It will compare the results with the latest commit on `main` and will alert if there are large deviations


chris-rock (approved):
Great addition @imilchev

PR #3161, opened 2024-01-30
🐛 avoid panics when trying to get s3 bucket location

arlimus:
Fixes https://github.com/mondoohq/cnspec/issues/1087


scottford-io (approved):
Thanks @arlimus 👍

PR #3168, opened 2024-01-31
🚤 reduce user fetch calls for slack conversation members

chris-rock:
Before this improvement we needed to fetch users individually. This change fetches users just once and then references them when we call members.

### DIFF
--- a/providers/slack/resources/conversations.go
+++ b/providers/slack/resources/conversations.go
@@ -176,6 +176,24 @@ func (s *mqlSlackConversation) members() ([]interface{}, error) {
 		return nil, errors.New("cannot retrieve new data while using a mock connection")
 	}
 
+	// before we fetch the members, we need to fetch all users so that we save the time for individual lookups
+	users, err := CreateResource(s.MqlRuntime, "slack.users", nil)
+	if err != nil {
+		return nil, err
+	}
+	mqlUsers := users.(*mqlSlackUsers)
+	userEntries := mqlUsers.GetList()
+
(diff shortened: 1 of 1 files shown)


providers/slack/resources/conversations.go
```diff
@@ -176,6 +176,24 @@ func (s *mqlSlackConversation) members() ([]interface{}, error) {
 		return nil, errors.New("cannot retrieve new data while using a mock connection")
 	}
 
+	// before we fetch the members, we need to fetch all users so that we save the time for individual lookups
+	users, err := CreateResource(s.MqlRuntime, "slack.users", nil)
```
imilchev, line 180: you could add an init function for `slack.user`. The init function would do this lookup based on an input argument `id`. An example of this approach can be found [here](https://github.com/mondoohq/cnquery/blob/0cacfbd7f0192409051f9faa2c42166bfef1b64d/providers/gcp/resources/compute.go#L1137-L1180).

If you have that init function prepared, then in here you could just call:
```go
res, err := CreateResource(g.MqlRuntime, "slack.user", map[string]*llx.RawData{
    "id":      llx.StringData(id),
})
if err != nil {
    return nil, err
}
return res.(*mqlSlackUser), nil
```

There are 2 benefits of this approach:
1. the user can also initialize a `slack.user` manually by doing `slack.user(id: 'userId')`
2. if we decide to have a separate asset for `slack.user` at some point, we already have the `init` function for it

chris-rock: I 100% agree with your suggestion. Will do that as followup.

PR #3177, opened 2024-02-01
🧹 refactor providers coordinator

imilchev:
The goal of this change is to allow the creation of nested provider coordinators. The concept is similar to the Go `context`. We have a `GlobalCoordinator`, which can have children and they have children too. When a parent coordinator shuts down, all children shutdown too.

Benefits of this approach:
- we can have 1 coordinator per root asset in an inventory
- provider instances are isolated for each coordinator
  - this means parallel scans can now work as expected. 1 scan cannot kill the providers of another
- we still have a global coordinator that will shutdown all children and all started providers

On the side I tried to cleanup the `distributeJob` function a bit by making a more readable structure for all the asset objects we are tossing around

### DIFF
--- a/cli/providers/providers.go
+++ b/cli/providers/providers.go
@@ -423,8 +423,12 @@ func setConnector(provider *plugin.Provider, connector *plugin.Connector, run fu
 			}
 		}
 
+		coordinator := providers.NewCoordinator()
+		defer coordinator.Shutdown()
+
 		// TODO: add flag to set timeout and then use RuntimeWithShutdownTimeout
-		runtime := providers.Coordinator.NewRuntime()
+		runtime := coordinator.NewRuntime()
+		defer runtime.Close()
 		if err = providers.SetDefaultRuntime(runtime); err != nil {
 			log.Error().Msg(err.Error())
 		}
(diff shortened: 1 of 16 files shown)


cli/providers/providers.go
```diff
@@ -423,8 +423,11 @@ func setConnector(provider *plugin.Provider, connector *plugin.Connector, run fu
 			}
 		}
 
+		defer providers.GlobalCoordinator.Shutdown()
+
 		// TODO: add flag to set timeout and then use RuntimeWithShutdownTimeout
-		runtime := providers.Coordinator.NewRuntime()
+		runtime := providers.GlobalCoordinator.NewRuntime()
+		defer runtime.Close()
```
jaym, line 430: you can't defer this and the previous one because of log.Fatal

providers/coordinator.go
```diff
@@ -26,20 +26,23 @@ import (
 
 var BuiltinCoreID = coreconf.Config.ID
 
-var Coordinator = coordinator{
+var GlobalCoordinator = Coordinator{
```
arlimus, line 29: Why expose the coordinator?

imilchev: because now you can call `GlobalCoordinator.NewChild()`. That one needs to return a `Coordinator`, which isn't possible if it's not exposed. If we don't want to expose the struct, then we should define an interface and return that

imilchev:
superseded #3218

PR #3190, opened 2024-02-02
🐛 fix crash on aws.iam.virtualMfaDevices & logGroup.metricsFilters alarms failure

vjeffrey:
fixes https://github.com/mondoohq/cnquery/issues/3179

fixes https://github.com/mondoohq/cnquery/issues/3178

### DIFF
--- a/providers/aws/resources/aws_iam.go
+++ b/providers/aws/resources/aws_iam.go
@@ -289,7 +289,7 @@ func (a *mqlAwsIam) virtualMfaDevices() ([]interface{}, error) {
 	devicesResp, err := svc.ListVirtualMFADevices(ctx, &iam.ListVirtualMFADevicesInput{})
 	if err != nil {
 		log.Error().Err(err).Msg("cannot gather virtual mfa devices info")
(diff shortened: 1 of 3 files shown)


providers/aws/resources/aws_iam.go
```diff
@@ -288,7 +288,9 @@ func (a *mqlAwsIam) virtualMfaDevices() ([]interface{}, error) {
 
 	devicesResp, err := svc.ListVirtualMFADevices(ctx, &iam.ListVirtualMFADevicesInput{})
 	if err != nil {
-		return nil, errors.Wrap(err, "could not gather aws iam virtual-mfa-devices")
```
chris-rock, line 291: any reason why we not set the error?

vjeffrey: ah ya we should do that, pushing the change now

chris-rock: With https://github.com/mondoohq/cnquery/pull/3191 a plain error would also work. This approach is also fine.

chris-rock (approved):
Thank you @vjeffrey

PR #3191, opened 2024-02-02
🐛 cache mql field errors

chris-rock:
Errors where not cached and slowed down the execution.


vjeffrey (approved):

arlimus (approved):
Massive catch!! thank you

PR #3192, opened 2024-02-03
🧹 refactor ms365

chris-rock:
This PR refactors the ms 365 to be much simpler and more manageable. It moves all the powershell code out of the connection and makes it part of the resources. 

In general we should not write code that runs for a long time in the init function. Therefore this code refactors that and uses the new capabilities to cache the data properly.

Additionally it adds better error code handling and sends more clear messages when authentication errors happen. This helps to avoid the strange powershell xml output whenever possible.

In best case the scan time is similar to the previous implementation. In combination with https://github.com/mondoohq/cnquery/pull/3191, the scan time is down from 12 minutes to 3 minutes when permission errors happen.

It also fixes https://github.com/mondoohq/cnquery/issues/2572 where the credential resolution was not properly handled when run was used.


arlimus (approved):
so many awesome improvements in this one!

PR #3193, opened 2024-02-05
⚡ re-use errors in gatherProcess info

arlimus:
In process, when it is not centrally initialized, errors may lead to the info-gathering to be re-run multiple times and not re-used.

Since `processes` initializes all objects, this generally only happens when we call an individual `process`. Additionally, if the process doesn't exist it will be centrally handled so the affected call won't run.


chris-rock (approved):
Thank you @arlimus

PR #3194, opened 2024-02-05
⭐ introduce sshd.config.blocks

arlimus:
This introduces support for querying individual blocks in SSHd configs. It's a more direct way of adressing feedback in
https://github.com/mondoohq/cnspec-policies/issues/340 by exposing the underlying block entirely, while also supporting aggregate values in the existing params structure.

Example: Let's assume we have an existing `/etc/ssh/sshd_config` on our system with a bunch of existing configuration. If we added a new match group at the end of the file like this:

```ini
Match Group sftp-users
X11Forwarding no
PermitRootLogin no
AllowTCPForwarding yes
```

We can now query this match block both explicitly and implicitly.

Implicitly it's (already) represented in the existing `params` field:

```coffee
> sshd.config.params.AllowTcpForwarding
"no,yes"
```

In the above example you can see, that we already had this field set above the match block with the value set to `no`. After adding our match group, it was additionally set to `yes`. The field aggregates both values.

This implicit access to config values has already existed in MQL as the default behavior.

With the new `blocks` field, we are extending implicit match block access to become explicit:

```coffee
> sshd.config.blocks
sshd.config.blocks: [
  0: sshd.config.matchBlock criteria=""
  1: sshd.config.matchBlock criteria="Group sftp-users"
]
```

This first match block is the default block, which is always present. It has no criteria set and applies to everything.

The second match block has a `criteria` field that shows it only matches for `Group sftp-users`.

You can easily access its configuration:

```coffee
> sshd.config.blocks { criteria params }
sshd.config.blocks: [
  0: {
    criteria: ""
    params: {
      AllowTcpForwarding: "no"
      ...
    }
  }
  1: {
    criteria: "Group sftp-users"
    params: {
      AllowTcpForwarding: "yes"
      PermitRootLogin: "no"
      X11Forwarding: "no"
    }
  }
]
```

In this example you can see that each block contains its own set of parameters. These are now restricted to the configuration of the block only. Thus the `AllowTcpForwarding` setting is not an aggregate of values anymore, it now only contains the value defined in the block.

Added Note: As a consequence of this change we are now also consistently structuring the `Match` field in the `sshd.config.params` structure to behave like all other fields: It combines any match group separated by commas:

```coffee
> sshd.config.params.Match
sshd.config.params[Match]: "Group sftp-users,User myservice"
```


chris-rock (approved):
Thank you @arlimus Works like a charm

PR #3202, opened 2024-02-05
🧹 refactor local scanner code

imilchev:
Simplify the code inside the `local_scanner.go`:

- remove multiple loops over assets
- remove multiple copies of a runtime for 1 asset
- expose discovery logic in a way that it can be directly used by `cnspec`


explorer/scan/discovery.go
```diff
@@ -0,0 +1,191 @@
...
+		// for all discovered assets, we apply mondoo-specific labels and annotations that come from the root asset
+		for _, a := range rootRuntime.Provider.Connection.Inventory.Spec.Assets {
+			runtime, err := providers.Coordinator.RuntimeFor(resolvedRootAsset, providers.DefaultRuntime())
+			if err != nil {
+				discoveredAssets.AddError(a, err)
+				continue
+			}
+			if err := runtime.SetRecording(recording); err != nil {
+				discoveredAssets.AddError(resolvedRootAsset, err)
+				runtime.Close()
```
jaym: it might be good to break some of this into smaller functions so you can use defer

explorer/scan/discovery.go
```diff
@@ -0,0 +1,191 @@
...
+				continue
+			}
+
+			resolvedAsset := runtime.Provider.Connection.Asset
+			prepareAsset(resolvedAsset, resolvedRootAsset, runtimeLabels)
+
+			// If the asset has been already added, we should close its runtime
+			if !discoveredAssets.Add(resolvedAsset, runtime) {
+				runtime.Close()
```
jaym: is the idea we keep a runtime open for each asset?

explorer/scan/local_scanner.go
```diff
@@ -184,295 +170,194 @@ func preprocessQueryPackFilters(filters []string) []string {
...
+		multiprogress = progress.NoopMultiProgressBars{}
+	}
+	return multiprogress, nil
+}
 
-		// for all discovered assets, we apply mondoo-specific labels and annotations that come from the root asset
-		for _, a := range runtime.Provider.Connection.GetInventory().GetSpec().GetAssets() {
-			a.AddMondooLabels(asset)
-			a.AddAnnotations(asset.GetAnnotations())
+func Batch[T any](list []T, batchSize int) [][]T {
```
jaym: might be a good one to unit test

explorer/scan/local_scanner.go
```diff
@@ -184,295 +170,194 @@ func preprocessQueryPackFilters(filters []string) []string {
...
-				Bundle:           job.Bundle,
-				Props:            job.Props,
-				QueryPackFilters: preprocessQueryPackFilters(job.QueryPackFilters),
-				Ctx:              ctx,
-				Reporter:         reporter,
-				ProgressReporter: p,
-				runtime:          runtime,
+			resp, err := services.SynchronizeAssets(ctx, &explorer.SynchronizeAssetsReq{
+				SpaceMrn: client.SpaceMrn,
+				List:     discoveredAssets.GetAssets(),
```
jaym: this looks like you're in the loop of batches and send up all discovered assets. can you double check

explorer/scan/discovery.go
imilchev: yes, that's also how the current code works (apart from the fact that it creates 2 runtimes per asset). The goal is to create the runtime during discovery and close it after the scan has finished

explorer/scan/discovery.go
imilchev: good idea! done!

explorer/scan/local_scanner.go
imilchev: yup, will do that. That one used to be in cnspec. It's now moved here. I will add a test for it

explorer/scan/local_scanner.go
imilchev: yup, missed that one...

explorer/scan/discovery_test.go
```diff
@@ -0,0 +1,215 @@
...
+
+		for _, asset := range discoveredAssets.Assets {
+			assert.Equal(t, inv.Spec.Assets[0].ManagedBy, asset.Asset.ManagedBy)
+		}
+	})
+
+	t.Run("set ci/cd labels", func(t *testing.T) {
+		inv := getInventory()
+		inv.Spec.Assets[0].Category = inventory.AssetCategory_CATEGORY_CICD
+		require.NoError(t, os.Setenv("GITHUB_ACTION", "go-test"))
```
jaym, line 215: might be good to undo changes you make to the environment after the test is done

utils/slicesx/batch_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBatch(t *testing.T) {
+	slice := []string{}
+	for i := 0; i < 100; i++ {
+		slice = append(slice, fmt.Sprintf("item-%d", i))
+	}
+
+	batches := Batch(slice, 10)
```
jaym, line 19: this test should probably test things other than evenly divisible batches

jaym (approved):

utils/slicesx/batch_test.go
imilchev: added 1 more test case

explorer/scan/discovery_test.go
imilchev: made sure the value is reset after the test is done

PR #3218, opened 2024-02-06
🧹 providers coordinator v2

imilchev:
The goal is to allow creating local coordinators with a parent coordinator. The way the code is setup is:
- providers are always managed by the parent coordinator. This means eventually all providers should end up in the global coordinator
- the local coordinator only keeps a reference to the started providers but are not managing them
- from the perspective of the global coordinator any provider started by a local coordinator is ephemeral. This means the local coordinators have dedicated providers for them. This allows scans to run independently of each other
- there is no direct dependency from the global coordinator to its children

### DIFF
--- a/explorer/scan/discovery.go
+++ b/explorer/scan/discovery.go
@@ -90,7 +90,7 @@ func (d *DiscoveredAssets) GetAssetsByPlatformID(platformID string) []*inventory
 		for _, c := range a.Children {
 			for _, p := range c.Asset.PlatformIds {
 				if platformID == "" || p == platformID {
-					assets = append(assets, a.Asset)
+					assets = append(assets, c.Asset)
 					break
 				}
 			}
--- a/explorer/scan/discovery.go
+++ b/explorer/scan/discovery.go
@@ -18,6 +18,12 @@ import (
 	"go.mondoo.com/cnquery/v10/providers-sdk/v1/upstream"
 )
 
+type RootAsset struct {
+	Coordinator providers.Coordinator
+	Asset       *inventory.Asset
+	Children    []*AssetWithRuntime
+}
+
 type AssetWithRuntime struct {
 	Asset   *inventory.Asset
 	Runtime *providers.Runtime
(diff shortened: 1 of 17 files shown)


explorer/scan/discovery.go
```diff
@@ -30,13 +36,26 @@ type AssetWithError struct {
 
 type DiscoveredAssets struct {
 	platformIds map[string]struct{}
-	Assets      []*AssetWithRuntime
-	Errors      []*AssetWithError
+	// Assets      []*AssetWithRuntime
```
jaym, line 39: should this line be removed

explorer/scan/discovery.go
```diff
@@ -30,13 +36,26 @@ type AssetWithError struct {
...
+	Errors []*AssetWithError
+
+	RootAssets map[*inventory.Asset]*RootAsset
+}
+
+func (d *DiscoveredAssets) AddRoot(root *inventory.Asset, coordinator providers.Coordinator, runtime *providers.Runtime) bool {
```
jaym, line 45: runtime isn't used. is that expected?

explorer/scan/discovery.go
```diff
@@ -146,15 +184,15 @@ func DiscoverAssets(ctx context.Context, inv *inventory.Inventory, upstream *ups
 	// if there is exactly one asset, assure that the --asset-name is used
 	// TODO: make it so that the --asset-name is set for the root asset only even if multiple assets are there
 	// This is a temporary fix that only works if there is only one asset
-	if len(discoveredAssets.Assets) == 1 && invAssets[0].Name != "" && invAssets[0].Name != discoveredAssets.Assets[0].Asset.Name {
-		log.Debug().Str("asset", discoveredAssets.Assets[0].Asset.Name).Msg("Overriding asset name with --asset-name flag")
-		discoveredAssets.Assets[0].Asset.Name = invAssets[0].Name
-	}
+	// if len(discoveredAssets.Assets) == 1 && invAssets[0].Name != "" && invAssets[0].Name != discoveredAssets.Assets[0].Asset.Name {
```
jaym, line 187: commented out code

arlimus:
Thank you for taking a stab at this! We are motivated by random provider shutdowns across the system and by supporting multiple scans at the same time. We'll have a separate call to go over a discussion of this, so we will keep this PR as draft until then.

imilchev:
we are fixing this by getting rid of ephemeral providers and making sure we have a more robust provider management system. Closing this one

PR #3222, opened 2024-02-06
🏃🏻‍♂️ faster aws connect and discovery

imilchev:
we have a similar flow for k8s. We don't need to do discovery every time we connect to an asset. This is currently happening for aws. With the current code a `Connect` call for an AWS asset takes about 1s. With this change the `Connect` call takes about 500ms

Please verify this works for all discoveries. I only tested this with s3 buckets and kms keys

### DIFF
--- a/providers/aws/resources/discovery.go
+++ b/providers/aws/resources/discovery.go
@@ -186,6 +186,11 @@ func Discover(runtime *plugin.Runtime, filters connection.DiscoveryFilters) (*in
 		Assets: []*inventory.Asset{},
 	}}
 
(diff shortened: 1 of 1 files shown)


vjeffrey:
i'm getting different results on this branch than released cnquery, the ec2 instance is missing, and some others too i need to figure out what. it's 3 (on this branch) vs 6 assets discovered with the same query

imilchev:
please re-test this. There was an issue with the discovery mechanism that was fixed. Maybe that fixes what you observed before

vjeffrey (approved):

PR #3223, opened 2024-02-06
🐛 properly handle errors from createRuntimeForAsset

imilchev:
the response of `createRuntimeForAsset` is `nil` when there is an error. We shouldn't use it when storing the errors.

Can't add a test for this yet because I cannot easily mock the coordinator. That will be possible once #3218 merges


czunker (approved):
Thanks @imilchev

PR #3242, opened 2024-02-08
⭐️ refactor providers code

imilchev:
- [x] extract runtimes management to base `Service`
- [x] extract `GetData` to base `Service`
- [x] extract `SetData` to base `Service`
- [x] extract `Shutdown` to base `Service`
- [x] implement `Disconnect` function in base `Service`
- [x] make sure `Disconnect` is called after an asset is scanned (on closing the runtime)

Things to verify:
**Make sure you build the corresponding provider locally and install it before testing**
- [x] aws ebs volume cleanup still working (done on Shutdown)
- [x] azure snapshot cleanup still working (done on Shutdown)
- [x] gcp snapshot cleanup still working (done on Shutdown)
- [x] closing tar connections. deleting the file when pulling a container image (done on Shutdown)
- [x] terraform provider cleanup. i think that's used for gitlab scans (done on Shutdown)


providers/k8s/connection/admission/connection.go
```diff
@@ -66,6 +66,10 @@ func (c *Connection) SupportedResourceTypes() (*resources.ApiResourceIndex, erro
 	return c.ManifestParser.SupportedResourceTypes()
 }
 
+func (c *Connection) SetID(id uint32) {
```
chris-rock: I am not sure we want to let someone outside of the provider overwrite the connection id

imilchev: It's gone now

chris-rock (commented):
Looks great to me. I am going to test this in different scenarios.

imilchev:
This should be rebased on main to make sure you are testing with all fixes

chris-rock (approved):
Thank you @imilchev

PR #3243, opened 2024-02-08
Add basic Azure doc references

misterpantz:


tas50:
Holding to test a better solution for line breaks

PR #3265, opened 2024-02-11
🐛 parse time formats case-insensitive

chris-rock:
fixes https://github.com/mondoohq/cnquery/issues/3207 by making the parsing of the time format case insensitive


arlimus (approved):
Thanks for the bugfix!

PR #3266, opened 2024-02-11
🧹 do not show the core provider as outdated since it cannot be updated without cnquery update

chris-rock:
Right now the core provider is displayed as outdated. This ignores the core provider in our status output since the user cannot update this provider without updating cnquery


imilchev (approved):
nit: the mock provider is also builtin so it may run into the same issue

PR #3272, opened 2024-02-12
🐛 make sure the os provider always returns unique platform ids

imilchev:
fixes https://github.com/mondoohq/cnquery/issues/3261


chris-rock (approved):
Thank you @imilchev

PR #3274, opened 2024-02-12
⭐️ support re-using MQL cache between provider connections

imilchev:
Allow re-using the MQL resource cache between provider connections. Connections can now have parents. When that is the case, the MQL resource cache from the parent is used for the child connection as well.

This PR only implements sharing MQL resource cache within a single provider. Cross-provider sharing isn't possible yet


arlimus (approved):
Very cool addition, great job on this!! :rocket: :tada: 

Nice additions to the testing, nice fixes for some of the error messages, the sharing of the connection details through the plugins is on point. Updating all the providers to use this is nice to see in action. I haven't tested this across version boundaries, but since the interface is only additive we can introduce this without a major bump. Changing the syncx.Map to be a pointer for the sharing makes sense.

I really only have a small nitpick, which is using a pointer for the parentId instead of a simple value.  But yeah, nitpick ;)

PR #3275, opened 2024-02-12
🐛 Fix resource lookup for `cpe`

czunker:
This fixes the error:
```
failed to compile query 'asset { name platform version arch ids labels cpes.map(uri) }': cannot find resource that is called by 'uri' of type core.cpe"
```

### DIFF
--- a/providers/os/resources/os.lr
+++ b/providers/os/resources/os.lr
@@ -18,7 +18,7 @@ alias os.unix.sshd = sshd
 
 extend asset {
   // Common Platform Enumeration (CPE) for the asset
-  cpes() []core.cpe
+  cpes() []cpe
   // Advisory & vulnerability report
   // Deprecated; will be removed in version 10.0
   // use vulnmgmt instead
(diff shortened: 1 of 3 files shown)


czunker:
@chris-rock / @arlimus Could you please have a look at whether this is as intended?

I couldn't find another field where we use `provider.resource` as identifier. Using only the resource name solved my problem, but I'm not sure whether the combination `provider.resource` should also work and I'm only working around another issue.

chris-rock:
I am not able to reproduce the issue. Everything works as it should be:

```
cnquery shell docker debian:11                                                                                                                                                                                                                                    
→ connected to Debian GNU/Linux 11 (bullseye)

cnquery> packages.list.first.cpes
packages.list.first.cpes: [
  0: cpe uri="cpe:2.3:a:adduser:adduser:3.118\\+deb11u1:all:*:*:*:*:*:*"
]
cnquery> 
```

I am closing this as cnquery works as expected.

PR #3276, opened 2024-02-12
🧹 Attach data disks with a delete opt. Add an option to skip detach/delete steps

preslavgerchev:
Attach the Azure data disks with a delete option. That means that if the VM gets deleted, the data disks get delete alongside.

Also add an option to skip detach/delete steps. This may be useful when we're performing a scan on a VM we know will get deleted soon after. In that case, manually detaching/deleting the data disk is redundant as we know this will happen when the VM itself gets deleted. To trigger this, add the `--skip-snapshot-cleanup` flag.

### DIFF
--- a/providers/azure/connection/azureinstancesnapshot/snapshot.go
+++ b/providers/azure/connection/azureinstancesnapshot/snapshot.go
@@ -214,12 +214,14 @@ func (sc *snapshotCreator) attachDisk(targetInstance instanceInfo, diskName, dis
 		return err
 	}
 	attachOpt := compute.DiskCreateOptionTypesAttach
+	deleteOpt := compute.DiskDeleteOptionTypesDelete
 	// the Azure API requires all disks to be specified, even the already attached ones.
 	// we simply attach the new disk to the end of the already present list of data disks
 	disks := targetInstance.vm.Properties.StorageProfile.DataDisks
 	disks = append(disks, &compute.DataDisk{
 		Name:         &diskName,
 		CreateOption: &attachOpt,
+		DeleteOption: &deleteOpt,
 		Lun:          &lun,
 		ManagedDisk: &compute.ManagedDiskParameters{
 			ID: &diskId,
--- a/providers/azure/provider/provider.go
+++ b/providers/azure/provider/provider.go
@@ -42,7 +42,7 @@ func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error)
 	subscriptionsToExclude := flags["subscriptions-exclude"]
 	certificatePath := flags["certificate-path"]
 	certificateSecret := flags["certificate-secret"]
-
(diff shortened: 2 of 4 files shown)


providers/azure/provider/provider.go
```diff
@@ -57,7 +57,10 @@ func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error)
 	if len(subscriptionsToExclude.Value) > 0 {
 		opts["subscriptions-exclude"] = string(subscriptionsToExclude.Value)
 	}
-
+	// the presence of the flag indicates that we should skip cleanup
+	if len(skipSnapshotCleanup.Value) > 0 {
```
czunker: This might be counterintuitive.

Setting `--skip-snapshot-cleanup=false` will result in the option being true.

providers/azure/connection/azureinstancesnapshot/snapshot.go
```diff
@@ -214,12 +214,14 @@ func (sc *snapshotCreator) attachDisk(targetInstance instanceInfo, diskName, dis
...
 	}
 	attachOpt := compute.DiskCreateOptionTypesAttach
+	deleteOpt := compute.DiskDeleteOptionTypesDelete
 	// the Azure API requires all disks to be specified, even the already attached ones.
 	// we simply attach the new disk to the end of the already present list of data disks
 	disks := targetInstance.vm.Properties.StorageProfile.DataDisks
 	disks = append(disks, &compute.DataDisk{
 		Name:         &diskName,
 		CreateOption: &attachOpt,
+		DeleteOption: &deleteOpt,
```
czunker, line 224: Nice find!

providers/azure/connection/azureinstancesnapshot/snapshot.go
```diff
@@ -214,12 +214,14 @@ func (sc *snapshotCreator) attachDisk(targetInstance instanceInfo, diskName, dis
 		return err
 	}
 	attachOpt := compute.DiskCreateOptionTypesAttach
+	deleteOpt := compute.DiskDeleteOptionTypesDelete
 	// the Azure API requires all disks to be specified, even the already attached ones.
 	// we simply attach the new disk to the end of the already present list of data disks
 	disks := targetInstance.vm.Properties.StorageProfile.DataDisks
 	disks = append(disks, &compute.DataDisk{
```
czunker, line 221: Just to be sure without reading the complete code:
This only affects disks we create for snapshots? Not, that we set the delete option for a disk the user created.

czunker (commented):
Nice flag and option!

providers/azure/connection/azureinstancesnapshot/snapshot.go
preslavgerchev: Yeah, we only set that to disks that we attach. Disks that are already there are unaffected

czunker (approved):
Thanks @preslavgerchev

PR #3278, opened 2024-02-13
🧹  Improving by adding resourcetypes to AWS config recorder (aws)

HRouhani:
Finally solved the problem:

Screenshot from 2024-02-22 15-31-13

It should be array based on following:

Screenshot from 2024-02-22 15-31-57

### DIFF
--- a/providers/aws/resources/aws_config.go
+++ b/providers/aws/resources/aws_config.go
@@ -73,13 +73,17 @@ func (a *mqlAwsConfig) getRecorders(conn *connection.AwsConnection) []*jobpool.J
 					recording = val.recording
 					lastStatus = val.lastStatus
 				}
+				resourceTypesInterface := make([]interface{}, len(r.RecordingGroup.ResourceTypes))
+				for i, resourceType := range r.RecordingGroup.ResourceTypes {
+					resourceTypesInterface[i] = resourceType
+				}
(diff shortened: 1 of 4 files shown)


providers/aws/resources/aws_config.go
```diff
@@ -73,12 +73,17 @@ func (a *mqlAwsConfig) getRecorders(conn *connection.AwsConnection) []*jobpool.J
...
+				for i, resourceType := range r.RecordingGroup.ResourceTypes {
+					resourceTypesInterface[i] = string(resourceType)
+				}
 				mqlRecorder, err := CreateResource(a.MqlRuntime, "aws.config.recorder",
 					map[string]*llx.RawData{
 						"name":                       llx.StringDataPtr(r.Name),
 						"roleArn":                    llx.StringDataPtr(r.RoleARN),
 						"allSupported":               llx.BoolData(r.RecordingGroup.AllSupported),
 						"includeGlobalResourceTypes": llx.BoolData(r.RecordingGroup.IncludeGlobalResourceTypes),
+						"resourceTypes":              llx.ArrayData(resourceTypesInterface, types.Any),
```
czunker: According to your definition in the lr file, you should use `types.String` instead of `types.Any`.

HRouhani: Thanks Christian. Done

PR #3291, opened 2024-02-13
🧹 disable os detection for registry images

imilchev:
this makes sure we aren't pulling the container unless we are running a scan


imilchev:
fixed in #3496

PR #3292, opened 2024-02-13
🐛 fix container scanning

imilchev:


explorer/scan/discovery_test.go
```diff
@@ -253,4 +253,19 @@ func TestDiscoverAssets(t *testing.T) {
 		require.NoError(t, err)
 		assert.Len(t, discoveredAssets.Assets, 1)
 	})
+
+	t.Run("container platform discovery", func(t *testing.T) {
```
imilchev: not sure if that's the best place to put this test but it's definitely important to have such tests... let's keep it here for now and we can break off real integration tests to another place later

imilchev: oh.. this needs to use a builtin os provider in order to work...

imilchev: moved it to the os provider

chris-rock (approved):
Thank you @imilchev

PR #3294, opened 2024-02-13
⭐️ allow to write sbom reports to a file

chris-rock:


apps/cnquery/cmd/sbom.go
```diff
@@ -100,13 +110,27 @@ var sbomCmdRun = func(cmd *cobra.Command, runtime *providers.Runtime, cliRes *pl
...
 		err := exporter.Render(&output, &bom)
 		if err != nil {
 			log.Fatal().Err(err).Msg("failed to render SBOM")
 		}
-		fmt.Println(output.String())
+
+		if outputTarget != "" {
+			filename := outputTarget
+			if len(boms) > 1 {
+				filename = fmt.Sprintf("%s-%d.%s", path.Base(outputTarget), i, path.Ext(outputTarget))
```
preslavgerchev, line 137: nit/question: this might be a problem in CI/CD setups, a user wouldn't expect that we change the filename. can we not write all boms to one file?

chris-rock: Mmh, the current format only has one report included, so we would need one file for a sbom.

preslavgerchev (approved):

PR #3295, opened 2024-02-13
🧹 add more integration testing for sbom scanning

chris-rock:
This PR adds integration testing for the sbom subcommand. It scans various container distros to make sure everything works as expected.

NOTE: that the tests are not enabled as default yet since the container fetching makes the pipeline slow. We work on improvements to use recordings which will make that a super fast test.


vjeffrey (approved):
🌟  MORE TESTS! yay!

PR #3296, opened 2024-02-14
🐛 ensure we catch all 403 access denied errors

vjeffrey:
especially in cases where users have region deny set up, this error can cause chaos!

fixes https://github.com/mondoohq/cnquery/issues/3246

### DIFF
--- a/providers/aws/resources/aws_dms.go
+++ b/providers/aws/resources/aws_dms.go
@@ -5,9 +5,9 @@ package resources
 
 import (
 	"context"
+	"errors"
 
 	"github.com/aws/aws-sdk-go-v2/service/databasemigrationservice"
-	"github.com/aws/aws-sdk-go-v2/service/databasemigrationservice/types"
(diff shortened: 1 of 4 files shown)


providers/aws/resources/aws_dms.go
```diff
@@ -31,7 +30,9 @@ func (a *mqlAwsDms) replicationInstances() ([]interface{}, error) {
 	}
 	// get all the results
 	for i := range poolOfJobs.Jobs {
-		res = append(res, poolOfJobs.Jobs[i].Result.([]types.ReplicationInstance)...)
+		if poolOfJobs.Jobs[i].Result != nil {
```
chris-rock, line 38: Should we check for an error here and propagate?

chris-rock (approved):

providers/aws/resources/aws_dms.go
vjeffrey: oh, ya we should! ill do that

PR #3298, opened 2024-02-14
⭐️ Allow skipping setup for azure snapshot connection

preslavgerchev:
Allow skipping creating/attaching the disk for the azure snapshot connection. This means that it is expected that the disk is already attached to the target VM.

To use this, specify the `--skip-snapshot-setup` and `--lun` parameters. The Azure API provides no indication as to where it will mount the disk, the only thing we know is its logical unit number (LUN). Therefore, by specifying the lun as a parameter, we can let the snapshot connection try and look it up. There may be multiple devices at one LUN so we look for a block that has no mounting points.

Specifying `--skip-snapshot-setup` will also implicitly set `--skip-snapshot-cleanup`. We cannot delete the disk if we do not know what it's azure ID is.

Here's an example of how the whole command looks like:
```
cnquery scan azure compute instance <id> --skip-snapshot-setup --lun 3
```

In addition, this PR also improves the naming logic for disks we create from cnquery by hashing the following params:
  * The target name
  * The target type
  * The resource group of the target
  * The subscription
  * The current time
 
This ensures that if we're performing parallel scans, no naming collisions can occur


chris-rock (approved):
Thank you @preslavgerchev

PR #3302, opened 2024-02-14
🧹 safe guard connections that should implement the Closer interface

imilchev:
Just make sure that we explicitly state that these connections implement the `plugin.Closer` interface. This will make sure that if in the future the interface changes, we will get compiler errors

I removed the `Close` functions from connections that didn't really implement any logic in them to avoid confusion. If we need those in the future we can re-add and implement the `Close` functions.


chris-rock (approved):
Thank you @imilchev

PR #3304, opened 2024-02-14
🧹 remove logic for ephemeral providers]

imilchev:
This is no longer used, so we can get rid of it


chris-rock (approved):
Thank you @imilchev

PR #3308, opened 2024-02-14
✨ refactor providers coordinator

imilchev:
- Remove runtimes from the coordinator when they are closed
- Shutdown any running providers that are no longer referenced by any active runtime
- Cleanup coordinator code
- Add tests!!!


go.mod
```diff
@@ -118,6 +117,8 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require go.uber.org/mock v0.4.0
```
arlimus: always baffles me why some editors don't group the dependencies :see_no_evil:

providers/coordinator.go
```diff
@@ -426,48 +236,199 @@ func (c *coordinator) unsafeSetAssetRuntime(asset *inventory.Asset, runtime *Run
...
+		if _, ok := usedProviders[id]; !ok {
+			log.Debug().Msg("shutting down unused provider " + p.Name)
+			if err := c.stop(p); err != nil {
+				log.Warn().Err(err).Str("provider", p.Name).Msg("failed to shut down provider")
+			}
+		}
+	}
+}
+
+func (c *coordinator) GetRunningProvider(id string, update UpdateProvidersConfig) (*RunningProvider, error) {
```
arlimus, line 273: Is it possible for 2 almost synchronous calls to be made to this function that cause the provider to be started twice? Asking because running may be nil on the request and startProvider is not in any sync lock, i.e. I wonder if two functions of startProvider can run before `runningByID` gets set

arlimus (changes requested):
Great cleanup, especially the small edges that I had overlooked in the original implementation. Only one thing left, then this one is good to go from my side

go.mod
imilchev: fixed that one

providers/coordinator.go
imilchev: nice catch! i made sure we have a lock for the whole duration of `GetRunningProvider`. That should make sure we don't run into such situations

arlimus (approved):
awesome, thank you for adding all the handling here! (also kudos for the `unsafe` naming ;) )

PR #3309, opened 2024-02-14
🧹  Draft: Enhancing the Retrieval and Handling of AccessKey Information in aws.iam.users (aws)

HRouhani:
It needs to be fixed, something is still wrong!

The issue is related to accessing the Status field in:

``` 
 aws.iam.users {accessKeys}
``` 

As mentioned here: https://github.com/mondoohq/cnspec-enterprise-policies/pull/712


HRouhani:
It has been addressed already!

PR #3314, opened 2024-02-14
🐛 make sure no detectors are used for containers

imilchev:


chris-rock (approved):
Great catch @imilchev

PR #3326, opened 2024-02-15
🧹 use 1 global resource schema instead of per runtime

imilchev:
Use 1 schema for the coordinator instead of creating copies for every runtime we start. That should improve our memory footprint significantly for large scans.

Scanning a cluster with 366 assets with cnspec 10.3.0:
telegram-cloud-photo-size-4-5940308416457195425-y

Scanning the same cluster with the changes from this branch:
telegram-cloud-photo-size-4-5940308416457195424-y

this builds on top of #3308


providers/runtime.go
```diff
@@ -243,7 +229,7 @@ func (r *Runtime) Connect(req *plugin.ConnectReq) error {
 	}
 
 	r.Recording().EnsureAsset(r.Provider.Connection.Asset, r.Provider.Instance.ID, r.Provider.Connection.Id, asset.Connections[0])
-	r.schema.prioritizeIDs(BuiltinCoreID, r.Provider.Instance.ID)
+	// r.schema.prioritizeIDs(BuiltinCoreID, r.Provider.Instance.ID)
```
imilchev, line 232: we cannot really do this with a global schema. It seems like before we were prioritizing the IDs with the specific provider for the asset coming in first. I am not sure if that would actually break anything. @arlimus not sure if you recall why we needed that

arlimus: Yeah this one is going to break things if we don't keep it around...

The reason is that some keywords are equally implemented in two different providers (like `vulnmgmt` in os and vsphere) to provide the appropriate level of functionality in either, without having to learn new keywords. It is functionally correct, but it forces us to detect the right provider depending on the current prioritization...

(Long-term this is not a problem, because all keywords will be anchored to a parent, which entirely removes the guessword, but we can't force this into v10 right now...)

There are a few ways to handle this. First we need the list of resource definitions that have the same identifier. Once we have those, we can prioritize when we look up the resource/field based on the current connection. 

(a) We can introduce a new field into ResourceInfo that we can later easily remove (once everything is anchored to parents). We can't modify any part of ResourceInfo because it may be different between providers (which is ok). We could however add another field to it to chain things for now: `ResourceInfo.Other` which is either `nil` if no other provider offers it or is the next available info. 

(b) The alternative is to change the schema lookup from a simple `map[string]ResourceInfo` into a `map[string][]ResourceInfo`. It requires a bit more work and rewrite.

providers-sdk/v1/resources/schema.go
```diff
@@ -3,15 +3,21 @@
 
 package resources
 
+type ResourcesSchema interface {
```
arlimus, line 6: nit: Since we are moving this under the `resources` namespace, could we name this `Schema` instead?

imilchev: There is already `Schema` struct in the proto. That's why I had to give the interface a different name, otherwise it clashes with the struct name

providers/runtime.go
imilchev: I found 2 cases around this:
- 1 resource that can be defined in multiple providers (not extending, just having the same resource in multiple places)
- resource extensions where the root is in 1 place and the extensions are potentially in multiple providers

To cover both cases I picked approach (a) and added `Others` field that stores all other definitions of the `ResourceInfo`. However, I had to add the same `Others` field to the `Field` struct, because fields can also exist in multiple providers. Just like the `vulnmgmt` field that exists in the `vsphere` and `os` providers.

I think both cases work now. I still need to clean up the code a bit and will definitely add tests for both cases. Took me quite a bit of time to understand how this should work...

providers-sdk/v1/resources/schema.go
arlimus: Gotcha, let's clean it up in a follow-up and get merged first :)

providers-sdk/v1/resources/resources.proto
```diff
@@ -40,6 +40,7 @@ message ResourceInfo {
   string min_mondoo_version = 25;
   string defaults = 26;
   string provider = 27;
+  repeated ResourceInfo others = 29;
```
arlimus, line 48: nit: would you mind adding a comment above this and the `others` field below:

```
// This field contains references to other providers with the same resource/field.
// Note: Please do not use this field, it is only temporary and will be removed
// in the future once binding resources are mandatory for all executions.
```

(if this merges first no worries, we can add it later)

providers/providers.go
```diff
@@ -58,6 +58,26 @@ func init() {
...
+
+	// Load the schema for all the active providers
+	providers, err := ListActive()
+	if err != nil {
+		log.Error().Err(err).Msg("failed to list active providers")
+		return
+	}
+
+	for name := range providers {
+		schema, err := Coordinator.LoadSchema(name)
```
arlimus, line 75: nit:

This code will read all schemas that are installed on the system by default on startup. For regular use of scan/run/shell I had originally avoided this, so that we only load schemas we need and dynamically load additional schemas (making startup more lightweight). 

Is it necessary to load all schemas for this change?
If that is the case and we require them here, would you mind adding a comment so that we change this back to dynamic loading later on (I'm sure I'll forget otherwise :see_no_evil: ) (something like `FIXME: dynamically load providers on startup`).

providers/runtime.go
arlimus: Saw that you added others to cover fields as well, that's awesome!

arlimus:
Yoho @imilchev , great updates!! This is super close now, enough for me to +1 from my side.
I left two comment and marked them as nits so that this PR isn't blocked :) Feel free to get another approval to finish this in case I'm not around (I'm back Tuesday morning PST time, which is a bit late)

imilchev:
Fixed the comments. Will merge this into #3308 now

PR #3327, opened 2024-02-15
🐛 Fix cpes resource definition in vSphere

czunker:
Without the import, the resources.json file was created with `core.cpe` as the resource for `cpes`. This failed on lookup when compiling a query.

Adding the import creates the resources.json file `cpe` as the resource for `cpes`. Then the lookup works again.


czunker:
Screenshot from 2024-02-15 14-17-06

This shows the difference in the json file between the os provider, where we had the import and the sphere provider were the import was missing.

chris-rock (approved):
Thank you @czunker

PR #3328, opened 2024-02-15
🧹 refactor npm parser

chris-rock:
This refactors the npm package parser. It does not expose it to mql yet.


providers/os/resources/npm/package_json.go
```diff
@@ -0,0 +1,195 @@
...
+// Explanation:
+//
+//	^: Asserts the start of the string.
+//	([^<]+): Captures one or more characters that are not <. This is for the author's name.
+//	\s+: Matches one or more whitespace characters.
+//	<([^>]+)>: Captures the email address within angle brackets.
+//	\s+: Matches one or more whitespace characters again.
+//	\(([^)]+)\): Captures the URL within parentheses.
+//	$: Asserts the end of the string.
+var authorPatter = regexp.MustCompile(`^([^<]+)\s+<([^>]+)>(?:\s+\(([^)]+)\))?$`)
```
preslavgerchev:
```suggestion
var authorPattern = regexp.MustCompile(`^([^<]+)\s+<([^>]+)>(?:\s+\(([^)]+)\))?$`)
```

providers/os/resources/npm/package_json.go
```diff
@@ -0,0 +1,195 @@
...
+		a.Email = author.Email
+		a.URL = author.URL
+		return nil
+	}
+
+	// try to unmarshal as string
+	err = json.Unmarshal(b, &authorStr)
+	if err == nil {
+		matches := authorPatter.FindStringSubmatch(authorStr)
+		if len(matches) == 4 {
```
preslavgerchev, line 74: i didnt comb through the regex, but can we not make this only capture the relevant bits? looks like we need 3 matches, not 4 ?

providers/os/resources/npm/package_json.go
```diff
@@ -0,0 +1,195 @@
...
+	DevDependencies map[string]string     `jsonn:"devDependencies"`
+	Repository      packageJsonRepository `json:"repository"`
+	Engines         map[string]string     `jsonn:"engines"`
+	CPU             []string              `json:"cpu"`
+	OS              []string              `json:"os"`
+}
+
+// packageJsonPeople represents the author of the package
+// https://docs.npmjs.com/cli/v10/configuring-npm/package-json#people-fields-author-contributors
+type packageJsonPeople struct {
```
preslavgerchev, line 37:
```suggestion
type packageJsonAuthor struct {
```

preslavgerchev (approved):

providers/os/resources/npm/package_json.go
chris-rock: 1 is the overall match

providers/os/resources/npm/package_json.go
chris-rock: We should not do this since the same struct is used for contributors as well. npm docs itself call this people so lets stick with that.

providers/os/resources/npm/package_json.go
chris-rock: will adjust, great catch.

providers/os/resources/npm/package_json.go
preslavgerchev: oh, right! makes sense

PR #3332, opened 2024-02-15
⭐️ support multiple versions of package lock file

chris-rock:
This PR extends the current package-lock parsing capability and adds more tests to validate that we are able to load its multiple versions.

### DIFF
--- a/providers/os/resources/npm/package_lock.go
+++ b/providers/os/resources/npm/package_lock.go
@@ -10,22 +10,64 @@ import (
 	"go.mondoo.com/cnquery/v10/providers-sdk/v1/upstream/mvd"
 )
 
-type PackageJsonLockEntry struct {
-	Version string `json:"version"`
-	Dev     bool   `json:"dev"`
+// packageLock is the struct to represent the package.lock file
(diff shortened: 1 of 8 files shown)


providers/os/resources/npm/package_lock.go
```diff
@@ -10,22 +10,64 @@ import (
...
+	Resolved  string             `json:"resolved"`
+	Integrity string             `json:"integrity"`
+	License   packageLockLicense `json:"license"`
+}
+
+type packageLockLicense []string
+
+// UnmarshalJSON is a custom unmarshaler for the packageLockLicense type. It allows to handle the license field
+// which could be either a string or an array.
+func (l *packageLockLicense) UnmarshalJSON(data []byte) (err error) {
```
preslavgerchev, line 47: I might have missed it but I don't see a test file where licenses is an array of string, can we add such a test case?
I only see `"license": "Artistic-2.0"`

chris-rock: I added a new test to cover the multiple licenses case

preslavgerchev (approved):

PR #3337, opened 2024-02-16
⭐️ generate purl & cpe for npm packages

chris-rock:
This adjusts the npm package parsing to enrich it Purl and CPE information.

### DIFF
--- a/providers/os/resources/npm/npm.go
+++ b/providers/os/resources/npm/npm.go
@@ -4,10 +4,12 @@
 package npm
 
 import (
-	"github.com/package-url/packageurl-go"
-	"go.mondoo.com/cnquery/v10/providers/os/resources/cpe"
 	"io"
 	"strings"
+
+	"github.com/package-url/packageurl-go"
+	"github.com/rs/zerolog/log"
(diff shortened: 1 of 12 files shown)


providers/os/resources/npm/npm.go
```diff
@@ -4,11 +4,53 @@
...
+		name,
+		version,
+		nil,
+		"").String()
+}
+
+func NewCpes(name string, version string) []string {
+	cpes := []string{}
+	cpeEntry, err := cpe.NewPackage2Cpe(name, name, version, "", "")
+	if err == nil && cpeEntry != "" {
```
preslavgerchev: can we have a case where there is no err but cpeEntry is an empty string? if not, we should make sure that the `NewPackage2Cpe` function returns an err

chris-rock: I agree that we should handle errors. For that specific case I would like to avoid it to ensure that the cpe generation is to accidentally stopping a package gathering. To compensate, I added more tests for the cpe generation. I add comment to make this more clear.

preslavgerchev (approved):

PR #3340, opened 2024-02-16
⭐️ include npm packages in sbom export

chris-rock:
The sbom export now includes os, pypi and npm packages.

```shell
> cnquery sbom
npm/which/^4.0.0
npm/write-file-atomic/^5.0.1
pypi/yarl/1.9.4
npm/yarn/1.22.21
macos/zoom.us/5.17.5 (29101)
```

### DIFF
--- a/sbom/sbom.go
+++ b/sbom/sbom.go
@@ -98,54 +98,52 @@ func GenerateBom(r *ReportCollectionJson) ([]Sbom, error) {
 					bom.Packages = append(bom.Packages, bomPkg)
 				}
 			}
-			if rb.PythonPackages != nil {
-				for _, pkg := range rb.PythonPackages {
-					bomPkg := &Package{
-						Name:    pkg.Name,
-						Version: pkg.Version,
-						Purl:    pkg.Purl,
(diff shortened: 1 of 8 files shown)


sbom/sbom.go
```diff
@@ -124,6 +124,26 @@ func GenerateBom(r *ReportCollectionJson) ([]Sbom, error) {
 						})
 					}
 
+					bom.Packages = append(bom.Packages, bomPkg)
+				}
+			}
+			if rb.NpmPackages != nil {
```
jaym: you don't need to check for nil before iterating:

>For an array, pointer to array, or slice value a, the index iteration values are produced in increasing order, starting at element index 0. If at most one iteration variable is present, the range loop produces iteration values from 0 up to len(a)-1 and does not index into the array or slice itself. For a nil slice, the number of iterations is 0.

jaym (approved):

sbom/sbom.go
chris-rock: Great point, useless if conditions removed.

PR #3349, opened 2024-02-17
🧹 Replace llx.StringData and llx.BoolData with llx.StringDataPtr and llx.BoolDataPtr in ms365.

preslavgerchev:
Use the new ptr functions to better represent absence of data. This leads to the following changes.

Before:
```
1: microsoft.user id="f50deeea-31f3-4644-8004-93b3c0cc4995" displayName="Mondoo Tools" mail=""
```

After:
```
1: microsoft.user id="f50deeea-31f3-4644-8004-93b3c0cc4995" displayName="Mondoo Tools" mail=null
```

Note the mail difference. Fields that come back as null from the API are now better represented as such in MQL.


chris-rock (approved):
Thank you @preslavgerchev Great improvement.

PR #3351, opened 2024-02-17
🧹 Replace llx.StringData and llx.BoolData with llx.StringDataPtr and llx.BoolDataPtr in azure.

preslavgerchev:
Same as #3349


chris-rock (approved):
Thank you @preslavgerchev

PR #3353, opened 2024-02-18
🐛 Fix a gcp resource field typo

tas50:
It's correct in the .lr file, but wrong where it's set.

Also fix a few other misc things and update the spellcheck matchers


chris-rock (approved):
Thank you @tas50

PR #3359, opened 2024-02-19
⭐️ integration tests for cnquery + providers + mql queries

chris-rock:
This is an initial attempt to extend the testing of cnquery binary in combination with the provider and mql queries. If that pattern works we can easily extend that to further providers and more queries. This would ensure that future releases do not accidentally break user experience.


test/providers/os_test.go
```diff
@@ -0,0 +1,187 @@
...
+	"sync"
+	"testing"
+)
+
+var once sync.Once
+
+// setup builds cnquery locally
+func setup() {
+	// build cnspec
+	if err := exec.Command("go", "build", "../../apps/cnquery/cnquery.go").Run(); err != nil {
```
czunker, line 23: Why not add this as an dependency to the make target? The same for the provider build.

This would safe us some code here and we do not have the same in multiple places.

chris-rock: I like the idea to move to the make task, we still would keep two steps here since cnquery install and provider install are separate.

czunker: I'm not sure I understand what you mean.

My idea was, to have this in the `Makefile`:
```
test/integration: providers/install cnquery/install
```

vjeffrey (approved):
we should merge these!

test/providers/os_test.go
chris-rock: I want to avoid that to make sure the go test always use the latest code

PR #3361, opened 2024-02-19
🐛 do not error out on failing to discover a k8s container

imilchev:
instead of erroring out, we should continue with the next container


providers/k8s/resources/discovery.go
```diff
@@ -822,12 +822,14 @@ func convertImagesToAssets(images map[string]ContainerImage) ([]*inventory.Asset
 
 		ref, err := name.ParseReference(i.resolvedImage, name.WeakValidation)
 		if err != nil {
-			return nil, err
+			log.Error().Err(err).Msg("failed to parse image reference")
```
chris-rock, line 825: That is a good temporary measure, we need to see how we can still report that correctly.

PR #3362, opened 2024-02-19
🐛 load core schema last

imilchev:
makes sure resources from the core provider cannot be overridden

### DIFF
--- a/providers/extensible_schema.go
+++ b/providers/extensible_schema.go
@@ -172,8 +172,16 @@ func (x *extensibleSchema) unsafeRefresh() {
 	res := resources.Schema{
 		Resources: map[string]*resources.ResourceInfo{},
 	}
-	for _, schema := range x.loaded {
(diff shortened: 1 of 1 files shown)


providers/extensible_schema.go
```diff
@@ -172,8 +172,16 @@ func (x *extensibleSchema) unsafeRefresh() {
 	res := resources.Schema{
 		Resources: map[string]*resources.ResourceInfo{},
 	}
-	for _, schema := range x.loaded {
-		res.Add(schema)
+
+	for id, schema := range x.loaded {
+		if id != BuiltinCoreID {
```
chris-rock: would be easier to read if we write:

```
if id == BuiltinCoreID {
  continue
}
res.Add(schema)

```

imilchev: fixed

chris-rock (approved):
Great 🚀

PR #3373, opened 2024-02-20
🐛 guard against nil aws account asset in discovery

vjeffrey:


providers/aws/resources/discovery.go
```diff
@@ -190,8 +190,10 @@ func Discover(runtime *plugin.Runtime, filters connection.DiscoveryFilters) (*in
 	if err != nil {
 		return nil, err
 	}
-
-	awsAccount := res.(*mqlAwsAccount)
+	var awsAccount *mqlAwsAccount
+	if res != nil {
```
chris-rock, line 194: would we not need to abort here? otherwise the followup call with a nil aws account probably also would fail

vjeffrey:
this seems to have been caused by a mismatch in code that was calling this code and updates in this codebase. with the latest of everything it works

PR #3376, opened 2024-02-21
🧹 ensure we grab the tags for the ssm instance during discovery

vjeffrey:
since tags is noted as an optional field, tags(), on ssm instances, we have to call get to actually fetch them


chris-rock (approved):
Thank you @vjeffrey

PR #3392, opened 2024-02-21
🐛 make sure gitlab discovers only unique assets

imilchev:
It is possible to setup gitlab in such a way that 1 project appears in multiple groups. The gitlab provider discovers assets starting from the root and traversing down finding everything along the way. It is possible that the same thing is found in 2 different places. In the end the provider should return a unique list of assets with no duplicates to make sure everything works as expected


chris-rock (approved):
Thank you @imilchev

PR #3406, opened 2024-02-22
🧹 Add annotations to all assets in the inventory.

preslavgerchev:
Ensure all assets in the inventory (whether discovered from the inventory or passed in as fallbacks) get annotated correctly.


chris-rock (approved):
Thank you @preslavgerchev

PR #3407, opened 2024-02-22
🧹 optimize github provider

imilchev:
- rely more on MQL cache
- use a memoizer attached to an MQL resource to store user requests so we don't fetch the same user multiple times


providers/github/connection/platform.go
```diff
@@ -142,8 +149,7 @@ func (c *GithubConnection) Repository() (*github.Repository, error) {
 
 	repoId := conf.Options["repository"]
 	if ownerId != "" && repoId != "" {
-		repo, _, err := c.Client().Repositories.Get(context.Background(), ownerId, repoId)
```
chris-rock, line 145: is there a place where we verify the data?

providers/github/resources/github.go
```diff
@@ -4,11 +4,44 @@
 package resources
 
 import (
+	"context"
 	"time"
 
 	"github.com/google/go-github/v57/github"
+	"github.com/kofalt/go-memoize"
```
chris-rock: if that code is not updated and third pary dependencies are not updated often, we may need to handle this as a util internally.

providers/github/resources/github.lr
```diff
@@ -4,6 +4,8 @@
 option provider = "go.mondoo.com/cnquery/v9/providers/github"
 option go_package = "go.mondoo.com/cnquery/v10/providers/github/resources"
 
+private github {}
```
chris-rock, line 7: great idea

providers/github/resources/github.go
```diff
@@ -4,11 +4,44 @@
...
 )
 
+// We use a global MQL resource for the connection to store the memoizer.
+// In this way we can cache requests for things that aren't directly attached to an MQL resource.
+// For example, we don't want to fetch the same users multiple times, for that we can use the memoizer.
+type mqlGithubInternal struct {
+	memoize *memoize.Memoizer
+}
+
+func GetUser(ctx context.Context, runtime *plugin.Runtime, conn *connection.GithubConnection, user string) (*github.User, error) {
```
chris-rock: can this become a private function?

providers/github/go.mod
```diff
@@ -9,6 +9,7 @@ toolchain go1.22.0
 require (
 	github.com/cockroachdb/errors v1.11.1
 	github.com/google/go-github/v57 v57.0.0
```
tas50: nit: while you're in here you might want to switch over to v59. These never get updated automatically due to how they version them.

providers/github/resources/github.go
imilchev: yup, done

providers/github/connection/platform.go
imilchev: not sure what you mean by this. We don't do in memory validation. We just call the github api eventually with this info

providers/github/resources/github.go
imilchev: I am not sure how to handle that... it has an MIT license, so not sure if we can just copy it

providers/github/go.mod
imilchev: updated that too

tas50: Thanks!

providers-sdk/v1/util/memoize/memoize.go
```diff
@@ -0,0 +1,48 @@
+package memoize
```
imilchev, line 7: this is the original file

providers-sdk/v1/util/memoize/memoize_test.go
```diff
@@ -0,0 +1,76 @@
+package memoize
```
imilchev, line 7: this one has been modified to use a different test library

chris-rock (approved):
Thank you @imilchev

PR #3412, opened 2024-02-22
🐛 Do not return the type of the predicate for if blocks

jaym:
When there were errors, `ifCallV2` would return the type of the predicate for the result of the if block.

That becomes problematic for a query like the following:
```
if (file('/etc/amazon/').exists) {
  return "AWS"
} else if ( users.any( name == 'ec2-user' || name == 'ec2-instance-connect' ) ) {
  return "AWS"
} else if ( machine.baseboard.manufacturer == 'Amazon EC2') {
  return "AWS"
} else if ( asset.platform == "cos" ) {
  return "GCP"
} else if ( file('/etc/google_instance_id').exists ) {
  return "GCP"
} else if ( file('/var/lib/waagent/').exists ) {
  return "Azure"
} else if ( file('/var/lib/dhcp/dhclient.enp1s0.leases').exists && file('/var/lib/dhcp/dhclient.enp1s0.leases').content.lines.contains( /unknown-245/ ) ) {
  return "Azure"
} else {
  return "Unknown"
}
```

If you cannot run `machine.baseboard.manufacturer`, for example if you're not root, an error is returned. The type of the result is boolean, when it should be string. You then get a message like the following:

```
failed to send datapoints error="1 error occurred:\n\t* failed to store data for \"HTCzrF9d2vMJ1iUZsoMcMwaC8XQ7bvQE5nLxC0I5cSBlLXawpmn74WyIEaTgXuUmz/C5iRSeedT3eBg9h6xx8A==\", types don't match: expected string, got bool\n\n"
```

### DIFF
--- a/llx/builtin_global.go
+++ b/llx/builtin_global.go
@@ -67,8 +67,11 @@ func ifCallV2(e *blockExecutor, f *Function, ref uint64) (*RawData, uint64, erro
 	max := len(f.Args)
 	for idx+2 < max {
 		res, dref, err := e.resolveValue(f.Args[idx], ref)
-		if err != nil || dref != 0 || res == nil {
+		if dref != 0 || (err != nil && res == nil) {
 			return res, dref, err
+		}
+		if res.Error != nil {
+			return NilData, 0, res.Error
 		}
 
 		if truthy, _ := res.IsTruthy(); truthy {


jaym:
i haven't been able to get this to fail with the mock tests. It also doesn't seem to happen all the time. For example, an error in `file.content` doesn't produce such an error

jaym:
I think I kind of understand it in a hand wavy way. In the case of `file.content`, which does print any error even though a file doesn't exist, we are able to create the file resource, and error out when we access the field `content`. This triggers some special code called `triggerChainError` which sets everything in the chain to the error value with no type. `machine.baseboard.manufacturer` produces an error much earlier. It produces an error on `machine.baseboard`, which is a resource. Since its not an error on a field, it doesn't go through the same `triggerChainError`.

As an aside, I don't really think we need `triggerChainError`. It really feels like the builtins should decide how to handle errors. For example, if errors were to be treated as false, `triggerChainError` would prevent that from happening because it would report an error for the entire chain before the builtin `if` could do its thing.

arlimus (approved):
Thank you for analyzing this, I can see that return types inside of the conditionals that are failing can lead to the whole thing to fall apart.  Let's get this merged first to move forward.

arlimus:
Let's come up with a decision around how we want to handle errors in conditionals. I'm somewhat in favor of treating them as `false` conditions and moving forward. 

Another approach could be to iterate all conditional branches to see if a truthy branch is found. If none is encountered, it could still opt to expose errors (as seen in other frameworks). The downside with this approach is that it will set user expectations to see errors, but all of a sudden hide them if a truthy branch is found. 

Ultimately it's a decision about the behavior of this code.

PR #3415, opened 2024-02-23
🧹 refactor os provider connections

chris-rock:
This PR is not doing any functional change, it just refactors the code for better readability and reuse. Especially the tar connection was cleaned up and separated from the docker images part.


imilchev (commented):
The code looks much nicer than before. I will do a bit more manual testing before approving this but so far it looks great! 🎉

imilchev (approved):
seems to work as expected

PR #3418, opened 2024-02-23
🏃Speed up azure provider by caching the azure subscription.

preslavgerchev:
By caching the azure subscription we avoid that it's being called every single time a query asks for it.

I ran some manual scans and the scan time goes from 55-60s to 18-20s for me


chris-rock (approved):
Thank you @preslavgerchev

PR #3421, opened 2024-02-24
🐛 handle case where ipv4 or ipv6 is disabled

chris-rock:
fixes https://github.com/mondoohq/cnquery/issues/2957

### DIFF
--- a/providers/os/resources/port.go
+++ b/providers/os/resources/port.go
@@ -253,7 +253,7 @@ func (p *mqlPorts) parseProcNet(path string, protocol string, users map[int64]*m
 	if errors.Is(err, os.ErrNotExist) {
 		return nil, nil
 	} else if err != nil {
-		return nil, errors.New("cannot access stat for " + path)
+		return nil, fmt.Errorf("cannot access stat for "+path+": %v", err)
 	}
 	if stat.IsDir() {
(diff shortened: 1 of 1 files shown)


providers/os/resources/port.go
```diff
@@ -242,13 +243,16 @@ func (p *mqlPorts) processesBySocket() (map[int64]*mqlProcess, error) {
...
 func (p *mqlPorts) parseProcNet(path string, protocol string, users map[int64]*mqlUser) ([]interface{}, error) {
 	conn := p.MqlRuntime.Connection.(shared.Connection)
 	fs := conn.FileSystem()
 	stat, err := fs.Stat(path)
-	if err != nil {
+	// if the file does not exist, we just return nil since no ports are open then
+	if errors.Is(err, os.ErrNotExist) {
+		return nil, nil
+	} else if err != nil {
 		return nil, errors.New("cannot access stat for " + path)
```
preslavgerchev: nit: can we wrap the err here? `cannot access stat for..` does not reveal what the actual problem is

preslavgerchev (approved):

providers/os/resources/port.go
chris-rock: Done!

PR #3424, opened 2024-02-26
try large runners for providers build

imilchev:
- user larger runners
- enable Go cache

https://github.com/mondoohq/cnquery/actions/runs/8047544409
https://github.com/mondoohq/cnquery/actions/runs/8048639872


chris-rock:
I close this for now. We can re-open when we need it.

PR #3426, opened 2024-02-26
✨ Include version in origin field for debian packages

czunker:
Some packages contain the version information for the source package. Where we can find it, we add it.

### DIFF
--- a/providers/os/resources/os.lr
+++ b/providers/os/resources/os.lr
@@ -614,11 +614,8 @@ package @defaults("name version") {
   // Common Platform Enumeration (CPE) for the package
   cpes []core.cpe
 
-  // Package origin (optional)
+  // Package origin (optional) (This includes the version when present)
   origin() string
-  // Package origin version (optional)
-  originVersion() string
-
 
   // Available version
   available string
--- a/providers/os/resources/os.lr
+++ b/providers/os/resources/os.lr
@@ -614,7 +614,7 @@ package @defaults("name version") {
   // Common Platform Enumeration (CPE) for the package
   cpes []core.cpe
 
-  // Package origin (optional) (This includes the version when present)
+  // Package origin, may includes version if available (optional)
   origin() string
 
   // Available version
--- a/providers/os/resources/os.lr
+++ b/providers/os/resources/os.lr
@@ -616,6 +616,9 @@ package @defaults("name version") {
 
   // Package origin (optional)
   origin() string
+  // Package origin version (optional)
+  originVersion() string
+
 
   // Available version
   available string
--- a/providers/os/resources/packages.go
+++ b/providers/os/resources/packages.go
@@ -8,13 +8,12 @@ import (
 	"regexp"
 	"sync"
 
-	"go.mondoo.com/cnquery/v10/types"
-
 	"github.com/rs/zerolog/log"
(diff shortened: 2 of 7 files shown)


czunker:
An exmaple:
```
cnquery run container image samba:bookworm -c "packages.where(origin == /samba/){ name version origin originVersion }" 
packages.where.list: [
  0: {
    name: "libldb2"
    version: "2:2.6.2+samba4.17.12+dfsg-0+deb12u1"
    originVersion: "2:4.17.12+dfsg-0+deb12u1"
    origin: "samba"
  }
  1: {
    name: "python3-ldb"
    version: "2:2.6.2+samba4.17.12+dfsg-0+deb12u1"
    originVersion: "2:4.17.12+dfsg-0+deb12u1"
    origin: "samba"
  }
]
```

providers/os/resources/os.lr
```diff
@@ -616,6 +616,9 @@ package @defaults("name version") {
 
   // Package origin (optional)
   origin() string
+  // Package origin version (optional)
+  originVersion() string
```
chris-rock: should this not be origin package that references another full package? We have the following options:

```
// dict returns name and version field
originPackage() dict

// return full package resource
originPackage() package
```

chris-rock (commented):
I added some comment about the MQL resource. Do we also need to adjust other operating systems that have origin support or is that just debian?

czunker:
> I added some comment about the MQL resource. Do we also need to adjust other operating systems that have origin support or is that just debian?

I checked Alpine. That does not provide a version, only the origin name.

providers/os/resources/os.lr
czunker: As discuessed, I now include the version in the origin field as it is shown in dpkg output:
```
samba (2:4.17.12+dfsg-0+deb12u1)
```

providers/os/resources/packages.go
```diff
@@ -5,10 +5,11 @@ package resources
 
 import (
 	"errors"
-	"go.mondoo.com/cnquery/v10/types"
 	"regexp"
 	"sync"
 
+	"go.mondoo.com/cnquery/v10/types"
+
```
chris-rock: nit: whitespace

providers/os/resources/os.lr
```diff
@@ -614,7 +614,7 @@ package @defaults("name version") {
   // Common Platform Enumeration (CPE) for the package
   cpes []core.cpe
 
-  // Package origin (optional)
+  // Package origin (optional) (This includes the version when present)
```
chris-rock:
```suggestion
  // Package origin, may includes version if available (optional)
```

providers/os/resources/packages/dpkg_packages.go
```diff
@@ -6,15 +6,16 @@ package packages
...
 	"io"
 	"os"
 	"regexp"
 	"strings"
 
+	"github.com/package-url/packageurl-go"
+	"go.mondoo.com/cnquery/v10/providers-sdk/v1/inventory"
+	"go.mondoo.com/cnquery/v10/providers/os/resources/cpe"
+	"go.mondoo.com/cnquery/v10/providers/os/resources/purl"
+
```
chris-rock: nit: whitespace

chris-rock (approved):
Thank you @czunker

PR #3433, opened 2024-02-27
🐛 terraform: avoid nil crash on non-existent block in module

arlimus:
Fixes https://github.com/mondoohq/cnquery/issues/3377


chris-rock (approved):
Thank you @arlimus

PR #3445, opened 2024-02-28
⭐️  Added Trust relationship policy  to the role (aws)

HRouhani:
Screenshot from 2024-02-28 16-54-23

Screenshot from 2024-02-28 16-54-44

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -852,6 +852,8 @@ private aws.iam.role @defaults("arn name") {
   tags map[string]string
   // Time when the role was created
   createDate time
+  // The policy document that grants an entity permission to assume the role
+  assumeRolePolicyDocument dict
 }
 
 // AWS IAM group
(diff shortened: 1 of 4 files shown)


providers/aws/resources/aws.lr
```diff
@@ -852,6 +852,8 @@ private aws.iam.role @defaults("arn name") {
   tags map[string]string
   // Time when the role was created
   createDate time
+  // The policy document that grants an entity permission to assume the role
+  assumeRolePolicyDocument dict
```
vjeffrey, line 856: i wonder if we shouldn't make this a real resource object?  is there a specific test you're targeting with this resource? what would the query look like with this as a dict vs a defined resource?

HRouhani: Thanks for your feedback. So far I wrote like following:

Screenshot from 2024-02-29 18-39-45

vjeffrey: ah, that makes sense. prob better to just leave as a dict

vjeffrey (approved):

PR #3455, opened 2024-02-29
🧹 do not fail s3 buckets resource when no location found, warn and continue

vjeffrey:
if the bucket is owned by a different account we can hit a 403 error


czunker (approved):
Thanks @vjeffrey

PR #3492, opened 2024-03-05
✨ use globally unique connection IDs for providers

imilchev:
Use globally unique connection IDs for providers. This allows us to call `Connect` multiple times for providers without having to create new connections in the process. It also helps us to do cross-provider connections in the future since 1 asset will always have 1 connection ID for all providers it talks to.

**New flow**
Connection IDs are assigned by the coordinator. If a connection has no ID yet, it gets one from the coordinator which makes sure it is globally unique.
- Made sure cloned inventories are cloned without the connection ID. This makes sure discovered assets get their own connection ID
- Made sure that we first check if a connection with this ID exists for the current provider. If there is one, we return that and we do not create a new connection
- This flow is only used for cnquery/cnspec clients that support it. Old clients will use the old way

**Old flow**
The old flow is maximally contained in the base `Service` struct that is reused by all providers. When we clean that up in v12, most of the changes will be in that struct. The old flow is triggered only when we attempt to create a runtime with connection ID 0. This indicates cnquery/cnspec is an old version and doesn't set IDs externally. The old flow will also create a new runtime for an asset that calls `Connect` for a second time

### DIFF
--- a/providers-sdk/v1/inventory/inventory.go
+++ b/providers-sdk/v1/inventory/inventory.go
@@ -417,7 +417,7 @@ func (cfg *Config) Clone(opts ...CloneOption) *Config {
 	}
 
 	clonedObject := proto.Clone(cfg).(*Config)
-
+	clonedObject.Id = 0
 	if cloneSettings.noDiscovery {
 		clonedObject.Discover = &Discovery{}
 	}
--- a/providers-sdk/v1/plugin/service.go
+++ b/providers-sdk/v1/plugin/service.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	llx "go.mondoo.com/cnquery/v10/llx"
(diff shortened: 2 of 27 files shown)


providers-sdk/v1/inventory/inventory.go
```diff
@@ -417,7 +417,7 @@ func (cfg *Config) Clone(opts ...CloneOption) *Config {
 	}
 
 	clonedObject := proto.Clone(cfg).(*Config)
-
+	clonedObject.Id = 0
```
imilchev, line 420: we aren't relying on the `Id` to be copied anywhere, where `Clone` is called. That's why I opt to make it default that the ID isn't copied. If in the future we have cases that we need to copy the ID too, we can add an option to clone the ID too and fix it with that:
```go
clone := inv.Clone(WithId())
```

arlimus (approved):
Awesome!! Thank you for creating globally unique conn IDs :star:

PR #3493, opened 2024-03-05
🤖 Automatically create release PR for providers when cnquery is released

czunker:


.github/workflows/release-providers.yml
```diff
@@ -0,0 +1,67 @@
...
+        uses: actions/setup-go@v5
+        with:
+          go-version: ">=${{ env.golang-version }}"
+          cache: false
+
+      - name: Check for provider updates
+        id: update-providers
+        run: |
+          shopt -s expand_aliases
+          alias version="go run providers-sdk/v1/util/version/version.go"
```
preslavgerchev:
```suggestion
          alias update="go run providers-sdk/v1/util/version/version.go"
```

if i go by the line below

czunker: Good catch. Thanks.

preslavgerchev (approved):

PR #3496, opened 2024-03-05
⭐️ delayed asset discovery

imilchev:


explorer/scan/discovery.go
```diff
@@ -200,7 +200,9 @@ func prepareAsset(a *inventory.Asset, rootAsset *inventory.Asset, runtimeLabels
 	a.AddMondooLabels(rootAsset)
 	a.AddAnnotations(rootAsset.GetAnnotations())
 	a.ManagedBy = rootAsset.ManagedBy
-	a.KindString = a.GetPlatform().Kind
+	if platform := a.GetPlatform(); platform != nil {
+		a.KindString = a.GetPlatform().Kind
```
czunker: nit: You could use `platform`.

providers/os/connection/container/image_connection.go
```diff
@@ -21,6 +22,12 @@ import (
 
 // NewImageConnection uses a container image reference as input and creates a tar connection
 func NewImageConnection(id uint32, conf *inventory.Config, asset *inventory.Asset, img v1.Image) (*tar.Connection, error) {
+	// FIXME: DEPRECATED, remove in v12.0 vv
+	// The SkipDiscovery flag should always be set from v12
```
czunker: nit: The comment should mention `DelayDiscovery`, I think.

providers/os/connection/container/image_connection_test.go
```diff
@@ -68,6 +70,32 @@ type dockerConnTest struct {
...
+	ref, err := name.ParseReference(alpineImage, name.WeakValidation)
+	require.NoError(t, err)
+
+	img, err := image.LoadImageFromRegistry(ref)
+	require.NoError(t, err)
+
+	inv := &inventory.Config{Options: map[string]string{}}
+	_, err = container.NewImageConnection(1, inv, &inventory.Asset{}, img)
+	require.NoError(t, err)
+	assert.True(t, inv.DelayDiscovery)
```
czunker, line 83: Could we add a check for the platform here and in the other test?

Here, we shouldn't have any platform information. In the other test we should know it is an alpine image.

explorer/scan/local_scanner.go
```diff
@@ -243,30 +246,39 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
...
+
+		services, err = explorer.NewRemoteServices(client.ApiEndpoint, client.Plugins, client.HttpClient)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	assetBatches := slicesx.Batch(discoveredAssets.Assets, 100)
 	for i := range assetBatches {
+		log.Info().Msg("synchronize assets")
```
czunker: This will show the message also in cases where we do not have an upstream config.

czunker: This should be somewhere between L270 and L304.

providers-sdk/v1/inventory/inventory.proto
```diff
@@ -197,6 +197,9 @@ message Config {
   // connection
   string platform_id = 26;
   repeated string capabilities = 29;
+  // Determines whether to delay discovery during the connection phase.
+  // Discovery will only happen if this is false
```
czunker: nit: To me this comment is a bit misleading. My understanding is, that discovery will only happen later.

czunker (commented):
Why or in which cases should this be used?

Is it intended, that this is not exposed as a flag?

imilchev:
> Why or in which cases should this be used?
> 
> Is it intended, that this is not exposed as a flag?

I am not sure I understand the question. This is an inventory setting. There is nothing for the user to specify here. The goal is that every provider knows which discovery takes  a lot memory, disk space or IO. It is up to the provider to decide whether delayed discovery should be used for an asset or not. If delayed discovery is used, then just 1 operation at a time will be active during a scan. That's not something the user should decide or care about

explorer/scan/discovery.go
imilchev: fixed

providers/os/connection/container/image_connection.go
imilchev: fixed

providers/os/connection/container/image_connection_test.go
imilchev: discovery is handled in the provider, not in the connection. At this point, we never have a platform regardless of the flag. There are tests that verify the provider sets the platform based on the flag in the provider tests

explorer/scan/local_scanner.go
imilchev: fixed

providers-sdk/v1/inventory/inventory.proto
imilchev: No, discovery will only happen if you call `Connect` and `DelayDiscovery = false`. I will update the comment to clarify this

czunker (approved):

PR #3513, opened 2024-03-06
Elasticsearch -> OpenSearch

misterpantz:
With Amazon's rebranding, updates "Elasticsearch" to "OpenSearch"


misterpantz:
Closing because we're going to make a new provider instead of changing this one, per @tas50.

PR #3518, opened 2024-03-07
🐛 fix incognito scans

### LINKED ISSUE #3502: Suddenly the --incognito flag seems not to work anymore with cnquery scan
**Describe the bug**
`cnquery scan aws` does not work anymore, with or without specifying a querypack.

**To Reproduce**

For AWS, gain AWS credentials first. Then, with or without an optional querypack:
`cnquery scan aws --incognito --querypack mondoo-incident-response-aws`

**Expected behavior**
The output should equal to the same command executed without the `--incognito` flag.

**Screenshots or CLI Output**
```cnquery scan aws --incognito --querypack mondoo-incident-response-aws
→ found a new version for 'aws' provider installed=10.2.7 latest=10.2.8
→ successfully installed aws provider path=/Users/hidden/.config/mondoo/providers/aws version=10.2.8
→ loaded configuration from /Users/hidden/.config/mondoo/mondoo.yml using source default

 AWS Account hidden (123456789012) ───────────────────────────────────────────────────────────────────────────────────────────────────────────────    X


Summary (1 assets)
==================

Target:     AWS Account hidden (123456789012)
error: cannot find any policy for this search
```

- OS: macOS
- OS Version: 14.2.1
- cnquery version
cnquery 10.3.4 (f380756543, 2024-02-17T08:12:07Z)

**Additional context**
This has been working until recently. Even `cnquery scan local --incognito` results in the same error as shown above.

imilchev:
- pass through upstream creds if available
- do not fetch the bundle for every asset
- use upstream bundle when upstream config is available (even if in incognito)

fixes #3502

### DIFF
--- a/apps/cnquery/cmd/scan.go
+++ b/apps/cnquery/cmd/scan.go
@@ -239,14 +239,6 @@ func getCobraScanConfig(cmd *cobra.Command, runtime *providers.Runtime, cliRes *
 
 	serviceAccount := opts.GetServiceCredential()
 	if serviceAccount != nil {
-		// TODO: determine if this needs migrating
-		// // determine information about the client
-		// sysInfo, err := sysinfo.GatherSystemInfo()
-		// if err != nil {
-		// 	log.Warn().Err(err).Msg("could not gather client information")
-		// }
-		// plugins = append(plugins, defaultRangerPlugins(sysInfo, opts.GetFeatures())...)
-
 		log.Info().Msg("using service account credentials")
 		conf.runtime.UpstreamConfig = &upstream.UpstreamConfig{
 			SpaceMrn:    opts.GetParentMrn(),
--- a/apps/cnquery/cmd/scan.go
+++ b/apps/cnquery/cmd/scan.go
@@ -237,31 +237,28 @@ func getCobraScanConfig(cmd *cobra.Command, runtime *providers.Runtime, cliRes *
 		conf.Inventory.ApplyCategory(inventory.AssetCategory_CATEGORY_CICD)
 	}
 
-	var serviceAccount *upstream.ServiceAccountCredentials
-	if !conf.IsIncognito {
-		serviceAccount = opts.GetServiceCredential()
-		if serviceAccount != nil {
-			// TODO: determine if this needs migrating
-			// // determine information about the client
-			// sysInfo, err := sysinfo.GatherSystemInfo()
-			// if err != nil {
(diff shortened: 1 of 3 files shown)


apps/cnquery/cmd/scan.go
```diff
@@ -237,31 +237,28 @@ func getCobraScanConfig(cmd *cobra.Command, runtime *providers.Runtime, cliRes *
...
-				Incognito:   conf.IsIncognito,
-				Creds:       serviceAccount,
-			}
-			providers.DefaultRuntime().UpstreamConfig = conf.runtime.UpstreamConfig
-		} else {
-			log.Warn().Msg("No credentials provided. Switching to --incognito mode.")
-			conf.IsIncognito = true
+	serviceAccount := opts.GetServiceCredential()
+	if serviceAccount != nil {
+		// TODO: determine if this needs migrating
```
chris-rock: i think we do not need this anymore, lets remove the comments

explorer/scan/local_scanner.go
```diff
@@ -561,13 +591,16 @@ func (s *localAssetScanner) runQueryPack() (*AssetReport, error) {
 	var conductor explorer.QueryConductor = s.services
 
 	log.Debug().Str("asset", s.job.Asset.Mrn).Msg("client> request bundle for asset")
-	assetBundle, err := hub.GetBundle(s.job.Ctx, &explorer.Mrn{Mrn: s.job.Asset.Mrn})
-	if err != nil {
-		return nil, err
+	// If we run in debug mode, download the asset bundle and dump it to disk
+	if val, ok := os.LookupEnv("DEBUG"); ok && (val == "1" || val == "true") {
```
chris-rock, line 595: Can we not use the same logic as in here https://github.com/mondoohq/cnquery/blob/1b98d7fd1826f6d851e7dba336437392a6b5ee22/logger/debug.go#L44-L46?

imilchev: we only dump this if the env var `DEBUG=1` is set. I think `log.Debug().Enabled()` is also `true` when we just have `--log-level debug`. I don't think we want to have these files in that case. 

That same condition exists in cnspec for dumping the bundle there too. I just used the same approach

chris-rock: temporarily ok, but we need to extract this into a function. Otherwise it will be hard to change that in the future https://github.com/mondoohq/cnquery/blob/main/logger/debug.go#L77

chris-rock (approved):

PR #3533, opened 2024-03-11
🧹 refactor config gathering for sbom cmd

chris-rock:
This PR harmonizes the config gathering. Since sbom subcommand does not support all flags, the flag parsing has been converted to viper so that it can be used even if a specific flag is not supported.

### DIFF
--- a/apps/cnquery/cmd/root.go
+++ b/apps/cnquery/cmd/root.go
@@ -9,6 +9,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/cockroachdb/errors"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


apps/cnquery/cmd/root.go
```diff
@@ -196,3 +204,103 @@ func GenerateMarkdown(dir string) error {
 
 	return nil
 }
+
+func getCobraScanConfig(cmd *cobra.Command, runtime *providers.Runtime, cliRes *plugin.ParseCLIRes) (*scanConfig, error) {
```
imilchev, line 208: this function can return an error but it seems like we are doing `os.Exit` and `log.Fatal()` instead. I think it's better to return an error and let the caller of the function decide what to do with that error

chris-rock: +1 good call. I'll adjust that

chris-rock: Code updated.

apps/cnquery/cmd/root.go
```diff
@@ -196,3 +204,97 @@ func GenerateMarkdown(dir string) error {
...
+	opts, err := config.Read()
+	if err != nil {
+		log.Fatal().Err(err).Msg("failed to load config")
```
imilchev: missed 1 here

imilchev (approved):

apps/cnquery/cmd/root.go
chris-rock: oh no, will adjust

PR #3535, opened 2024-03-11
🐛 fix gcp init functions panic

imilchev:
some of the init functions in the GCP provider weren't properly checking for missing asset IDs, which could result in provider panics

also dataproc config was trying to set a non-existent field


mm-weber (approved):
Tried and tested.
LGTM

mm-weber:
Thanks @imilchev !

PR #3540, opened 2024-03-12
⭐ add TechnologyUrl to platform + AssetUrlTrees to provider

arlimus:
1. Providers are extended to define their AssetUrlTrees. These are part of the config and can be loaded statically. This allows developers to define and extend existing AssetUrlTrees. Note that we currently maintain a `/technology=X` asset tree. We are in the process of adding all providers to this grouping strategy.
2. Extend the inventory definition of the asset `Platform` to include the TechnologyUrl. While assets may belong to a multitude of URLs, the platform definition only focuses on the technology URL.
3. Implement both for the `os` provider. Note: Initially this was designed to define the entire tree of URLs, including all OS names and versions. I noticed, however, that the structure of the tree doesn't change by different OS families, names, or versions, at all. Thus, we use `*` to allow for arbitrary terms, as these terms won't have any impact on the remaining URL.


providers/os/detector/detector.go
```diff
@@ -11,10 +11,57 @@ import (
...
+	if platform.Kind == "container-image" {
+		platform.TechnologyUrlSegments = []string{
+			// technology, kind
+			"container", platform.Kind,
+		}
+	} else {
+		platform.TechnologyUrlSegments = []string{"os"}
+	}
+
+	platform.TechnologyUrlSegments = append(platform.TechnologyUrlSegments,
```
jaym, line 70: is it possible platform name or platform version is not available?

arlimus: technically no, but since there is a longer tail of code sitting behind this, let's be safe than sorry for this use-case and prevent that from happening; PR updated

chris-rock (approved):

PR #3547, opened 2024-03-12
🐛 fix init for GCP resources

imilchev:
multiple resources didn't properly return errors on init if not found. That caused panics


preslavgerchev (approved):

mm-weber:
Tried and tested.
Works!
Thanks @imilchev !

PR #3549, opened 2024-03-12
🐛 Checkout history for provider release

czunker:


czunker:
Now, the workflow knows the history: https://github.com/mondoohq/cnquery/actions/runs/8248523654/job/22559022125

Before it showed:
```
! looks like there is no previous version in your commit history
```

chris-rock (approved):

PR #3551, opened 2024-03-12
Fix BuildQueries

jaym:
This rewrites BuildQueries. The indended algorithm is as follows:
- Find all the initial nodes that could match any of the filters
- Starting at these nodes, walk to the root of the tree
  - If during the walk, all filters end up matching, its a valid branch that needs to be returned
  - This probably tries to be too clever. The actual implementation keeps a memo table, tracking each filter that matches the (node,value) pair. The current node matches all the filters its parent (node,value) matches, so look that up / compute it. Walk through the filters and see if any of them match the current one.
  - The memo table at the end of that process contains each possible node that can match and which filters it matched. If it matched all of them, build the branch.

### DIFF
--- a/providers-sdk/v1/inventory/asset_url.go
+++ b/providers-sdk/v1/inventory/asset_url.go
@@ -45,6 +45,10 @@ type AssetUrlSchema struct {
 	keys map[string][]*AssetUrlBranch
 }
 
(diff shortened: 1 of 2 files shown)


arlimus:
Suprseded by https://github.com/mondoohq/cnquery/pull/3556

PR #3556, opened 2024-03-13
🐛 fix multiple parameters in AssetURL

arlimus:
We returned too many queries from the list. This approach fixes the underlying bug by:
1. making sure all requested parameters are in the query and
2. removing nodes from the search tree once a descendent is found

### DIFF
--- a/providers-sdk/v1/inventory/asset_url_test.go
+++ b/providers-sdk/v1/inventory/asset_url_test.go
@@ -105,7 +105,7 @@ func TestAddSubtree(t *testing.T) {
 		assert.Equal(t, []AssetUrlChain{
 			{
 				KV{"technology", "aws"},
-				KV{"account", "*"},
+				KV{"account", "123"},
 				KV{"service", "ec2"},
 				KV{"family", "windows"},
 				KV{"platform", "windows server"},
--- a/providers-sdk/v1/inventory/asset_url_test.go
+++ b/providers-sdk/v1/inventory/asset_url_test.go
@@ -71,7 +71,7 @@ func TestAddSubtree(t *testing.T) {
 		}, keys)
 	})
 
-	t.Run("build a query with referenced tree positions", func(t *testing.T) {
+	t.Run("build query with 1 params and 2 results", func(t *testing.T) {
 		err := root.RefreshCache()
 		require.NoError(t, err)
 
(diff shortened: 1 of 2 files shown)


providers-sdk/v1/inventory/asset_url_test.go
```diff
@@ -93,4 +93,40 @@ func TestAddSubtree(t *testing.T) {
...
+		require.NoError(t, err)
+
+		queries := root.BuildQueries([]KV{
+			{Key: "account", Value: "123"},
+			{Key: "platform", Value: "windows server"},
+		})
+		assert.Equal(t, []AssetUrlChain{
+			{
+				KV{"technology", "aws"},
+				KV{"account", "*"},
```
jaym: rereading this test, it seems that this might not be what we want. In this case, we've set account to something specific and we should probably preserve that

jaym: perhaps something like this:
```
diff --git a/providers-sdk/v1/inventory/asset_url.go b/providers-sdk/v1/inventory/asset_url.go
index 796d44a86..94d22b323 100644
--- a/providers-sdk/v1/inventory/asset_url.go
+++ b/providers-sdk/v1/inventory/asset_url.go
@@ -369,7 +369,7 @@ func filterArr[T comparable](arr []T, entry T) []T {
        return arr
 }

-func filterKV(kvs []KV, key string, value string) []KV {
+func filterKV(kvs []KV, key string, value string) ([]KV, string) {
        found := -1
        for i := range kvs {
                if kvs[i].Key == key && (value == "*" || kvs[i].Value == value) {
@@ -378,10 +378,11 @@ func filterKV(kvs []KV, key string, value string) []KV {
                }
        }
        if found == -1 {
-               return kvs
+               return kvs, ""
        }
+       boundValue := kvs[found].Value
        kvs[found] = kvs[len(kvs)-1]
-       return kvs[:len(kvs)-1]
+       return kvs[:len(kvs)-1], boundValue
 }

 // Build a parent query from a leaf in an asset url schema and a given value
@@ -399,7 +400,11 @@ func buildParentQuery(leaf *AssetUrlBranch, value string, requiredKVs []KV) Asse
                        Value: curValue,
                }

-               requiredKVs = filterKV(requiredKVs, cur.Key, curValue)
+               var boundValue string
+               requiredKVs, boundValue = filterKV(requiredKVs, cur.Key, curValue)
+               if boundValue != "" {
+                       res[cur.Depth-1].Value = boundValue
+               }
                curValue = cur.ParentValue
                cur = cur.Parent
        }
diff --git a/providers-sdk/v1/inventory/asset_url_test.go b/providers-sdk/v1/inventory/asset_url_test.go
index 84bd1e273..98fef0500 100644
--- a/providers-sdk/v1/inventory/asset_url_test.go
+++ b/providers-sdk/v1/inventory/asset_url_test.go
@@ -105,7 +105,7 @@ func TestAddSubtree(t *testing.T) {
                assert.Equal(t, []AssetUrlChain{
                        {
                                KV{"technology", "aws"},
-                               KV{"account", "*"},
+                               KV{"account", "123"},
                                KV{"service", "ec2"},
                                KV{"family", "windows"},
                                KV{"platform", "windows server"},
@@ -130,3 +130,46 @@ func TestAddSubtree(t *testing.T) {
                }, queries)
        })
 }
 ```

arlimus: Good point, i'll adjust

jaym (approved):

PR #3572, opened 2024-03-14
🧹 make ansible inventory parser more restrict

chris-rock:
**Restricter Ansible Inventory Parsing**

This implementation makes the ansible inventory parser more restrict. It tests the passed inventory data and returns a proper error. 

While the following inventory is a valid ansible inventory, cnquery & cnspec only parse fully resolved inventories.

```yaml
all:
  hosts:
    ub2204:
      ansible_host: 34.69.247.108
      ansible_user: chris
      ansible_ssh_private_key_file: ~/.ssh/id_ed25519
```

Instead of a silent pass-through where the inventory could not be parsed and therefore resolves 0 assets, we return a proper error now. This also prevents that the inventory falls back to local machine.

```shell
> cnspec scan --inventory-format-ansible --inventory-file ansible-inventory.yaml
→ load inventory inventory-file=providers-sdk/v1/inventory/ansibleinventory/testdata/ansible-inventory.yaml
→ use ansible inventory
FTL failed to parse inventory error="could not parse inventory: unsupported ansible inventory, use `ansible-inventory -i {inventory} --list` to convert it first"
```

This message highlights the solution. We need to convert the ansible inventory first with the `ansible-inventory` command.

**Handle relative paths for identities eg. ~/.ssh/.id_rsa**

Before this PR we could not handle `~/.ssh/.id_rsa` properly. This issue in combination with the case above resulted in unexpected user behavior. This PR changes that and resolves `~` to the home directory.

### DIFF
--- a/explorer/scan/discovery.go
+++ b/explorer/scan/discovery.go
@@ -74,7 +74,7 @@ func (d *DiscoveredAssets) GetAssetsByPlatformID(platformID string) []*AssetWith
 func DiscoverAssets(ctx context.Context, inv *inventory.Inventory, upstream *upstream.UpstreamConfig, recording llx.Recording) (*DiscoveredAssets, error) {
 	im, err := manager.NewManager(manager.WithInventory(inv, providers.DefaultRuntime()))
 	if err != nil {
-		return nil, errors.New("failed to resolve inventory for connection")
+		return nil, errors.New("failed to resolve inventory for connection: " + err.Error())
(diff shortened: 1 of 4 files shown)


chris-rock:
also related to https://github.com/mondoohq/cnspec/issues/1124

czunker (approved):
Works as described.

Thanks @chris-rock

PR #3588, opened 2024-03-18
🧹 add tags to ec2 instances

vjeffrey:
these labels are useful for users, and were included in v8, but were removed when moving to v9. adding them back in


czunker (approved):
Thanks @vjeffrey

PR #3589, opened 2024-03-18
🐛 ensure ec2 instances get a name when scanned

vjeffrey:
default to instance id for the name, give it the name if we get the tag

fixes https://github.com/mondoohq/cnspec/issues/1144


czunker (approved):
Thanks @vjeffrey

PR #3594, opened 2024-03-18
🐛 fix unset fromPort/toPort and port on aws resources

arlimus:
When the fromPort and toPort on security groups are not set, they are not internally treated as 0, but rather -1 (which is similar to unset). Technically we could even expose the unset state, but we should reserve that for the future.

I also updated a few other occurances of port to make it show up as -1. The code for number handling in AWS responses has been further simplified.


vjeffrey (approved):
nice, thank you!

PR #3596, opened 2024-03-19
🐛 fix aws iam {attachedRoles {*}} query

vjeffrey:
fixes https://github.com/mondoohq/cnquery/issues/3582

### DIFF
--- a/providers/aws/resources/aws_iam.go
+++ b/providers/aws/resources/aws_iam.go
@@ -1229,12 +1229,23 @@ func initAwsIamRole(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[
 		}
 
 		role := resp.Role
+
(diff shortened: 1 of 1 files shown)


providers/aws/resources/aws_iam.go
```diff
@@ -1229,12 +1229,23 @@ func initAwsIamRole(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[
 		}
 
 		role := resp.Role
+
+		var policyDocumentMap map[string]interface{}
+		if err == nil && role != nil && role.AssumeRolePolicyDocument != nil {
```
czunker: nit: `err` will be nil here, otherwise the func would have returned in L1228.

czunker (approved):
Thanks @vjeffrey

PR #3601, opened 2024-03-19
🐛 fix upstream scans

imilchev:
seems like we always need to get the bundle when running a cnquery scan... the implementation here is much different from the one in cnspec and doesn't work without the bundle


chris-rock (approved):
Thank you @imilchev

PR #3613, opened 2024-03-20
🧹 refactor os users initialization and lookups

imilchev:
This is a different approach for managing the os users. Instead of caching and locking things, we can just use MQL and it's init functions. That still means we are going to be looping over the whole list of users instead of using a map in some cases. However, the code is simple and it matches what we do with resources in other providers.

Will need further testing to confirm it works correctly


imilchev:
we will do this at a later point

PR #3614, opened 2024-03-20
🐛 check if users exist before trying to use them

imilchev:
prevents panics if the users for a port or a group don't exist for some reason


tas50:
Tested on my failing system

jaym (approved):

PR #3616, opened 2024-03-21
⭐️ collect running kernel in sbom

chris-rock:


sbom/sbom.mql.yaml
```diff
@@ -12,10 +12,16 @@ packs:
         mql: asset { name platform version arch ids labels cpes.map(uri) }
       - uid: mondoo-sbom-packages
         title: Retrieve list of installed packages
-        mql: packages { name version purl cpes.map(uri) arch origin format files.map(path) }
```
czunker: When we remove this here, we also need to remove it in the backend.

chris-rock: Nice catch, that is an accident. Just accelerated testing. Will remove.

chris-rock:
@czunker ready for re-review

czunker (approved):
Thanks @chris-rock

PR #3633, opened 2024-03-22
🧹 Add function to load the asset url schema for all known providers

jaym:
add a function to make the asset url schema from the providers

### DIFF
--- a/providers-sdk/v1/inventory/asset_url.go
+++ b/providers-sdk/v1/inventory/asset_url.go
@@ -149,8 +149,32 @@ func (a *AssetUrlSchema) Add(branch *AssetUrlBranch) error {
 	}
 
 	branch.setDepth(found.Depth + 1)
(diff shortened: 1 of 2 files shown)


providers-sdk/v1/inventory/asset_url.go
```diff
@@ -149,10 +149,34 @@ func (a *AssetUrlSchema) Add(branch *AssetUrlBranch) error {
 	}
 
 	branch.setDepth(found.Depth + 1)
-	found.Values[lastKey] = branch
+	found.Values[lastKey] = copyBranch(branch)
```
jaym, line 152: we don't own these, so we copy them. If we don't copy them, then creating an `NewAssetUrlSchema` produces garbage on the second go around as this code modifies it in a way that is not repeatable.

chris-rock (approved):

PR #3635, opened 2024-03-22
Add more fields to aws.autoscaling.group and aws.cloudfront.distribution

tas50:
- cnames for cloudfront distributions
- desiredCapacity for autoscaling
- maxInstanceLifetime for autoscaling


providers/aws/resources/aws.lr
```diff
@@ -1456,6 +1460,8 @@ private aws.cloudfront.distribution @defaults("domainName status") {
   enabled bool
   // The price class of the distribution
   priceClass string
+  // CNAMEs aliases if any for this distribution
+  cnames []dict
```
imilchev: shouldn't this be `[]string`?

tas50: yep

imilchev (approved):

PR #3645, opened 2024-03-26
🐛 fix private image scanning for k8s containers

imilchev:
scanning private images from a k8s cluster was broken since merging https://github.com/mondoohq/cnquery/pull/2719 The implementation is not correct because it is not providing any pull secrets for resolving the containers. On top of that, we should try to avoid direct dependencies between providers.

This PR reverts #2719. In case the operator tests are breaking, we should fix them instead of changing the discovery code


providers/vsphere/provider/provider_test.go
```diff
@@ -164,5 +153,5 @@ func TestVsphereDiscovery(t *testing.T) {
 	}, nil)
 	require.NoError(t, err)
 	assert.NotNil(t, resp.Asset)
-	assert.Equal(t, 8, len(resp.Inventory.Spec.Assets)) // api + esx + vm
+	// assert.Equal(t, 8, len(resp.Inventory.Spec.Assets)) // api + esx + vm
```
czunker, line 156: nit: shouldn't this be 1, when only discovering the API?

chris-rock (approved):

czunker (approved):
I tried it with the images in the mondoo-operator namespace in edge. Asset names are the same. LGTM.

Thanks for enabling and fixing all these tests! :tada:

providers/vsphere/provider/provider_test.go
imilchev: apparently not. That is something @chris-rock suggested because this test wasn't working initially

PR #3646, opened 2024-03-26
🐛 fix docker image scan for remote images

imilchev:
makes sure we can scan docker images that aren't present locally


chris-rock (approved):
Works like a charm. Thank you @imilchev

PR #3660, opened 2024-03-26
 🧹 Reduce the number of times we try to get platform ident

jaym:
We were getting platform identity information multiple times. We would
 connect, then see the asset's platform not set, and redo it.

### DIFF
--- a/providers/os/provider/provider_test.go
+++ b/providers/os/provider/provider_test.go
@@ -39,10 +39,8 @@ func TestLocalConnectionIdDetectors(t *testing.T) {
 	require.Contains(t, connectResp.Asset.IdDetector, ids.IdDetector_Hostname)
 	require.Contains(t, connectResp.Asset.IdDetector, ids.IdDetector_CloudDetect)
 	require.NotContains(t, connectResp.Asset.IdDetector, ids.IdDetector_SshHostkey)
-	// here we have the hostname twice, as platformid and stand alone
-	// This get's cleaned up later in the code
-	// FIXME: this should only be 1
-	require.Len(t, connectResp.Asset.PlatformIds, 2)
+
+	require.Len(t, connectResp.Asset.PlatformIds, 1)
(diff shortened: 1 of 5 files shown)


providers/os/provider/provider_test.go
```diff
@@ -106,8 +104,8 @@ func TestLocalConnectionIdDetectors_DelayedDiscovery(t *testing.T) {
 	require.Contains(t, connectResp.Asset.IdDetector, ids.IdDetector_Hostname)
 	require.Contains(t, connectResp.Asset.IdDetector, ids.IdDetector_CloudDetect)
 	require.NotContains(t, connectResp.Asset.IdDetector, ids.IdDetector_SshHostkey)
-	// Now the platformIDs are cleaned up
-	require.Len(t, connectResp.Asset.PlatformIds, 2)
```
jaym, line 110: This list had a thing that wasn't a platform id in it. It was just the hostname directly

providers/os/provider/provider_test.go
```diff
@@ -92,7 +90,7 @@ func TestLocalConnectionIdDetectors_DelayedDiscovery(t *testing.T) {
 	require.Contains(t, connectResp.Asset.IdDetector, ids.IdDetector_CloudDetect)
 	require.NotContains(t, connectResp.Asset.IdDetector, ids.IdDetector_SshHostkey)
 	require.Len(t, connectResp.Asset.PlatformIds, 1)
-	require.Nil(t, connectResp.Asset.Platform)
+	require.Nil(t, connectResp.Asset.Platform) // Why does this need to be nil
```
jaym: I don't know why this needs to be nil

imilchev: if we have the `DelayDiscovery` set to true, then the platform is not set because we haven't performed the "discover" step

imilchev (approved):

PR #3668, opened 2024-03-28
⭐ resources explorer API

arlimus:
1. **Migrate resources explorer to cnquery**
   It was originally added to cnspec when we released v10. However, that concept doesn't hold up well if we think about the overall usage with both cnquery and cnspec. Thus, the API was moved.

2. **Create a separate service for the ResourcesExplorer**
   We had the option to add it to the QueryHub or QueryConductor. However, it is best served as a standalone service. This allows users to create separately scalable services, i.e. services for ingress and egress. For this use-case particularly they can be separately routed while keeping small.

3. **Restructure recordings**
   They now live in their own package and more clearly showcase local multi-asset recordings, upstream data, and single-asset recordings. There are more structural changes coming to further simplify this code, but this is a first step to unpack it. The words are also more aligned with go's naming scheme. (Feels like proto due to its name flattening has led us to some bad naming which is hard to eliminate)

Follow-ups:
- we currently infer the primary provider from the asset, however it would be better to store it explicitly

Pairs with https://github.com/mondoohq/cnspec/pull/1206


cli/providers/providers.go
```diff
@@ -454,7 +455,7 @@ func setConnector(provider *plugin.Provider, connector *plugin.Connector, run fu
 		}
 		doRecord := record != ""
 
-		recording, err := providers.NewRecording(recordingPath, providers.RecordingOptions{
+		recording, err := recording.NewWithFile(recordingPath, recording.RecordingOptions{
```
chris-rock, line 458: I like the new naming

chris-rock (commented):
Thank you @arlimus

chris-rock (approved):

PR #3686, opened 2024-04-01
🔁 use retry client for aws connection

vjeffrey:


providers/aws/connection/connection.go
```diff
@@ -335,3 +343,32 @@ func (h *AwsConnection) Regions() ([]string, error) {
 	h.clientcache.Store("_regions", &CacheEntry{Data: regions})
 	return regions, nil
 }
+
+// zeroLogAdapter is the adapter for retryablehttp is outputting debug messages
+type zeroLogAdapter struct{}
```
chris-rock, line 348: we may want to move this to the provider sdk so we can reuse it for many providers.

chris-rock (approved):

PR #3697, opened 2024-04-03
⭐️ define proto for cnspec report

chris-rock:


cli/reporter/cnquery_report.proto
```diff
@@ -0,0 +1,34 @@
...
+  string type = 3;
+  string platform_name = 4;
+}
+
+message DataValues {
+  map <string, DataValue> values = 1;
+}
+
+message DataValue {
+  string mrn = 1;
```
chris-rock: remove mrn

apps/cnquery/cmd/plugin.go
```diff
@@ -197,7 +197,7 @@ func (c *cnqueryPlugin) RunQuery(conf *run.RunQueryConfig, runtime *providers.Ru
 		if conf.Format != "json" {
 			sh.PrintResults(code, results)
 		} else {
-			reporter.BundleResultsToJSON(code, results, out)
+			_ = reporter.CodeBundleToJSON(code, results, out)
```
preslavgerchev, line 200: why discard the result here?

cli/reporter/proto.go
```diff
@@ -0,0 +1,126 @@
...
+				queryMrnIdx[query.CodeId] = query.Mrn
+			}
+		}
+	}
+
+	// fill in assets
+	for assetMrn, a := range data.Assets {
+		pAsset := &Asset{
+			Mrn:  a.Mrn,
+			Name: a.Name,
```
preslavgerchev, line 46: do we also want the platform name and type here?

cli/reporter/cnquery_report.proto
```diff
@@ -0,0 +1,33 @@
...
+message Report {
+  map <string, Asset> assets = 1;
+  map <string, DataValues> data = 2;
+  map <string, string> errors = 3;
+}
+
+message Asset {
+  string mrn = 1;
+  string name = 2;
+  string type = 3;
```
preslavgerchev: whats type representing?

cli/reporter/proto.go
```diff
@@ -0,0 +1,126 @@
...
+		assetMrn := prettyPrintString(id)
+		errorMsg := errStatus.Message
+		protoReport.Errors[assetMrn] = errorMsg
+	}
+
+	return protoReport, nil
+}
+
+func (r *Report) ToJSON() ([]byte, error) {
+	return protojson.Marshal(r)
```
preslavgerchev, line 113: does this intentionally use `protojson` instead of just `json?`

cli/reporter/proto.go
chris-rock: We do not have that information right now, in cnquery asset. We can add that later if required.

cli/reporter/proto.go
chris-rock: Should make it more safe for proto. Output should be the same.

apps/cnquery/cmd/plugin.go
chris-rock: we only discard the error. output is rendered into out writer

cli/reporter/cnquery_report.proto
chris-rock: We do not use it right now, will remove

preslavgerchev (approved):

PR #3698, opened 2024-04-04
🧹 Unify az auth methods

preslavgerchev:
* Move https://github.com/mondoohq/cnspec/blob/main/cli/reporter/azure_service_bus_handler.go#L90 here so we can reuse this
* Combine the `auth` packages from ms365 and azure into one function in `providers-sdk/v1/utils/azauth`. The code was identical so there was no sense to have this separate

This means all az auth logic is now in a single module


go.mod
```diff
@@ -418,14 +418,22 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
 
-require github.com/patrickmn/go-cache v2.1.0+incompatible
+require (
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.11.1
```
chris-rock, line 422: Is there a way to avoid adding the azure dependency to core cnquery? eg. could ms365 reference the azure provider go mod and use go code from there?

preslavgerchev: discussed, we will solve this at a later point

chris-rock: We also deem this ok since it is not pulling the whole azure sdk.

chris-rock (approved):
Thank you @preslavgerchev

PR #3712, opened 2024-04-05
🧹 Replace deprecated azure dependencies.

preslavgerchev:
I noticed that `go mod` warns about those being deprecated so I replaced them all


chris-rock (approved):
Nice

PR #3714, opened 2024-04-07
⭐️ check for enabled gcp services

chris-rock:
This change makes better use of GCPs service API. It checks if the service is enabled before MQL tries to fetch data (which will return in errors).

```
error: googleapi: Error 403: Compute Engine API has not been used in project example-test before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/compute.googleapis.com/overview?project=example-test then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.
Details:
[
  {
    "@type": "type.googleapis.com/google.rpc.Help",
    "links": [
      {
        "description": "Google developers console API activation",
        "url": "https://console.developers.google.com/apis/api/compute.googleapis.com/overview?project=example-test"
      }
]
```

### DIFF
--- a/providers/gcp/resources/access_approval.go
+++ b/providers/gcp/resources/access_approval.go
@@ -33,6 +33,15 @@ func (g *mqlGcpProject) accessApprovalSettings() (*mqlGcpAccessApprovalSettings,
 	}
 	id := g.Id.Data
 
+	serviceEnabled, err := g.isServiceEnabled(service_accessapproval)
+	if err != nil {
+		return nil, err
+	}
+	if !serviceEnabled {
+		g.AccessApprovalSettings.State = plugin.StateIsSet | plugin.StateIsNull
+		return nil, nil
+	}
+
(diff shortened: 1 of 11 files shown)


providers/gcp/resources/access_approval.go
```diff
@@ -33,6 +33,15 @@ func (g *mqlGcpProject) accessApprovalSettings() (*mqlGcpAccessApprovalSettings,
 	}
 	id := g.Id.Data
 
+	serviceEnabled, err := g.isServiceEnabled(service_accessapproval)
+	if err != nil {
+		return nil, err
+	}
+	if !serviceEnabled {
+		g.AccessApprovalSettings.State = plugin.StateIsSet | plugin.StateIsNull
```
imilchev, line 41: I know in cases where we return nil and not error at some point further down the code, the provider panics. Is this the way to make sure MQL isn't trying to dereference a nil pointer?

chris-rock: Yes, this prevents the panic because it sets the state explicitly.

imilchev (approved):

PR #3715, opened 2024-04-07
🐛 fix folder discovery

chris-rock:
The discovery for folder was not properly implemented for the new plugin mechanism. With this PR users can discover all project by using the folder id:

```
> cnspec shell  gcp folder 12345567
```

```
> cnspec shell  gcp folder 12345567                                                                                                                                                                                                                                                                                                                                                                                                                    

    Available assets                       
                                           
  > 1. GCP Folder 12345567 (gcp-folder)
    2. luna-example (gcp-project)  
```


providers/gcp/resources/discovery.go
```diff
@@ -153,6 +184,50 @@ func discoverOrganization(conn *connection.GcpConnection, gcpOrg *mqlGcpOrganiza
...
+func discoverFolder(conn *connection.GcpConnection, gcpFolder *mqlGcpFolder) ([]*inventory.Asset, error) {
+	assetList := []*inventory.Asset{}
+
+	if stringx.ContainsAnyOf(conn.Conf.Discover.Targets, DiscoveryAll, DiscoveryAuto, DiscoveryProjects) {
+		projects := gcpFolder.GetProjects()
+		if projects.Error != nil {
+			return nil, projects.Error
+		}
+
+		projectList := projects.Data.GetList() // resolve all projects including nested
```
imilchev, line 234: don't remember this fully but wasn't it possible to have a folder inside a folder in GCP? If that's the case, then I guess with this code we don't discover nested folders, but only nested projects

chris-rock: That is ok, we can tackle that later.

PR #3724, opened 2024-04-08
⭐ Basic asset url for aws

jaym:


providers/aws/connection/platform.go
```diff
@@ -6,23 +6,25 @@ package connection
...
 	return &inventory.Platform{
-		Name:    "aws",
-		Title:   "AWS Account",
-		Kind:    "api",
-		Runtime: "aws",
+		Name:                  "aws",
+		Title:                 "AWS Account",
+		Kind:                  "api",
+		Runtime:               "aws",
+		TechnologyUrlSegments: []string{"aws", accountId, "other"},
```
jaym: I'm not sure how we want the asset url for things like accounts themselves to look. I don't think we can have something like `tech=aws/account=accountid[/service=whatever]`, where service is optional

chris-rock (approved):

PR #3727, opened 2024-04-08
🧹 Remove scope from gcp urls

jaym:
The format is now "/gcp/projectid/servicename/...". Project resources get a special "servicename" call "project"


chris-rock (approved):

chris-rock (approved):
Thank you @jaym

PR #3734, opened 2024-04-09
⭐️ github asset urls

chris-rock:


providers/github/config/config.go
```diff
@@ -31,4 +32,35 @@ var Config = plugin.Provider{
...
+				"organization": {
+					Key:   "organization",
+					Title: "Organization",
+					Values: map[string]*inventory.AssetUrlBranch{
+						"organization": nil,
+						"*": {
+							Key:   "repository",
+							Title: "Repository",
+							Values: map[string]*inventory.AssetUrlBranch{
+								// any valid version for the OS
```
czunker: Looks like a Copy & Paste error.

providers/github/connection/platform.go
```diff
@@ -19,26 +19,29 @@ const (
...
+		Runtime:               "github",
+		TechnologyUrlSegments: []string{"github", "user"},
+	}
+	GithubOrgPlatform = inventory.Platform{
+		Name:                  "github-org",
+		Title:                 "GitHub Organization",
+		Family:                []string{"github"},
+		Kind:                  "api",
+		Runtime:               "github",
+		TechnologyUrlSegments: []string{"github", "org"},
```
czunker: I'm not sure about this:
Shouldn't "org" match something in the UrlTree? There we have "organization".

providers/github/connection/platform.go
```diff
@@ -58,20 +61,40 @@ type RepositoryId struct {
...
 
 	if userId := conf.Options["user"]; userId != "" {
-		return GithubUserPlatform, nil
+
+		return NewGithubUserPlatform(userId), nil
 	}
 
 	if repo := conf.Options["repository"]; repo != "" {
-		return GithubRepoPlatform, nil
+		owner := conf.Options["owner"]
```
czunker, line 70: Do we check somewhere else, that "owner" is set in the options?

providers/github/connection/platform.go
```diff
@@ -19,26 +19,29 @@ const (
...
-		Family:  []string{"github"},
-		Kind:    "api",
-		Runtime: "github",
+	GithubRepoPlatform = inventory.Platform{
+		Name:                  "github-repo",
+		Title:                 "GitHub Repository",
+		Family:                []string{"github"},
+		Kind:                  "api",
+		Runtime:               "github",
+		TechnologyUrlSegments: []string{"github", "repo"},
```
czunker: I'm not sure about this:
Shouldn't "repo" match something in the UrlTree? There we have "repository".

providers/github/connection/platform.go
chris-rock: That is a flexible entry the config just sets the schema

providers/github/connection/platform.go
chris-rock: It is, "*" means that all values that do not have "org" are repositories

providers/github/connection/platform.go
chris-rock: we can remove them here since we set them in the New* functions

providers/github/connection/platform.go
chris-rock: we may need more checking, but that would need to be done earlier https://github.com/mondoohq/cnquery/blob/main/providers/github/provider/provider.go#L167-L176

czunker (approved):
Thanks @chris-rock

PR #3738, opened 2024-04-09
🧹 Refactor auth opts for registries. Pass through opts when pulling registry images

preslavgerchev:
* Move all `remote.Options` code to `auth/options`
 * Ensure that if we pass external opts, those are also passed all the way down to the image fetching
 * Remove redundant code

### DIFF
--- a/providers/os/resources/discovery/container_registry/registry.go
+++ b/providers/os/resources/discovery/container_registry/registry.go
@@ -4,12 +4,8 @@
 package container_registry
 
 import (
-	"crypto/tls"
 	"fmt"
-	"net"
-	"net/http"
 	"net/url"
-	"time"
 
 	"github.com/cockroachdb/errors"
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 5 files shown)


providers/os/resources/discovery/container_registry/resolver.go
```diff
@@ -50,12 +47,11 @@ func (r *Resolver) Resolve(ctx context.Context, root *inventory.Asset, conf *inv
 	if err == nil {
 		log.Debug().Str("image", conf.Host).Msg("detected container image in container registry")
 
-		remoteOpts := AuthOption(conf.Credentials, credsResolver)
-		// we need to disable default keychain auth if an auth method was found
-		if len(remoteOpts) > 0 {
```
preslavgerchev, line 55: this was redundant, if we ensure that the auth opt always returns then we dont need the `DisableKeychainAuth` since we can have the `AuthOption` decide whats the best auth method by looking at the creds

providers/os/resources/discovery/container_registry/registry.go
```diff
@@ -164,7 +121,7 @@ func (a *DockerRegistryImages) ListRepository(repoName string) ([]*inventory.Ass
 			return nil, fmt.Errorf("parsing reference %q: %v", repoWithTag, err)
 		}
 
-		a, err := a.toAsset(ref, nil)
+		a, err := a.toAsset(ref, nil, opts...)
```
preslavgerchev, line 124: pass opts thru to ensure these are respected

chris-rock (approved):
Very nice cleanup

PR #3751, opened 2024-04-15
🛑 switch to json v2 output format

chris-rock:
This PR enables the v2 json format as standard. The new format is accessible via:

```
cnquery scan local --json
```

or 

```
cnquery scan local --output json
```

For backwards-compatibility, the old format is available via:

```
cnquery scan local --output json-v1
```


arlimus (approved):
Great to see this move forward!

PR #3752, opened 2024-04-15
✨Add trace-id to sbom proto.

preslavgerchev:
Pass the asset's `trace_id` through to the sbom asset.

Also ensure `GenerateSbom` returns a list of sbom pointers


sbom/sbom.go
```diff
@@ -9,10 +9,11 @@ import (
 	_ "embed"
 	"encoding/json"
 	"fmt"
-	"go.mondoo.com/cnquery/v10/cli/reporter"
 	"strings"
 	"time"
 
+	"go.mondoo.com/cnquery/v10/cli/reporter"
+
```
chris-rock: nit: newline

chris-rock (approved):
Thank you @preslavgerchev

sbom/sbom.go
preslavgerchev: nice catch, fixed

PR #3775, opened 2024-04-17
🐛 fix provider shutdown race condition

imilchev:
We have a race condition where providers can be killed at the moment a scan is starting. Then we see errors like: https://github.com/mondoohq/mondoo-operator/issues/1091 This problem occurs specifically when we can start concurrent scans with a single cnspec instance. Currently, that is only when using `serve-api` command.

The issue is that the providers list in runtime isn't using a lock so we can run into race conditions. I changed the implementation to only kill providers if there are no active runtimes, since at the moment we don't have a thread-safe way to compare the providers for each runtime

### DIFF
--- a/providers/coordinator.go
+++ b/providers/coordinator.go
@@ -203,17 +203,9 @@ func (c *coordinator) RemoveRuntime(runtime *Runtime) {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


arlimus (approved):
This should clear things up, thank you!

PR #3777, opened 2024-04-18
🧹 use new filters

chris-rock:
This updates the pack to use the latest filter structure to avoid warnings on command line:

```
! Found an old use of filters (as a list of strings). This will be removed in the next major version. Please migrate to:
- mql: filter 1
- mql: filter 2
```


arlimus (approved):
Thank you!!

PR #3778, opened 2024-04-18
🐛 return the rpm version including the release when static analysis is used

chris-rock:
For versions like `attr 2.4.48-1.ph4` we only returned `attr 2.4.48` and `-1.ph4` was missing.


czunker (approved):

czunker:
I checked with `cnspec vuln container image photon:4.0-20221224`

This changes the vuln report, but to me it looks like it fixes it. I checked it with `PHSA-2022-4.0-0153`.
Before, we reported this advisory, but now we don't.

PR #3792, opened 2024-04-18
⭐️ support env variables as credential type for inventories

chris-rock:
This is especially useful when used in CI/CD environments. Users have a pre-defined inventory file with one or multiple entries but do not want to store the secrets in there. Instead the CI/CD run defines the secret via an environment variable.

```yaml
spec:
  assets:
  - connections:
    - type: slack
      credentials:
      - type: env
        env: CUSTOM_SLACK_TOKEN
```

You can use the inventory in combination with the env variable then:

```shell
export CUSTOM_SLACK_TOKEN="token"
cnquery scan --inventory-file intentoryyml 
```


arlimus (approved):
Lgtm!

PR #3794, opened 2024-04-19
⭐️ vsphere asset urls

chris-rock:


tas50:
@chris-rock What's left on this one? It would be great to get these out of other before next week.

PR #3801, opened 2024-04-20
🧹 os: re-use conf-type keywords

arlimus:
A simple update that uses pre-defined keywords instead of strings, for those cases where we already extensively have keywords.


chris-rock (approved):
Nice cleanup. Thank you @arlimus

PR #3805, opened 2024-04-22
🧹 default to MQL asset context

arlimus:
We had this feature-flag since v7, aimed to make it the default in v8. Until now, it has not been mainlined.

This is changing today. Compiling with context is generally very helpful and provides nice auto-complete on embedded resources (similar to how you can access embedded fields in golang structs).

There is more work to be done, namely that we want to tie the context of the connection to the MQL execution itself. This will allow us to call fields without constantly refering to the global context in the future. (i.e. instead of `docker.file.sth`, once you connect to a Dockerfile you can just call `sth`). For this reason, we retain the feature-flag and will complete its functionality in this major release.

Note: I deactivated one level of functionality which had been degraded since v9, namely: cross-provider calling of MQL queries. We are restoring this with provider provider initialization (which hasn't been added yet). The release of v9 had removed this because providers were then individually packaged and structured and we hadn't yet had the time to migrate this functionality. Without the change in this PR, these types of calls/usages all resulted in a panic, jfyi

### DIFF
--- a/featureflags.go
+++ b/featureflags.go
@@ -66,7 +66,8 @@ const (
 	// MQLAssetContext feature flag
 	//
 	// start: v7.0
-	// end:   v8.0
+	// updated: v11.0  We default to embed flags and re-use this feature for actual asset context
+	// end:   v12.0
 	MQLAssetContext
 
 	// ErrorsAsFailures feature flag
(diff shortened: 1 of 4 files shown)


chris-rock (approved):
Thank you @arlimus

PR #3806, opened 2024-04-22
⭐ Dockerfile support

arlimus:
Start to add Dockerfile support. You can now:

```
cnquery ACTION docker file DOCKERFILE
```

Example:

```
cnquery shell docker file Dockerfile

cnquery scan docker file Dockerfile.dev
```

This exposes new keyworeds, which are attached to `docker.file`.

## 1. Commands

All instructions are accessible in a structured way. Please note that a Dockerfile may have [multiple stages](https://docs.docker.com/reference/dockerfile/#from).

```coffee
> docker.file.stages{*}
docker.file.stages: [
  0: {
    entrypoint: docker.file.run script="cnquery"
    add: []
    run: [
      0: docker.file.run script="apk update &&    apk add ca-certificates wget tar &&    rm -rf /var/cache/apk/*"
    ]
    from: docker.file.from name="root" image="alpine" tag="3.19"
    env: {}
    copy: [
      0: docker.file.addcopy src=[
        0: "cnquery"
      ] dst="/usr/local/bin"
    ]
    cmd: docker.file.run script="help"
  }
  1: {
    entrypoint: null
    add: []
...
```

## 2. Instructions

An alternative is to access low-level instructions. These are great if you just want to interact with all the instructions without thinking about their context.

```coffee
> docker.file.instructions
docker.file.instructions: [
  0: {
    original: "FROM alpine:3.19 AS root"
  }
  1: {
    original: "RUN apk update &&    apk add ca-certificates wget tar &&    rm -rf /var/cache/apk/*"
  }
...
```

## 3. File access

Finally, a `docker.file` has the regular `file` resource embedded. This means you can use anything that is accessible in file (given that https://github.com/mondoohq/cnquery/pull/3805 merges before this). 

```coffee
> docker.file.path
docker.file.file.path: "Dockerfile"
```


utils/urlx/urlx.go
```diff
@@ -0,0 +1,33 @@
...
+
+package urlx
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ParseGitSshUrl retrieves the
+func ParseGitSshUrl(url string) (string, string, string, error) {
```
chris-rock, line 12: Can we add a few tests so its easier to understand how this parses. eg. with github url, gitlab url and custom git url

chris-rock (approved):
Woot! Nice addtion.

utils/urlx/urlx.go
arlimus: +1, done

PR #3811, opened 2024-04-22
Update titles of gcp.* resources for docs

tas50:
Improve the titles of the docs pages we generate to align with the "Google Cloud" branding


providers/gcp/resources/gcp.lr
```diff
@@ -102,7 +102,7 @@ gcp.project @defaults("name") {
   services() []gcp.service
   // List of recommendations
   recommendations() []gcp.recommendation
-  // GCP GKE resources
+  // Google GKE resources
```
misterpantz: I'm just curious why no "cloud" here.

misterpantz:
```suggestion
  //  Google Kubernetes Engine (GKE)  resources
```

tas50: Let me revert that one. I wasn't tackling the property descriptions

misterpantz (approved):

PR #3825, opened 2024-04-23
🧹 Add documentation for docker resources

tas50:
More docs generation


tas50:
@arlimus I added this descriptions the same way I did with the aws descriptions in #3824, but when running `make lr/docs/markdown` the extra values are not getting added to the docs. Any idea what I'm doing wrong here?

tas50:
This was not being honored due to an incorrect path in the cnquery makefile. That was fixed in #3843

chris-rock (approved):

PR #3828, opened 2024-04-24
✨ dockerfile.expose

arlimus:
1. Adds support for the `EXPOSE` keyword in dockerfiles. Adds them as a list of entries and defaults the protocol if not set
2. Fixes a bug where IDs were not set for a range of docker resources
3. Removes the warnings for unsupported keywords in Dockerfiles. We already know what can be supported by looking at the fields we expose.

Looking deeper into the library too, we have a bit more work to properly support some of the contextual keywords...

### DIFF
--- a/providers/os/resources/dockerfile.go
+++ b/providers/os/resources/dockerfile.go
@@ -170,6 +170,7 @@ func (p *mqlDockerFile) stage2resource(stage instructions.Stage) (*mqlDockerFile
 	}
 
 	env := map[string]any{}
+	var expose []any
 	var runs []any
 	var copy []any
 	var add []any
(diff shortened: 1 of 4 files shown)


providers/os/resources/dockerfile.go
```diff
@@ -220,6 +222,7 @@ func (p *mqlDockerFile) stage2resource(stage instructions.Stage) (*mqlDockerFile
 				src[i] = v.SourcesAndDest.SourcePaths[i]
 			}
 			resource, err := CreateResource(p.MqlRuntime, "docker.file.add", map[string]*llx.RawData{
+				"__id":  llx.StringData(p.locationID(v.Location())),
```
imilchev, line 225: Why is this __id needed? Have we used such a thing before?

arlimus: This is only done because we don't have an `id()` method and we don't have line context on the child resources yet, but definitely will add it. I'd love to pull the ID-generation in once we have line context back. I took a bit of a shortcut here :)

providers/os/resources/os.lr.manifest.yaml
```diff
@@ -143,6 +143,11 @@ resources:
       dst: {}
       src: {}
     min_mondoo_version: latest
+  docker.file.expose:
+    fields:
+      port: {}
+      protocol: {}
+    min_mondoo_version: latest
```
tas50:
```suggestion
    min_mondoo_version: 9.0.0
```

providers/os/resources/os.lr.manifest.yaml
```diff
@@ -143,6 +143,11 @@ resources:
       dst: {}
       src: {}
     min_mondoo_version: latest
```
tas50, line 145:
```suggestion
    min_mondoo_version: 9.0.0
```

chris-rock (approved):
Thank you @arlimus

PR #3844, opened 2024-04-24
Expand docs LR files

tas50:
- Add minimum mondoo release for atlassian to be 9.0.0
- Change all 9.0.3 min versions to be just 9.0.0
- Add more Azure links
- Fix incorrect Azure link titles
- Fix some bad formatting in Atlassian


misterpantz (approved):
Good stuff!!

PR #3845, opened 2024-04-24
🧹 make lr docs yaml file loading for markdown generation more robust

tas50:
The way the filestat was being checked if a user gave an incorrect path we'd love that the file was not provided and move on. These should always be provided so let's just not include this check entirely.

I also improved a bit of text while I was in here and fixed incorrect help

### DIFF
--- a/providers-sdk/v1/lr/cli/cmd/markdown.go
+++ b/providers-sdk/v1/lr/cli/cmd/markdown.go
@@ -69,18 +69,19 @@ var markdownCmd = &cobra.Command{
 		}
 
 		var lrDocsData docs.LrDocs
+
 		docsFilepath, _ := cmd.Flags().GetString("docs-file")
-
-		content, err := os.ReadFile(docsFilepath)
-		if err != nil {
-			log.Fatal().Err(err).Msg("could not read file " + docsFilepath)
+		if docsFilepath != "" { // as soon as a path has been provided, we try to load the file
+			content, err := os.ReadFile(docsFilepath)
+			if err != nil {
+				log.Fatal().Err(err).Msg("could not read file " + docsFilepath)
+			}
+			err = yaml.Unmarshal(content, &lrDocsData)
+			if err != nil {
+				log.Fatal().Err(err).Msg("could not load yaml data")
+			}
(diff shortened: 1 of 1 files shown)


providers-sdk/v1/lr/cli/cmd/markdown.go
```diff
@@ -175,7 +171,7 @@ func (l *lrSchemaRenderer) renderToc(packName string, description string, resour
 
 	// render content
 	builder.WriteString("# Mondoo " + packName + " Resource Pack Reference\n\n")
-	builder.WriteString("In this pack:\n\n")
+	builder.WriteString("Resources included in this pack:\n\n")
```
tas50, line 176: @misterpantz can you let me know if this works for you. This goes on the main resource page pages like https://mondoo.com/docs/mql/resources/gcp-pack/

misterpantz: Yep! That works just great, thank you.

chris-rock:
Can you help me understand why this is required?

providers-sdk/v1/lr/cli/cmd/markdown.go
```diff
@@ -70,22 +70,18 @@ var markdownCmd = &cobra.Command{
...
-			}
-			err = yaml.Unmarshal(content, &lrDocsData)
-			if err != nil {
-				log.Fatal().Err(err).Msg("could not load yaml data")
-			}
 
-			log.Info().Int("resources", len(lrDocsData.Resources)).Msg("loaded docs from " + docsFilepath)
-		} else {
-			log.Info().Msg("no docs file provided")
+		content, err := os.ReadFile(docsFilepath)
```
chris-rock: Instead of making this required, we should change the check logic. If a path is provided, we check that it exists. So we only load the yaml when the path is != ""

chris-rock: I updated the code to ensure we do not silently ignore a file that was provided with a wrong path. The current logic still treats the yaml as optional since it is not required to generate markdown.

chris-rock (approved):
Thank you @tas50

PR #3859, opened 2024-04-25
✨ add creation date to ec2 images

vjeffrey:
```
cnquery> aws.ec2.instances.where(image.createdAt < time.now + time.day*-365) { instanceId region image { id createdAt}}
```

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -2568,6 +2568,8 @@ private aws.ec2.image @defaults("arn") {
   ownerId string
   // Alias for the image owner
   ownerAlias string
+  // Date the image was created
+  createdAt time
 }
 
 // Amazon EC2 instance device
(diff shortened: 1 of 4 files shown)


providers/aws/resources/aws.lr
```diff
@@ -2568,6 +2568,8 @@ private aws.ec2.image @defaults("arn") {
   ownerId string
   // Alias for the image owner
   ownerAlias string
+  // Date the image was created
+  creationDate time
```
tas50: We don't have this totally standardized, but most of these are using the name `createdAt`. It would be good to make this that same name so we don't have more to deprecated in the future.

providers/aws/resources/aws.lr.manifest.yaml
```diff
@@ -947,6 +947,8 @@ resources:
     fields:
       architecture: {}
       arn: {}
+      creationDate:
+        min_mondoo_version: latest
```
```suggestion
        min_mondoo_version: 9.0.0
```

providers/aws/resources/aws.lr
vjeffrey: updated!

tas50 (approved):

PR #3864, opened 2024-04-26
🐛⭐ Overhaul SSHd config parsing

arlimus:
This is a massive overhaul of the parser to address a series of issues we have identified.

1. Match groups are now properly identified **across** include paths. We have extensively tested how SSHd handles the various edge-cases and have adjusted our parser accordingly. It now properly parses the different scenarios of match groups with or without include statements and adds them to the affected subgroups.

2. The content field is now deprecated. This is an old remnant and at this point it is no longer providing the best version of the raw SSHd config anymore. The reason are the changes listed above where include and match statements actually behave differently based on their context and cannot just be aggregated into a single content file. Instead please use the already provided `file` and `files` fields which both have `content` as their subfields.

3. Multiple statements are now correctly treated in params. In the case of SSHd, the first statement usually wins (with a couple of edge-cases which are still aggregated - and those have been added here as well).

4. Include statements now work with relative and absolute paths. We previously only supported relative paths, i.e. files inside of `/etc/ssh`. This limitation is no longer in place.

Fixes https://github.com/mondoohq/cnquery/issues/3316


chris-rock (approved):
Thank you @arlimus

PR #3874, opened 2024-04-29
🐛 fix coordinator race for shutting down providers

imilchev:
In #3775 a race condition in providers shutdown logic has been addressed. The issue described there almost never happens now but there are still some cases where we see the error: `ERR error encountered while scanning stdout error="read |0: file already closed"`

I went through the code and the only other reason where things may go wrong is if we have unprocessed runtimes at the moment we check for provider cleanup. I think to fix this case we need to make sure that there are no runtimes and no unprocessed runtimes before killing any providers.


czunker (approved):
Thanks @imilchev

PR #3887, opened 2024-04-30
🐛 Ensure we use the proper runtime for shell connections, do not store assets with no platform

preslavgerchev:
Ensure we use the right runtime to start the shell. Also ensure we cannot get panics in the recording by ensuring we do not store assets with no asset platform.


apps/cnquery/cmd/shell.go
```diff
@@ -155,7 +154,6 @@ func StartShell(runtime *providers.Runtime, conf *ShellConfig) error {
 
 	// when we close the shell, we need to close the backend and store the recording
 	onCloseHandler := func() {
-		runtime.Close()
```
preslavgerchev, line 158: this is already being called in the shell Close function:
https://github.com/mondoohq/cnquery/blob/main/cli/shell/shell.go#L312

providers-sdk/v1/recording/recording.go
```diff
@@ -355,6 +357,10 @@ func (r *recording) findAssetConnID(asset *inventory.Asset) (int, string) {
 func (r *recording) EnsureAsset(asset *inventory.Asset, providerID string, connectionID uint32, conf *inventory.Config) {
 	found, _ := r.findAssetConnID(asset)
 
+	if asset.Platform == nil {
```
preslavgerchev, line 359: one example is connecting to a cloud provider: we return an 'empty' asset initially from discovery which has nothing set. this ensures we do not break the scan/shell

a better fix here would be to ensure we return only the right assets from discovery but this is still a nice safeguard

providers-sdk/v1/recording/recording.go
```diff
@@ -135,6 +135,8 @@ func NewWithFile(path string, opts RecordingOptions) (llx.Recording, error) {
 				Path:            path,
 				prettyPrintJSON: opts.PrettyPrintJSON,
 				doNotSave:       opts.DoNotSave,
+				assets:          map[uint32]*Asset{},
+				Assets:          []*Asset{},
```
imilchev: nit: you don't really need to initialize the slice here, `nil` would also do the job

imilchev (approved):

PR #3888, opened 2024-04-30
✨ Add support for Dockerfile LABEL

### LINKED ISSUE #3814: Expose `label` instruction in `docker.file`
**Describe the bug**
The `label` instruction is important for writing best practices policies. For example our best practices policy should check that no label starts with `org.label-schema` as these have been replaced with `org.opencontainers.image` labels.

Exposing each k/v pair would allow us to write checks against the content of labels like this.

preslavgerchev:
Fixes #3814 
 * Add `docker.file.stage.labels`
 * Rename `dockerfile.go` to `docker_file.go`, vscode was thinking that this is a pure dockerfile so highlighting was off
 * Add test for parsing a dockerfile

### DIFF
--- a/providers/os/resources/os.lr
+++ b/providers/os/resources/os.lr
@@ -776,7 +776,7 @@ docker.file.stage @defaults("from.name") {
   file docker.file
   // Environment variables configured in the Dockerfile
   env map[string]string
-  // Labels configured in the Dockerfile
+  // LABEL instructions in the Dockerfile
   labels map[string]string
   // RUN instructions in this Dockerfile
   run []docker.file.run


vjeffrey (approved):

providers/os/resources/os.lr
```diff
@@ -776,6 +776,8 @@ docker.file.stage @defaults("from.name") {
   file docker.file
   // Environment variables configured in the Dockerfile
   env map[string]string
+  // Labels configured in the Dockerfile
```
tas50:
```suggestion
  // LABEL instructions in the Dockerfile
```

tas50 (approved):

PR #3889, opened 2024-04-30
🐛 Replace fsconn with local connection for docker files, allow running commands

### LINKED ISSUE #3813: Error in `docker.file.file`
**Describe the bug**
When running `docker.file.file` I get an error and permissions are not shown

```
cnquery shell docker file dokken-images/almalinux-8/Dockerfile
→ connected to Dockerfile

cnquery> docker.file.path
docker.file.file.path: "Dockerfile"
cnquery> docker.file.file
1 error occurred:
	* cannot find users manager
docker.file.file: file path="Dockerfile" size=1564 permissions.string=no data available
```

preslavgerchev:
Fixes #3813
 * Replace the FS conn with a local connection as we need this to able to fully resolve the docker file's properties (such as the owner).
 * Inject the local OS's family into the dockerfile asset to help the local connection run the right commands to be able to fetch users
 * Improve error messages that come from the users manager

### DIFF
--- a/providers/os/connection/docker/dockerfile_connection.go
+++ b/providers/os/connection/docker/dockerfile_connection.go
@@ -11,7 +11,7 @@ import (
 	"path/filepath"
 
 	"go.mondoo.com/cnquery/v11/providers-sdk/v1/inventory"
-	"go.mondoo.com/cnquery/v11/providers/os/connection/fs"
+	"go.mondoo.com/cnquery/v11/providers/os/connection/local"
 	"go.mondoo.com/cnquery/v11/providers/os/connection/shared"
 	"go.mondoo.com/cnquery/v11/utils/multierr"
 	"go.mondoo.com/cnquery/v11/utils/urlx"
@@ -20,11 +20,11 @@ import (
 var _ shared.Connection = &DockerfileConnection{}
 
 type DockerfileConnection struct {
-	*fs.FileSystemConnection
+	*local.LocalConnection
 	Filename string
 }
 
-func NewDockerfile(id uint32, conf *inventory.Config, asset *inventory.Asset) (*DockerfileConnection, error) {
+func NewDockerfileConnection(id uint32, conf *inventory.Config, asset *inventory.Asset, localConn *local.LocalConnection, localFamily []string) (*DockerfileConnection, error) {
 	if conf == nil {
 		return nil, errors.New("missing configuration to create dockerfile connection")
 	}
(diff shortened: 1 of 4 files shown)


providers/os/connection/docker/dockerfile_connection.go
```diff
@@ -53,21 +53,17 @@ func NewDockerfile(id uint32, conf *inventory.Config, asset *inventory.Asset) (*
...
-
 	asset.Platform = &inventory.Platform{
 		Name:    "dockerfile",
 		Title:   "Dockerfile",
 		Family:  []string{"docker"},
 		Kind:    "code",
 		Runtime: "docker",
 	}
+	// this helps with running commands against the local connection
+	asset.Platform.Family = append(asset.Platform.Family, localFamily...)
```
preslavgerchev, line 64: i am not fully sure how good of a fix that is.

We need to run local cmds to be able to evaluate `file.user` and `file.group`. To be able to do so we need to know what OS we're dealing with.
 This will not work if we start supporting scanning dockerfiles by pulling the contents down from somewhere.
 
 To be clear, we only need this to support evaluating the user, group and permissions of the file. The rest does not require the local connection

providers/os/connection/docker/dockerfile_connection.go
```diff
@@ -20,11 +20,11 @@ import (
 var _ shared.Connection = &DockerfileConnection{}
 
 type DockerfileConnection struct {
-	*fs.FileSystemConnection
+	*local.LocalConnection
 	Filename string
 }
 
-func NewDockerfile(id uint32, conf *inventory.Config, asset *inventory.Asset) (*DockerfileConnection, error) {
+func NewDockerfileConnection(id uint32, conf *inventory.Config, asset *inventory.Asset, localConn *local.LocalConnection, localFamily []string) (*DockerfileConnection, error) {
```
imilchev, line 27: is it possible to discover dockerfiles in ssh, tar or container connections? if yes, then we cannot pass a local connection here

preslavgerchev: this connection is only initiated if we shell into a dockerfile, i.e. if we run `cnquery shell docker file Dockerfile`. I think at this point this code can only work with docker file that are present locally.

the docker file MQL resource itself can still work fine in an SSH connection or any provider that supports running commands, those providers are not even connected to that PR or the `DockerfileConnection`:
```
➜ cnquery shell ssh azureuser@<ip> --identity-file <file>
→ connected to Ubuntu 22.04.4 LTS

cnquery> docker.file('Dockerfile')
docker.file: docker.file file.path="Dockerfile" instructions.length=8 stages.length=2
cnquery> docker.file('Dockerfile').file
docker.file.file: file path="Dockerfile" size=370 permissions.string="-rw-rw-r--"
cnquery> docker.file('Dockerfile').file.user
docker.file.file.user: user name="azureuser" uid=1000 gid=1000
```

chris-rock: We also want to scan docker files from other locations going forward, eg. stored in git. Therefore local is not a preferred solution.

chris-rock: Discussed with @arlimus: we want to set file if it is a local connection, otherwise we leave `file` nil

PR #3891, opened 2024-04-30
🐛 do not reset named asset to empty name

vjeffrey:


providers/os/provider/detector.go
```diff
@@ -77,7 +77,10 @@ func (s *Service) detect(asset *inventory.Asset, conn shared.Connection) error {
 		log.Debug().Msg("run cloud platform detector")
 		if id, name, related := aws.Detect(conn, asset.Platform); id != "" {
 			asset.PlatformIds = append(asset.PlatformIds, id)
-			asset.Name = name
+			if name != "" {
```
preslavgerchev, line 80: do we know why the return value here is empty? it might be worth looking into why `Detect` returns no meaningful value here

vjeffrey: ill see if i can figure that out..the weird is that when i was testing this, the asset.name we had when we got to this part of the code was already the instance id...which u would think is what we're trying to fetch here :/

vjeffrey: the name can only be fetched from that path if the name instance metadata option is on. we're still getting the id here, which is what we already have, but it is expected that name is empty

preslavgerchev (approved):

PR #3893, opened 2024-04-30
🧹 add static credentials option to aws connection

vjeffrey:


providers/aws/connection/connection.go
```diff
@@ -172,6 +179,11 @@ func parseFlagsForConnectionOptions(m map[string]string) []ConnectionOption {
 		externalId := m["external-id"]
 		o = append(o, WithAssumeRole(cfg, role, externalId))
 	}
+
+	if len(creds) > 0 {
```
preslavgerchev, line 183: is there a use case where there might be more creds here and we'd need to iterate over them instead of taking the first one?

vjeffrey: i can't think of one, since one would override the other

preslavgerchev (approved):

PR #3904, opened 2024-05-02
Update github provider to the latest sdk

tas50:
59 -> 61. The way they version this doesn't get pulled in via
   dependabot.


chris-rock (approved):
Thank you @tas50

PR #3906, opened 2024-05-02
⭐️ support filtering discovered github repos

imilchev:
We can now filter out github repos using globs:
```
cnquery scan github org podkrepi-bg --repos-exclude api,frontend
```

```
cnquery scan github org podkrepi-bg --repos api,frontend
```

### DIFF
--- a/providers/github/resources/discovery.go
+++ b/providers/github/resources/discovery.go
@@ -4,6 +4,10 @@
 package resources
 
 import (
+	"strings"
+
+	"github.com/gobwas/glob"
+	"github.com/rs/zerolog/log"
 	"go.mondoo.com/cnquery/v11/llx"
 	"go.mondoo.com/cnquery/v11/providers-sdk/v1/inventory"
 	"go.mondoo.com/cnquery/v11/providers-sdk/v1/plugin"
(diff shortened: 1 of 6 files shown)


providers/github/resources/discovery.go
```diff
@@ -200,3 +206,55 @@ func getMqlGithubUser(runtime *plugin.Runtime, userName string) (*mqlGithubUser,
...
+	return nsFilter
+}
+
+func (f *ReposFilter) skipRepo(namespace string) bool {
+	// anything explicitly specified in the list of includes means accept only from that list
+	if len(f.include) > 0 {
+		for _, ns := range f.include {
+			g, err := glob.Compile(ns)
+			if err != nil {
+				log.Error().Err(err).Msg("failed to compile glob")
```
czunker: nit: I think we should mention that a glob for a repo name failed.

providers/github/resources/discovery_test.go
czunker, line 1: Could you please also add a test where both options are specified?

imilchev: added a test

czunker (approved):
Thanks @imilchev

PR #3911, opened 2024-05-02
🧹 github provider improvements

imilchev:
- fix memoizer for users
- add memoizer for org
- fix panics for committer and author being empty

### DIFF
--- a/providers/github/resources/github.go
+++ b/providers/github/resources/github.go
@@ -31,7 +31,7 @@ func getUser(ctx context.Context, runtime *plugin.Runtime, conn *connection.Gith
 	if g.memoize == nil {
 		g.memoize = memoize.NewMemoizer(30*time.Minute, 1*time.Hour)
 	}
-	res, err, _ := g.memoize.Memoize("user", func() (interface{}, error) {
+	res, err, _ := g.memoize.Memoize("user-"+user, func() (interface{}, error) {
 		log.Debug().Msgf("fetching user %s", user)
 		user, _, err := conn.Client().Users.Get(ctx, user)
 		return user, err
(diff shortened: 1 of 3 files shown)


providers/github/resources/github.go
```diff
@@ -31,7 +31,7 @@ func getUser(ctx context.Context, runtime *plugin.Runtime, conn *connection.Gith
 	if g.memoize == nil {
 		g.memoize = memoize.NewMemoizer(30*time.Minute, 1*time.Hour)
 	}
-	res, err, _ := g.memoize.Memoize("user", func() (interface{}, error) {
+	res, err, _ := g.memoize.Memoize("user"+user, func() (interface{}, error) {
```
preslavgerchev: nit: 
```suggestion
	res, err, _ := g.memoize.Memoize("user-"+user, func() (interface{}, error) {
```
to help with readability (if we need to debug, etc)

providers/github/resources/github_repo.go
```diff
@@ -584,7 +584,7 @@ func newMqlGithubCommit(runtime *plugin.Runtime, rc *github.RepositoryCommit, ow
 		}
 	}
 
-	if rc.Author != nil {
+	if rc.Author != nil && rc.Author.ID != nil && rc.Author.Login != nil {
```
preslavgerchev, line 587: what are the panics that require those changes?
`llx.IntDataPtr` should have a nil check and fall back to returning the MQL equivalent of nil

imilchev: these things get passed to the init function for user and that function then crashes, we shouldn't be trying to create a user if such didn't exist for that commit... this is just github's weird way of saying there's no author itseems... the author field is set but all props are `nil`

preslavgerchev (approved):

PR #3924, opened 2024-05-05
Update away from deprecated GitLab field

tas50:
Keep our field the same, but avoid this deprecated call

### DIFF
--- a/providers/gitlab/resources/gitlab.go
+++ b/providers/gitlab/resources/gitlab.go
@@ -41,7 +41,7 @@ func initGitlabGroup(runtime *plugin.Runtime, args map[string]*llx.RawData) (map
 	args["requireTwoFactorAuthentication"] = llx.BoolData(grp.RequireTwoFactorAuth)
 	args["preventForkingOutsideGroup"] = llx.BoolData(grp.PreventForkingOutsideGroup)
 	args["mentionsDisabled"] = llx.BoolData(grp.MentionsDisabled)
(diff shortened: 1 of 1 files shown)


providers/gitlab/resources/gitlab.go
```diff
@@ -39,7 +39,7 @@ func initGitlabGroup(runtime *plugin.Runtime, args map[string]*llx.RawData) (map
 	args["requireTwoFactorAuthentication"] = llx.BoolData(grp.RequireTwoFactorAuth)
 	args["preventForkingOutsideGroup"] = llx.BoolData(grp.PreventForkingOutsideGroup)
 	args["mentionsDisabled"] = llx.BoolData(grp.MentionsDisabled)
-	args["emailsDisabled"] = llx.BoolData(grp.EmailsDisabled)
+	args["emailsDisabled"] = llx.BoolData(!grp.EmailsEnabled)
```
chris-rock, line 44: we should also add the new field and mark the old one as deprecated

chris-rock (dismissed):

providers/gitlab/resources/gitlab.go
tas50: I'd prefer to just update the call under the hood and keep our fields the same. I don't think we need to match our fields to their underlying API 1:1 in every case. The fields are named pretty consistently which is nice and it wouldn't benefit the user to switch them unless they remove all the "disabled" calls in the future.

PR #3926, opened 2024-05-05
WIP: Add new fields to GH repos + handle null bools

tas50:
Add new fields + handle a few merge options that are bools with null values if not set

WIP as I'm digging into some odd behavior with some existing bools


imilchev:
please undraft this once you think it is ready

tas50:
If someone in engineering would like to pick this up feel free. Closing for now due to time constraints.

PR #3940, opened 2024-05-08
⭐️ inventory template support

chris-rock:
When cnquery is used in CI/CD pipelines, users want to leverage the inventory file which allows a lot of pre-configuration. If you want to scan different assets for different environments you also want to re-use the same template but allow some custom variables so that usage in CI/CD is easier. This PR introduces inventory templates. Just define a template and use the `getenv` function to access environment variables:

```yaml
spec:
  assets:
    - name: {{ getenv "ASSET_NAME" }}
      connections:
        - type: local
          discover:
            targets:
              - auto
      annotations:
        application:  {{ getenv "APPLICATION" }}
```

This inventory can now be used as following:

```shell
ASSET_NAME="my_asset" APPLICATION="my-super-app" cnquery scan --inventory-template template.yaml
```

### DIFF
--- a/apps/cnquery/cmd/root.go
+++ b/apps/cnquery/cmd/root.go
@@ -214,15 +214,11 @@ func getCobraScanConfig(cmd *cobra.Command, runtime *providers.Runtime, cliRes *
 	config.DisplayUsedConfig()
 
 	props := viper.GetStringMapString("props")
-	annotations := viper.GetStringMapString("annotation")
 
 	// merge the config and the user-provided annotations with the latter having precedence
 	optAnnotations := opts.Annotations
 	if optAnnotations == nil {
 		optAnnotations = map[string]string{}
-	}
-	for k, v := range annotations {
-		optAnnotations[k] = v
 	}
 
 	assetName := viper.GetString("asset-name")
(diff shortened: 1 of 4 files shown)


apps/cnquery/cmd/root.go
```diff
@@ -214,16 +214,12 @@ func getCobraScanConfig(cmd *cobra.Command, runtime *providers.Runtime, cliRes *
 	config.DisplayUsedConfig()
 
 	props := viper.GetStringMapString("props")
-	annotations := viper.GetStringMapString("annotation")
```
chris-rock, line 217: the binding was wrong, so this never worked before

apps/cnquery/cmd/scan.go
```diff
@@ -97,15 +100,15 @@ To manually configure a query pack, use this:
...
-		_ = viper.BindPFlag("props", cmd.Flags().Lookup("props"))
-
 		// for all assets
 		_ = viper.BindPFlag("incognito", cmd.Flags().Lookup("incognito"))
 		_ = viper.BindPFlag("insecure", cmd.Flags().Lookup("insecure"))
 		_ = viper.BindPFlag("querypacks", cmd.Flags().Lookup("querypack"))
 		_ = viper.BindPFlag("sudo.active", cmd.Flags().Lookup("sudo"))
 		_ = viper.BindPFlag("record", cmd.Flags().Lookup("record"))
+		// NOTE: we may "annotation" to "annotations" to align it with the internal config struct
+		_ = viper.BindPFlag("annotations", cmd.Flags().Lookup("annotation"))
```
chris-rock, line 110: required since the options use "annotations"

imilchev (approved):
LGTM!

Issue #2883, opened 2023-12-19
Bump k8s to v0.29.0 and remove `k8s.podSecurityPolicies`


czunker:
**Describe the bug**
```
{"level":"info","file":"providers/k8s/resources/k8s.lr.manifest.yaml","time":"2023-12-19T15:56:03Z","message":"write file"}
--> [k8s] generate CLI json
go: downloading k8s.io/apiextensions-apiserver v0.29.0
go: downloading k8s.io/kube-openapi v0.0.0-20231214164306-ab13479f8bf8
go: downloading github.com/go-openapi/swag v0.22.5
go: downloading github.com/go-openapi/jsonreference v0.20.3
go: downloading github.com/go-openapi/jsonpointer v0.20.1
# go.mondoo.com/cnquery/v9/providers/k8s/resources
Error: resources/podsecuritypolicy.go:19:22: undefined: policyv1beta1.PodSecurityPolicy
Error: resources/podsecuritypolicy.go:31:39: undefined: policyv1beta1.PodSecurityPolicy
Error: resources/podsecuritypolicy.go:64:34: cannot infer T (/home/runner/_work/cnquery/cnquery/providers-sdk/v1/util/convert/map.go:7:24)
Error: resources/podsecuritypolicy.go:68:34: cannot infer T (/home/runner/_work/cnquery/cnquery/providers-sdk/v1/util/convert/map.go:7:24)
make: *** [Makefile:310: providers/build/k8s] Error 1
Error: Process completed with exit code 2.
```

**To Reproduce**

1. Bump k8s to v0.29.0
2. Run pipeline
3. Note the error

**Expected behavior**
Should work and pass pipeline.

czunker:
Hold for now. But at some point we have to upgrade.

czunker:
PSPs were removed in v1.25:
https://kubernetes.io/docs/concepts/security/pod-security-policy/

tas50:
Closing: k8s API upgraded to v0.35.1 and podSecurityPolicies fully removed from k8s provider.

PR #3954, opened 2024-05-10
🧹 Remove k8s PSPs and bump deps to v0.30

czunker:
Fixed: #2883


tas50:
@czunker I did the PSP side of this in the v12 branch, but we need to remove the version pin and make sure everything works on the latest library once v12 merges.

PR #3959, opened 2024-05-10
🧹 Deprecate k8s PSPs and bump deps to latest v0.28 patch version

czunker:
Preparation for #2883


czunker:
Deprecate before we can remove it in v12 with https://github.com/mondoohq/cnquery/pull/3954

chris-rock (approved):
Thank you @czunker

PR #3964, opened 2024-05-10
:bug: docker file `Dockerfile` not found

afiune:
Fixes https://github.com/mondoohq/cnquery/issues/3962

* Renamed the file because my editor was complaining about the name being `dockerfile*`
* Returned the full absolute path instead of the relative path
* Added tests that cover this change

### DIFF
--- a/providers/os/connection/docker/docker_file_connection.go
+++ b/providers/os/connection/docker/docker_file_connection.go
@@ -66,7 +66,7 @@ func NewDockerfileConnection(_ uint32,
 		return nil, errors.New("no inventory connections")
 	}
 
-	if url, ok := asset.Connections[0].Options["ssh-url"]; ok {
+	if url, ok := conf.Options["ssh-url"]; ok {
 		domain, org, repo, err := urlx.ParseGitSshUrl(url)
 		if err != nil {
 			return nil, err
--- a/providers/os/connection/docker/docker_file_connection.go
+++ b/providers/os/connection/docker/docker_file_connection.go
@@ -24,7 +24,9 @@ type DockerfileConnection struct {
 	Filename string
 }
 
-func NewDockerfileConnection(id uint32, conf *inventory.Config, asset *inventory.Asset, localConn *local.LocalConnection, localFamily []string) (*DockerfileConnection, error) {
+func NewDockerfileConnection(_ uint32,
+	conf *inventory.Config, asset *inventory.Asset,
+	localConn *local.LocalConnection, localFamily []string) (*DockerfileConnection, error) {
 	if conf == nil {
 		return nil, errors.New("missing configuration to create dockerfile connection")
 	}
@@ -44,8 +46,6 @@ func NewDockerfileConnection(id uint32, conf *inventory.Config, asset *inventory
 		return nil, err
 	}
 
-	// if we have a regular file, we need to point the fs.Connection to
-	// look at the folder instead and store the filename separately
 	var filename string
 	if !stat.IsDir() {
 		filename = filepath.Base(absSrc)
@@ -62,6 +62,10 @@ func NewDockerfileConnection(id uint32, conf *inventory.Config, asset *inventory
 	}
 	// this helps with running commands against the local connection
 	asset.Platform.Family = append(asset.Platform.Family, localFamily...)
+
+	if len(asset.GetConnections()) == 0 { // prevents panics by accesssing `asset.Connections[0]`
+		return nil, errors.New("no inventory connections")
+	}
 
 	if url, ok := asset.Connections[0].Options["ssh-url"]; ok {
 		domain, org, repo, err := urlx.ParseGitSshUrl(url)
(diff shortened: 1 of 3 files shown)


providers/os/connection/docker/docker_file_connection.go
```diff
@@ -63,6 +63,10 @@ func NewDockerfileConnection(id uint32, conf *inventory.Config, asset *inventory
 	// this helps with running commands against the local connection
 	asset.Platform.Family = append(asset.Platform.Family, localFamily...)
 
+	if len(asset.GetConnections()) == 0 { // prevents panics by accesssing `asset.Connections[0]`
```
chris-rock: asset.Connections should not be used here, use `conf` directly since it is passed in

providers/os/connection/docker/docker_file_connection.go
```diff
@@ -63,6 +63,10 @@ func NewDockerfileConnection(id uint32, conf *inventory.Config, asset *inventory
...
+		return nil, errors.New("no inventory connections")
+	}
+
 	if url, ok := asset.Connections[0].Options["ssh-url"]; ok {
```
chris-rock: should use `conf` instead of `asset.Connections[0]`, please check if there are more places where the `asset.Connections[0]` is used

providers/os/connection/docker/docker_file_connection.go
```diff
@@ -86,7 +90,7 @@ func NewDockerfileConnection(id uint32, conf *inventory.Config, asset *inventory
 
 	conn := &DockerfileConnection{
 		LocalConnection: localConn,
-		Filename:        filename,
+		Filename:        filepath.Join(conf.Path, filename),
```
chris-rock: Why can't we use `absSrc` directly?

providers/os/connection/docker/docker_file_connection.go
```diff
@@ -24,7 +24,9 @@ type DockerfileConnection struct {
 	Filename string
```
chris-rock: We should add a comment that this has the absolute path to the dockerfile

providers/os/connection/docker/docker_file_connection.go
afiune: @chris-rock I can change that, though there are some places where we populate data into `asset.Connections`, should I remove that? 🤔

providers/os/connection/docker/docker_file_connection.go
afiune: Should we be modifying `conf` instead of `asset` here? For connections, platformId, etc.?

providers/os/connection/docker/docker_file_connection.go
afiune: because `absSrc` gets overwritten to only the directory, not the full path.

afiune:
@chris-rock I did another pass, but I think you need to check if we need to avoid setting `asset.Connections[0].PlatformId` or leave it.

providers/os/connection/docker/docker_file_connection.go
chris-rock: Yes, we should do:

```
conf.PlatformId = platformID
asset.PlatformIds = []string{platformID}
```

@imilchev Can you confirm that?

chris-rock: I also think we should not do `if len(asset.GetConnections()) == 0 { // prevents panics by accesssing `asset.Connections[0]`. This case should never happen. If it happens, we need to fix the root cause instead of the symptom.

@imilchev please confirm

imilchev: connections can never be 0, this is confirmed. That's because the connection determines which provider to be called. If we ended up in a provider, we had a connection set

imilchev: for the question about

```go
conf.PlatformId = platformID
asset.PlatformIds = []string{platformID}
```

`conf` is actually equal (and also the same pointer as `asset.Connections[0]`. We can see that here https://github.com/mondoohq/cnquery/blob/main/providers/os/provider/provider.go#L305-L306

providers/os/connection/docker/docker_file_connection.go
```diff
@@ -86,7 +91,9 @@ func NewDockerfileConnection(id uint32, conf *inventory.Config, asset *inventory
 
 	conn := &DockerfileConnection{
 		LocalConnection: localConn,
-		Filename:        filename,
+		// here we must use the absolute path of the Dockerfile so
+		// that we find the file downstream
+		Filename: absSrc,
```
imilchev: I know @preslavgerchev spent some time looking into the file paths for dockerfiles. He may be a better option to evaluate this PR

providers/os/connection/docker/docker_file_connection.go
afiune: Thank you for that context, I'll remove this if statement.

providers/os/connection/docker/docker_file_connection.go
```diff
@@ -21,10 +21,14 @@ var _ shared.Connection = &DockerfileConnection{}
 
 type DockerfileConnection struct {
 	*local.LocalConnection
+	// Filename must be the absolute path of the Dockerfile so
+	// that we find the file downstream
 	Filename string
```
preslavgerchev:
```suggestion
	FileAbsSrc string
```
i think that makes more sense given the comment above it too

preslavgerchev (approved):
lgtm, one small nit

PR #3974, opened 2024-05-13
🐛 Ensure asset platform is not nil in sysinfo.

preslavgerchev:


providers-sdk/v1/sysinfo/sysinfo.go
```diff
@@ -51,8 +54,17 @@ func Get() (*SystemInfo, error) {
 		}
 	}
 
-	sysInfo.Platform = platform
+	// FIXME: if IdentifyPlatform returns an err, we get a nil platform this will cause a nil pointer dereference in the next step (hostname)
+	// FIXME: we should ensure IdentifyPlatform's error is handled properly
+	if platform == nil {
```
chris-rock, line 59: I think we need to gracefully handle the case that we cannot retrieve the sys info. While not optimal, we should not stop execution when we cannot determine the host platform.

preslavgerchev: I am closing this PR as it's not needed anymore for the use case it was trying to fix

PR #3976, opened 2024-05-13
🧹 Drop deprecated v8 kind.

preslavgerchev:
We should ensure this is fully migrated out everywhere else before merging.


tas50:
@preslavgerchev Is this something we'd want to ship in v12?

PR #3981, opened 2024-05-14
⭐️ add container-proxy flag for scanning containers

imilchev:
we can now specify a container proxy for scanning containers:
```
cnquery scan docker ubuntu:latest --container-proxy http://localhost:3128
```

this is different from the `api-proxy` setting, which is used only to talk to the Mondoo platform. The `container-proxy` flag is used only for resolving and pulling containers


chris-rock (approved):
Thank you @imilchev

PR #3984, opened 2024-05-14
✨ set container-proxy flag via k8s provider for container discovery

imilchev:
we can do:
```
cnspec scan k8s --discover container-images --container-proxy http://localhost:3128
```


czunker (approved):
LGTM.
Thanks @imilchev

PR #3992, opened 2024-05-14
🐛 Return error if no platform information

jaym:
`SystemInfo.Get` used to return an error if it couldn't get the platform information.
This regressed in https://github.com/mondoohq/cnquery/pull/3660 where nil platform info stopped returning an error


vjeffrey (approved):
thank you! as mentioned in https://github.com/mondoohq/cnquery/pull/3974/files , we may need to revisit this to not error out when platform is nil, but this fixes the panic we're seeing, so let's get this in

PR #4006, opened 2024-05-16
🧹 Discover new azure flexible postgresql/mysql servers as part of the 'all' discovery.

preslavgerchev:


tas50 (approved):
Tested locally. Works great.

PR #4008, opened 2024-05-16
✨ Enable SBOM on Windows

czunker:


sbom/sbom.mql.yaml
```diff
@@ -21,7 +22,8 @@ packs:
         mql: npm.packages { name version purl cpes.map(uri) files.map(path) }
       - uid: mondoo-sbom-kernel-installed
         filters:
-          - mql: asset.family.contains('linux')
-          - mql: asset.runtime != 'container' && asset.kind != 'container' && asset.kind != 'container-image'
+          - mql: |
```
chris-rock, line 25: why do we need to join this? We can set multiple filters. Is the current approach not working?

sbom/sbom.mql.yaml
```diff
@@ -6,6 +6,7 @@ packs:
     name: Mondoo SBOM
     filters:
       - mql: asset.family.contains("unix")
+      - mql: asset.family.contains('windows')
```
chris-rock, line 9: Multiple entries should be and connected not OR, so this should not work

sbom/sbom.mql.yaml
czunker:
```
filters:
  - mql: ...
  - mql: ...
```
These filters are ORed. On Windows, this resulted in problems with the kernel query.
Windows doesn't match the first filter, but the second.

sbom/sbom.mql.yaml
czunker: Then, we have a bug in the filters. See comment on the kernel query.

sbom/sbom.mql.yaml
chris-rock: My fault, they are or connected

chris-rock (approved):
Thank you @czunker

sbom/sbom.mql.yaml
czunker: As discussed: https://github.com/mondoohq/cnspec/pull/1304

PR #4009, opened 2024-05-16
🧹 Prevent panic for unsupported operating systems

czunker:
Follow-up to #4008

### DIFF
--- a/sbom/sbom.go
+++ b/sbom/sbom.go
@@ -72,6 +72,12 @@ func GenerateBom(r *reporter.Report) ([]*Sbom, error) {
 
 		// extract os packages and python packages
 		dataPoints := r.Data[mrn]
+		if dataPoints == nil {
+			bom.Status = Status_STATUS_FAILED
+			bom.ErrorMessage = "no data points found"
+			boms = append(boms, bom)
+			continue
+		}
 		for k := range dataPoints.Values {
 			dataValue := dataPoints.Values[k]
 			jsondata, err := reporter.JsonValue(dataValue.Content)


czunker:
I tested this, by removing the linux filter from the sbom policy and run it on my local linux system:
```
FTL failed to parse bom error="no data points found"
Process 111189 has exited with status 1
```

sbom/sbom.go
```diff
@@ -72,6 +72,9 @@ func GenerateBom(r *reporter.Report) ([]*Sbom, error) {
 
 		// extract os packages and python packages
 		dataPoints := r.Data[mrn]
+		if dataPoints == nil {
+			return nil, fmt.Errorf("no data points found")
```
chris-rock: If we have no data points, this is not an error in the conversion. We should just return the current constructed bom then without any additional package data.

czunker: When the operating system is not supported, the complete sbom query pack is not executed and we don't have any data:
```

Process 158394 has exited with status 0
```
Shouldn't we at least show some info?

chris-rock: Great idea. What we could do  is, we set the error state, add a message and at least fill in the asset information. Here is the failed state https://github.com/mondoohq/cnquery/blob/main/sbom/sbom.proto#L38 that we can set here https://github.com/mondoohq/cnquery/blob/main/sbom/sbom.proto#L65. We can set the error message here https://github.com/mondoohq/cnquery/blob/main/sbom/sbom.proto#L81

chris-rock: I think we want to store the error in the report but not error the report generation.

czunker: This way the CLI still does not show any error, but the json output includes it:
```
{
  "generator": {
    "vendor": "Mondoo, Inc.",
    "name": "cnquery",
    "url": "https://mondoo.com"
  },
  "timestamp": "2024-05-21T08:31:32Z",
  "status": 3,
  "asset": {
    "name": "X1",
    "platform": {}
  },
  "error_message": "no data points found"
}
```

chris-rock (approved):
Thank you for the updated error handling.

PR #4017, opened 2024-05-16
🐛 don't keep closed unprocessed runtimes in coordinator

imilchev:
we are currently leaking memory for long-running cnquery/cnspec instances because the coordinator is keeping around closed unprocessed runtimes. This PR makes sure we throw away these runtimes

### DIFF
--- a/providers/coordinator.go
+++ b/providers/coordinator.go
@@ -153,6 +153,9 @@ func (c *coordinator) unsafeRefreshRuntimes() {
 	var remaining []*Runtime
 	for i := range c.unprocessedRuntimes {
 		rt := c.unprocessedRuntimes[i]
+		if rt.isClosed {
+			continue
+		}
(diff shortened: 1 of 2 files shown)


providers/coordinator.go
```diff
@@ -153,7 +153,7 @@ func (c *coordinator) unsafeRefreshRuntimes() {
 	var remaining []*Runtime
 	for i := range c.unprocessedRuntimes {
 		rt := c.unprocessedRuntimes[i]
-		if asset := rt.asset(); asset == nil || !c.unsafeSetAssetRuntime(asset, rt) {
+		if asset := rt.asset(); (asset == nil || !c.unsafeSetAssetRuntime(asset, rt)) && !rt.isClosed {
```
preslavgerchev: nit:
can we check for `rt.isClosed` before even getting the asset? if closed we can just continue

preslavgerchev (approved):

PR #4020, opened 2024-05-16
Updates aws.account.id resource

scottford-io:
This PR updates the `aws.account.id` resource to only return the account ID. The previous version concatenates the string "aws.account/" to the account id. That string is redundant because the resource is already `aws.account.id` 

## OLD VERSION:
```
cnspec> aws.account.id 
aws.account.id: "aws.account/177043759666"
``` 

### NEW VERSION:
```
cnquery> aws.account.id
aws.account.id: "177043759666"
```

### DIFF
--- a/providers/aws/resources/aws_account.go
+++ b/providers/aws/resources/aws_account.go
@@ -16,7 +16,7 @@ import (
 )
 
 func (a *mqlAwsAccount) id() (string, error) {
-	return "aws.account/" + a.Id.Data, nil
+	return a.Id.Data, nil
 }
 
 func (a *mqlAwsAccount) aliases() ([]interface{}, error) {


providers/aws/resources/aws_account.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 func (a *mqlAwsAccount) id() (string, error) {
-	return "aws.account/" + a.Id.Data, nil
```
vjeffrey, line 19: just a heads up, there are a couple other adjustments around discovery that we will need to do if this change goes in

tas50: This change feels worth it IMO. The old output didn't really make sense. This output is better for asset inventory and customer usage. We should definitely understand the impact of this breaking change though anywhere else in the product.

PR #4024, opened 2024-05-17
 ⭐️  Adding new resource for TeamsMessagingPolicy (MS365)

HRouhani:
Screenshot from 2024-05-17 14-45-58

### DIFF
--- a/providers/ms365/resources/ms365.lr
+++ b/providers/ms365/resources/ms365.lr
@@ -442,13 +442,13 @@ private ms365.sharepointonline.site {
 
 // Microsoft 365 Teams
 ms365.teams {
-  // CS Teams client configuration
+  // Teams client configuration
   csTeamsClientConfiguration() dict
-  // CS Teams tenant federated configuration
+  // Teams tenant federated configuration
   csTenantFederationConfiguration() ms365.teams.tenantFederationConfig
-  // CS Teams meeting policy configuration
+  // Teams meeting policy configuration
   csTeamsMeetingPolicy() ms365.teams.teamsMeetingPolicyConfig
-  // CS Teams message policy configuration
+  // Teams message policy configuration
   csTeamsMessagingPolicy() ms365.teams.teamsMessagingPolicyConfig
 }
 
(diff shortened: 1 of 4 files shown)


providers/ms365/resources/ms365.lr.manifest.yaml
```diff
@@ -328,12 +328,21 @@ resources:
...
         min_mondoo_version: 9.0.0
+      csTeamsMessagingPolicy:
+        min_mondoo_version: 9.0.0
       csTenantFederationConfiguration:
         min_mondoo_version: 9.0.0
     min_mondoo_version: 5.15.0
     platform:
       name:
       - microsoft365
+  ms365.teams.TeamsMessagingPolicyConfig:
```
preslavgerchev: Can you remove this?

providers/ms365/resources/ms365.lr
```diff
@@ -492,4 +494,10 @@ private ms365.teams.teamsMeetingPolicyConfig {
   allowExternalParticipantGiveRequestControl bool
   // Whether users can report security concerns
   allowSecurityEndUserReporting bool
+}
+
+// Teams meeting policy configuration
```
tas50, line 499:
```suggestion
// Microsoft 365 Teams meeting policy configuration
```

providers/ms365/resources/ms365.lr
```diff
@@ -448,6 +448,8 @@ ms365.teams {
   csTenantFederationConfiguration() ms365.teams.tenantFederationConfig
   // CS Teams meeting policy configuration
   csTeamsMeetingPolicy() ms365.teams.teamsMeetingPolicyConfig
+  // CS Teams message policy configuration
+  csTeamsMessagingPolicy() ms365.teams.teamsMessagingPolicyConfig
```
tas50, line 452: We should probably consider not using the CS here since that goes back to the legacy Skype Communications Server (CS). The cmdlets in PowerShell use the names, but that doesn't mean we need to follow that same naming.

providers/ms365/resources/ms365.lr
HRouhani: done. Thanks

providers/ms365/resources/ms365.lr.manifest.yaml
HRouhani: sure, did it!

providers/ms365/resources/ms365.lr
HRouhani: removed CS, Thanks

PR #4036, opened 2024-05-20
🐛 prevent hang when HelloRetryRequest is returned

jaym:
It was possible for the following query to hang
```
tls.ciphers
```

This happened in cases where the server response with a `HelloRetryRequest`. This looks like a `ServerHello` until you check the random field of that message.

We ended up hanging because we thought the server replied successfully with a hello and were waiting for more data, and the server was waiting for us to write.

### DIFF
--- a/providers/network/resources/tlsshake/tlsshake.go
+++ b/providers/network/resources/tlsshake/tlsshake.go
@@ -309,12 +309,38 @@ func (s *Tester) parseAlert(data []byte, conf *ScanConfig) error {
 	return nil
 }
 
+// helloRetryRequestRandom is the value of the random field in a ServerHello
+// message that indicates a HelloRetryRequest. We must check for this value
+// in order to handle HelloRetryRequests correctly. Otherwise, we will
+// incorrectly interpret the message as a ServerHello. The server will
+// wait for us to send a new ClientHello with the correct parameters. And we
+// will be stuck waiting for data that will never come.
+var helloRetryRequestRandom = []byte{
+	0xcf, 0x21, 0xad, 0x74, 0xe5, 0x9a, 0x61, 0x11,
+	0xbe, 0x1d, 0x8c, 0x02, 0x1e, 0x65, 0xb8, 0x91,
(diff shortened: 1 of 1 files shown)


providers/network/resources/tlsshake/tlsshake.go
```diff
@@ -309,12 +309,24 @@ func (s *Tester) parseAlert(data []byte, conf *ScanConfig) error {
 	return nil
 }
 
+var helloRetryRequestRandom = []byte{
```
czunker, line 318: nit: Perhaps add a comment or rename it.

Only the comment in the PR helped me understand that this is not random data, but the random message in the response.

providers/network/resources/tlsshake/tlsshake.go
```diff
@@ -309,12 +309,24 @@ func (s *Tester) parseAlert(data []byte, conf *ScanConfig) error {
...
 
-	idx += 2 + 32
-	// handshake tls version (2), which we don't need yet (we will look at it in the extension if necessary)
-	// random (32), which we don't need
+	idx += 2
+	random := data[idx : idx+32]
+	if bytes.Equal(random, helloRetryRequestRandom) {
+		// The server wants us to retry. However, we hardcode a bunch of inputs
+		// so we can't really retry.
+		return errors.New("unexpected HelloRetryRequest")
```
czunker: Is it possible to extract the key_share or supported_versions from the response extensions?

That might help the next we hit this, and we need to figure out what else besides `secp256r1` is needed.

czunker (approved):

providers/network/resources/tlsshake/tlsshake.go
jaym: I've added the key share group to the error message. We already seem to do something with supported versions

providers/network/resources/tlsshake/tlsshake.go
jaym: Added a comment

czunker (approved):
Really nice! Thanks @jaym

PR #4038, opened 2024-05-21
🐛 Assume connection reset means the server rejects all ciphersuites

jaym:
When testing for supported ciphersuites, some servers fail the check with:
```
read: connection reset by peer
```

This seems to consistently happen on these servers. And it seems to happen on requests where the only ciphersuites in the `ClientHello` are unsupported ones


czunker (approved):
LGTM

Thanks @jaym

PR #4041, opened 2024-05-21
⭐️ github app authentication

imilchev:
Allow scanning github with app authentication:
```
cnquery scan github org mondoohq --app-id 1234 --app-installation-id 123456 --app-private-key <path-to-pem>
```


providers/github/provider/provider.go
```diff
@@ -46,17 +46,35 @@ func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error)
...
+	isAppAuth := false
+	if appId, ok := req.Flags[connection.OPTION_APP_ID]; ok && len(appId.Value) > 0 {
+		conf.Options[connection.OPTION_APP_ID] = string(appId.Value)
+
+		installId := req.Flags[connection.OPTION_APP_INSTALLATION_ID]
+		conf.Options[connection.OPTION_APP_INSTALLATION_ID] = string(installId.Value)
+
+		pk := req.Flags[connection.OPTION_APP_PRIVATE_KEY]
+		conf.Options[connection.OPTION_APP_PRIVATE_KEY] = string(pk.Value)
+		isAppAuth = true
```
chris-rock, line 62: If the app-id is set but installation id or private key are missing we should return an error

imilchev: that's done in the connection, not here. I wanted to not duplicate the logic for checking if everything is set in the `ParseCLI`

chris-rock: That makes a lot of sense.

chris-rock (approved):
Thank you @imilchev

PR #4049, opened 2024-05-22
⭐️ Add device connection

preslavgerchev:
This PR adds the initial version of the device connection. It's currently only supported on linux.

To run a best-effort (guess the device) scan:
```
cnquery scan device
```

To specify the actual device you want:
```
cnquery scan device --device-name /dev/sda
```

To specify a device by its LUN (specifically useful for azure vms):
```
cnquery scan device --lun 1
```

The connection also supports injecting platform-ids:
```
cnquery scan device --lun 1 --platform-ids 1,2,3
```


providers/os/connection/device/device_manager.go
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package device
+
+import "go.mondoo.com/cnquery/v11/providers/os/connection/device/shared"
+
+type DeviceManager interface {
```
preslavgerchev, line 8: we can change this interface in the future to make this work better with mac/windows if needed

providers/os/connection/snapshot/volumemounter.go
```diff
@@ -26,12 +26,12 @@ type VolumeMounter struct {
 	// where we tell AWS to attach the volume; it doesn't necessarily get attached there, but we have to reference this same location when detaching
 	VolumeAttachmentLoc string
 	opts                map[string]string
-	cmdRunner           *LocalCommandRunner
+	CmdRunner           *LocalCommandRunner
 }
 
 func NewVolumeMounter(shell []string) *VolumeMounter {
```
preslavgerchev, line 32: we can move this entirely to `device/linux` once we migrate the snapshot connections (AWS, Azure, GCP)

providers/os/connection/device/device_connection.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func getDeviceManager(conf *inventory.Config) (DeviceManager, error) {
+	shell := []string{"sh", "-c"}
+	if runtime.GOOS == "darwin" {
+		return nil, errors.New("device manager not implemented for darwin")
+	}
+	if runtime.GOOS == "windows" {
+		// shell = []string{"powershell", "-c"}
+		return nil, errors.New("device manager not implemented for windows")
```
preslavgerchev, line 38: adding support for windows should be as simple as implementing the device manager and then ensuring it gets created here

chris-rock: nice preparation

providers/os/connection/device/device_connection.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func NewDeviceConnection(connId uint32, conf *inventory.Config, asset *inventory.Asset) (*DeviceConnection, error) {
+	manager, err := getDeviceManager(conf)
+	if err != nil {
+		return nil, err
+	}
+	log.Debug().Str("manager", manager.Name()).Msg("device manager created")
+
+	mi, err := manager.IdentifyBlock(conf.Options)
```
chris-rock: Should we call this function `IdentifyBlockDevice`?

chris-rock: Going forward we should expect that this returns a list of block devices.

providers/os/connection/device/device_connection.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	log.Debug().Str("device", mi.DeviceName).Msg("identified block for mounting")
+
+	res := &DeviceConnection{
+		Connection:    plugin.NewConnection(connId, asset),
+		deviceManager: manager,
+		asset:         asset,
+	}
+
+	scanDir, err := manager.Mount()
```
chris-rock: I think we should pass in the device we want to mount to make this more re-usable. We can also do this later when we add discovery for multiple block devices.

chris-rock (approved):
Thank you @preslavgerchev

vjeffrey (approved):
👍 thank you!!

providers/os/connection/device/device_connection.go
preslavgerchev: great idea, adjusted, the code now returns a list. the connection stills works with a single device but we can adjust this when needed.

providers/os/connection/device/device_connection.go
preslavgerchev: great idea, adjusted. i will clean up the actual volume mounter too in another PR-

PR #4085, opened 2024-05-23
🧹 Rework azure snapshot connection to use device connection.

preslavgerchev:
Reworks the azure snapshot connection almost entirely.
 - All the flags around skipping setup are now dropped. This connection now only supports selecting a target (disk, snapshot, instance), cloning that and mounting onto the scanner VM.
 - Delete all code that requires running local commands. The needed properties for mounting are determined entirely via the API
 - Embed `DeviceConnection` into the azure connection. It now only holds the job of identifying the VM its on and making a disk for the target we're scanning.


providers/azure/connection/azureinstancesnapshot/instance.go
```diff
@@ -0,0 +1,58 @@
...
+	resourceGroup  string
+	instanceName   string
+	location       string
+	bootDiskId     string
+	zones          []*string
+	// Attach the entire VM response as well
+	vm compute.VirtualMachine
+}
+
+func (s *instanceInfo) getFirstAvailableLun() (int, error) {
```
preslavgerchev, line 26: we no longer run cmds, we simply use the azure API to tell us what LUN we can attach the disk to

imilchev (approved):

PR #4088, opened 2024-05-23
🧹 Update AWS snapshot connection to use device connection.

preslavgerchev:
~~Putting this one in draft as it needs testing~~


providers/aws/provider/provider.go
```diff
@@ -203,15 +203,6 @@ func (s *Service) connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 
 		case string(awsec2ebsconn.EBSConnectionType):
 			conn, err = awsec2ebsconn.NewAwsEbsConnection(connId, conf, asset)
-			if conn.Asset() != nil && len(conn.Asset().Connections) > 0 && conn.Asset().Connections[0].Options["mounted"] != "" {
```
preslavgerchev, line 206: im not sure why this was needed. the ebs conn seems to work without it just fine

vjeffrey: were you checking with builtin provider or not builtin? we saw some weird behaviour in the past that was diff between the two scenarios

preslavgerchev: built in. i can try testing with the provider binary separate too

preslavgerchev: Just tested, no issues with a non-built in provider

PR #4089, opened 2024-05-23
Improve test in GitHub app scanning help

tas50:
- GitHub calls these apps not applications.
- Make sure it's clear we're asking for the path to a pem file not the pem contents as a string


misterpantz (approved):
thanks, @tas50!

PR #4101, opened 2024-05-24
🧹 Add new methods for detecting and mounting in the volume mounter.

preslavgerchev:
Split the detection of the partition we're mounting from the actual mounting process. To not make any changes to existing code that's using the volume mounter, I've added two new functions there:  `GetDeviceForMounting` and `MountP` (artition)

The device manager's interface was also reworked:
 * Changed `IdentifyBlockDevice` to `IdentifyMountTargets` to better account for the fact we can have multiple targets (even if unimplemented)
 * `Mount` now returns the scan dir where it was mounted

### DIFF
--- a/providers/os/connection/snapshot/volumemounter.go
+++ b/providers/os/connection/snapshot/volumemounter.go
@@ -35,13 +35,13 @@ func NewVolumeMounter(shell []string) *VolumeMounter {
 	}
 }
 
+// we should try and migrate this towards MountP where we specifically mount a target partition.
+// the detection of the partition should be done in the caller (separately from Mount)
 func (m *VolumeMounter) Mount() error {
-	err := m.createScanDir()
+	_, err := m.createScanDir()
 	if err != nil {
 		return err
 	}
-	// we should consider dropping this if VolumeAttachmentLoc is set. we need to also add FsType but
-	// otherwise that means we're listing the devices twice
 	fsInfo, err := m.getFsInfo()
 	if err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


providers/os/connection/snapshot/volumemounter.go
```diff
@@ -102,7 +120,37 @@ func (m *VolumeMounter) getFsInfo() (*fsInfo, error) {
 	return nil, err
 }
 
-func (m *VolumeMounter) mountVolume(fsInfo *fsInfo) error {
+// GetDeviceForMounting iterates through all the partitions of the target and returns the first one that matches the filters
+// If target is empty, it will return the first unnamed block device (best-effort guessing)
+// E.g. if target is "sda", it will return the first partition of the block device "sda" that satisfies the filters
+func (m *VolumeMounter) GetDeviceForMounting(target string) (*PartitionInfo, error) {
```
preslavgerchev, line 126: this is now super explicit in what it does. if we have a target, we only look for mountable partitions there. if the target is empty only then we do best-effort guessing.

preslavgerchev: we can also extend this method to support different filters on the block devices

preslavgerchev: maybe we should call this `GetTargetForMounting` 🤔

providers/os/connection/snapshot/volumemounter.go
```diff
@@ -53,18 +53,36 @@ func (m *VolumeMounter) Mount() error {
 	return m.mountVolume(fsInfo)
 }
 
-func (m *VolumeMounter) createScanDir() error {
+// Mounts a specific partition and returns the directory it was mounted to
+func (m *VolumeMounter) MountP(partition *PartitionInfo) (string, error) {
```
preslavgerchev, line 57: this no longer runs `lsblk` again, it simply tries to mount what we're telling it to

providers/os/connection/snapshot/volumemounter.go
```diff
@@ -102,7 +120,37 @@ func (m *VolumeMounter) getFsInfo() (*fsInfo, error) {
...
+	data, err := io.ReadAll(cmd.Stdout)
+	if err != nil {
+		return nil, err
+	}
+	blockEntries := BlockDevices{}
+	if err := json.Unmarshal(data, &blockEntries); err != nil {
+		return nil, err
+	}
+	if target == "" {
+		// TODO: i dont know what the difference between GetUnnamedBlockEntry and GetUnmountedBlockEntry is
```
vjeffrey, line 146: lol neither do i and i think i wrote them 😆 
i'm glad we're revisiting this code :)

providers/os/connection/snapshot/volumemounter.go
vjeffrey: i like it! i'm either way on the name, both are clear

vjeffrey: we still have `func (m *VolumeMounter) getFsInfo() (*PartitionInfo, error) {` in the code, which is v similar to this new func - is that intentional until we can test things more?

preslavgerchev: yep! i didnt wanna touch the old code but it should go away soon

PR #4104, opened 2024-05-24
🧹 Pull in replace dependencies from provider go.mods when using configure

preslavgerchev:
If the provider we're pulling in as a built in has specific pinned dependencies, ensure we also pull those in the main `go.mod` file.

Else we might pull newer versions of the provided dependency, which might cause failures in compilation, e.g.
the azure provider pins a specific `armsecurity` package here:  https://github.com/mondoohq/cnquery/blob/main/providers/azure/go.mod#L6

With this change, we also pull in this `replace` statement when building azure as a built-in provider.

Before:
```
module go.mondoo.com/cnquery/v11

replace go.mondoo.com/cnquery/v11/providers/azure => ./providers/azure

go 1.22

toolchain go1.22.0
```

After:
```
module go.mondoo.com/cnquery/v11

replace go.mondoo.com/cnquery/v11/providers/azure => ./providers/azure

replace github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/security/armsecurity => github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/security/armsecurity v0.13.0

go 1.22

toolchain go1.22.0
```

### DIFF
--- a/providers-sdk/v1/util/configure/configure.go
+++ b/providers-sdk/v1/util/configure/configure.go
@@ -18,6 +18,7 @@ import (
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
 	"go.mondoo.com/cnquery/v11/logger"
+	"golang.org/x/mod/modfile"
 	"sigs.k8s.io/yaml"
 )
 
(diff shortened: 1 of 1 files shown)


afiune (approved):
Sweeet man!

afiune:
@preslavgerchev having approve this PR, we might need to work on https://github.com/mondoohq/cnspec to
apply a similar strategy like you did here because if we try to use `cnquery` from `cnspec` with these replace
statements, they don't get updated over there.

PR #4105, opened 2024-05-26
⭐️ aws cloudformation provider

chris-rock:
This PR adds support for AWS Cloudformation templates and AWS SAM templates.

```
cnquery shell cloudformation providers/cloudformation/testdata/cloudformation.yaml
→ loaded configuration from /Users/chris/.config/mondoo/mondoo.yml using source default
→ connected to AWS Cloudformation
  ___ _ __   __ _ _   _  ___ _ __ _   _ 
 / __| '_ \ / _` | | | |/ _ \ '__| | | |
| (__| | | | (_| | |_| |  __/ |  | |_| |
 \___|_| |_|\__, |\__,_|\___|_|   \__, |
  mondoo™      |_|                |___/  interactive shell

cnquery> asset { name platform }
asset: {
  platform: "cloudformation"
  name: "Cloudformation Static Analysis cloudformation"
}

cnquery> cloudformation.template { * }
cloudformation.template: {
  mappings: {
    RegionMap: {
      ap-northeast-1: {
        AMI: "ami-06cd52961ce9f0d85"
      }
      ap-southeast-1: {
        AMI: "ami-08569b978cc4dfa10"
      }
      ap-southeast-2: {
        AMI: "ami-09b42976632b27e9b"
      }
      eu-west-1: {
        AMI: "ami-047bb4163c506cd98"
      }
      sa-east-1: {
        AMI: "ami-07b14488da8ea02a0"
      }
      us-east-1: {
        AMI: "ami-0ff8a91507f77f867"
      }
      us-west-1: {
        AMI: "ami-0bdb828fd58c52235"
      }
      us-west-2: {
        AMI: "ami-a0cfeed8"
      }
    }
  }
  parameters: {
    EnvType: {
      AllowedValues: [
        0: "prod"
        1: "dev"
        2: "test"
      ]
      ConstraintDescription: "must specify prod, dev, or test."
      Default: "test"
      Description: "Environment type."
      Type: "String"
    }
  }
  resources: [
    0: cloudformation.resource name="EC2Instance"
    1: cloudformation.resource name="MountPoint"
    2: cloudformation.resource name="NewVolume"
  ]
  globals: {}
  metadata: {}
  conditions: {
    CreateDevResources: {
      Fn::Equals: [
        0: {
          Ref: "EnvType"
        }
        1: "dev"
      ]
    }
    CreateProdResources: {
      Fn::Equals: [
        0: {
          Ref: "EnvType"
        }
        1: "prod"
      ]
    }
  }
[...]

PR #4108, opened 2024-05-27
🧹 Make path an arg, not a flag for the FS conn.

preslavgerchev:
Aligns the provider with #4105 by having the path as an arg.  Passing it in as a flag is still supported, but deprecated


chris-rock (approved):
Thank you @preslavgerchev

PR #4110, opened 2024-05-27
🧹 Rename functions in blockdevices.go. Add more tests, sort suitable partitions by size before choosing one.

preslavgerchev:
* Rename `GetDeviceForMounting` to `GetMountablePartition`
* Rename `GetBlockByEntryName` to `GetMountablePartitionByDevice`
* Change the cmd we run to be more extensible, using `lsblk -o OPTS` allows us to include things easily.
* Sort all suitable partitions and pick the largest one. In the case where we have non-mounted, non-boot partitions and we cannot distinguish between them, we simply take the biggest one
* Update tests

### DIFF
--- a/providers/os/connection/snapshot/blockdevices.go
+++ b/providers/os/connection/snapshot/blockdevices.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"sort"
 	"strings"
 
 	"github.com/rs/zerolog/log"
@@ -19,12 +20,12 @@ type BlockDevices struct {
 
 type BlockDevice struct {
 	Name       string        `json:"name,omitempty"`
-	FsType     string        `json:"FsType,omitempty"`
+	FsType     string        `json:"fstype,omitempty"`
 	Label      string        `json:"label,omitempty"`
 	Uuid       string        `json:"uuid,omitempty"`
 	MountPoint string        `json:"mountpoint,omitempty"`
 	Children   []BlockDevice `json:"children,omitempty"`
-	FsUse      string        `json:"fsuse%,omitempty"`
+	Size       int           `json:"size,omitempty"`
 }
 
 type PartitionInfo struct {
@@ -33,7 +34,7 @@ type PartitionInfo struct {
 }
 
 func (cmdRunner *LocalCommandRunner) GetBlockDevices() (*BlockDevices, error) {
-	cmd, err := cmdRunner.RunCommand("lsblk -f --json")
+	cmd, err := cmdRunner.RunCommand("lsblk -bo NAME,SIZE,FSTYPE,MOUNTPOINT,LABEL,UUID --json")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


providers/os/connection/snapshot/blockdevices.go
```diff
@@ -19,12 +20,12 @@ type BlockDevices struct {
 
 type BlockDevice struct {
 	Name       string        `json:"name,omitempty"`
-	FsType     string        `json:"FsType,omitempty"`
+	FsType     string        `json:"fstype,omitempty"`
```
chris-rock, line 23: Is that caused by the fact that we switch to specific fields?

providers/os/connection/snapshot/blockdevices.go
```diff
@@ -33,7 +34,7 @@ type PartitionInfo struct {
 }
 
 func (cmdRunner *LocalCommandRunner) GetBlockDevices() (*BlockDevices, error) {
-	cmd, err := cmdRunner.RunCommand("lsblk -f --json")
+	cmd, err := cmdRunner.RunCommand("lsblk -bo NAME,SIZE,FSTYPE,MOUNTPOINT,LABEL,UUID --json")
```
chris-rock, line 37: Just for my understanding, why are we adding those additional arguments? Are those fields are not covered with the default output?

providers/os/connection/snapshot/blockdevices.go
```diff
@@ -71,36 +73,61 @@ func (blockEntries BlockDevices) GetRootBlockEntry() (*PartitionInfo, error) {
 	return nil, errors.New("target volume not found on instance")
 }
 
-func (blockEntries BlockDevices) GetBlockEntryByName(name string) (*PartitionInfo, error) {
-	log.Debug().Str("name", name).Msg("get matching block entry")
+// Searches all the partitions in the target device and finds one that can be mounted. It must be unmounted, non-boot partition
+// If multiple partitions meet this criteria, the largest one is returned.
+func (blockEntries BlockDevices) GetMountablePartitionByDevice(device string) (*PartitionInfo, error) {
```
chris-rock, line 77: I like the rename to device

chris-rock (approved):
Thank you @preslavgerchev

providers/os/connection/snapshot/blockdevices.go
preslavgerchev: Nope, if you do `lsblk -f --json` the size isnt there. This also lets us extend this in the future. For example, some partitions have a `PARTLABEL` which we can use (e.g. in Azure some partitions can be labeled with the BOOT PARTLABEL so we can omit those)

providers/os/connection/snapshot/blockdevices.go
preslavgerchev: I am really not sure. This is also `fstype` if you run `lsblk -f --json`. Don't know why this was specified like that

providers/os/connection/snapshot/blockdevices.go
preslavgerchev: This just makes it easier for us in the future since all we neeed is a an extra column and an extra field in the struct.

PR #4115, opened 2024-05-28
🧹 Split device discovery from partition discovery in blockdevices.go

preslavgerchev:
Split device discovery from partition discovery entirely. This means that the code is now more predictable and makes less assumptions.

 1. When we have a device name or a LUN arg, we first identify the specific device
 2. We search for partitions *only* in that device. We don't loop around the other devices
 3. Deprecate the old code. We need to clean it after we upgrade the AWS snapshot connection.


providers/os/connection/device/linux/device_manager.go
```diff
@@ -104,27 +104,43 @@ func (c *LinuxDeviceManager) identifyViaLun(lun int) (*snapshot.PartitionInfo, e
...
 	}
-
-	var target string
+	blockDevices, err := c.volumeMounter.CmdRunner.GetBlockDevices()
+	if err != nil {
+		return nil, err
+	}
+	var device snapshot.BlockDevice
+	var deviceErr error
 	// if we have exactly one device present at the LUN we can directly point the volume mounter towards it
```
afiune: this needs to be update? we, on top, look for the block device.

providers/os/connection/device/linux/device_manager.go
```diff
@@ -104,27 +104,43 @@ func (c *LinuxDeviceManager) identifyViaLun(lun int) (*snapshot.PartitionInfo, e
...
-	return c.volumeMounter.GetMountablePartition(deviceName)
+	blockDevices, err := c.volumeMounter.CmdRunner.GetBlockDevices()
+	if err != nil {
+		return nil, err
+	}
+
+	// if we don't have a device name we can just return the first non-boot, non-mounted partition.
+	// this is a best-guess approach
+	if deviceName == "" {
+		// TOOD: we should rename/simplify this method
```
afiune: typo + we have more comments like this one.

`providers/os/connection/snapshot/volumemounter.go:138`
```
// TODO: i dont know what the difference between GetUnnamedBlockEntry and GetUnmountedBlockEntry is
// we need to simplify those
```

Do we have a place to track these TODOs?

providers/os/connection/device/linux/lun.go
```diff
@@ -57,7 +57,7 @@ func filterScsiDevices(scsiDevices scsiDevices, lun int) []scsiDeviceInfo {
 // there can be multiple devices mounted at the same LUN.
 // the LUN so we need to find all the blocks, mounted at that LUN. then we find the first one
 // that has no mounted partitions and use that as the target device. this is a best-effort approach
```
afiune, line 59: could we please update this comment?

providers/os/connection/device/linux/lun.go
```diff
@@ -69,25 +69,27 @@ func findMatchingDeviceByBlock(scsiDevices scsiDevices, blockDevices *snapshot.B
...
-		mounted := false
 		for _, ch := range b.Children {
 			if len(ch.MountPoint) > 0 && ch.MountPoint != "" {
 				log.Debug().Str("name", ch.Name).Msg("device connection> has mounted partitons, skipping")
-				mounted = true
-			}
-			if !mounted {
-				target = "/dev/" + b.Name
+			} else {
+				targetDevice = b
```
afiune: why don't we stop processing blocks after we have found the target device?

I feel we are missing something here.

providers/os/connection/device/linux/lun_test.go
```diff
@@ -109,31 +109,4 @@ func TestFindDeviceByBlock(t *testing.T) {
 		_, err := findMatchingDeviceByBlock(devices, blockDevices)
 		assert.Error(t, err)
 	})
-	t.Run("empty target as all blocks are mounted", func(t *testing.T) {
```
afiune: any reason why to remove this test? specifically the part for when multiple blocks are mounted.

providers/os/connection/snapshot/blockdevices.go
```diff
@@ -123,6 +124,58 @@ func (blockEntries BlockDevices) GetMountablePartitionByDevice(device string) (*
 	return &PartitionInfo{Name: devFsName, FsType: partitions[0].FsType}, nil
 }
 
+// Searches for a device by name
+func (blockEntries BlockDevices) FindDevice(device string) (BlockDevice, error) {
```
afiune: nit, since we are searching for device "name", can we either rename the function or just update the argument?
```suggestion
func (blockEntries BlockDevices) FindDevice(name string) (BlockDevice, error) {
```

providers/os/connection/snapshot/blockdevices.go
```diff
@@ -123,6 +124,58 @@ func (blockEntries BlockDevices) GetMountablePartitionByDevice(device string) (*
...
+	}
+	for i := range blockEntries.BlockDevices {
+		d := blockEntries.BlockDevices[i]
+		log.Debug().Str("name", d.Name).Interface("children", d.Children).Interface("mountpoint", d.MountPoint).Msg("found block device")
+		fullDeviceName := "/dev/" + d.Name
+		if device != fullDeviceName { // check if the device name matches
+			if secondName == "" {
+				continue
+			}
+			if secondName != fullDeviceName { // check if the device name matches the second name option (sdh and xvdh are interchangeable)
```
afiune: nit; we can simplify this to only one `if` statement
```suggestion
			// check if the device name matches the second name option (sdh and xvdh are interchangeable)
			if secondName == "" || secondName != fullDeviceName {
```

providers/os/connection/snapshot/blockdevices.go
```diff
@@ -123,6 +124,58 @@ func (blockEntries BlockDevices) GetMountablePartitionByDevice(device string) (*
...
+
+	if len(partitions) == 0 {
+		return nil, fmt.Errorf("no suitable partitions found on device %s", device.Name)
+	}
+
+	// sort the candidates by size, so we can pick the largest one
+	sortPartitionsBySize(partitions)
+
+	// return the largest partition. we can extend this to be a parameter in the future
+	devFsName := "/dev/" + partitions[0].Name
```
afiune, line 172: If we only sort the partitions to return the largest one, if feels a waste of processing.

I know that we won't have a huge number of partitions so, processing here is low-cost, but if feels that the sorting is unnecessary.

providers/os/connection/snapshot/volumemounter.go
```diff
@@ -138,8 +138,13 @@ func (m *VolumeMounter) GetMountablePartition(device string) (*PartitionInfo, er
 		// TODO: i dont know what the difference between GetUnnamedBlockEntry and GetUnmountedBlockEntry is
 		// we need to simplify those
 		return blockDevices.GetUnnamedBlockEntry()
+	} else {
```
afiune: This is the end of the function. You don't need this `else`.

providers/os/connection/device/linux/device_manager.go
preslavgerchev: it says that we have exactly one device that is found through the LUN lookup, which is what the `if` check does

providers/os/connection/device/linux/lun_test.go
preslavgerchev: great catch, ive reverted it

providers/os/connection/snapshot/blockdevices.go
preslavgerchev: i can also replace this with iterating over and keeping track of the one with the largest size. this just felt like less code

providers/os/connection/device/linux/lun.go
preslavgerchev: yep, we can just return. adjusted

providers/os/connection/device/linux/device_manager.go
preslavgerchev: i will clean them up once I migrate AWS's snapshot connection (theres already a PR i need to update)

providers/os/connection/device/linux/lun.go
preslavgerchev: what would you like to have updated? it still is relevant to the code it refers to

providers/os/connection/device/linux/device_manager.go
afiune: This is absolutely a nitpick, my point here was that, we added an extra step here, we are no longer just passing the volume path directly, we now try to find the block.

That's all.

preslavgerchev: ah, i see your point, ive adjusted it

PR #4119, opened 2024-05-29
Add cloudformation provider to a few other places

tas50:
Make sure we have the new cloudformation provider everywhere.


chris-rock (approved):
Thank you @tas50

PR #4131, opened 2024-05-29
🐛 fix k8s manifest scanning bugs

imilchev:
- make sure we discover all objects that we can from manifests
- make sure asset names are correct for objects with no specified namespace


czunker (approved):
LGTM. Thanks @imilchev

PR #4133, opened 2024-05-29
⭐️ github k8s manifests discovery

imilchev:
Add the possibility to discover k8s manifests in a github repo:
```
cnquery scan github repo mondoohq/X --discover repository,k8s-manifests
```

We are only checking if the github repo has `yaml` or `yml` files. If the repo has at least one of them, we clone the repo and ask the k8s provider to find possible k8s manifests in those files. Hidden yaml files and yaml files in hidden folders aren't considered

- extract git clone code from terraform provider and add to a reusable function in the providers sdk
- implement `k8s-manifests` discovery flag for github provider
- implement git scanning for k8s provider
- make sure asset discovery code is recursive (until now we only treated the top level asset as root and anything under it was considered an asset that needs to be scanned. With this change, that assumption is no longer true, so the discovery code had to be adjusted)


providers-sdk/v1/plugin/git.go
```diff
@@ -0,0 +1,111 @@
...
+
+	path, closer, err := gitClone(gitUrl)
+	if err != nil {
+		return "", nil, err
+	}
+	return path, closer, nil
+}
+
+func gitClone(gitUrl string) (string, func(), error) {
+	cloneDir, err := os.MkdirTemp(os.TempDir(), "gitClone")
```
chris-rock: should we use something like `cnquery-git-clone`

providers/github/resources/discovery.go
```diff
@@ -319,3 +338,64 @@ func hasTerraformHcl(client *github.Client, repo *mqlGithubRepository) (bool, er
...
+				Credentials: creds,
+				Discover:    &inventory.Discovery{Targets: []string{"auto"}},
+			}},
+		})
+	}
+	return res, nil
+}
+
+// hasYaml will check if the repository contains YAML files
+func hasYaml(client *github.Client, repo *mqlGithubRepository) (bool, error) {
```
chris-rock, line 376: nice

chris-rock (approved):
Thank you @imilchev

PR #4144, opened 2024-05-30
🧹 improve provider release version detect

imilchev:
we wrongfully detect versions for some providers because we `grep` for `Version`. Some of the providers have a flag that is called "Version", which conflicts. Searching for `Version:` should limit the results to just the Go struct


imilchev:
tested the change here: https://github.com/mondoohq/cnquery/actions/runs/9301302756/job/25599139213

seems to work correctly

chris-rock (approved):

PR #4167, opened 2024-06-02
More consistent help commands

tas50:
- Align some wording
- Remove periods
- Fix some bad spacing


providers/atlassian/config/config.go
```diff
@@ -50,31 +50,31 @@ For SCIM, you receive both the token and the directory-id from Atlassian when yo
 					Long:    "admin-token",
 					Type:    plugin.FlagType_String,
 					Default: "",
-					Desc:    "Provide Atlassian admin API token (used for Atlassian admin).",
```
tas50, line 53: @misterpantz I removed these since we're not using them everywhere else, but let me know if you want to start all the option flags with Provide or Specify and I can add that everywhere

chris-rock (dismissed):
Thank you @tas50

chris-rock (approved):
Thank you @tas50

PR #4173, opened 2024-06-03
🐛 set os kind in connect only path

jaym:


providers/os/provider/provider.go
```diff
@@ -509,6 +510,14 @@ func (s *Service) connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 		return nil, err
 	}
 
+	if asset.Platform.Kind == "" {
+		asset.Platform.Kind = "baremetal"
+	}
+	if asset.Connections[0].Runtime == "vagrant" {
```
chris-rock: why do we set this?

chris-rock (approved):

PR #4181, opened 2024-06-04
🧹 Run OS detection before running the ID detectors for the device connection

preslavgerchev:
We run `DetectOS` first before we run `IdentifyPlatform`.  `DetectOS` was already being called inside of `IdentifyPlatform` so this does not change a lot.

It does enable us to scan an asset where we've identified the platform but the ID detectors were unable to pick up a platform-id. If one is provided via the `--platform-ids` param and the OS detection was successful we can still scan the asset.


afiune (approved):
Wishing there could be some tests for this code, but this looks like a good cleanup, so let's get it in and maybe later we can think of ways to add tests.

PR #4192, opened 2024-06-05
🧹 Allow re-running the goreleaser job

czunker:
With this change, it can override the existing release files. This can help in case something goes wrong.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -340,3 +340,5 @@ docker_manifests:  # https://goreleaser.com/customization/docker_manifest/
       - mondoo/{{ .ProjectName }}:latest-arm64v8-rootless
       - mondoo/{{ .ProjectName }}:latest-armv6-rootless
       - mondoo/{{ .ProjectName }}:latest-armv7-rootless
+release:
+  replace_existing_artifacts: true


.goreleaser.yml
```diff
@@ -340,3 +340,6 @@ docker_manifests:  # https://goreleaser.com/customization/docker_manifest/
       - mondoo/{{ .ProjectName }}:latest-arm64v8-rootless
       - mondoo/{{ .ProjectName }}:latest-armv6-rootless
       - mondoo/{{ .ProjectName }}:latest-armv7-rootless
+release:
+  replace_existing_artifacts: true
+  mode: keep-existing
```
imilchev: do we need this when we are replacing existing artifgacts?

czunker: Yes. It is only about the release notes, not the artifacts:
> # What to do with the release notes in case there the release already exists.
>  #
>  # Valid options are:
>  # - `keep-existing`: keep the existing notes
>  # - `append`: append the current release notes to the existing notes
>  # - `prepend`: prepend the current release notes to the existing notes
>  # - `replace`: replace existing notes
>  #
>  # Default is `keep-existing`.
>  mode: append

https://goreleaser.com/customization/release/?h=replace_existing_artifacts#github

Perhaps not, now that I see the default.

czunker: I removed it.

imilchev (approved):

PR #4196, opened 2024-06-05
🐛 shutdown unused providers

imilchev:
i noticed we never kill the unused providers in the scan api mode because we always have a couple of empty unprocessed runtimes (that are used for initializing the scanner). Those shouldn't block cleaning up unused providers

### DIFF
--- a/providers/coordinator.go
+++ b/providers/coordinator.go
@@ -206,8 +206,17 @@ func (c *coordinator) RemoveRuntime(runtime *Runtime) {
 		}
 	}
 
-	// Shutdown any providers that are not being used anymore
-	if len(c.runtimes) == 0 && len(c.unprocessedRuntimes) == 0 {
+	// Shutdown any providers that are not being used anymore.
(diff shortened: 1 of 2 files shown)


imilchev:
Note: I am running this version on my cluster now. If no errors show up in the next few hours, we can merge the PR

providers/coordinator.go
```diff
@@ -207,7 +207,13 @@ func (c *coordinator) RemoveRuntime(runtime *Runtime) {
 	}
 
 	// Shutdown any providers that are not being used anymore
-	if len(c.runtimes) == 0 && len(c.unprocessedRuntimes) == 0 {
+	uprocessedRuntimeWithProviders := false
```
vjeffrey, line 213: can you add a comment here, either in this pr or as a followup, to say why we need this check?

vjeffrey (approved):

providers/coordinator.go
imilchev: done

PR #4203, opened 2024-06-06
⭐️ Add support for static analysis of Windows systems 

preslavgerchev:
This PR adds support for identifying Windows disks, mounted on a given FS as assets.
 * Move everything registry related under `providers/os/registry`. This allows both the connections and the OS MQL resources to use the same code.
 * Cleanup `providers/os/resources`, the MQL resources now only reference `providers/os/registry` and use that.
 * Add support for calling `RegLoadKeyW` and `UnLoadRegKeyW` syscalls that help us mount static registry files as subkeys under HKLM in the registry.
 * Add detection of `version`, `build`, `platform`, and hostname for Windows filesystems


providers/os/resources/windows/registrykey_powershell.go
```diff
@@ -69,131 +61,3 @@ ConvertTo-Json -compress $properties
 func GetRegistryKeyChildItemsScript(path string) string {
```
chris-rock: should we move this code to the registry package as well or was there a reason why we left this out?

providers/os/detector/detector_all.go
```diff
@@ -703,36 +704,83 @@ var windows = &PlatformResolver{
 	Name:     "windows",
 	IsFamily: false,
 	Detect: func(r *PlatformResolver, pf *inventory.Platform, conn shared.Connection) (bool, error) {
-		data, err := win.GetWmiInformation(conn)
-		if err != nil {
-			log.Debug().Err(err).Msg("could not gather wmi information")
-			return false, nil
-		}
+		if conn.Capabilities().Has(shared.Capability_RunCommand) {
+			data, err := win.GetWmiInformation(conn)
```
chris-rock: I wonder if we extract the individual detections (runtime and static) into to functions so the detect function is easier to read?

preslavgerchev: great point, i extracted those to a new `detector_windows` file

providers/os/resources/windows/registrykey_powershell.go
preslavgerchev: yep, moved. it was a leftover, let's keep all registry powershell related stuff in one place

chris-rock (approved):

providers/os/detector/detector_win.go
```diff
@@ -0,0 +1,91 @@
...
+	"go.mondoo.com/cnquery/v11/providers-sdk/v1/inventory"
+	"go.mondoo.com/cnquery/v11/providers/os/connection/shared"
+	win "go.mondoo.com/cnquery/v11/providers/os/detector/windows"
+	"go.mondoo.com/cnquery/v11/providers/os/registry"
+)
+
+func runtimeWindowsDetector(pf *inventory.Platform, conn shared.Connection) (bool, error) {
+	data, err := win.GetWmiInformation(conn)
+	if err != nil {
+		log.Debug().Err(err).Msg("could not gather wmi information")
```
afiune, line 19: why are we not returning the error here? 

Is it safe for us to access `data` even if there is an error?

providers/os/detector/detector_win.go
```diff
@@ -0,0 +1,91 @@
...
+		pf.Labels = map[string]string{}
+	}
+	pf.Labels["windows.mondoo.com/product-type"] = data.ProductType
+
+	// optional: try to get the ubr number (win 10 + 2019)
+	current, err := win.GetWindowsOSBuild(conn)
+	if err == nil && current.UBR > 0 {
+		pf.Build = strconv.Itoa(current.UBR)
+	} else {
+		log.Debug().Err(err).Msg("could not parse windows current version")
```
afiune, line 41: if `err` is `nil` but `urrent.UBR` is empty. We will panic here, right?

providers/os/detector/detector_win.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	err = rh.LoadSubkey(registry.Software, fi.Path)
+	if err != nil {
+		log.Debug().Err(err).Msg("could not load SOFTWARE registry key file")
+		return false, nil
+	}
+
+	pf.Name = "windows"
+	productName, err := rh.GetRegistryItemValue(registry.Software, "Microsoft\\Windows NT\\CurrentVersion", "ProductName")
+	if err == nil {
```
afiune, line 70: nit: if not nil, do we want to log this out?

providers/os/registry/registryhandler.go
```diff
@@ -0,0 +1,103 @@
...
+	return &RegistryHandler{
+		registries: make(map[string]string),
+	}
+}
+
+// Loads the given registry file into the registry handler under a subkey, generated by buildSubKeyPath.
+// The subkey file is indicated by the filepath parameter.
+// Only known registry files (see KnownRegistryFiles) can be loaded.
+func (r *RegistryHandler) LoadSubkey(registryId, filepath string) error {
+	if runtime.GOOS != "windows" {
```
afiune, line 35: nit; why not define this function also inside the files `registryhandler_unix.go` and `registryhandler_windows.go` instead?

providers/os/registry/registryhandler_windows.go
```diff
@@ -0,0 +1,50 @@
...
+	keyPtr, err := syscall.UTF16PtrFromString(key)
+	if err != nil {
+		return err
+	}
+	pathPtr, err := syscall.UTF16PtrFromString(path)
+	if err != nil {
+		return err
+	}
+	ret, _, err := regLoadKey.Call(syscall.HKEY_LOCAL_MACHINE, uintptr(unsafe.Pointer(keyPtr)), uintptr(unsafe.Pointer(pathPtr)))
+	// the Microsoft docs indicate that the return value is 0 on success
```
afiune, line 31: this is funny, what is the `err` for then? 😂

afiune: also, what if ret is `0` but `err` is nil?

Would the caller of `LoadRegistrySubkey()` be confused if it was a success of failure?

providers/os/registry/registryhandler_windows.go
```diff
@@ -0,0 +1,50 @@
...
+func UnloadRegistrySubkey(key string) error {
+	keyPtr, err := syscall.UTF16PtrFromString(key)
+	if err != nil {
+		return err
+	}
+
+	ret, _, err := regUnloadKey.Call(syscall.HKEY_LOCAL_MACHINE, uintptr(unsafe.Pointer(keyPtr)))
+	// the Microsoft docs indicate that the return value is 0 on success
+	if ret != 0 {
+		return err
```
afiune, line 47: same here, if ret is not zero and err is nil, would the caller of UnloadRegistrySubkey think that was a success?

providers/os/registry/registrykey.go
```diff
@@ -1,73 +1,119 @@
...
 
-// RegistryKeyItem represents a registry key item and its properties
-const getRegistryKeyItemScript = `
-$path = '%s'
-$reg = Get-Item ('Registry::' + $path)
-if ($reg -eq $null) {
-  Write-Error "Could not find registry key"
-  exit 1
+// derived from "golang.org/x/sys/windows/registry"
+// see https://github.com/golang/sys/blob/master/windows/registry/value.go#L17-L31
```
afiune, line 19: Niiiiiice!

providers/os/registry/registrykey_powershell.go
```diff
@@ -0,0 +1,95 @@
...
+		return nil, err
+	}
+
+	var items []RegistryKeyItem
+	err = json.Unmarshal(data, &items)
+	if err != nil {
+		return nil, err
+	}
+
+	return items, nil
```
afiune: nit; there is no need to check if err is nil
```suggestion
	return items, err
```

providers/os/registry/registrykey_powershell.go
```diff
@@ -0,0 +1,95 @@
...
+		return nil, err
+	}
+
+	var children []RegistryKeyChild
+	err = json.Unmarshal(data, &children)
+	if err != nil {
+		return nil, err
+	}
+
+	return children, nil
```
afiune: same
```suggestion
	return children, err
```

providers/os/resources/packages/packages.go
```diff
@@ -109,6 +109,7 @@ func ResolveSystemPkgManager(conn shared.Connection) (OperatingSystemPkgManager,
 		pm = &AlpinePkgManager{conn: conn, platform: asset.Platform}
 	case asset.Platform.Name == "macos": // mac os family
 		pm = &MacOSPkgManager{conn: conn}
+		// TODO: add a new manager here for fs conn that lists registry directly
```
afiune: when will we do this? or better said, is there a place we are tracking TODOs?

providers/os/id/hostname/hostname.go
```diff
@@ -65,5 +66,37 @@ func Hostname(conn shared.Connection, pf *inventory.Platform) (string, bool) {
...
+		key, err := rh.GetRegistryItemValue(registry.System, "ControlSet001\\Control\\ComputerName\\ComputerName", "ComputerName")
+		if err == nil {
+			return key.Value.String, true
+		} else {
+			// we also can try ControlSet002 as a fallback
+			key, err := rh.GetRegistryItemValue(registry.System, "ControlSet002\\Control\\ComputerName\\ComputerName", "ComputerName")
+			if err == nil {
+				return key.Value.String, true
+			}
+		}
```
afiune: no need to have this `else`

```suggestion
		}
		
		log.Debug().Err(err).Msg("unable to read windows registry, trying ControlSet002 fallback")
		key, err := rh.GetRegistryItemValue(registry.System, "ControlSet002\\Control\\ComputerName\\ComputerName", "ComputerName")
		if err == nil {
			return key.Value.String, true
		}
```

providers/os/detector/detector_win.go
preslavgerchev: whoops, i left that one out, great catch

providers/os/detector/detector_win.go
preslavgerchev: yep, fixed

providers/os/registry/registryhandler.go
preslavgerchev: It also has some additional logic besides making the syscall so I figured it best belongs to the struct which handles the syscalls too (and storing which registries are loaded)

providers/os/registry/registryhandler_windows.go
preslavgerchev: The err is to be thrown when the `ret` is not 0. There's always an err, even on success, so we cannot rely on the usual `err != nil` check:
https://go.dev/src/syscall/dll_windows.go (see 151)

providers/os/resources/packages/packages.go
preslavgerchev: i will remove this TODO, this isnt the way i was planning on fixing that either way, its a leftover

afiune (approved):
# YESSS!!!!

PR #4204, opened 2024-06-06
🧹 Use the registry directly to fetch Windows packages

preslavgerchev:
After #4203 
Instead of running a powershell script, query the registry directly for the installed apps.

### DIFF
--- a/providers/os/resources/packages/windows_packages.go
+++ b/providers/os/resources/packages/windows_packages.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"runtime"
 	"time"
 
 	"github.com/cockroachdb/errors"
@@ -14,6 +15,7 @@ import (
 	"go.mondoo.com/cnquery/v11/providers-sdk/v1/inventory"
 	"go.mondoo.com/cnquery/v11/providers/os/connection/shared"
 	"go.mondoo.com/cnquery/v11/providers/os/detector/windows"
+	"go.mondoo.com/cnquery/v11/providers/os/registry"
 	"go.mondoo.com/cnquery/v11/providers/os/resources/cpe"
 	"go.mondoo.com/cnquery/v11/providers/os/resources/powershell"
 )
@@ -80,6 +82,16 @@ const (
 	Drivers
 	Defender
 )
+
+const installedAppsScript = `
+Get-ItemProperty (@(
+  'HKLM:\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\*',
+  'HKCU:\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\*',
+  'HKLM:\\SOFTWARE\\Wow6432Node\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\*',
+  'HKCU:\\SOFTWARE\\Wow6432Node\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\*'
+) | Where-Object { Test-Path $_ }) |
+Select-Object -Property DisplayName,DisplayVersion,Publisher,EstimatedSize,InstallSource,UninstallString | ConvertTo-Json -Compress
+`
 
 var (
 	WINDOWS_QUERY_HOTFIXES      = `Get-HotFix | Select-Object -Property Status, Description, HotFixId, Caption, InstalledOn, InstalledBy | ConvertTo-Json`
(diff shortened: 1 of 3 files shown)


providers/os/resources/packages/windows_packages.go
```diff
@@ -203,15 +204,78 @@ func (w *WinPkgManager) Format() string {
...
-Get-ItemProperty (@(
-  'HKLM:\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\*',
-  'HKCU:\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\*',
-  'HKLM:\\SOFTWARE\\Wow6432Node\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\*',
-  'HKCU:\\SOFTWARE\\Wow6432Node\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\*'
-) | Where-Object { Test-Path $_ }) |
-Select-Object -Property DisplayName,DisplayVersion,Publisher,EstimatedSize,InstallSource,UninstallString | ConvertTo-Json -Compress
-`
+func (w *WinPkgManager) getInstalledApps() ([]Package, error) {
+	pkgs := []string{
```
chris-rock, line 219: Maybe I am missing something here, but how would that work with remote ssh scan of a Windows machine?

preslavgerchev: i had missed that. we need both in that case, determining which one to use based on the connection type. i will fix this

preslavgerchev: @chris-rock added that back. we now run directly through the registry when using `local` conn and as a fallback: we still use the powershell

providers/os/resources/packages/windows_packages.go
```diff
@@ -203,15 +214,90 @@ func (w *WinPkgManager) Format() string {
...
+			continue
+		}
+		for _, c := range children {
+			// for each package the information is contained as children of that registry's key
+			p := w.getPackageFromRegistryKey(c)
+			if p == nil {
+				continue
+			}
+			packages = append(packages, *p)
+		}
```
afiune, line 245: I have seen this same for loop in other places, I wonder if there is a way to abstract this later for two reasons:

* have a single place where we locate packages, so we don't have to update multiple functions
* use interfaces? so we can write some unit tests! 😄

providers/os/resources/packages/windows_packages.go
```diff
@@ -203,15 +214,90 @@ func (w *WinPkgManager) Format() string {
...
+		if i.Key == "DisplayName" {
+			displayName = i.Value.String
+		}
+		if i.Key == "DisplayVersion" {
+			displayVersion = i.Value.String
+		}
+		if i.Key == "Publisher" {
+			publisher = i.Value.String
+		}
```
afiune: nit: would a `switch` statement be better here?

providers/os/resources/packages/windows_packages.go
```diff
@@ -203,15 +214,90 @@ func (w *WinPkgManager) Format() string {
...
+
+	if uninstallString == "" {
+		return nil
+	}
+
+	cpeWfn := ""
+	if displayName != "" && displayVersion != "" {
+		cpeWfn, err = cpe.NewPackage2Cpe(publisher, displayName, displayVersion, "", "")
+		if err != nil {
+			log.Debug().Err(err).Str("name", displayName).Str("version", displayVersion).Msg("could not create cpe for windows app package")
```
afiune, line 294: is this CPE optional? or why we don't error out here? 🤔

providers/os/resources/packages/windows_packages.go
```diff
@@ -203,15 +214,90 @@ func (w *WinPkgManager) Format() string {
...
+	}
+	return packages, nil
+}
+
+func (w *WinPkgManager) getInstalledApps() ([]Package, error) {
+	if w.conn.Type() == shared.Type_Local {
+		return w.getLocalInstalledApps()
+	}
+
+	cmd, err := w.conn.RunCommand(powershell.Wrap(installedAppsScript))
```
afiune: Just for my own knowledge, whey we switched the use of `Encode()` to `Wrap()` here?

providers/os/resources/packages/windows_packages.go
```diff
@@ -245,19 +329,22 @@ func (w *WinPkgManager) List() ([]Package, error) {
...
-	hotfixes, err := ParseWindowsHotfixes(cmd.Stdout)
-	if err != nil {
-		return nil, errors.Wrapf(err, "could not parse hotfix results")
-	}
-	hotfixAsPkgs := HotFixesToPackages(hotfixes)
-
-	pkgs = append(pkgs, hotfixAsPkgs...)
+	if !canRunCmd {
+		log.Debug().Msg("cannot run command on windows, skipping hotfix list")
+	} else {
```
afiune: nit; if you return in the above `if`, you can avoid this `else`

(I think you know I like to reduce indentation to the limits 😂, up to you to agree or disagree 😅  )

afiune (approved):
A few comments but looks pretty nice

providers/os/resources/packages/windows_packages.go
preslavgerchev: I believe it is, yeah. Ive copied over the last implementation of how we fetch that.

providers/os/resources/packages/windows_packages.go
preslavgerchev: that sounds like a good idea

providers/os/resources/packages/windows_packages.go
preslavgerchev: i will revert, unintentional

providers/os/resources/packages/windows_packages.go
preslavgerchev: yep, actually i need to only check this once and return if false

providers/os/resources/packages/windows_packages.go
preslavgerchev: i dont think this is exactly what you meant here but: i pulled out the registry key -> package transformation into a separate function that is easily testable. we now fetch the children of the registry key and we pass them in so that the `packageFromRegistryKey` doesnt do any calls on its own. added tests too

providers/os/resources/packages/windows_packages.go
chris-rock: Perfect, thank you

PR #4205, opened 2024-06-06
✨ Fetch registry packages for Windows filesystem connections

preslavgerchev:
After #4204 

```
PS C:\Users\mondoo> ./cnquery run fs --path F:\ -c 'packages'
Flag --path has been deprecated, has been deprecated
Flag --path has been deprecated, has been deprecated
WRN using builtin provider for os
WRN no path provided as an arg, looking for --path flag
WRN using builtin provider for os
- no Mondoo configuration file provided, using defaults
WRN using builtin provider for os
packages.list: [
  0: package name="Microsoft Edge" version="124.0.2478.97"
]
```

### DIFF
--- a/providers/os/resources/packages/windows_packages.go
+++ b/providers/os/resources/packages/windows_packages.go
@@ -224,20 +224,15 @@ func (w *WinPkgManager) getLocalInstalledApps() ([]Package, error) {
 	}
 	packages := []Package{}
 	for _, r := range pkgs {
-		// we get all the packages, found under the pkgs paths
 		children, err := registry.GetNativeRegistryKeyChildren(r)
 		if err != nil {
 			continue
 		}
 		for _, c := range children {
-			// for each package the information is contained as items of that registry's key,
-			// so we request the items under the fully qualified path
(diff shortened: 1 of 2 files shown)


providers/os/resources/packages/windows_packages.go
```diff
@@ -245,13 +245,58 @@ func (w *WinPkgManager) getInstalledApps() ([]Package, error) {
...
+	packages := []Package{}
+	for _, r := range pkgs {
+		// we get all the packages, found under the pkgs paths
+		children, err := rh.GetNativeRegistryKeyChildren(registry.Software, r)
+		if err != nil {
+			continue
+		}
+		for _, c := range children {
+			// for each package the information is contained as children of that registry's key
+			p := w.getPackageFromRegistryKey(c)
```
preslavgerchev: use the same logic as the `localInstalledApps`, only difference is we let the registry handler tell us where the keys are mounted via `GetNativeRegistryKeyChildren`

afiune (approved):

PR #4207, opened 2024-06-07
✨ Added aws account tags query

Pauti:


providers/aws/resources/aws.lr.manifest.yaml
```diff
@@ -83,6 +83,8 @@ resources:
       id: {}
       organization:
         min_mondoo_version: 6.11.0
+      tags:
+        min_mondoo_version: latest
```
```suggestion
        min_mondoo_version: 9.0.0
```

providers/aws/resources/aws.lr
```diff
@@ -23,6 +23,8 @@ aws.account @defaults("id") {
   aliases() []string
   // Information about the associated organization, if any
   organization() aws.organization
+  // Tags on the Account
```
tas50:
```suggestion
  // Tags on the account
```

providers/aws/resources/aws_account.go
```diff
@@ -75,6 +75,35 @@ func (a *mqlAwsOrganization) accounts() ([]interface{}, error) {
...
+	conn := c.MqlRuntime.Connection.(*connection.AwsConnection)
+	client := conn.Organizations("") // no region for orgs, use configured region
+
+	input := &organizations.ListTagsForResourceInput{
+		ResourceId: &c.Id.Data,
+	}
+
+	tags := make(map[string]interface{})
+	for {
+		res, err := client.ListTagsForResource(context.TODO(), input)
```
vjeffrey, line 92: could you please add a small comment here and also maybe a small comment in the lr regarding permissions? The API states this is only accessible to management accounts and delegated admins, so many users will encounter a perm denied on this resource field (which is fine, but we should explain it)

from the API:
```
// This operation can be called only from the organization's management account or
// by a member account that is a delegated administrator for an Amazon Web Services
// service.
```

vjeffrey:
@Pauti thank you for contributing this change! looks great!

everything looks great, i'll get this merged in once you add in the permissions comment, thanks again!

vjeffrey (approved):

vjeffrey:
looks like there are conflicts with man, once those are resolved we can get this merged in

Pauti:
I resolved the conflicts and kept the newer versions.

PR #4208, opened 2024-06-07
✨ Add basic support for scanning windows disk drives via device conn.

preslavgerchev:
Add support for identifying the drive letter where the windows partition is located by only specifying the LUN (logical unit number) or the serial number of the disk.

```
./cnquery.exe shell device --serial-number 123455
```

```
./cnquery.exe shell device --lun 2
```

There's also support for setting a disk online if it's found to be offline. This only happens if we that this is the target disk that we're searching for. Once scanning is done, the disk is set back to offline.

### DIFF
--- a/providers/os/connection/device/windows/device_manager.go
+++ b/providers/os/connection/device/windows/device_manager.go
@@ -17,10 +17,12 @@ const (
 )
 
 type WindowsDeviceManager struct {
-	cmdRunner       *snapshot.LocalCommandRunner
-	opts            map[string]string
+	cmdRunner *snapshot.LocalCommandRunner
+	opts      map[string]string
+	// indicates if the disk we've targeted has been set to online. we use this to know if we need to put it back offline once we're done
 	diskSetToOnline bool
-	diskIndex       int
+	// if we've set the disk online, we need to know the index to set it back offline
+	diskIndex int
 }
 
 func NewWindowsDeviceManager(shell []string, opts map[string]string) (*WindowsDeviceManager, error) {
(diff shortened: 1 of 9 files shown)


afiune (commented):
About to test this!

providers/os/connection/device/windows/device_manager.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	partitionInfo := &snapshot.PartitionInfo{
+		Name:   partition.DriveLetter,
+		FsType: "Windows",
+	}
+	return []*snapshot.PartitionInfo{partitionInfo}, nil
+}
+
+// validates the options provided to the device manager
+func validateOpts(opts map[string]string) error {
```
afiune, line 82: can we test this? What if these options doesn't exist inside the `map[]`?

providers/os/connection/device/windows/disk_drive.go
```diff
@@ -0,0 +1,198 @@
...
+		if err != nil {
+			return nil, err
+		}
+		return []*diskPartition{partition}, nil
+	}
+
+	return partitions, nil
+}
+
+func filterDiskDrives(drives []*diskDrive, opts map[string]string) (*diskDrive, error) {
```
afiune, line 156: same here, this can be unit tested.

afiune (approved):

providers/os/connection/device/windows/device_manager.go
preslavgerchev: great point, tests aded

PR #4211, opened 2024-06-10
🧹 add k8s services to discovery

vjeffrey:
Screenshot 2024-06-10 at 16 16 52

### DIFF
--- a/providers/k8s/resources/discovery.go
+++ b/providers/k8s/resources/discovery.go
@@ -40,6 +40,7 @@ const (
 	DiscoveryAdmissionReviews = "admissionreviews"
 	DiscoveryIngresses        = "ingresses"
 	DiscoveryNamespaces       = "namespaces"
+	DiscoveryServices         = "services"
 )
 
 type NamespaceFilterOpts struct {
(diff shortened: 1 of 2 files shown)


providers/k8s/resources/discovery.go
```diff
@@ -933,6 +996,10 @@ func createPlatformData(objectKind, runtime string) (*inventory.Platform, error)
 		platformData.Family = append(platformData.Family, "k8s-workload")
 		platformData.Name = "k8s-cronjob"
 		platformData.Title = "Kubernetes CronJob"
+	case "Service":
+		platformData.Family = append(platformData.Family, "k8s-workload")
```
czunker: A `service` is not really a workload. Perhaps we need a new family for this.

czunker: @imilchev Any ideas? Or should we go with `k8s-service` as its own family, like in `k8s-ingress`.

czunker:
Gave this a try. LGTM:
```
cnquery run k8s --namespaces "mondoo" -c "asset{ name platform }" 
...
asset: {
  platform: "k8s-service"
  name: "mondoo/service1"
}
asset: {
  platform: "k8s-service"
  name: "mondoo/service2"
}
...
```

providers/k8s/resources/discovery.go
imilchev: I think you are right, this should be `k8s-service`

imilchev (approved):

PR #4217, opened 2024-06-10
Fix docs generation for Snowflake

tas50:
Command typo here caused a failure


misterpantz (approved):
In case my approval is enough... thank yoU!

PR #4231, opened 2024-06-13
Update the platform name for Azure SQL Database servers

tas50:
See https://en.wikipedia.org/wiki/Microsoft_Azure_SQL_Database


czunker (approved):
Thanks @tas50

PR #4236, opened 2024-06-14
⭐️ allow finding files by permissions

imilchev:
When we search for riles with `RunCommand` capability, we use `find`. There we can provide permissions filters. In the cases, where we have no `RunCommand` capability we are manually looping over the filesystem files. 

The manual loop we have doesn't check for permissions matching. This is problematic for some queries that we have in our policies. For example:
```
files.find(from: "/", type: "file", xdev: false, permissions: 0002).all(path == "")
files.find(from: "/", type: "directory", xdev: false, permissions: 0002).all(permissions.sticky == true)
```
If we try to execute these queries in a filesystem scan (where no `RunCommand` capability is present), we will be adding all the files on the filesystem as an mql resource. This is very CPU and memory intensive and we have already encountered issues with that. This PR makes sure we match permissions when running a manual find. That significantly improves the performance for the query on my test system.

The easiest way to test this is to run `cnspec shell fs --path /` on a Linux machine. In the shell, try the first line of the query:
```
files.find(from: "/", type: "file", xdev: false, permissions: 0002).all(path == "")
```

### DIFF
--- a/providers/os/fs/find_files.go
+++ b/providers/os/fs/find_files.go
@@ -9,6 +9,8 @@ import (
 	"os"
 	"regexp"
 	"strings"
+
+	"golang.org/x/sys/unix"
 )
 
 func FindFiles(iofs fs.FS, from string, r *regexp.Regexp, typ string, perm *uint32) ([]string, error) {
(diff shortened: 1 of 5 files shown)


providers/os/fs/find_files.go
```diff
@@ -11,17 +11,26 @@ import (
...
-		if err != nil {
-			if errors.Is(err, os.ErrPermission) {
-				return nil
-			}
+		if err := handleFsError(err); err != nil {
 			return err
 		}
 		if matcher.Match(p, d.Type()) {
+			// If we have a permission filter, match on permissions too
+			if perm != nil {
```
czunker: Shouldn't this be part of the matcher?

czunker:
Performance is looking better:
Screenshot from 2024-06-14 11-12-58

But I still get one error:
```
[failed] files.find.all()
  error: readdirent /proc/1506537/net: invalid argument
```

Perhaps we can also skip this? The directory exists, I get the same error when I try to access it without cnquery:
```
sudo ls -l /proc/1506537/net 
ls: reading directory '/proc/1506537/net': Invalid argument
total 0
```

czunker (approved):

PR #4251, opened 2024-06-17
🐛 fix gcp subnetwork navigation from gke resources

### LINKED ISSUE #3871: Type error fetching subnetwork of GKE cluster
**Describe the bug**
cnquery shows a type error instead of the subnetwork value in the GKE resource

```
cnquery shell gcp project 253146780735

cnquery> gcp.project.gke.clusters.first{*}.networkConfig
1 error occurred:
	* [gcp] cannot set 'region' in resource 'gcp.project.computeService.subnetwork', type does not match
gcp.project.gke.clusters.first.networkConfig: {
  dnsConfig: {
    clusterDns: "CLOUD_DNS"
    clusterDnsDomain: "cluster.local"
    clusterDnsScope: "CLUSTER_SCOPE"
  }
  defaultSnatStatus: {
    disabled: false
  }
  serviceExternalIpsConfig: {
    enabled: false
  }
  network: gcp.project.computeService.network name="default"
  networkPath: "projects/1234/global/networks/default"
  privateIpv6GoogleAccess: "PRIVATE_IPV6_GOOGLE_ACCESS_UNSPECIFIED"
  id: "gcp.project.gkeService.cluster/1234/networkConfig"
  enableL4IlbSubsetting: false
  datapathProvider: "ADVANCED_DATAPATH"
  subnetwork: [gcp] cannot set 'region' in resource 'gcp.project.computeService.subnetwork', type does not match
  subnetworkPath: "projects/1234/regions/us-west4/subnetworks/default"
  enableIntraNodeVisibility: true
}
cnquery>
```

imilchev:
Fixes #3871

### DIFF
--- a/providers/gcp/resources/compute.go
+++ b/providers/gcp/resources/compute.go
@@ -42,32 +42,34 @@ func initGcpProjectComputeService(runtime *plugin.Runtime, args map[string]*llx.
 	return args, nil, nil
 }
 
-type mqlGcpProjectComputeServiceInternal struct {
-	serviceEnabled bool
-}
-
 func (g *mqlGcpProject) compute() (*mqlGcpProjectComputeService, error) {
 	if g.Id.Error != nil {
 		return nil, g.Id.Error
 	}
 	projectId := g.Id.Data
 
-	res, err := CreateResource(g.MqlRuntime, "gcp.project.computeService", map[string]*llx.RawData{
+	res, err := NewResource(g.MqlRuntime, "gcp.project.computeService", map[string]*llx.RawData{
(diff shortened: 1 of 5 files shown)


czunker (commented):
The type mismatch is fixed.

But now, I get another error:
```
1 error occurred:
        * subnetwork not found
```

I tried this with our edge cluster, and there definitely is a subnet. I can see it in the Google console.

imilchev:
When running the GCP provider we need to use the project ID, not the project number as that is not working

czunker (approved):
When using the project ID, it works.

PR #4253, opened 2024-06-17
⭐ allow inventory-provided identity option to load base aws credentials for assume role

vjeffrey:
allow users to provide an initial identity for the AWS config by providing an identity as part of the inventory

this is useful in cases where role-based connection is desired and there are no local credentials present

<example coming soon>

### DIFF
--- a/providers/aws/connection/awsutil.go
+++ b/providers/aws/connection/awsutil.go
@@ -1,0 +1,188 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package connection
+
+import (
+	"context"
+	"errors"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"golang.org/x/oauth2"
+	"google.golang.org/api/idtoken"
+)
+
+type TokenSourceProvider func(ctx context.Context) (oauth2.TokenSource, error)
+
+func AwsWithIdentityProvider(opts map[string]string) (*aws.Config, error) {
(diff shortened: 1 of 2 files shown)


providers/aws/connection/awsutil.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"golang.org/x/oauth2"
+	"google.golang.org/api/idtoken"
+)
+
+type TokenSourceProvider func(ctx context.Context) (oauth2.TokenSource, error)
+
+func AwsWithIdentityProvider(opts map[string]string) (*aws.Config, error) {
```
preslavgerchev, line 20: this func is called `AwsWithIdentityProvider` but internally it calls the `GcpInstanceIdentity`. Can we rename it to make it clear this only works with GCP?

providers/aws/connection/connection.go
```diff
@@ -254,6 +255,25 @@ func WithAssumeRole(defaultCfg aws.Config, roleArn string, externalId string) Co
 	}
 }
 
+const (
```
preslavgerchev, line 258: nit: can you define those at the top? makes searching for them easier

providers/aws/connection/connection.go
```diff
@@ -254,6 +255,25 @@ func WithAssumeRole(defaultCfg aws.Config, roleArn string, externalId string) Co
...
+	GcpInstanceIdentityKey string = "gcp-audience"
+	IdentityRoleArnKey     string = "identity-role-arn"
+)
+
+func LoadAwsConfig(c *AwsConnection, opts map[string]string) (aws.Config, error) {
+	// if an identity is provided, that means we need to load that identity config first
+	// to establish base credentials with aws, and then go on to load the aws config
+	// with the given options like normal
+	if opts[GcpInstanceIdentityKey] != "" {
+		_, err := AwsWithIdentityProvider(opts)
```
preslavgerchev, line 268: why is the return value here ignored?

PR #4254, opened 2024-06-17
🧹 ensure we always read the fine grained assets FF for AWS

vjeffrey:
I'm not sure why I needed the second commit/why that test failed..


czunker:
> I'm not sure why I needed the second commit/why that test failed..

Preslav looked into it: https://github.com/mondoohq/cnquery/pull/4260

The checksum failed because of changed defaults.

czunker (approved):

PR #4255, opened 2024-06-18
🐛 Assign epoch in RPM static package manager.

preslavgerchev:
This helps us do better vulnerability assessment when performing static analysis of RPM packages.

Thank you @czunker for the help with this one


czunker (approved):
Thanks @preslavgerchev 

I tried this with an `registry.access.redhat.com/ubi8/ubi:8.6` image.

PR #4225, opened 2024-06-12
Improve defaults for Azure SQL resources + fix some branding

tas50:
```coffee
cnquery> azure.subscription.sql.servers.first
azure.subscription.sql.servers.first: azure.subscription.sqlService.server name="timtimtimtim" location="eastus" properties.version="12.0" properties.state="Ready" properties.fullyQualifiedDomainName="timtimtimtim.database.windows.net"

cnquery> azure.subscription.sql.servers.first.databases
azure.subscription.sql.servers.first.databases: [
  0: azure.subscription.sqlService.database name="master" edition="System" zoneRedundant=false status="Online" creationDate=2024-06-12 12:20:08.777 -0700 PDT maxSizeBytes=53687091200 earliestRestoreDate=null
]
```

PR #4260, opened 2024-06-18
🧹 Fix mqlc checksum test

preslavgerchev:
#4225  added new defaults for the resource that this test uses


czunker (approved):
Thanks @preslavgerchev

PR #4270, opened 2024-06-19
🏃 Improve azure provider performance

preslavgerchev:
* Use parent connections to take advantage of MQL cache.
* Ensure we do not discover subs multiple times, that's redundant


czunker (approved):
:rocket: 

Thanks @preslavgerchev

PR #4278, opened 2024-06-20
🐛 Safeguard GCP connection in gcp init resources.

preslavgerchev:


providers/gcp/resources/common.go
```diff
@@ -41,7 +41,10 @@ type assetIdentifier struct {
 }
 
 func getAssetIdentifier(runtime *plugin.Runtime) *assetIdentifier {
-	conn := runtime.Connection.(*connection.GcpConnection)
+	conn, ok := runtime.Connection.(*connection.GcpConnection)
+	if !ok {
+		return nil
```
czunker, line 46: Should we extend this func to also return an `error`?

czunker (commented):
Just curious: What happened that we need this now?

preslavgerchev:
> Just curious: What happened that we need this now?

Using a GCP snapshot connection right now is a bit weird: We use the GCP provider and we return a file system connection. There are some query filters that do query GCP resources, e.g. `gcp.organization.labels` (this is a totally made up example). Since we're coming from an GCP provider, this query can be compiled so it will be tried. However, the connection underneath is not a `GcpConnection` so we get the panic. Long term I believe we want to fix this differently but that's a good fix for the time being

providers/gcp/resources/common.go
preslavgerchev: There are already code paths in the same func that return nil (when no matching platform exists) so I assume returning nil here would be ok too

czunker:
> > Just curious: What happened that we need this now?
> 
> Using a GCP/Azure/AWS snapshot connection right now is a bit weird: We use the GCP provider and we return a file system connection. There are some query filters that do query GCP resources, e.g. `gcp.organization.labels` (this is a totally made up example). Since we're coming from an GCP provider, this query can be compiled so it will be tried. However, the connection underneath is not a `GcpConnection` so we get the panic. Long term I believe we want to fix this differently but that's a good fix for the time being

Thanks for the explanation. That's really wired.

czunker (approved):

PR #4286, opened 2024-06-21
✨ extend google workspace apps resource with admin data

mm-weber:


providers/google-workspace/resources/google-workspace.lr
```diff
@@ -175,6 +175,7 @@ private googleworkspace.role @defaults("name") {
 // Google Workspace apps reports
 private googleworkspace.report.apps {
   drive() []googleworkspace.report.activity
```
tas50, line 178: Both of these need a description so they get documented properly

mm-weber: Thanks @tas50 ! Fixed it.

imilchev (approved):

PR #4291, opened 2024-06-24
🐛 fix readdirnames for docker scans

imilchev:
Fixes #8322

To trigger the issue use the following Dockerfile:
```Dockerfile
FROM wordpress

RUN touch /root/known_host
RUN mkdir /var/www/.ssh
RUN ln -s /root/known_hosts /var/www/.ssh/known_hosts
```

Or just pull the image from my repo: `ivanmilchev/test:loop` (I provided amd64 and arm64 images)

1. Run the image: `docker run -d ivanmilchev/test:loop`
2. Get the running container ID with `docker ps`
3. Run a container scan for the running container `cnspec scan docker <<CONTAINER_ID>>` (it must be cnspec. The issue isn't present in cnquery)

With the updated provider from this PR, the scan should finish successfully

### DIFF
--- a/providers/os/connection/docker/file.go
+++ b/providers/os/connection/docker/file.go
@@ -106,9 +106,12 @@ func (f *File) Readdirnames(n int) ([]string, error) {
 	}
 
 	// extract names
-	basenames := make([]string, len(directories))
-	for i := range directories {
-		basenames[i] = filepath.Base(directories[i])
+	basenames := []string{}
+	for _, dir := range directories {
+		if dir == "" {
+			continue
+		}
+		basenames = append(basenames, filepath.Base(dir))
 	}
 	return basenames, nil
 }


providers/os/connection/docker/file.go
```diff
@@ -106,8 +106,11 @@ func (f *File) Readdirnames(n int) ([]string, error) {
 	}
 
 	// extract names
-	basenames := make([]string, len(directories))
+	basenames := []string{}
 	for i := range directories {
+		if directories[i] == "" {
```
imilchev: The issue is that if there is that when listing directories and parsing the output of `find`, the last string in `directories` can be empty string. If we provide an empty string to `filepath.Base` it returns `.`, which creates a loop

providers/os/connection/docker/file.go
```diff
@@ -106,8 +106,11 @@ func (f *File) Readdirnames(n int) ([]string, error) {
...
+			continue
+		}
 		basenames[i] = filepath.Base(directories[i])
```
czunker: Doesn't this leave the array with empty entries (`""`)?

Let's assume we skip `i == 2`. Then, the `basenames` array wouldn't have an entry there.
Perhaps it works, because we also do not call `filepath.Base` on this...

czunker (commented):
The fix itself works. Perhaps just a minor uncertainty left.

providers/os/connection/docker/file.go
imilchev: The array isn’t initialized with the size of the directories. That’s why it doesn’t contain empty entries

czunker: Something is wrong here. I'm not sure why this fixes the issue, but this doesn't work.

I extracted the loop:
https://play.golang.com/p/zVgALhKWlWU
And I get a panic:
```
panic: runtime error: index out of range [0] with length 0
```
The array isn't initialized. It works with `append`. But with the example image, this means that `basenames` is empty after the loop. I'm not sure this is what we expect.

czunker (approved):
Thanks @imilchev

PR #4293, opened 2024-06-24
🧹 add a test for ec2 instance labels

vjeffrey:
just some cleanup - make the labels constants and add a small test

### DIFF
--- a/providers/aws/resources/discovery_conversion.go
+++ b/providers/aws/resources/discovery_conversion.go
@@ -6,6 +6,7 @@ package resources
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	ssmtypes "github.com/aws/aws-sdk-go-v2/service/ssm/types"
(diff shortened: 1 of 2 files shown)


providers/aws/resources/discovery_conversion.go
```diff
@@ -247,14 +281,17 @@ func addConnectionInfoToEc2Asset(instance *mqlAwsEc2Instance, accountId string,
...
+		platformDetails: instance.PlatformDetails.Data,
+		instanceType:    instance.InstanceType.Data,
+		accountId:       accountId,
+		instanceId:      instance.InstanceId.Data,
+		launchTime:      instance.LaunchTime.Data,
+	}, asset)
 
 	if instance.GetImage().Data != nil {
-		asset.Labels["mondoo.com/image"] = instance.GetImage().Data.Id.Data
+		asset.Labels[MondooImageLabelKey] = instance.GetImage().Data.Id.Data
```
czunker: Could we also add this to `addMondooLabels` to ave all in one place?

`LaunchTime` also has a `nil` check inside `addMondooLabels`. I guess this would be the same?

providers/aws/resources/discovery_test.go
```diff
@@ -136,3 +137,26 @@ func TestFilters(t *testing.T) {
 	}))
 	require.False(t, shouldScanEcsContainerInstances(connection.DiscoveryFilters{}))
 }
+
+func TestAddConnInfoToEc2Instances(t *testing.T) {
```
czunker, line 142: :tada: Thanks for the tests.

czunker (approved):
Thansk @vjeffrey

PR #4296, opened 2024-06-26
Skip protection rules retrieval for unprotected branches

### LINKED ISSUE #3905: Update to get GitHub branch protection w/o being an admin
**Describe the bug**
The latest SDK v61 updated in #3904 allows us to get branch detection information w/o an API call and without being an admin. We should flip to that so we can more reliably get this information and avoid the API call.

See https://github.com/google/go-github/pull/3095

slntopp:
Resolves #3905 

---

Leaving Protection Rules retrieval as is, since Github API doesn't seem to return full information about Protection rules in GetBranch and just stub instead.
attempted with the tokens (and multiple orgs):
1. Fine-Grained (administration read-only)
2. Fine-Grained (everything read-only)
3. Classic (admin read-only)

Example:
```json
* * *
  "protected": true,
  "protection": {
    "enabled": true,
    "required_status_checks": {
      "enforcement_level": "off",
      "contexts": [

      ],
      "checks": [

      ]
    }
  },
  "protection_url": "https://api.github.com/repos/mondoohq/cnquery/branches/main/protection"
}
```

### DIFF
--- a/providers/github/resources/github_repo.go
+++ b/providers/github/resources/github_repo.go
@@ -453,6 +453,12 @@ func (g *mqlGithubBranch) protectionRules() (*mqlGithubBranchprotection, error) 
 	conn := g.MqlRuntime.Connection.(*connection.GithubConnection)
 	var err error
 
+	// if the branch is not protected, we don't need to fetch the protection rules
+	if !g.IsProtected.Data {
+		g.ProtectionRules.State = plugin.StateIsSet | plugin.StateIsNull
+		return nil, nil
+	}
+
 	if g.RepoName.Error != nil {
 		return nil, g.RepoName.Error
 	}


czunker (commented):
Nice, this will save us some of the API calls.

I think, we can extend this PR to not need admin privileges for the protection rules.

This comment explains why we do not get all the data:
https://github.com/google/go-github/pull/3095/files#diff-7a0544e44d81e972917728b13efd3fa55e7f9d1bd8edb745216494fa1e178f93R1007-R1012
We fetch the branches via `ListBranches`.

By replacing the `GetBranchProtection` call https://github.com/mondoohq/cnquery/pull/4296/files#diff-29c8fa98ab76550d36b0bc88beda748f5930e3c94ad5feb88f94ef24c6f5a050R484
with a `GetBranch` call we should get the data w/o admin.

czunker:
@slntopp and I verified that the `GetBranch` API call is not reporting all the data that the `GetBranchProtection` call returns.

czunker (approved):
Thanks @slntopp

PR #4298, opened 2024-06-27
:octocat: handle GitHub API rate limits

slntopp:
based on: https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit

### DIFF
--- a/providers/github/connection/connection.go
+++ b/providers/github/connection/connection.go
@@ -9,10 +9,14 @@ import (
 	"net/url"
 	"os"
 	"strconv"
+	"time"
 
 	"github.com/bradleyfalzon/ghinstallation/v2"
 	"github.com/cockroachdb/errors"
 	"github.com/google/go-github/v62/github"
+	"github.com/hashicorp/go-retryablehttp"
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
 	"go.mondoo.com/cnquery/v11/providers-sdk/v1/inventory"
 	"go.mondoo.com/cnquery/v11/providers-sdk/v1/plugin"
 	"go.mondoo.com/cnquery/v11/providers-sdk/v1/vault"
@@ -131,7 +135,7 @@ func newGithubAppClient(conf *inventory.Config) (*github.Client, error) {
 		return nil, errors.New("app-private-key is required for GitHub App authentication")
 	}
 
-	return github.NewClient(&http.Client{Transport: itr}), nil
+	return github.NewClient(newGithubRetryableClient(&http.Client{Transport: itr})), nil
 }
 
 func newGithubTokenClient(conf *inventory.Config) (*github.Client, error) {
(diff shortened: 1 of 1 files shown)


providers/github/connection/connection.go
```diff
@@ -161,5 +165,73 @@ func newGithubTokenClient(conf *inventory.Config) (*github.Client, error) {
...
+
+	if httpClient == nil {
+		httpClient = http.DefaultClient
+	}
+	retryClient.HTTPClient = httpClient
+
+	retryClient.CheckRetry = func(ctx context.Context, resp *http.Response, err error) (bool, error) {
+		// Default Retry Policy would not retry on 403 (adding 429 for good measure)
+		if resp.StatusCode == 403 || resp.StatusCode == 429 {
+			// Primary and Seconday rate limit
```
czunker: As I understood the linked docs, there may be another case for the secondary limit.
The `retry-after` header might be set before we get `x-ratelimit-remaining` with zero.

I don't know why there are two steps for the secondary limit, but I read it we should take the other header into account:
> If the retry-after response header is present, you should not retry your request until after that many seconds has elapsed.

providers/github/connection/connection.go
```diff
@@ -161,5 +165,73 @@ func newGithubTokenClient(conf *inventory.Config) (*github.Client, error) {
...
+		}
+		return retryablehttp.DefaultRetryPolicy(ctx, resp, err)
+	}
+	retryClient.Backoff = func(min, max time.Duration, attemptNum int, resp *http.Response) time.Duration {
+		if resp.StatusCode == 403 || resp.StatusCode == 429 {
+			// Primary and Seconday rate limit
+			if resp.Header.Get("x-ratelimit-remaining") == "0" {
+				unix, err := strconv.ParseInt(resp.Header.Get("x-ratelimit-reset"), 10, 64) // x-ratelimit-reset	- The time at which the current rate limit window resets, in UTC epoch seconds
+				if err != nil {                                                             // Must be impossible to hit errors here, but just in case
+					return time.Second * 8 // Github starts with 8s jailtime
```
czunker: Where did you find this? I only found a one minute pause on the secondary limit.

Do we need to handle primary and secondary separately?

czunker (commented):
Nice!

Just some minor things on the secondary limit.

providers/github/connection/connection.go
slntopp: I understood it differently tbh, that in addition they send that one for compatibility.
But it might be a good idea to explicitly handle it, I'll update (otherwise it should be caught by Default Policy)

providers/github/connection/connection.go
slntopp: Couldn't find at this point.
What would you suggest for error handling?

I don't think so, no, it'll be reflected in wait time. But, I think we could add the maximum wait time and throw error after that, because otherwise the scan can be stuck up to an hour and provide no data

czunker: Ok, I understood your comment in a way that it is some default GitHub setting.

I like the idea with the max wait time.

czunker (approved):

PR #4300, opened 2024-06-28
🐛 fix aws iam users attachedpolicies

vjeffrey:
fixes https://github.com/mondoohq/cnquery/issues/3206

### DIFF
--- a/providers/aws/resources/aws_iam.go
+++ b/providers/aws/resources/aws_iam.go
@@ -856,7 +856,7 @@ func (a *mqlAwsIamUser) attachedPolicies() ([]interface{}, error) {
 			mqlAwsIamPolicy, err := CreateResource(a.MqlRuntime, "aws.iam.policy",
 				// this isn't correct for the id, this is the name. we need to remove "id" from the policy type.
 				// it is creating a conflict bc we cannot make it optional, as it then bumps up against the internal id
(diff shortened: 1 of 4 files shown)


providers/aws/resources/aws_iam.go
```diff
@@ -852,10 +853,10 @@ func (a *mqlAwsIamUser) attachedPolicies() ([]interface{}, error) {
 
 		for i := range userAttachedPolicies.AttachedPolicies {
 			attachedPolicy := userAttachedPolicies.AttachedPolicies[i]
-
 			mqlAwsIamPolicy, err := CreateResource(a.MqlRuntime, "aws.iam.policy",
-				map[string]*llx.RawData{"arn": llx.StringDataPtr(attachedPolicy.PolicyArn)},
-			)
+				// this isn't correct for the id, this is the name. we need to remove "id" from the policy type.
+				// it is creating a conflict bc we cannot make it optional, as it then bumps up against the internal id
+				map[string]*llx.RawData{"arn": llx.StringDataPtr(attachedPolicy.PolicyArn), "id": llx.StringDataPtr(attachedPolicy.PolicyName)})
```
vjeffrey: is anyone has ideas on how to better handle this, I'm all ears!

czunker: Just an idea, I'm not sure what happens:
What if you set the "id" field to `PolicyArn`. Then it would at least be the same data as what is returned by `id()`.

vjeffrey: 👍  I can do that

czunker (approved):

PR #4303, opened 2024-07-01
⭐ github.packages resource

slntopp:


providers/github/resources/github_package.go
```diff
@@ -12,3 +22,96 @@ func (g *mqlGithubPackage) id() (string, error) {
...
+func (g *mqlGithubPackages) id() (string, error) {
+	return "github.packages", nil
+}
+
+func (g *mqlGithubPackages) list() ([]any, error) {
+	return newMqlGithubPackages(g.MqlRuntime, nil)
+}
+
+func (g *mqlGithubPackages) public() ([]any, error) {
+	visibility := "public"
```
czunker: nit: Perhaps use some `const` for different visibilities?

providers/github/resources/github_package.go
```diff
@@ -12,3 +22,96 @@ func (g *mqlGithubPackage) id() (string, error) {
...
+}
+
+func newMqlGithubPackages(runtime *plugin.Runtime, visibility *string) ([]any, error) {
+	conn := runtime.Connection.(*connection.GithubConnection)
+	orgId, err := conn.Organization()
+	if err != nil {
+		return nil, err
+	}
+
+	pkgTypes := []string{"npm", "maven", "rubygems", "docker", "nuget", "container"}
```
czunker, line 62: Is this what we are interested in or what GitHub provides?

providers/github/resources/github_package.go
```diff
@@ -12,3 +22,96 @@ func (g *mqlGithubPackage) id() (string, error) {
...
+	for i := range pkgTypes {
+		listOpts := &github.PackageListOptions{
+			ListOptions: github.ListOptions{PerPage: paginationPerPage},
+			PackageType: github.String(pkgTypes[i]),
+			Visibility:  visibility,
+		}
+
+		var allPackages []*github.Package
+		for {
+			packages, resp, err := conn.Client().Organizations.ListPackages(context.Background(), orgId.Name, listOpts)
```
czunker, line 73: I don't know how this is intended to be used.
But this line makes the output of this command a bit confusing to me:
```
cnquery run github repo mondoohq/packer-plugin-cnspec -c "github.packages"
```
I'd expected only the packages of the specified repo, but I got all across the org. Is this how it should be?

providers/github/resources/github_package.go
slntopp: This is all kinds of packages GitHub provides

providers/github/resources/github_package.go
slntopp: Packages are not linked to repos, packages are always stored under account/org

So scan still would be with `github org some_org`, just it won't fetch the org

providers/github/resources/github_package.go
slntopp: defined `const`s

imilchev (approved):
LGTM

providers/github/resources/github_package.go
czunker: Ah, thanks for the background.

czunker (approved):
Thanks @slntopp

PR #4306, opened 2024-07-01
⭐️ add subnets to rds db instances

vjeffrey:
Screenshot 2024-07-01 at 15 59 54

### DIFF
--- a/providers/aws/resources/aws_rds.go
+++ b/providers/aws/resources/aws_rds.go
@@ -163,8 +163,7 @@ func (a *mqlAwsRdsDbinstance) subnets() ([]interface{}, error) {
 		}
 		return res, nil
 	}
-	a.Subnets.State = plugin.StateIsNull | plugin.StateIsSet
-	return nil, nil
+	return nil, errors.New("no subnets found for RDS DB instance")
 }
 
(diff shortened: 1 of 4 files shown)


providers/aws/resources/aws_rds.go
```diff
@@ -141,6 +143,30 @@ func (a *mqlAwsRds) getDbInstances(conn *connection.AwsConnection) []*jobpool.Jo
...
+			if err != nil {
+				a.Subnets.State = plugin.StateIsNull | plugin.StateIsSet
+				return nil, err
+			}
+			res = append(res, sub)
+		}
+		return res, nil
+	}
+	a.Subnets.State = plugin.StateIsNull | plugin.StateIsSet
+	return nil, nil
```
czunker: I'm wondering whether this should return an error.
I don't know this AWS service from personal use, so I skimmed over the docs. I found this one:
> Your VPC must have a DB subnet group that you create.

Could it be possible we cannot access this data?

vjeffrey: oh, interesting, ill check 👍

vjeffrey: ya, it seems we should always have this info

czunker (approved):

PR #4309, opened 2024-07-02
🐛 Github: access personal repos: 

slntopp:
redundant error return with personal account as the repo owner(not an org)

### DIFF
--- a/providers/github/resources/github_repo.go
+++ b/providers/github/resources/github_repo.go
@@ -155,7 +155,6 @@ func initGithubRepository(runtime *plugin.Runtime, args map[string]*llx.RawData)
 				}
 				user = obj.(*mqlGithubUser)
 			}
-			return nil, nil, err
 		} else {
 			org = obj.(*mqlGithubOrganization)
 		}


providers/github/resources/github_repo.go
```diff
@@ -155,7 +155,6 @@ func initGithubRepository(runtime *plugin.Runtime, args map[string]*llx.RawData)
 				}
 				user = obj.(*mqlGithubUser)
 			}
-			return nil, nil, err
```
vjeffrey, line 158: was this not intended to return the err if the err is not a 404? (like what we would do if we had an else statement there)

slntopp: Yes, but API thows 404 in the case repo owner is a user, hence the handling before this return
There is exactly same return above for the case user isn't found as well

So, if org not found -> attempt to find user -> if user not found -> return error as is

Example for non-existent repo
Screenshot 2024-07-02 at 18 56 52

vjeffrey: thanks for the explanation!

vjeffrey (approved):

PR #4315, opened 2024-07-06
Make sure K8s service discovery is in help

tas50:
This adds it to `scan --help` output under `--discovery`


czunker (approved):
Thanks @tas50

PR #4319, opened 2024-07-08
🧹 mark private resources as private

slntopp:


providers/os/resources/os.lr.manifest.yaml
```diff
@@ -131,6 +131,10 @@ resources:
     refs:
     - title: What is a container?
       url: https://docs.docker.com/guides/walkthroughs/what-is-a-container/
+  docker.exposes:
```
tas50: @slntopp should this be included?

slntopp: good catch, was testing smth, cleaned it up

tas50 (approved):

PR #4325, opened 2024-07-09
✨Run status check on login errors

slntopp:
Resolves https://github.com/mondoohq/cnspec/issues/1342

### DIFF
--- a/apps/cnquery/cmd/login.go
+++ b/apps/cnquery/cmd/login.go
@@ -62,7 +62,12 @@ You remain logged in until you explicitly log out using the 'logout' subcommand.
 		splay, _ := cmd.Flags().GetInt("splay")
 		err := register(token, annotations, timer, splay)
 		if err != nil {
-			defer StatusCmd.RunE(cmd, args)
+			defer func() {
+				err := StatusCmd.RunE(cmd, args)
+				if err != nil {
+					log.Warn().Err(err).Msg("could not run status command")
+				}
+			}()
 		}
 		return err
 	},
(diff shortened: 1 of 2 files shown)


imilchev:
can you fix the linter issue?

slntopp:
@imilchev done

apps/cnquery/cmd/login.go
```diff
@@ -60,7 +60,16 @@ You remain logged in until you explicitly log out using the 'logout' subcommand.
 		annotations, _ := cmd.Flags().GetStringToString("annotation")
 		timer, _ := cmd.Flags().GetInt("timer")
 		splay, _ := cmd.Flags().GetInt("splay")
-		return register(token, annotations, timer, splay)
+		err := register(token, annotations, timer, splay)
+		if err != nil {
+			defer func() {
+				err := StatusCmd.RunE(cmd, args)
```
imilchev: it's a bit weird to run 1 command from another. Can you extract whatever is in the status command into a separate function and just call that function here and in `StatusCmd.Run`. I don't think you need the `args` at all for checking the status

slntopp: done

apps/cnquery/cmd/status.go
```diff
@@ -135,6 +70,81 @@ Status sends a ping to Mondoo Platform to verify the credentials.
...
+
+	s.Client.LatestVersion = latestVersion
+
+	// check valid agent authentication
+	plugins := []ranger.ClientPlugin{}
+
+	// try to load config into credentials struct
+	credentials := opts.GetServiceCredential()
+	if credentials != nil && len(credentials.Mrn) > 0 {
+		s.Client.ParentMrn = credentials.ScopeMrn
```
imilchev: I am not sure if this breaks anything. What if I have an old service account that still has `ParentMrn` instead of `ScopeMrn`. My integration will be broken in that case and I would need to re-create my service account. I suggest not changing this now because we don't know what the impact will be for customers

imilchev (approved):

PR #4348, opened 2024-07-16
✨ Add Azure ID and Azure sub external IDs to the sbom proto.

preslavgerchev:


sbom/sbom.proto
```diff
@@ -118,6 +118,17 @@ enum ExternalIDType {
   // EXTERNAL_ID_TYPE_AWS_ORG denotes the organization to which the AWS Account
   // and all subresources belong.
   EXTERNAL_ID_TYPE_AWS_ORG = 3;
+
+  // EXTERNAL_ID_TYPE_AZURE_SUB represents an Azure subscription identifier. 
+  // This type is used when the external ID refers specifically to an Azure
+  // subscription number, which is a unique identifier for an Azure 
+  // subscription.
+  EXTERNAL_ID_TYPE_AZURE_SUB = 4;
```
vjeffrey, line 126: what about tenant? should that information be included as an external id?

vjeffrey (approved):

PR #4356, opened 2024-07-17
🐛 fix: scoring system json marshal

slntopp:
MarshalJSON was returning values unquoted, i.e. like:

```json
{
  "scoring_system": decayed
}
```


imilchev (approved):

imilchev (approved):
LGTM

PR #4365, opened 2024-07-18
⭐️ Adding the resources for approval settings (Gitlab)

HRouhani:
Screenshot from 2024-07-18 17-53-05

Important:

we need to figure out why I need to run several time gitlab.project {*} until it works.
Maybe something with initialization which I tried to use and still did not work

Screenshot from 2024-07-18 18-11-16

### DIFF
--- a/providers/gitlab/resources/gitlab.go
+++ b/providers/gitlab/resources/gitlab.go
@@ -121,3 +121,30 @@ func initGitlabProject(runtime *plugin.Runtime, args map[string]*llx.RawData) (m
 	args = getGitlabProjectArgs(project)
 	return args, nil, nil
 }
+
+// New function to fetch project approval settings
+func (p *mqlGitlabProject) approvalSettings() ([]interface{}, error) {
+	conn := p.MqlRuntime.Connection.(*connection.GitLabConnection)
+
+	projectID := int(p.Id.Data)
+	approvalConfig, _, err := conn.Client().Projects.GetApprovalConfiguration(projectID)
+	if err != nil {
+		return nil, err
+	}
+
+	approvalSettings := map[string]*llx.RawData{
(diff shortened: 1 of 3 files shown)


providers/gitlab/resources/gitlab.go
```diff
@@ -121,3 +121,30 @@ func initGitlabProject(runtime *plugin.Runtime, args map[string]*llx.RawData) (m
...
+		"mergeRequestsDisableCommittersApproval":    llx.BoolData(approvalConfig.MergeRequestsDisableCommittersApproval),
+		"requirePasswordToApprove":                  llx.BoolData(approvalConfig.RequirePasswordToApprove),
+	}
+
+	mqlApprovalSettings, err := CreateResource(p.MqlRuntime, "gitlab.project.approvalSettings", approvalSettings)
+	if err != nil {
+		return nil, err
+	}
+
+	return []interface{}{mqlApprovalSettings}, nil
```
vjeffrey: it looks like maybe approval settings should be singular? as in `approvalSettings() []gitlab.project.approvalSettings` could be `approvalSettings() gitlab.project.approvalSettings`, since there's only ever one?

if that's true, I think you can change this return to `mqlApprovalSettings.(*mqlGitlabProjectApprovalSettings)` and then set the return of the function to `*mqlGitlabProjectApprovalSettings` after changing `[]gitlab.project.approvalSettings` to `gitlab.project.approvalSettings`

HRouhani: Thanks, that's true, there is only one approval settings for the whole project. Changed it as requested.

vjeffrey (approved):

providers/gitlab/resources/gitlab.go
vjeffrey: nice, thank you!

PR #4367, opened 2024-07-19
⭐️  Adding support to fetch merge methods (Gitlab)

HRouhani:
Screenshot from 2024-07-19 17-36-11

### DIFF
--- a/providers/gitlab/resources/gitlab.lr
+++ b/providers/gitlab/resources/gitlab.lr
@@ -84,6 +84,8 @@ gitlab.project @defaults("fullName visibility webURL") {
   requirementsEnabled bool
   // Approval rules for the project
   approvalRules() []gitlab.project.approvalRule
+  // Merge methods for the project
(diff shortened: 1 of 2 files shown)


vjeffrey (approved):

providers/gitlab/resources/gitlab.lr
```diff
@@ -82,4 +82,12 @@ gitlab.project @defaults("fullName visibility webURL") {
   autoDevopsEnabled bool
 	// Whether the requirements feature is enabled
   requirementsEnabled bool
+  // Merge methods for the project
+  mergeMethod() gitlab.project.mergeMethod
 }
+
+// GitLab project merge method
+gitlab.project.mergeMethod @defaults("method") {
```
vjeffrey: are you expecting to have more fields in this resource in the future? 
if not, `mergeMethod()` could just become a string

HRouhani: You are right, it is better keet it simple. done. Thanks

PR #4387, opened 2024-07-22
Integrations Vault: handle Overwrite secret data on Set

slntopp:


providers-sdk/v1/vault/gcpberglas/berglas.go
```diff
@@ -139,11 +139,12 @@ func (v *Vault) Set(ctx context.Context, cred *vault.Secret) (*vault.SecretID, e
 		return nil, err
 	}
 
-	_, err = c.Create(ctx, &berglas.StorageCreateRequest{
-		Bucket:    v.bucket,
-		Object:    cred.Key,
-		Plaintext: cred.Data,
-		Key:       v.kmsKeyID,
+	_, err = c.Update(ctx, &berglas.StorageUpdateRequest{
```
imilchev, line 142: does that create if the key doesn't exist?

slntopp: that's what documentation suggests
if that's not what API is actually doing, we can rewrite it like aws secrets manager (create -> if AlreadyExists -> Update)

ref: https://pkg.go.dev/github.com/GoogleCloudPlatform/berglas@v1.0.3/pkg/berglas#StorageUpdateRequest

imilchev (approved):

PR #4397, opened 2024-07-22
Add in boilerplate from the latest resource updates

tas50:
These are getting updated every time we build, but weren't added to the repo


HRouhani (approved):
Thanks Tim, totally forgotten about those boilerplate files!

PR #4398, opened 2024-07-23
⭐️ Adding the support to fetch the project members and their Roles (Gitlab)

HRouhani:
Screenshot from 2024-07-23 10-48-51


vjeffrey (approved):
nice, thank you!

PR #4400, opened 2024-07-23
⭐️ Adding the support to fetch the webhooks and the ssl config (Gitlab)

HRouhani:
Screenshot from 2024-07-23 12-58-49


vjeffrey (approved):
⭐️ thank you!

PR #4407, opened 2024-07-25
🧹  Adding timer and retry to fetching Approvalsettings (Gitlab)

HRouhani:
It will solve the issue mentioned here:  https://github.com/mondoohq/cnquery/pull/4365

### DIFF
--- a/providers/gitlab/resources/gitlab.go
+++ b/providers/gitlab/resources/gitlab.go
@@ -7,7 +7,6 @@ import (
 	"encoding/base64"
 	"fmt"
 	"strconv"
-	"time"
 
 	"github.com/xanzy/go-gitlab"
 	"go.mondoo.com/cnquery/v11/llx"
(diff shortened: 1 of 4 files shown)


vjeffrey (changes requested):
sorry I didn't catch this before! this actually isn't a retry problem, it's a naming thing - the private resource is named the exact same thing as the resource you're calling - renaming the resource to `gitlab.project.approvalsettings` from `gitlab.project.approvalSettings` should help with the errors you've seen here

HRouhani:
> sorry I didn't catch this before! this actually isn't a retry problem, it's a naming thing - the private resource is named the exact same thing as the resource you're calling - renaming the resource to `gitlab.project.approvalsettings` from `gitlab.project.approvalSettings` should help with the errors you've seen here

Thanks @vjeffrey  My bad, totally confused. Tested again and all fine now

vjeffrey (approved):
⭐️

PR #4410, opened 2024-07-26
🐛 Ensure ec2 instance name is always set as asset name if present.

preslavgerchev:


chris-rock (approved):
Thank you @preslavgerchev

PR #4414, opened 2024-07-26
Fix confluence shared connection type

tas50:
I'm assuming this is a typo.


preslavgerchev:
nice catch @tas50, i updated the rest of the provider code to ensure we reference this variable everywhere

preslavgerchev (approved):

PR #4431, opened 2024-07-29
✨ focus k8s scanning on top-level workloads

slntopp:


providers/k8s/resources/discovery.go
```diff
@@ -1157,3 +1178,17 @@ func assetName(ns, name string) string {
 	}
 	return ns + "/" + name
 }
+
+func ownerReferencesFilter(refs []metav1.OwnerReference, filter map[string]struct{}) bool {
```
preslavgerchev: can we change the filter to a `[]string` instead? i see no reason why it should be a map. you can even do a `string...` param to allow for something like `ownerReferenceFilters(obj,"DaemonSet","CronJob")`

preslavgerchev: you can then either build the map internally or just use the `stringx.Contains` func which can check if the `ref.Kind` is in that list

slntopp: Map is just golang's idiomatic way to implement Set, which does save unnecessary iterations when searching for a string inside a slice.

imilchev: agree to what you say @slntopp. I think in this case it is more clear and easy to use a slice. We are the only ones using this. In this specific case, it also wouldn't matter if we do more iterations since those slices are going to be very small. I think the code will be easier to read if we use a slice

slntopp: @imilchev @preslavgerchev no problem, done

providers/k8s/resources/discovery.go
```diff
@@ -1170,13 +1170,13 @@ func assetName(ns, name string) string {
 	return ns + "/" + name
 }
 
-func OwnerReferencesFilter(refs []metav1.OwnerReference, filter map[string]struct{}) bool {
+func OwnerReferencesFilter(refs []metav1.OwnerReference, filter ...string) bool {
```
preslavgerchev, line 1173: nit: can those be private funcs?

preslavgerchev (approved):

providers/k8s/resources/discovery.go
preslavgerchev: you can also build the map internally in the function. this was more about readability when looking at the func from the outside. but yes, we shouldn't worry about maps at the current point

providers/k8s/resources/discovery.go
imilchev: we want to re-use them in the mondoo-operator as it needs to do the same filtering

slntopp:
Thanks @preslavgerchev @imilchev :)

PR #4434, opened 2024-07-29
Fix the way existing assets are found for recording

jaym:


arlimus (approved):
Kudos, let's explore the matching of assets in the recordings more

PR #4435, opened 2024-07-29
⭐   Adding support to get the MongoDB and PostgreSQL cosmodb Types (Azure)

HRouhani:
Screenshot from 2024-07-29 18-49-59

Based on the current capabilities of the Azure SDK for Go, it appears that the kind, tags, and specially **properties** attributes are not directly available for Cosmodb type MongoDB and PostgreSQL server groups through the SDK.

### DIFF
--- a/providers/azure/resources/cosmosdb.go
+++ b/providers/azure/resources/cosmosdb.go
@@ -48,41 +48,46 @@ func (a *mqlAzureSubscriptionCosmosDbService) accounts() ([]interface{}, error) 
 	res := []interface{}{}
 
 	// Fetch resources of different types - other than MongoDB and PostgreSQL
-	err := fetchCosmosDBAccounts(ctx, a.MqlRuntime, conn, subId, &res)
+	cosmosAccounts, err := fetchCosmosDBAccounts(ctx, a.MqlRuntime, conn, subId)
 	if err != nil {
 		return nil, err
 	}
+	res = append(res, cosmosAccounts...)
 
-	err = fetchMongoDBClusters(ctx, a.MqlRuntime, conn, subId, &res)
+	mongoAccounts, err := fetchMongoDBAndPostgreSQLAccounts(ctx, a.MqlRuntime, conn, subId, "Microsoft.DocumentDB/mongoClusters")
 	if err != nil {
 		return nil, err
 	}
+	res = append(res, mongoAccounts...)
 
-	err = fetchPostgreSQLServerGroups(ctx, a.MqlRuntime, conn, subId, &res)
+	postgresAccounts, err := fetchMongoDBAndPostgreSQLAccounts(ctx, a.MqlRuntime, conn, subId, "Microsoft.DBforPostgreSQL/serverGroupsv2")
 	if err != nil {
 		return nil, err
 	}
+	res = append(res, postgresAccounts...)
 
 	return res, nil
 }
 
-func fetchCosmosDBAccounts(ctx context.Context, runtime *plugin.Runtime, conn *connection.AzureConnection, subId string, res *[]interface{}) error {
+func fetchCosmosDBAccounts(ctx context.Context, runtime *plugin.Runtime, conn *connection.AzureConnection, subId string) ([]interface{}, error) {
 	accClient, err := cosmosdb.NewDatabaseAccountsClient(subId, conn.Token(), &arm.ClientOptions{
 		ClientOptions: conn.ClientOptions(),
 	})
 	if err != nil {
-		return err
+		return nil, err
 	}
+
+	res := []interface{}{}
 	pager := accClient.NewListPager(&cosmosdb.DatabaseAccountsClientListOptions{})
 	for pager.More() {
 		page, err := pager.NextPage(ctx)
 		if err != nil {
-			return err
+			return nil, err
 		}
 		for _, account := range page.Value {
 			properties, err := convert.JsonToDict(account.Properties)
 			if err != nil {
-				return err
+				return nil, err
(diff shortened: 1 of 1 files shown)


providers/azure/resources/cosmosdb.go
```diff
@@ -75,10 +96,92 @@ func (a *mqlAzureSubscriptionCosmosDbService) accounts() ([]interface{}, error)
...
+		if err != nil {
+			return err
+		}
+		for _, account := range page.Value {
+			properties, err := convert.JsonToDict(account.Properties)
+			if err != nil {
+				return err
+			}
+
+			mqlResource, err := CreateResource(runtime, "azure.subscription.cosmosDbService.account",
```
tas50, line 137: Won't be need this in the .lr file?

providers/azure/resources/cosmosdb.go
```diff
@@ -75,10 +96,92 @@ func (a *mqlAzureSubscriptionCosmosDbService) accounts() ([]interface{}, error)
...
```
preslavgerchev: i am a bit confused here: the function says its fetching cluster, but we're creating cosmosdb accounts. is this a wrong mql resource here?

providers/azure/resources/cosmosdb.go
```diff
@@ -42,29 +43,49 @@ func (a *mqlAzureSubscriptionCosmosDbServiceAccount) id() (string, error) {
...
+
+	err = fetchPostgreSQLServerGroups(ctx, a.MqlRuntime, conn, subId, &res)
+	if err != nil {
+		return nil, err
+	}
+
+	return res, nil
+}
+
+func fetchCosmosDBAccounts(ctx context.Context, runtime *plugin.Runtime, conn *connection.AzureConnection, subId string, res *[]interface{}) error {
```
preslavgerchev: can't we have one function that takes a `type` and filters, based on that? it seems the code is almost identical, except for the fact we're requesting different entities. we can use a param here

preslavgerchev:
```suggestion
func fetchCosmosDBAccounts(ctx context.Context, runtime *plugin.Runtime, conn *connection.AzureConnection, subId string) ([]interface{},error) {
```
we should just return the list here, no need to pass a pointer to it to fill it in

providers/azure/resources/cosmosdb.go
HRouhani: yes usually MondoDB and PostgreSQL are referred as cluster as a logical group, but here in cosmodb context better to stick to the account naming. Updated

providers/azure/resources/cosmosdb.go
HRouhani: correct, mergged MondoDB and PostgreSQL to one function. Thanks

HRouhani: OK, I changed overal to return the list. Thanks

providers/azure/resources/cosmosdb.go
HRouhani: No, already exists, we just append for other types!

providers/azure/resources/cosmosdb.go
```diff
@@ -82,3 +108,46 @@ func (a *mqlAzureSubscriptionCosmosDbService) accounts() ([]interface{}, error)
...
+func fetchMongoDBAndPostgreSQLAccounts(ctx context.Context, runtime *plugin.Runtime, conn *connection.AzureConnection, subId string, resourceType string) ([]interface{}, error) {
+	resClient, err := armresources.NewClient(subId, conn.Token(), &arm.ClientOptions{
+		ClientOptions: conn.ClientOptions(),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	res := []interface{}{}
+	filter := "resourceType eq '" + resourceType + "'"
```
preslavgerchev:
```suggestion
	filter := fmt.Sprintf("resourceType eq '%s'", resourceType)
```

providers/azure/resources/cosmosdb.go
```diff
@@ -82,3 +108,46 @@ func (a *mqlAzureSubscriptionCosmosDbService) accounts() ([]interface{}, error)
 	}
 	return res, nil
 }
+
+func fetchMongoDBAndPostgreSQLAccounts(ctx context.Context, runtime *plugin.Runtime, conn *connection.AzureConnection, subId string, resourceType string) ([]interface{}, error) {
```
preslavgerchev: can we rename this func name? to something like `fetchDbAccountsByType`?

providers/azure/resources/cosmosdb.go
HRouhani: done, good point

providers/azure/resources/cosmosdb.go
HRouhani: done! no worries

preslavgerchev:
@HRouhani, have you tested if we get back other postgresql and mongodb instances when running this from the `cosmosdb` resource? What if I create a flexible postgresql server, will that end up in the results when querying cosmodb accounts?

HRouhani:
> @HRouhani, have you tested if we get back other postgresql and mongodb instances when running this from the `cosmosdb` resource? What if I create a flexible postgresql server, will that end up in the results when querying cosmodb accounts?

Very good question. No it is not possible:

the PostgreSQL has different resource type which is :  Microsoft.DBforPostgreSQL/flexibleServers

and MondoDB Atlas has different API, need to get it from Market place first.

preslavgerchev (approved):

PR #4449, opened 2024-07-30
GitHub repo discovery performance

chris-rock:
In the past we did some optimizations to fetch repositories only once. While this makes sense when we do discovery there are cases where fetching all repos from an org or user does not make sense. A specific case is when I just want to scan a single repo. 

```
cnquery shell github repo microsoftgraph/msgraph-sdk-go
```

That call should not fetch all repos of an org. It makes scanning an org for large organizations slower. This PR fixes it. It changes the code to leverage a cache. If that cache exists (eg. because a user called `github.organization.repositories`) then its used.

**before**

```
DEBUG=1 cnquery shell github repo microsoftgraph/msgraph-sdk-go
DBG performing request method=GET url=https://api.github.com/zen
DBG fetching organization microsoftgraph
DBG performing request method=GET url=https://api.github.com/orgs/microsoftgraph
DBG performing request method=GET url=https://api.github.com/orgs/microsoftgraph/repos?per_page=100&type=all
DBG performing request method=GET url=https://api.github.com/orgs/microsoftgraph/repos?page=2&per_page=100&type=all
DBG performing request method=GET url=https://api.github.com/orgs/microsoftgraph/repos?page=3&per_page=100&type=all
```

**after**

```
DBG fetching organization microsoftgraph
DBG performing request method=GET url=https://api.github.com/orgs/microsoftgraph
DBG performing request method=GET url=https://api.github.com/repos/microsoftgraph/msgraph-sdk-go
```

### DIFF
--- a/providers/github/resources/github.go
+++ b/providers/github/resources/github.go
@@ -53,8 +53,8 @@ func getOrg(ctx context.Context, runtime *plugin.Runtime, conn *connection.Githu
 	}
 	res, err, _ := g.memoize.Memoize("org-"+name, func() (interface{}, error) {
 		log.Debug().Msgf("fetching organization %s", name)
-		user, _, err := conn.Client().Organizations.Get(ctx, name)
-		return user, err
+		org, _, err := conn.Client().Organizations.Get(ctx, name)
+		return org, err
 	})
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


vjeffrey (approved):
🎉

PR #4453, opened 2024-07-30
⭐️ go package to create `tmp.File()`

afiune:
This go package should be used across `cnquery` to generate temporary files and directories.

`tmp.File()` creates a new temporary file.

By default `File()` uses the default directory to create a new temporary file, to change the temporary directory use the environment variable `MONDOO_TMP_DIR`

### DIFF
--- a/apps/cnquery/cmd/scan.go
+++ b/apps/cnquery/cmd/scan.go
@@ -47,9 +47,15 @@ func init() {
 	_ = scanCmd.Flags().MarkHidden("inventory-domainlist")
 
 	// bundles, packs & incognito mode
-	_ = scanCmd.Flags().Bool("incognito", false, "Run in incognito mode. Do not report scan results to Mondoo Platform")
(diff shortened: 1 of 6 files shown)


apps/cnquery/cmd/scan.go
```diff
@@ -47,9 +47,15 @@ func init() {
...
-	_ = scanCmd.Flags().StringSliceP("querypack-bundle", "f", nil, "Path to local query pack file")
+	_ = scanCmd.Flags().Bool("incognito",
+		false,
+		"Run in incognito mode. Do not report scan results to Mondoo Platform")
+	_ = scanCmd.Flags().StringSlice("querypack",
+		nil,
+		"Set the query packs to execute. This requires `querypack-bundle`. You can specify multiple UIDs")
+	_ = scanCmd.Flags().StringSliceP("querypack-bundle",
+		"f",
+		nil, "Path to local query pack file")
```
afiune, line 58: that was my linter, I can revert if it is annoying 😅

chris-rock (approved):
Great improvement @afiune

PR #4468, opened 2024-08-01
🐛 Use unique ids for mondoo spaces. Improve provider code

preslavgerchev:
If we're connecting with an org config, this won't work. Switch to using the space's MRN from the MQL resource.

### DIFF
--- a/providers/mondoo/provider/provider.go
+++ b/providers/mondoo/provider/provider.go
@@ -8,8 +8,6 @@ import (
 	"errors"
 	"fmt"
 
-	"go.mondoo.com/cnquery/v11/llx"
-	"go.mondoo.com/cnquery/v11/mrn"
 	"go.mondoo.com/cnquery/v11/providers-sdk/v1/inventory"
 	"go.mondoo.com/cnquery/v11/providers-sdk/v1/plugin"
 	"go.mondoo.com/cnquery/v11/providers/mondoo/connection"
@@ -81,11 +79,7 @@ func (s *Service) connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 			return nil, err
 		}
 
-		err = fillAsset(upstream.SpaceMrn, asset)
-		if err != nil {
-			return nil, err
-		}
-		asset.Connections[0].Id = connId
+		fillAsset(conn, asset)
 
 		return plugin.NewRuntime(
 			conn,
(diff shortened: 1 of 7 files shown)


providers/mondoo/resources/mondoo_space.go
```diff
@@ -39,7 +57,7 @@ func (m *mqlMondooSpace) assets() ([]any, error) {
 		} `graphql:"assets(spaceMrn: $spaceMrn)"`
 	}
 	vars := map[string]any{
-		"spaceMrn": mondoogql.String(conn.Upstream.SpaceMrn),
+		"spaceMrn": mondoogql.String(m.Mrn.Data),
```
preslavgerchev, line 60: this means we use the mql MRN to fetch data. this means if we connect via a mondoo org this still works

providers/mondoo/resources/mondoo_organization.go
```diff
@@ -5,12 +5,29 @@ package resources
...
 	mondoogql "go.mondoo.com/mondoo-go"
 )
 
+func initMondooOrganization(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error) {
+	conn := runtime.Connection.(*connection.Connection)
+	if conn.Type != connection.ConnTypeOrganization {
+		return nil, nil, errors.New("cannot initialize mondoo.organization, invalid connection")
+	}
+	args["mrn"] = llx.StringData(conn.Upstream.SpaceMrn)
+	args["name"] = llx.StringData(connection.MrnBasenameOrMrn(conn.Upstream.SpaceMrn))
```
chris-rock, line 22: That is the space id, right?

providers/mondoo/provider/provider.go
```diff
@@ -145,55 +124,7 @@ func fillAsset(mrnStr string, asset *inventory.Asset) error {
...
-	_, err = resources.CreateResource(runtime, "mondoo.client", map[string]*llx.RawData{
-		"mrn": llx.StringData(conn.Upstream.Creds.Mrn),
-	})
-	if err != nil {
-		return err
-	}
-
-	spaceMrn := conn.Upstream.SpaceMrn
-	if asset.Platform.Name == "mondoo-space" {
-		_, err = resources.CreateResource(runtime, "mondoo.space", map[string]*llx.RawData{
```
preslavgerchev, line 179: this stopped working once i added `id()` funcs to mondoo.space and mondoo.organization

i didnt dive into why, but i did move this to be init funcs for the mql resources

providers/mondoo/resources/mondoo_space.go
```diff
@@ -5,14 +5,32 @@ package resources
...
 )
 
+func initMondooSpace(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error) {
+	conn := runtime.Connection.(*connection.Connection)
+	if conn.Type != connection.ConnTypeSpace {
+		return nil, nil, errors.New("cannot initialize mondoo.space, invalid connection")
+	}
+	args["mrn"] = llx.StringData(conn.Upstream.SpaceMrn)
+	// TODO: we should fetch this from the API, mrn basename != name
+	args["name"] = llx.StringData(connection.MrnBasenameOrMrn(conn.Upstream.SpaceMrn))
```
chris-rock, line 25: I think we should turn that into a resolver to fetch the space details.

providers/mondoo/resources/mondoo_organization.go
preslavgerchev: this is either the org id or the space id, regardless of the fact we call it `SpaceMrn`, it all depends on what the mondoo config that we're connecting with has

providers/mondoo/resources/mondoo_space.go
preslavgerchev: +1, thats the TODO above that line

lets do it in a follow-up PR

chris-rock (approved):
Thank you @preslavgerchev

PR #4470, opened 2024-08-01
🧹   Set default values to -1 for non-existent registry keys Instead of null  (Windows)

HRouhani:
Screenshot from 2024-08-01 14-18-23

When exists is Flase, it means that the Key does not exist, and it gets **-1** for other stuff!


chris-rock:
We should not set arbitrary values. Generally this leads to more problems than in solves. Right now this needs to be solved via:

```
value := registrykey.property(path: 'HKEY_LOCAL_MACHINE\Software\Policies\Microsoft\WindowsFirewall\DomainProfile\Logging', name: 'LogFileSize').value 

if ( value != empty) {
    value >= 16384
}
```

PR #4473, opened 2024-08-01
✨ use github-client internal rate-limits timer

slntopp:
Github Client is handling rate-limits internally out of the box, which means if rate limits are reached, client will return fake 403 API Rate limits response before even trying out retryable client.
Unfortunately we can't use `bypassRateLimitCheck` strategy, because it's private, but we can use `SleepUntilPrimaryRateLimitResetWhenRateLimited` instead, which should be doing just the same job as the retryable client.
Secondary limits must still be handled by retryable client.

### DIFF
--- a/providers/github/connection/connection.go
+++ b/providers/github/connection/connection.go
@@ -68,6 +68,8 @@ func NewGithubConnection(id uint32, asset *inventory.Asset) (*GithubConnection, 
 		}
 	}
 
+	// set the context so github client can handle backoff
+	// (default behaviour is to send fake 403 response bypassing the retry logic)
 	ctx := context.WithValue(context.Background(), github.SleepUntilPrimaryRateLimitResetWhenRateLimited, true)
 
 	// perform a quick call to verify the token's validity.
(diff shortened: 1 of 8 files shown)


providers/github/connection/connection.go
```diff
@@ -67,8 +68,10 @@ func NewGithubConnection(id uint32, asset *inventory.Asset) (*GithubConnection,
 		}
 	}
 
+	ctx := context.WithValue(context.Background(), github.SleepUntilPrimaryRateLimitResetWhenRateLimited, true)
```
chris-rock, line 73: Can we add a comment with your research findings, so that this context is written down in code?

chris-rock (commented):
Looks great @slntopp Lets add some comments and we are good to go.

providers/github/connection/connection.go
slntopp: Added the comment

chris-rock (approved):
Thank you @slntopp

PR #4478, opened 2024-08-04
⭐ sample

arlimus:
Introduce the `sample` command, which extracts a number of samples from any list or map:

```
> [1,2,3,4,5,6,7,8,9,10].sample(3)
sample: [
  0: 2
  1: 10
  2: 9
]
```

Any time you run this, you'll (most likely) get a different set of results.

```
> {'a': 1, 'b': 2, 'c': 3, 'd': 4, 'e': 5}.sample(3)
sample: {
  a: 1
  b: 2
  e: 5
}
```

Duplicates are not allowed, unless of course the original data has duplicates. Ultimately every entry is only picked once.

```
> users.sample(3)
users.sample: [
  0: user id = user/972/deepin-one
  1: user id = user/991/systemd-network
  2: user id = user/973/systemd-timesync
]

### DIFF
--- a/llx/builtin.go
+++ b/llx/builtin.go
@@ -537,6 +537,7 @@ func init() {
 			"keys":                            {f: dictKeysV2, Label: "keys"},
 			"values":                          {f: dictValuesV2, Label: "values"},
 			"where":                           {f: dictWhere, Label: "where"},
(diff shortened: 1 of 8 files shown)


tas50:
Compliance time!

chris-rock (approved):
Thank you @arlimus Nice addition.

PR #4484, opened 2024-08-05
🐛 allow cross-provider containers from k8s

### LINKED ISSUE #4472: k8s.deployments ERROR: rpc error: code = Unknown desc = cannot find connection type k8s
**Describe the bug**
image

Running the command `k8s.deployments.where( namespace != "kube-system" ) { containers {*}}`  returns `rpc error: code = Unknown desc = cannot find connection type k8s`

**To Reproduce**

1. Open the shell on a k8s cluster: `cnspec shell k8s`
2. Run command: `k8s.deployments.where( namespace != "kube-system" ) { containers {*}}`

imilchev:
This is not a proper fix but a workaround to make sure the specific command works. We will need to re-visit cross-provider resources at some point.

Fixes #4472

### DIFF
--- a/providers/os/provider/provider.go
+++ b/providers/os/provider/provider.go
@@ -347,7 +347,7 @@ func (s *Service) connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 		var err error
 
 		switch conf.Type {
-		case shared.Type_Local.String():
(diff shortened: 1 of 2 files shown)


arlimus (approved):
Looks good, let's keep this as a temporary workaround until we streamline cross-provider calling

PR #4494, opened 2024-08-06
🐛 gitlab: fix GetProject for projects in subgroups

slntopp:
PathEscape is handled by `go-gitlab` for at least past 2 years: https://github.com/xanzy/go-gitlab/blob/f4f54fa4a42d2004e5b269921578bfe529ab19a0/projects.go#L525

> Escaping escaped string would produce '/' -> '%2F' -> '%252F'


chris-rock (approved):
Thank you @slntopp

PR #4495, opened 2024-08-07
🧹 Check the pricing tier for the Servers plan when verifying if Azure's Defender for Servers is enabled

preslavgerchev:
The latest CIS implementation indicates that we can check the pricing tier for the defender for servers plan. IF set to `standard` it's considered enabled. This PR adjusts the `defenderForServers` check to take that into account


chris-rock (approved):
Thank you @preslavgerchev

PR #4511, opened 2024-08-08
🧹 update proto generated files

chris-rock:


shared/proto/cnquery_grpc.pb.go
```diff
@@ -166,6 +181,13 @@ type UnsafeOutputHelperServer interface {
 }
 
 func RegisterOutputHelperServer(s grpc.ServiceRegistrar, srv OutputHelperServer) {
+	// If the following call pancis, it indicates UnimplementedOutputHelperServer was
```
tas50, line 184:
```suggestion
	// If the following call panics, it indicates UnimplementedOutputHelperServer was
```

shared/proto/cnquery_grpc.pb.go
```diff
@@ -75,6 +79,13 @@ type UnsafeCNQueryServer interface {
 }
 
 func RegisterCNQueryServer(s grpc.ServiceRegistrar, srv CNQueryServer) {
+	// If the following call pancis, it indicates UnimplementedCNQueryServer was
```
tas50, line 82:
```suggestion
	// If the following call panics, it indicates UnimplementedCNQueryServer was
```

vjeffrey (approved):

shared/proto/cnquery_grpc.pb.go
chris-rock: Already fixed upstream https://github.com/grpc/grpc-go/pull/7456/files, nothing we can do since this is auto-generated.

PR #4513, opened 2024-08-09
🧹 use protoc action consistently

chris-rock:
Use the protoc action consistently.

### DIFF
--- a/.github/workflows/pr-test-generated-files.yaml
+++ b/.github/workflows/pr-test-generated-files.yaml
@@ -37,15 +37,14 @@ jobs:
             echo "Install GCC on the runner."; 
           fi
 
-      # Note we do not use apt install -y protobuf-compiler` since it is too old
+        # Note we do not use apt install -y protobuf-compiler` since it is too old
+      - name: Install Protoc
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-test-generated-files.yaml
```diff
@@ -37,15 +37,14 @@ jobs:
             echo "Install GCC on the runner."; 
           fi
 
-      # Note we do not use apt install -y protobuf-compiler` since it is too old
+        # Note we do not use apt install -y protobuf-compiler` since it is too old
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
preslavgerchev, line 44: why do we need the repo-token here?

preslavgerchev (approved):

.github/workflows/pr-test-generated-files.yaml
chris-rock: Because the action pulls the binaries from a github repo. protoc is only released on github

PR #4520, opened 2024-08-10
Fallback to Get-Ciminstance based get-computerinfo

jaym:
If `OsProductType` is null, fallback to a custom implementation of `Get-ComputerInfo`. There are versions of powershell where if winrm is not enabled, `Get-ComputerInfo` does not work properly: See https://github.com/PowerShell/PowerShell/issues/3080


chris-rock (approved):
Thank you @jaym

PR #4521, opened 2024-08-10
🐛  fix guard duty pagination bug and harmonize list findings functions

chris-rock:


atomic111 (approved):
@chris-rock ty for the fix

PR #4525, opened 2024-08-11
🧹 change lr default version to 9.0.0

chris-rock:
This PR switches the default version for generated docs from latest to 9.0.0 since the resources are backwards compatible now.


chris-rock:
@tas50 This should reduce the review workload for the docs file dramatically :-)

tas50 (approved):

tas50:
Beautiful!

PR #4536, opened 2024-08-12
If there is no init system, return empty services

jaym:
If there is no init system, like in containers, we dont have something
managing services to list. This change modifies the code to return an
empty list in the case we do not detect `/sbin/init`

### DIFF
--- a/providers/os/resources/services/manager.go
+++ b/providers/os/resources/services/manager.go
@@ -40,6 +40,17 @@ type OSServiceManager interface {
 	List() ([]*Service, error)
 }
 
+type noopOsServiceManager struct {
+}
+
+func (n *noopOsServiceManager) Name() string {
+	return "none"
+}
+
+func (n *noopOsServiceManager) List() ([]*Service, error) {
+	return nil, nil
+}
+
 var amazonlinux1version = regexp.MustCompile(`^201\d`)
 
 func ResolveManager(conn shared.Connection) (OSServiceManager, error) {
@@ -50,7 +61,18 @@ func ResolveManager(conn shared.Connection) (OSServiceManager, error) {
 		return nil, errors.New("cannot find OS information for package detection")
 	}
 
+	useNoopInit := false
+	if asset.Platform.IsFamily("linux") {
+		// If we're on linux, check if there is no init system. If there is no init system,
+		// we don't have managed services. This happens in containers.
+		if _, err := conn.FileInfo("/sbin/init"); err != nil {
+			useNoopInit = true
+		}
+	}
+
 	switch {
+	case useNoopInit:
+		osm = &noopOsServiceManager{}
 	case asset.Platform.IsFamily("arch"): // arch family
 		osm = ResolveSystemdServiceManager(conn)
 	case asset.Platform.Name == "amazonlinux":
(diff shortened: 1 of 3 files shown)


providers/os/resources/services/manager.go
```diff
@@ -99,16 +99,21 @@ func ResolveManager(conn shared.Connection) (OSServiceManager, error) {
...
-		rv := detector.ParseOsVersion(asset.Platform.Version)
-		v, err := rv.MajorAtoi()
-		if err != nil {
-			return nil, errors.New("unknown debian version: " + asset.Platform.Version)
-		}
-
-		if v < 7 {
+		if _, err := conn.FileInfo("/.dockerenv"); err == nil {
+			// The debian docker image does not use systemd
 			osm = &SysVServiceManager{conn: conn}
```
tas50: This feels very fragile since many different containers end up being detected as Debian. Is there a way we can detect sys-v vs systemd automatically?

chris-rock: Agreed, it is probably more suitable to look for an existing `systemd` package. We may even need to look for a sysv package since that switch to sysv can also happen on VMs https://lecorbeausvault.wordpress.com/2022/02/07/debian-switching-init-system-easily-openrc-sysvinit-runit/

jaym: do you have any examples of those. Do any of them use systemd?

jaym: in a vm, we can check `/sbin/init` and see if it points to systemd

atomic111:
@jaym i have test it on the container. this is the result

```bash
services
x cannot parse sysv runlevel line=/etc/rcS.d/S01hwclock.sh
services.list: [
  0: service name="hwclock.sh" running=false enabled=false type="sysv"
  1: service name="nginx" running=true enabled=true type="sysv"
  2: service name="nginx-debug" running=true enabled=false type="sysv"
]
```

atomic111:
this is the way to load the mfa-exam.zip 

```bash
cat mfa-exam.zip | docker load
```

jaym:
@atomic111 maybe the expected result should be an empty list. There is no systemd or sysvinit in the container. pid 1 is just nginx

atomic111:
this is the list of process that are running

```
lrwxrwxrwx 1 root  root  0 Aug 13 15:40 /proc/1/exe -> /run/rosetta/rosetta
lrwxrwxrwx 1 nginx nginx 0 Aug 13 17:19 /proc/28/exe
lrwxrwxrwx 1 nginx nginx 0 Aug 13 17:19 /proc/29/exe
lrwxrwxrwx 1 nginx nginx 0 Aug 13 17:19 /proc/30/exe
lrwxrwxrwx 1 nginx nginx 0 Aug 13 17:19 /proc/31/exe
lrwxrwxrwx 1 root  root  0 Aug 13 17:19 /proc/390/exe -> /run/rosetta/rosetta
lrwxrwxrwx 1 root  root  0 Aug 13 17:19 /proc/self/exe -> /usr/bin/ls
lrwxrwxrwx 1 root  root  0 Aug 13 17:19 /proc/thread-self/exe -> /usr/bin/ls
```

atomic111:
from the service perspective it should be empty list

tas50:
Agree with Patrick on this one. Jay and I chatted about this. In the case of a container there are no services the way there are on an upstart/sys-v/systemd system. We need to return that 0 service list without throwing an error to the user.

jaym:
I've modified this to detect `/sbin/init`. if its not there we do not try to list services

chris-rock (approved):
Thank you @jaym

PR #4539, opened 2024-08-12
✨ add serial number to asset ids

slntopp:


providers/os/id/platform.go
```diff
@@ -161,6 +162,17 @@ func GatherPlatformInfo(conn shared.Connection, pf *inventory.Platform, idDetect
 			}, hostErr
 		}
 		return &PlatformInfo{}, nil
+	case idDetector == ids.IdDetector_SerialNumber:
+		serial, err := serialnumber.SerialNumber(conn, pf)
```
chris-rock, line 171: Have we tested that for macos, windows and linux?

providers/os/id/platform.go
```diff
@@ -52,7 +53,7 @@ func IdentifyPlatform(conn shared.Connection, p *inventory.Platform, idDetectors
 		// fallback to default id detectors
 		switch conn.Type() {
 		case shared.Type_Local:
-			idDetectors = []string{ids.IdDetector_Hostname, ids.IdDetector_CloudDetect}
+			idDetectors = []string{ids.IdDetector_Hostname, ids.IdDetector_SerialNumber, ids.IdDetector_CloudDetect}
```
chris-rock: I think we want to do some more real-life testing before we enable this as default.

providers/os/id/platform.go
slntopp: This is taken from the existing resource - `machine.system.serial` - which works on all aforementioned OS's
My tests so far show that the on the systems which don't provide that data (for example CI containers) serial evaluates to empty string, so it's omitted in the `asset.ids`

providers/os/id/platform.go
slntopp: As expected, if host can't provide serial name - it's just returned empty, therefore is not used as asset id.
Asset id is actually asset.ids, so server seeks for and uses the id server prefers (hostname at the moment).
If asset returns two or more ids this doesn't influence existing processes.

Tested on:
- MacOS
- Darwin Container
- Linux Container
- Linux Desktop
- Windows Desktop
- Windows VM

vjeffrey:
i tested this pr by running a scan on my local mac, changing my hostname, and then scanning again. the hostname value was updates as expected but i had the same serial number recorded in both cases (as expected). 
i was testing to see if there any problems reporting in to the server if we report diff hostnames along with the serial id, and i didnt run into any issues
@jaym just for peace of mind, can you confirm it's fine to have two assets reporting in that have one asset id that is the same and one asset id that is different?

jaym:
> just for peace of mind, can you confirm it's fine to have two assets reporting in that have one asset id that is the same and one asset id that is different?

No that does not seem ok. 

Here's a scenario. We have an asset `A` with platform id `a` that has scanned. Then we have an asset `B` with platform id `b` that has scanned.

Now, asset `A` scans with `[a, shared]`. It is successful. If asset `B` scans with `[b,shared]` it will fail because it matches asset `B` through platform id `b` and asset `A` through platform id `shared.

slntopp:
Isn't detectors (therefore ids) order predefined?

atomic111:
@slntopp @vjeffrey 

Can we read the asset name from 

```
scutil --get ComputerName
```

Because then the hostname is not changing.

slntopp:
@atomic111 id detectors aren't OS specific and using existing resources from os provider where they can, so i'm not sure about this one. 
Using serial number seems better either way, because unlike hostname it can't be changed on any platform

atomic111:
@jaym when can we merge this

jaym:
I think we need a way to prioritize platform ids if we want to take this change, otherwise it will be a breaking change. We will get into a state where if there are already duplicate representations of an asset where the hostname is flopping around, it will succeed once with a specific hostname in the platform id. Once it comes back with a different hostname in the platform id, the asset will not be able to scan

jaym (approved):

PR #4540, opened 2024-08-12
🐛 fix name for AWS Timestream

chris-rock:


providers/aws/resources/aws.lr.manifest.yaml
```diff
@@ -3011,46 +3011,15 @@ resources:
     platform:
       name:
       - aws
-  aws.timestream:
```
tas50, line 3014: ???

chris-rock: That was an old generation. The root switched to aws.timestream.liveanalytics

tas50 (approved):

PR #4548, opened 2024-08-13
🐛 Fix endless KMS retrieval loop.

preslavgerchev:
If an err happens when we're fetching KMS key, we continue and do not break. This means we get to retry again, however if the error is consistent (say a 403) we will never exit this loop. Instead, let's break out of the loop if we encounter an error.


providers/aws/resources/aws_kms.go
```diff
@@ -65,7 +65,7 @@ func (a *mqlAwsKms) getKeys(conn *connection.AwsConnection) []*jobpool.Job {
 					if Is400AccessDeniedError(err) {
 						log.Warn().Str("region", regionVal).Msg("error accessing region for AWS API")
 					}
-					continue
+					break
```
chris-rock, line 68: Can we check if we have more of those case?

chris-rock (approved):
Thank you @preslavgerchev

PR #4566, opened 2024-08-16
🧹 return the correct value for aws account id when connected to subresource

vjeffrey:
e.g. an s3 bucket, iam user

we were actually returning the arn of the iam user and s3 bucket in those cases, because we were grabbing the asset identifier ARN value.

this ensures we always return the account id.

tested with `cnspec shell aws `, `cnspec shell aws --discover iam-user`, `cnspec shell aws --discover s3-bucket`

### DIFF
--- a/providers/aws/resources/aws_account.go
+++ b/providers/aws/resources/aws_account.go
@@ -6,7 +6,6 @@ package resources
 import (
 	"context"
 	"errors"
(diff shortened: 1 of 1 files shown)


czunker (approved):
LGTM
Thanks @vjeffrey

PR #4572, opened 2024-08-17
⭐️ Microsoft 365 first party apps filtering

chris-rock:
When you want to verify enterprise apps and their permissions, it is essential to focus on the right enterprise applications. Therefore we introduce a new field `isFirstParty` to service principals:

```
cnquery> microsoft.serviceprincipals.where( isFirstParty == false)
microsoft.serviceprincipals.where: [
  0: microsoft.serviceprincipal name="chris-ms365"
  1: microsoft.serviceprincipal name="chris-mac-test"
]
```

### DIFF
--- a/providers/ms365/resources/serviceprincipals.go
+++ b/providers/ms365/resources/serviceprincipals.go
@@ -16,6 +16,11 @@ import (
 	"go.mondoo.com/cnquery/v11/utils/stringx"
 )
 
+const (
+	MicrosoftEntraTenantID = "f8cdef31-a31e-4b4a-93e4-5f571e91255a"
+	MicrosoftTenantID      = "72f988bf-86f1-41af-91ab-2d7cd011db47"
+)
+
 func (m *mqlMicrosoftServiceprincipal) id() (string, error) {
 	return m.Id.Data, nil
 }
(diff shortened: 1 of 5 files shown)


providers/ms365/resources/builtin_applications.go
```diff
@@ -0,0 +1,1868 @@
...
+	MicrosoftTenantID      = "72f988bf-86f1-41af-91ab-2d7cd011db47"
+)
+
+type KnownApplicationId struct {
+	Name  string
+	AppId string
+}
+
+// Apps owned by "f8cdef31-a31e-4b4a-93e4-5f571e91255a"
+// https://learn.microsoft.com/en-us/troubleshoot/azure/entra/entra-id/governance/verify-first-party-apps-sign-in#application-ids-of-commonly-used-microsoft-applications
```
tas50: Is there any plan for how we'll keep this updated over time?

chris-rock: Replaced with new implementation. Found another way to implement this more reliably.

providers/ms365/resources/serviceprincipals.go
```diff
@@ -16,6 +16,11 @@ import (
 	"go.mondoo.com/cnquery/v11/utils/stringx"
 )
 
+const (
+	MicrosoftEntraTenantID = "f8cdef31-a31e-4b4a-93e4-5f571e91255a"
+	MicrosoftTenantID      = "72f988bf-86f1-41af-91ab-2d7cd011db47"
```
preslavgerchev, line 23: can we add some doc here? unclear what those ids are

vjeffrey (approved):

providers/ms365/resources/serviceprincipals.go
chris-rock: Comment added

PR #4573, opened 2024-08-18
⭐️ improve microsoft.tenant resource

chris-rock:
This PR makes the access to the Microsoft 365 tenant information more easy:

```javascript
microsoft.tenant
microsoft.tenant: microsoft.tenant name="example"
```

We also provide easy access to the subscribed SKU:

```javascript
cnquery> microsoft.tenant.subscriptions.where(status == "Enabled").map(skuPartNumber)
microsoft.tenant.subscriptions.where.map: [
  0: "FLOW_FREE"
  1: "DEVELOPERPACK_E5"
]
```

To verify that a specific subscription is active:

```javascript
cnquery> microsoft.tenant.subscriptions.where(status == "Enabled").map(skuPartNumber).one("DEVELOPERPACK_E5")
[ok] value: true
```

Alternatively you can check for multiple sku, eg. where you want to ensure that you have E5 or P2 enabled:

```javascript
microsoft.tenant.subscriptions.where(status == "Enabled").map(string(skuPartNumber)).any( _.in(["DEVELOPERPACK_E5", "AAD_PREMIUM_P2"]))
```

Note that the field following fields have been deprecated

- `displayName` is renamed to `name`
- `createdDateTime` is renamed to `createdAt`
- `microsoft.organizations` is deprecated in favor of the new `microsoft.tenant`


providers/ms365/resources/ms365.lr
```diff
@@ -28,20 +30,30 @@ microsoft {
...
   // Organization verified domains
   verifiedDomains []dict
   // Whether password hash sync is enabled for hybrid deployments
   onPremisesSyncEnabled bool
+  // Tenant creation date
+  createdAt time
+  // Tenant type
+  type string
+  // Commercial subscription
+  subscriptions() []dict
```
preslavgerchev, line 56: not scope for this PR but maybe w can think of tying this with `azure.subscription`, if possible

providers/ms365/resources/tenant.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	ctx := context.Background()
+	resp, err := graphClient.Organization().Get(ctx, &organization.OrganizationRequestBuilderGetRequestConfiguration{})
+	if err != nil {
+		return "", transformError(err)
+	}
+	tenantDomainName := ""
+
+	for _, org := range resp.GetValue() {
+		org.GetId()
```
preslavgerchev: this looks like a leftover

providers/ms365/resources/ms365.lr
```diff
@@ -4,10 +4,12 @@
...
 option go_package = "go.mondoo.com/cnquery/v11/providers/ms365/resources"
 
+alias microsoft.organization = microsoft.tenant
+
 // Microsoft
 microsoft {
-  // List of organizations
-  organizations() []microsoft.organization
+  // Deprecated: use `microsoft.tenant` instead
+  organizations() []microsoft.tenant
```
tas50, line 12: Should we be introducing a `tenants` field here as well to that there's a similar way to get all tenants?

providers/ms365/resources/ms365.lr
chris-rock: That is not the same as an azure subscription. Completely different thing :-)

providers/ms365/resources/ms365.lr
chris-rock: We only have one tenant, not required since users can just type `microsoft.tenant`

providers/ms365/resources/tenant.go
chris-rock: indeed, I am going to remove it

providers/ms365/resources/ms365.lr
```diff
@@ -28,20 +30,30 @@ microsoft {
   tenantDomainName() string
 }
 
-// Microsoft organization
-private microsoft.organization @defaults("displayName") {
+//  Microsoft Entra tenant
+microsoft.tenant @defaults("name") {
```
chris-rock, line 34: @tas50 the resource is public now so it can be called directly

tas50: LGTM then

preslavgerchev (approved):

PR #4579, opened 2024-08-19
🧹 token validation errors handling

### LINKED ISSUE #4550: cnspec login does not veriy the token
**Describe the bug**

**Shows status when login errors**

Use an clearly invalid token: `cnspec login -t invalidtoken --config test.yml` and the output should just show that the token has failed. The output should not show `cnspec status` and the results of `cnspec help` in combination. This needs to be fixed for both `cnquery` and `cnspec`. The generated output is confusing.

```
~ $> cnspec login -t invalidtoken --config test.yml
! could not read the token error="token is malformed: token contains an invalid number of segments"
→ Platform:		macos
→ Version:		14.5
→ Hostname:		ChristoHartmann.fritz.box
→ IP:			192.168.178.57
→ Time:			2024-08-13T15:41:21+02:00
→ Version:		v11.13.2 (API Version: unstable)
→ Latest Version:	11.17.0
! A newer version is available
→ API ConnectionConfig:	https://us.api.mondoo.com
→ API Status:		SERVING
→ API Time:		2024-08-13T13:41:21Z
→ API Version:		11
! API versions do not match, please update the client
x client is not registered
x The Mondoo Platform credentials provided at test.yml didn't successfully authenticate with Mondoo Platform. Please re-authenticate with Mondoo Platform. To learn how, read https://mondoo.com/docs/cnspec/cnspec-adv-install/registration/.
Error: failed to log in client: failed to log in client: failed to do request: Post "https:///AgentManager/RegisterAgent": http: no Host in request URL
Usage:
  cnspec login [flags]

Aliases:
  login, register

Flags:
      --annotation stringToString   Set the client annotations (default [])
      --api-endpoint string         Set the Mondoo API endpoint
  -h, --help                        help for login
      --name string                 Set asset name
      --splay int                   Randomize the timer by up to this many minutes
      --timer int                   Set the scan interval in minutes
  -t, --token string                Set a client registration token

Global Flags:
      --api-proxy string   Set proxy for communications with Mondoo API
      --auto-update        Enable automatic provider installation and update (default true)
      --config string      Set config file path (default $HOME/.config/mondoo/mondoo.yml)
      --log-level string   Set log level: error, warn, info, debug, trace (default "info")
  -v, --verbose            Enable verbose output
  ```

**Do not show `token will expire at Mon, 01 Jan 0001 00:00:00 UTC` for non expiring token**

The validation mechanism is not correct. I can pass in an arbitrary JWT token and pass the initial validation.

```
jwt decode eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c

Token header
------------
{
  "typ": "JWT",
  "alg": "HS256"
}

Token claims
------------
{
  "iat": 1516239022,
  "name": "John Doe",
  "sub": "1234567890"
}
```

And then try:

```
cnquery login -t eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c --config test.yml                                                                                                                                                                                                                                                             
→ token will expire at Mon, 01 Jan 0001 00:00:00 UTC
```

**Improve token validation**

To ensure we not just have a valid JWT token, but also one we understand we should check that the `scope` or the `space` claim exist:

```
Token header
------------
{
  "typ": "JWT",
  "alg": "ES384"
}

Token claims
------------
{
  "api_endpoint": "https://us.api.mondoo.com",
  "aud": [
    "mondoo"
  ],
  "desc": "test",
  "exp": 1723557652,
  "iat": 1723557052,
  "iss": "mondoo/ams",
  "labels": null,
  "nbf": 1723557052,
  "owner": "",
  "scope": "//captain.api.mondoo.app/spaces/beautiful-wiles-12345",
  "space": "//captain.api.mondoo.app/spaces/beautiful-wiles-12345",
  "sub": "//captain.api.mondoo.app/users/123456"
}
```

If it is not a token that includes either `scope` or `space` claim, then issue the following error:

```
cnquery login -t token--config test.yml
x The token is not a valid token to register this client with Mondoo platform
```

**Expected behavior**

- Improved validation of token
- More clear CLI output that makes it easier to understand what went wrong.

### PR DESCRIPTION
**Pull request by slntopp**
Resolves #4550

### DIFF
--- a/apps/cnquery/cmd/login.go
+++ b/apps/cnquery/cmd/login.go
@@ -27,7 +27,7 @@ import (
 )
 
 var (
-	tokenValidationErr = errors.New("The token is not a valid token to register this client with Mondoo platform")
+	tokenValidationErr = errors.New("The token is not a valid token to register this client with Mondoo Platform")
 	tokenExpiredErr    = errors.New("The token is expired")
 )
 
--- a/apps/cnquery/cmd/login.go
+++ b/apps/cnquery/cmd/login.go
@@ -26,6 +26,11 @@ import (
 	"go.mondoo.com/ranger-rpc/status"
 )
 
+var (
+	tokenValidationErr = errors.New("The token is not a valid token to register this client with Mondoo platform")
+	tokenExpiredErr    = errors.New("The token is expired")
+)
+
 func init() {
 	rootCmd.AddCommand(LoginCmd)
 	LoginCmd.Flags().StringP("token", "t", "", "Set a client registration token")
@@ -63,6 +68,12 @@ You remain logged in until you explicitly log out using the 'logout' subcommand.
 		apiEndpointOverride, _ := cmd.Flags().GetString("api-endpoint")
 		err := register(token, annotations, timer, splay, apiEndpointOverride)
 		if err != nil {
+			if err == tokenValidationErr {
+				log.Error().Msg(err.Error())
+
+				// Prevents help message from being printed
+				return nil
+			}
 			defer func() {
 				s, err := checkStatus()
 				if err != nil {
@@ -70,6 +81,13 @@ You remain logged in until you explicitly log out using the 'logout' subcommand.
 				}
 				s.RenderCliStatus()
 			}()
+
+			if err == tokenExpiredErr {
+				log.Error().Msg(err.Error())
+
+				// Prevents help message from being printed
+				return nil
+			}
 		}
 		return err
 	},
@@ -105,14 +123,22 @@ func register(token string, annotations map[string]string, timer int, splay int,
 		claims, err := upstream.ExtractTokenClaims(token)
 		if err != nil {
 			log.Warn().Err(err).Msg("could not read the token")
+			return tokenValidationErr
 		} else {
 			if len(claims.Description) > 0 {
 				log.Info().Msg("token description: " + claims.Description)
 			}
 			if claims.IsExpired() {
-				log.Warn().Msg("token is expired")
+				return tokenExpiredErr
+			} else if claims.Expiry == nil {
+				log.Warn().Msg("token does not contain an expiry date")
 			} else {
 				log.Info().Msg("token will expire at " + claims.Claims.Expiry.Time().Format(time.RFC1123))
+			}
+			if claims.Space == "" {
+				log.Warn().
+					Msg("token does not contain a space")
+				return tokenValidationErr
 			}
 
 			// use the api endpoint from the token if not overridden via flag
--- a/apps/cnquery/cmd/login.go
+++ b/apps/cnquery/cmd/login.go
@@ -19,6 +19,7 @@ import (
 	cnquery_providers "go.mondoo.com/cnquery/v11/providers"
 	"go.mondoo.com/cnquery/v11/providers-sdk/v1/sysinfo"
 	"go.mondoo.com/cnquery/v11/providers-sdk/v1/upstream"
+	"go.mondoo.com/cnquery/v11/providers-sdk/v1/upstream/health"
 	rangerUtils "go.mondoo.com/cnquery/v11/utils/ranger"
 	"go.mondoo.com/ranger-rpc"
 	"go.mondoo.com/ranger-rpc/codes"
@@ -75,11 +76,27 @@ You remain logged in until you explicitly log out using the 'logout' subcommand.
 				return nil
 			}
 			defer func() {
-				s, err := checkStatus()
-				if err != nil {
-					log.Warn().Err(err).Msg("could not run status command")
+				opts, optsErr := config.Read()
+				if optsErr != nil {
+					log.Error().Err(optsErr).Msg("could not load configuration")
+					return
 				}
-				s.RenderCliStatus()
+
+				httpClient, err := opts.GetHttpClient()
+				if err != nil {
+					log.Error().Err(optsErr).Msg("failed to set up Mondoo API client")
+					return
+				}
+
+				upstreamStatus, err := health.CheckApiHealth(httpClient, opts.UpstreamApiEndpoint())
+				if err != nil {
+					log.Error().Err(err).Msg("could not check upstream health")
+					return
+				}
+
+				for _, warn := range upstreamStatus.Warnings {
+					log.Warn().Msg(warn)
+				}
 			}()
 
 			if err == tokenExpiredErr {

### CODE REVIEW

File: apps/cnquery/cmd/login.go [hunk 1]
```diff
@@ -63,13 +68,26 @@ You remain logged in until you explicitly log out using the 'logout' subcommand.
 		apiEndpointOverride, _ := cmd.Flags().GetString("api-endpoint")
 		err := register(token, annotations, timer, splay, apiEndpointOverride)
 		if err != nil {
+			if err == tokenValidationErr {
```
slntopp, line 72: Prevents check command from being run

apps/cnquery/cmd/login.go
```diff
@@ -63,13 +68,26 @@ You remain logged in until you explicitly log out using the 'logout' subcommand.
...
+			}
 			defer func() {
 				s, err := checkStatus()
 				if err != nil {
 					log.Warn().Err(err).Msg("could not run status command")
 				}
 				s.RenderCliStatus()
 			}()
+
+			if err == tokenExpiredErr {
```
slntopp, line 102: Doesn't prevent check command from running to check possible "clock skew" as per https://github.com/mondoohq/cnspec/issues/1342

apps/cnquery/cmd/login.go
```diff
@@ -105,15 +123,23 @@ func register(token string, annotations map[string]string, timer int, splay int,
...
+			return tokenValidationErr
 		} else {
 			if len(claims.Description) > 0 {
 				log.Info().Msg("token description: " + claims.Description)
 			}
 			if claims.IsExpired() {
-				log.Warn().Msg("token is expired")
+				return tokenExpiredErr
+			} else if claims.Expiry == nil {
+				log.Warn().Msg("token does not contain an expiry date")
```
slntopp, line 151: Adds handling for tokens with absent `exp` scope

apps/cnquery/cmd/login.go
```diff
@@ -26,6 +26,11 @@ import (
 	"go.mondoo.com/ranger-rpc/status"
 )
 
+var (
+	tokenValidationErr = errors.New("The token is not a valid token to register this client with Mondoo platform")
```
tas50:
```suggestion
	tokenValidationErr = errors.New("The token is not a valid token to register this client with Mondoo Platform")
```

tas50 (dismissed):

apps/cnquery/cmd/login.go
czunker: I don't think we need to care about the complete status at this point. We already know that the token would not work.

When you have an expired token and no clock skew:
- You get this error message
- clock skew test would past
- But you get another error message here: https://github.com/mondoohq/cnquery/blob/main/apps/cnquery/cmd/status.go#L129

When you have an expired token and clock skew:
- You get this error
- You get the status error about the clock skew
- And you get the error from L129

As I understand the code and your comment:
We know the token has expired. Therefore, we can not use it to perform a full status check, including upstream. We might only check the clock skew and the latest version to provide the user with as much actionable information as possible. What do you think?

apps/cnquery/cmd/login.go
```diff
@@ -105,15 +123,23 @@ func register(token string, annotations map[string]string, timer int, splay int,
...
 			}
 			if claims.IsExpired() {
-				log.Warn().Msg("token is expired")
+				return tokenExpiredErr
+			} else if claims.Expiry == nil {
+				log.Warn().Msg("token does not contain an expiry date")
 			} else {
 				log.Info().Msg("token will expire at " + claims.Claims.Expiry.Time().Format(time.RFC1123))
 			}
+			if claims.Space == "" {
```
czunker, line 155: The issue also mentions checking for `scope`.

slntopp: There is no `scope` claim in the claims

apps/cnquery/cmd/login.go
slntopp: Clock skew is detected by making the request (HealthProbe in this case) and comparing local clock to the server clock. Health probe should work with invalid token as well, and would highlight which API endpoint is attempted (issue we also had).

apps/cnquery/cmd/login.go
czunker: Perhaps the issue means this?
https://github.com/mondoohq/cnquery/blob/main/cli/config/config.go#L272-L287

slntopp: This is read from yml config / env variables and used if there is no token provided
Do you think we should check it like `if (config).Scope == "" && (tokenClaims).Space == "" { ... }` here?

czunker: I double-checked the issue. It is explicitly stating the claim and also has an example:
> If it is not a token that includes either scope or space claim, then issue the following error

@chris-rock : Can you please provide steps how you produced the token?

slntopp: Current Claims definition

```go
type VerifyClaim struct { // size=192 (0xc0)
    jwt.Claims
    CustomTokenClaims
}
// Embedded fields:
Issuer         string            // through Claims            
Subject        string            // through Claims            
Audience       jwt.Audience      // through Claims            
Expiry         *jwt.NumericDate  // through Claims            
NotBefore      *jwt.NumericDate  // through Claims            
IssuedAt       *jwt.NumericDate  // through Claims            
ID             string            // through Claims            
Space          string            // through CustomTokenClaims 
Description    string            // through CustomTokenClaims 
ApiEndpoint    string            // through CustomTokenClaims 
Labels         map[string]string // through CustomTokenClaims 
Owner          string            // through CustomTokenClaims 
CertValidUntil time.Time         // through CustomTokenClaims
```

czunker: Oh, ok. :+1:

czunker (approved):
Thanks @slntopp

PR #4587, opened 2024-08-20
🧹 improve the sagemaker notebook instance & iam mfadevice resource

vjeffrey:


providers/aws/resources/aws_iam.go
```diff
@@ -303,38 +303,46 @@ func (a *mqlAwsIam) virtualMfaDevices() ([]interface{}, error) {
...
+		awsIamUser, err := NewResource(a.MqlRuntime, "aws.iam.user", map[string]*llx.RawData{
+			"arn":  llx.StringDataPtr(a.cacheUserArn),
+			"name": llx.StringDataPtr(a.cacheUserName),
+		})
+		if err != nil {
+			return nil, err
+		}
+		return awsIamUser.(*mqlAwsIamUser), nil
+	}
+	a.User.State = plugin.StateIsNull
```
preslavgerchev: do we need to set this too in case we hit the cache?

preslavgerchev (approved):

PR #4591, opened 2024-08-22
make provider reconnectable

jaym:
Makes providers reconnectable. Keeps track of all connect requests to a provider. If it crashes, and is restarted, all the connect requests are replayed. Disconnected connections are also recreated because they can be a parent connect, and its non-existence can cause a child to fail to connect

### DIFF
--- a/providers/running_provider.go
+++ b/providers/running_provider.go
@@ -87,7 +87,7 @@ func (c *connectionGraph) topoSort() []uint32 {
 	return sorted
 }
 
-func (c *connectionGraph) garbageCollect() {
+func (c *connectionGraph) garbageCollect() []uint32 {
 	sorted := c.topoSort()
 
 	keep := map[uint32]struct{}{}
@@ -95,12 +95,16 @@ func (c *connectionGraph) garbageCollect() {
 		keep[node] = struct{}{}
 	}
 
+	collected := []uint32{}
 	for node := range c.nodes {
 		if _, ok := keep[node]; !ok {
+			collected = append(collected, node)
 			delete(c.nodes, node)
 			delete(c.edges, node)
 		}
 	}
+
+	return collected
 }
 
 type ReconnectFunc func() (pp.ProviderPlugin, *plugin.Client, error)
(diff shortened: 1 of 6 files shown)


imilchev:
I think this is going to be leaking connections for long running instance (serve mode). If we always reconnect all connections, they won't be closed properly. The scanner will only close connections that were just scanned.

I think this may be solved easily. If you remove the `connectRequest` when `Disconnect` is called. We call `Disconnect` when we "gracefully" shut down the connection. If that happens, it should be safe to drop the `connectRequest` too for this connection. Parent - child connections are tracked already so it's safe to assume we won't call `Disconnect` for a parent before the child is scanned

jaym:
I've updated things so that we keep track of which connections rely on which other connections. This information is used to reconnect. If we disconnect a connection and something else relies on it, we keep around the reconnection information until the thing that relies on it is disconnected

providers/running_provider.go
```diff
@@ -4,22 +4,249 @@
...
+	"google.golang.org/protobuf/proto"
 )
 
+type connectionGraphNode struct {
+	connected bool
+	data      connectReq
+}
+
+type connectionGraph struct {
+	nodes map[uint32]connectionGraphNode
```
imilchev, line 49: could you add some more comments what we store in these properties? It's not immediately clear for me when I see it, so it may help us in case we need to change this in the future

imilchev: it would also help to describe in text how this is intended to work and put it in a comment:
- when do we add connections to the graph?
- when do we remove connections from the graph?

imilchev (commented):
One general question that I have is what happens in the following case:
1. I connect to the provider with a root asset
2. I connect to the provider with child asset of that root
3. I scan the root
4. Root connection is closed
5. Provider crashes
6. We trigger a restart + reconnect

After step 6 is done what connection will be open for the restarted provider?

jaym:
> One general question that I have is what happens in the following case:
> 
>     1. I connect to the provider with a root asset
> 
>     2. I connect to the provider with child asset of that root
> 
>     3. I scan the root
> 
>     4. Root connection is closed
> 
>     5. Provider crashes
> 
>     6. We trigger a restart + reconnect
> 
> 
> After step 6 is done what connection will be open for the restarted provider?

So assuming there is another asset that needs to be scanned, and it has a connection that depends on the root, first the root connection is recreated, and then the next assets connection. https://github.com/mondoohq/cnspec/pull/1412 has the change to try to make sure the providers are up before starting an asset scan.

The root connection is then disconnected automatically once all the connections that depend on it are disconnected

providers/running_provider.go
jaym: Sure, will add comments here

imilchev (approved):

PR #4620, opened 2024-08-30
🧹 improve aws s3 buckets resource

vjeffrey:
i also wanted to switch to using HeadBucket to get the location but the api apparently requires that api to be called for the correct region to get a response, so it ends up being more api calls to do so


providers/aws/resources/aws.lr
```diff
@@ -1717,7 +1717,7 @@ aws.s3 @defaults("buckets") {
 }
 
 // Amazon S3 bucket
-private aws.s3.bucket @defaults("name location public") {
```
tas50: Is there to do this without losing the location?

vjeffrey: unfortunately i dont think so..
:/

if aws returned the region on the list response this wouldnt be a problem, the extra api call is annoying

chris-rock: Could we derive the location from the arn?

chris-rock (approved):

providers/aws/resources/aws.lr
vjeffrey: unfortunately it's not in the arn for s3. i could also remove the location change stuff and just leave the rest of the optimization in

vjeffrey: chatted with tim, im gonna bring back the location by default part

PR #4622, opened 2024-09-02
✨ use go 1.23

imilchev:


slntopp:
Why are we updating version in `go.mod`?
> That would make dependant software and packages with older go versions incompatible

imilchev:
> Why are we updating version in `go.mod`?
> 
> > That would make dependant software and packages with older go versions incompatible

that's a good point but we always did it with previous updates as well. I think this is something we can discuss internally for future update but I don't want to block the current upgrade.

PR #4624, opened 2024-09-02
🧹 Improving the keyVault resources to cover automatic key rotation policy

HRouhani:
- Ensure Automatic Key Rotation is Enabled Within Azure Key Vault for the Supported Services

Screenshot from 2024-09-02 15-57-47

### DIFF
--- a/providers/azure/resources/azure.lr
+++ b/providers/azure/resources/azure.lr
@@ -1542,6 +1542,8 @@ private azure.subscription.keyVaultService.key @defaults("kid keyName") {
   version() string
   // List of key versions
   versions() []azure.subscription.keyVaultService.key
+  // Auto-rotation enabled status
+  autoRotationEnabled bool
 }
 
 // Azure Key Vault certificate
--- a/providers/azure/resources/azure.lr
+++ b/providers/azure/resources/azure.lr
@@ -1514,6 +1514,16 @@ private azure.subscription.keyVaultService.vault @defaults("vaultName type vault
   secrets() []azure.subscription.keyVaultService.secret
   // Vault diagnostic settings
   diagnosticSettings() []azure.subscription.monitorService.diagnosticsetting
+  // Auto-rotation enabled status for all keys
+  autorotation() []azure.subscription.keyVaultService.key.autorotation
+}
+
+// Azure Key Vault key auto-rotation
+private azure.subscription.keyVaultService.key.autorotation @defaults("enabled") {
+  // Key ID (Key Identifier)
+  kid string
+  // Auto-rotation enabled status
+  enabled bool
 }
 
 // Azure Key Vault key
@@ -1542,8 +1552,6 @@ private azure.subscription.keyVaultService.key @defaults("kid keyName") {
   version() string
   // List of key versions
   versions() []azure.subscription.keyVaultService.key
-  // Auto-rotation enabled status
-  autoRotationEnabled bool
 }
 
 // Azure Key Vault certificate
--- a/providers/azure/resources/keyvault.go
+++ b/providers/azure/resources/keyvault.go
@@ -189,8 +189,62 @@ func (a *mqlAzureSubscriptionKeyVaultServiceVault) keys() ([]interface{}, error)
 		}
 
 		for _, entry := range page.Value {
+			mqlAzure, err := CreateResource(a.MqlRuntime, "azure.subscription.keyVaultService.key",
+				map[string]*llx.RawData{
+					"kid":           llx.StringDataPtr((*string)(entry.KID)),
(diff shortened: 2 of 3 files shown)


providers/azure/resources/azure.lr
```diff
@@ -1538,6 +1538,8 @@ private azure.subscription.keyVaultService.key @defaults("kid keyName") {
   version() string
   // List of key versions
   versions() []azure.subscription.keyVaultService.key
+  // Auto-rotation enabled status
```
vjeffrey: could we make this an optional field and only call the api if needed? (add the parens here and move the call to a function)

misterpantz: If you do keep it like it is, please change the description to:

```suggestion
  // Whether auto-rotation is enabled
```

Thanks!

HRouhani: @vjeffrey I did as you requested, however I still could not solve the uniqueness of the keys:

Screenshot from 2024-09-04 14-30-09

vjeffrey: oh, is there maybe an id function missing? i haven't re-checked the code, i'll look when i get online in a while, but usually that's an id thing

HRouhani: Thanks VJ, I already added id function as here:
```
func (a *mqlAzureSubscriptionKeyVaultServiceKeyAutorotation) keyName() (string, error) {
	id := a.Kid.Data
	kvid, err := parseKeyVaultId(id)
	if err != nil {
		return "", err
	}

	return kvid.Name, nil
}
```

providers/azure/resources/keyvault.go
```diff
@@ -211,6 +211,71 @@ func (a *mqlAzureSubscriptionKeyVaultServiceVault) keys() ([]interface{}, error)
...
 }
 
+func (a *mqlAzureSubscriptionKeyVaultServiceKeyAutorotation) keyName() (string, error) {
+	id := a.Kid.Data
+	kvid, err := parseKeyVaultId(id)
+	if err != nil {
+		return "", err
+	}
+
+	return kvid.Name, nil
```
vjeffrey, line 221: is the name guaranteed unique? if not, could just use the id itself here

vjeffrey (approved):

providers/azure/resources/keyvault.go
HRouhani: yes, it is always unique!

providers/azure/resources/azure.lr
```diff
@@ -1510,6 +1510,16 @@ private azure.subscription.keyVaultService.vault @defaults("vaultName type vault
   secrets() []azure.subscription.keyVaultService.secret
   // Vault diagnostic settings
   diagnosticSettings() []azure.subscription.monitorService.diagnosticsetting
+  // Auto-rotation enabled status for all keys
+  autorotation() []azure.subscription.keyVaultService.key.autorotation
```
preslavgerchev, line 1518: shouldn't this be under the key resource and not the vault resource?

providers/azure/resources/azure.lr
preslavgerchev: the function should be named `id()` for this to work. alternatively you can just pass in the id directly in the args as `__id`

providers/azure/resources/azure.lr
HRouhani: I am happy you raised that, it was our plan, but in this way we get following error:

```
# go.mondoo.com/cnquery/v11/providers/azure/resources
resources/azure.lr.go:14246:12: c.autorotation undefined (type *mqlAzureSubscriptionKeyVaultServiceKey has no field or method autorotation, but does have field Autorotation)
make[1]: *** [Makefile:339: providers/build/azure] Error 1
``` 

if we do:
```
// Azure Key Vault key
private azure.subscription.keyVaultService.key @defaults("kid keyName") {
  // Key ID
  kid string
  // Key tags
  tags map[string]string
  // Whether the key is managed
  managed bool
  // Whether the key is enabled
  enabled bool
  // Date the key begins to be usable
  notBefore time
  // Date the key expires
  expires time
  // Key creation time
  created time
  // Key last update time
  updated time
  // Key recovery level
  recoveryLevel string
  // Key name
  keyName() string
  // Key version
  version() string
  // List of key versions
  versions() []azure.subscription.keyVaultService.key
  // Auto-rotation enabled status for all keys
  autorotation() []azure.subscription.keyVaultService.key.autorotation
}
``

providers/azure/resources/azure.lr
HRouhani: @preslavgerchev  awesome, Thanks. It works now. Appreciate you help.

PR #4629, opened 2024-09-03
🐛 Handle Chassis returned as slice

slntopp:


providers/os/resources/smbios/win.go
```diff
@@ -14,7 +14,7 @@ import (
 const smbiosWindowsScript = `
 $bios = Get-WmiObject -class Win32_Bios
 $baseboard = Get-WmiObject Win32_BaseBoard
-$chassis = Get-WmiObject Win32_SystemEnclosure
+$chassis = @(Get-WmiObject Win32_SystemEnclosure)
```
slntopp, line 17: Note on PowerShell syntax:
- The `@()` syntax is used to (force) create an array in PowerShell.

slntopp:
Can be tested with this script:

slntopp:
Also worth noting, as documentation suggests, all `Get-WmiObject` calls may return multiple instances (slice)

https://powershell.one/wmi/root/cimv2/win32_systemenclosure

providers/os/resources/smbios/win.go
```diff
@@ -112,11 +112,11 @@ func (s *WindowsSmbiosManager) Info() (*SmBiosInfo, error) {
 			SerialNumber: winBios.BaseBoard.SerialNumber,
 			Version:      winBios.BaseBoard.Version,
 		},
-		ChassisInfo: ChassisInfo{
-			Vendor:       winBios.Chassis.Manufacturer,
-			Model:        toString(winBios.Chassis.Model),
-			Version:      winBios.Chassis.Version,
-			SerialNumber: winBios.Chassis.SerialNumber,
+		ChassisInfo: ChassisInfo{ // TODO: Might want to make this a slice
+			Vendor:       winBios.Chassis[0].Manufacturer,
```
preslavgerchev, line 120: can we check if `winBios.Chassis` is not empty? else this will panic

providers/os/resources/smbios/win.go
```diff
@@ -112,11 +116,11 @@ func (s *WindowsSmbiosManager) Info() (*SmBiosInfo, error) {
 			SerialNumber: winBios.BaseBoard.SerialNumber,
 			Version:      winBios.BaseBoard.Version,
 		},
-		ChassisInfo: ChassisInfo{
-			Vendor:       winBios.Chassis.Manufacturer,
-			Model:        toString(winBios.Chassis.Model),
-			Version:      winBios.Chassis.Version,
-			SerialNumber: winBios.Chassis.SerialNumber,
+		ChassisInfo: ChassisInfo{ // TODO: Might want to make this a slice
```
preslavgerchev, line 119: can this be nil? if `winBios.Chassis` is 0, we may as well jst set this to nil. we're currently filling in an empty struct that holds no value. if this cannot be nil, then thats ok

slntopp: It can't technically be, since it's not a pointer
Changing the type at this point would be even more breaking than making it slice, which would've been better solution imo

preslavgerchev (approved):

PR #4632, opened 2024-09-05
⭐️ Adding conditional Access - Named location part

HRouhani:
- Ensure Trusted Locations Are Defined 

Screenshot from 2024-09-10 17-11-22

### DIFF
--- a/providers/ms365/resources/conditional-access.go
+++ b/providers/ms365/resources/conditional-access.go
@@ -8,6 +8,7 @@ import (
 	"log"
 
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
+	"go.mondoo.com/cnquery/v11/llx"
 	"go.mondoo.com/cnquery/v11/providers/ms365/connection"
 )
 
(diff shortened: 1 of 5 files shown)


providers/ms365/resources/ms365.lr
```diff
@@ -56,6 +56,12 @@ microsoft.tenant @defaults("name") {
   subscriptions() []dict
 }
 
+// Microsoft Conditional Access Policies
+private microsoft.conditionalAccess {
+  // Return the name of the first named location
```
misterpantz:
```suggestion
  // First named location
```

providers/ms365/resources/conditional-access.go
```diff
@@ -0,0 +1,59 @@
...
+				if err != nil {
+					return nil, err
+				}
+				locationDetails = append(locationDetails, locationInfo)
+			}
+		}
+	}
+
+	if len(locationDetails) == 0 {
+		log.Println("No named locations are defined.")
```
jaym: i think we use zerolog for logging in this project. Also, if this is a normal condition it probably doesn't need to be logged

HRouhani: Done, Thanks for the feedback

jaym (approved):

PR #4633, opened 2024-09-06
⭐️ Adding consent settings  - ms365 

HRouhani:
Screenshot from 2024-09-06 17-27-16

### DIFF
--- a/providers/ms365/resources/ms365.lr
+++ b/providers/ms365/resources/ms365.lr
@@ -501,7 +501,10 @@ microsoft.policies {
   adminConsentRequestPolicy() dict
   // Permission grant policies
   permissionGrantPolicies() []dict
+  // Consent policy settings
+  consentPolicySettings() dict
 }
+
 
 // Deprecated: use `microsoft.roles` instead
 microsoft.rolemanagement {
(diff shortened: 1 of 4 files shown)


providers/ms365/resources/ms365.lr
```diff
@@ -487,8 +487,11 @@ microsoft.policies {
   adminConsentRequestPolicy() dict
   // Permission grant policies
   permissionGrantPolicies() []dict
+  // Consent policy settings
+  consentPolicySettings() dict
 }
 
+
```
tas50, line 508:
```suggestion
```

vjeffrey (approved):

providers/ms365/resources/ms365.lr
```diff
@@ -487,6 +487,8 @@ microsoft.policies {
   adminConsentRequestPolicy() dict
   // Permission grant policies
   permissionGrantPolicies() []dict
+  // Consent policy settings
+  consentPolicySettings() dict
```
tas50, line 505: @HRouhani how hard would it be to do this with a dedicated resource instead so we'd have documentation for each field the user could query?

HRouhani: It was my initial plan, but not easy as I couldn't find the proper document to see the type of each field, plus it is very dynamic what we get and if we want to add all of them, we need to add over 20 fields.

PR #4652, opened 2024-09-10
Remove documention for removed fields

tas50:
These fields have been removed already


chris-rock (commented):
Thank you for the cleanup

chris-rock (approved):

PR #4663, opened 2024-09-16
🧹 guard against nil tag keys or values in aws resources

vjeffrey:


providers/aws/resources/aws.lr
```diff
@@ -967,15 +967,15 @@ private aws.sagemaker.notebookinstance @defaults("arn name") {
   // Name of the notebook instance
   name string
   // Details about the notebook
-  details() aws.sagemaker.notebookinstance.details
+  details() aws.sagemaker.notebookinstancedetails
```
preslavgerchev, line 970: this might be a breaking change for stored assets/resolved policies, im not 100% though

vjeffrey: @jaym do you know? ^^

we're not changing anything about how we call the resource, but we are changing the actual resource name

vjeffrey: chatted with jay in discord, he said this would get fixed on re-scan if it was an issue, so this is a non-breaking change

preslavgerchev (approved):

PR #4666, opened 2024-09-18
✨ Improve ec2 discovery: do not fetch instances twice, use tags to filter

preslavgerchev:
* use passed in tags as filters
* use passed in instance ids as filters

### DIFF
--- a/providers/aws/resources/aws_ec2.go
+++ b/providers/aws/resources/aws_ec2.go
@@ -764,35 +764,17 @@ func (a *mqlAwsEc2) instances() ([]interface{}, error) {
 	return res, nil
 }
 
-func (a *mqlAwsEc2) getImdsv2Instances(ctx context.Context, svc *ec2.Client, filterName string) ([]ec2types.Reservation, error) {
+func (a *mqlAwsEc2) getEc2Instances(ctx context.Context, svc *ec2.Client, tags map[string]string) ([]ec2types.Reservation, error) {
 	res := []ec2types.Reservation{}
 	nextToken := aws.String("no_token_to_start_with")
 	params := &ec2.DescribeInstancesInput{
-		Filters: []ec2types.Filter{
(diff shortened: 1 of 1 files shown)


providers/aws/resources/aws_ec2.go
```diff
@@ -911,7 +872,7 @@ func (a *mqlAwsEc2) gatherInstanceInfo(instances []ec2types.Reservation, imdsvVe
 				"ebsOptimized":          llx.BoolDataPtr(instance.EbsOptimized),
 				"enaSupported":          llx.BoolDataPtr(instance.EnaSupport),
 				"httpEndpoint":          llx.StringData(string(instance.MetadataOptions.HttpEndpoint)),
-				"httpTokens":            llx.StringData(httpTokens),
+				"httpTokens":            llx.StringData(string(instance.MetadataOptions.HttpTokens)),
```
preslavgerchev: this is why we were doing it twice, it's now returned as a response

providers/aws/resources/aws_ec2.go
```diff
@@ -933,6 +896,10 @@ func (a *mqlAwsEc2) gatherInstanceInfo(instances []ec2types.Reservation, imdsvVe
 				"tpmSupport":            llx.StringDataPtr(instance.TpmSupport),
 			}
 
+			if instance.MetadataOptions != nil {
+				args["httpEndpoint"] = llx.StringData(string(instance.MetadataOptions.HttpEndpoint))
```
vjeffrey, line 900: should we make these fields optional in the lr?

vjeffrey (approved):

providers/aws/resources/aws_ec2.go
preslavgerchev: i dont think we can make those optional in the LR, but i did ensure that they are nil if the metadata struct is also nil (setting them to llx.NilData`

PR #4671, opened 2024-09-21
✨ add resource for azure functions & WAF config

vjeffrey:
```
azure.subscription { network  { applicationGateways { wafConfiguration }}}
```

```
azure.subscription { web { apps { functions { * } } }}
```

### DIFF
--- a/providers/azure/resources/azure.lr
+++ b/providers/azure/resources/azure.lr
@@ -70,6 +70,20 @@ azure.subscription @defaults ("name") {
   iot() azure.subscription.iotService
 }
 
+// Azure function
+private azure.subscription.webService.function @defaults("name") {
+  // ID of the function
+  id string
+  // Name of the function
+  name string
+  // Type of function
+  type string
+  // Kind of function
+  kind string
+  // Properties for the function
+  properties dict
+}
+
 // Azure resource group
 private azure.subscription.resourcegroup @defaults("name location") {
   // Resource group ID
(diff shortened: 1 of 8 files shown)


providers/azure/resources/azure.lr.manifest.yaml
```diff
@@ -1470,6 +1483,10 @@ resources:
       properties: {}
       tags: {}
       type: {}
+      wafConfig:
```
tas50: is this intention to have both?

providers/azure/resources/azure.lr
```diff
@@ -846,6 +860,22 @@ azure.subscription.networkService.applicationGateway @defaults("id name location
   properties dict
   // Gets the attached application firewall policy
   policy() azure.subscription.networkService.applicationFirewallPolicy
+  // WAF configurations
+  wafConfiguration() []azure.subscription.networkService.wafConfig
+}
+
+// Azure Application Firewall Config
+azure.subscription.networkService.wafConfig @defaults("id name") {
```
tas50: if type is just a string it might be helpful here too

providers/azure/resources/azure.lr
```diff
@@ -70,6 +70,20 @@ azure.subscription @defaults ("name") {
   iot() azure.subscription.iotService
 }
 
+// Azure function
+private azure.subscription.webService.function @defaults("name") {
```
tas50: type may be helpful here too

providers/azure/resources/azure.lr.manifest.yaml
preslavgerchev: nope, likely a rename that snuck its way into the manifest, removed

preslavgerchev (approved):

PR #4672, opened 2024-09-22
⭐️ Optimized fields for github provider

jaym:
- There are now fields to fetch the code of conduct, security file, and support file. These should create less garbage in cnquery
- There is a specific field for admin collaborators so we do not need to list all the collaborators if we just care about admins

### DIFF
--- a/providers/github/resources/github.lr
+++ b/providers/github/resources/github.lr
@@ -350,6 +350,8 @@ private github.repository @defaults("fullName") {
   codeOfConduct() github.file
   // Repository support file
   support() github.file
+  // Repository security file
+  security() github.file
 }
 
 // GitHub license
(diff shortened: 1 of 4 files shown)


providers/github/resources/github.lr
```diff
@@ -344,6 +346,12 @@ private github.repository @defaults("fullName") {
   closedIssues() []github.issue
   // Repository license
   license() github.license
+  // Repository code of conduct
+  codeOfConduct() github.file
```
tas50: Can we get file in the names so it's more obvious what you're getting back from the field name

slntopp (approved):

providers/github/resources/github.lr
```diff
@@ -380,6 +388,8 @@ private github.file @defaults("name type") {
   content() string
   // File download URL
   downloadUrl string
+  // File exists in repository
```
misterpantz:
```suggestion
  // Whether the file exists in the repository
```
This is just the style we follow for bools.

PR #4691, opened 2024-09-25
🧹 Remove discovery filters parameter from Discover

VasilSirakov:
#### In this PR:
- Removed `DiscoveryFilters` from `Discover(...)` function and use the parameters from the `AwsConnection` instead.
- Fixed parsing of inventory config discovery filters into `DiscoveryFilters` struct.
- Added new cases for ECS filters to be specified from inventory because they were only passed down from the removed parameter in `Discovery(...)`.
- Removed duplicate filtering of EC2 instances inside the AWS discovery that was already done by `GetInstances()`.
- Added corresponding exclude filters in `Ec2DiscoveryFilters`.
- Filter out EC2 instances if they match the values specified in the exclude filters. `ExcludeRegions` happens before the call to AWS, `ExcludeTags` and `ExcludeInstanceIds` happen after (AWS doesn't support server-side filtering for those).
- Added unit tests for a couple of functions.

#### Note:
Changing the signature for `Discovery` means everyone using the AWS provider and wanting to filter discovered instances must specify these filters in their inventory config discovery filters map.

### DIFF
--- a/providers/aws/resources/aws_ec2.go
+++ b/providers/aws/resources/aws_ec2.go
@@ -803,6 +803,9 @@ func (a *mqlAwsEc2) getInstances(conn *connection.AwsConnection) []*jobpool.Job 
 	if len(conn.Filters.Ec2DiscoveryFilters.Regions) > 0 {
 		regions = conn.Filters.Ec2DiscoveryFilters.Regions
 	}
+	for _, regionToExclude := range conn.Filters.Ec2DiscoveryFilters.ExcludeRegions {
+		regions = removeElement(regions, regionToExclude)
+	}
 	for _, region := range regions {
 		regionVal := region
 		f := func() (jobpool.JobResult, error) {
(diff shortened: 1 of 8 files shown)


providers/aws/resources/aws_ec2.go
```diff
@@ -803,6 +803,9 @@ func (a *mqlAwsEc2) getInstances(conn *connection.AwsConnection) []*jobpool.Job
 	if len(conn.Filters.Ec2DiscoveryFilters.Regions) > 0 {
 		regions = conn.Filters.Ec2DiscoveryFilters.Regions
 	}
+	for _, regionToExclude := range conn.Filters.Ec2DiscoveryFilters.ExcludeRegions {
+		regions = removeElement(regions, regionToExclude)
```
preslavgerchev: can we just adjust this to a single function that takes the regions and the regions to exclude and returns a new list?

VasilSirakov:
After a chat with Preslav, we decided it'd be nicer to have the filters as comma separated values as opposed to individual keys. So, we went from `{ ec2:instance-id:id1 = id1 , ec2:instance-id:id2 = id2 }` to `{ ec2:instance-ids = id1,id2 }`. Did that for all applicable filters excl. tags.

Also, added the parsing of CLI flags to opts and now the CLI should have the same functionality.

vjeffrey (approved):
nice, thank you!

PR #4692, opened 2024-09-25
🧹 improve cvss version detection

chris-rock:
Improves the CVSS version detection for vector strings without a prefix score like `CVSS:3.1/AV:N/AC:L/PR:H/UI:N/S:U/C:L/I:L/A:N`. Going forward we also need to add the calculation of the score itself.


czunker (approved):

czunker:
Thanks @chris-rock

PR #4693, opened 2024-09-25
🧹 Allow printing data queries without an assessment

jaym:
Boolean data queries will have an assessment. When we display them as data queries, we do not want an assessment

### DIFF
--- a/cli/printer/mql.go
+++ b/cli/printer/mql.go
@@ -17,15 +17,7 @@ import (
 	"golang.org/x/exp/slices"
 )
 
-// Results prints a full query with all data points
-// NOTE: ensure that results only contains results that match the bundle!
-func (print *Printer) Results(bundle *llx.CodeBundle, results map[string]*llx.RawResult) string {
-	assessment := llx.Results2Assessment(bundle, results)
-
-	if assessment != nil {
-		return print.Assessment(bundle, assessment)
-	}
-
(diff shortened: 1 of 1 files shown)


cli/printer/mql.go
```diff
@@ -38,6 +30,18 @@ func (print *Printer) Results(bundle *llx.CodeBundle, results map[string]*llx.Ra
...
+// Results prints a full query with all data points
+// NOTE: ensure that results only contains results that match the bundle!
+func (print *Printer) Results(bundle *llx.CodeBundle, results map[string]*llx.RawResult) string {
+	assessment := llx.Results2Assessment(bundle, results)
+
+	if assessment != nil {
+		return print.Assessment(bundle, assessment)
+	}
+
+	return print.Datas(bundle, results)
```
preslavgerchev, line 42: just to double check: is this PR just a restructure/cleanup or are there functional changes?

from what I see the only thing that happens here is that we've moved the for loop into its own `Datas` func that still gets called if `assessment` is nil

preslavgerchev: ah, i see the use case in https://github.com/mondoohq/cnspec/pull/1438 for extracting that

preslavgerchev (approved):

cli/printer/mql.go
jaym: This just splits the function into 2 so that it can be used from cnspec

PR #4694, opened 2024-09-25
✨Providers SDK: Templatable License Header

slntopp:
- generate license header from template
- added flags respectively
- add test assertion for license


providers-sdk/v1/lr/headers.go
```diff
@@ -0,0 +1,31 @@
+package lr
+
+import (
+	"strings"
+	"text/template"
+)
+
+const defaultLicenseHeaderTpl = `{{.LineStarter}} Copyright (c) Mondoo, Inc.
```
preslavgerchev, line 8: do we need the line starter to be a separate, specifiable param? why not just bake it in?

slntopp: We have yaml and we have go files that are generated, so `//` vs `#` for comments
Imo this is also more extendable if we every need that in the future

preslavgerchev (approved):

PR #4696, opened 2024-09-26
🧹 Refactor computerInfo Function to Use PowerShell Command Directly instead of PSGetComputerInfo (windows)

HRouhani:
The `PSGetComputerInfo`  no longer returns any results. This PR refactors the `computerInfo` function to directly use a PowerShell command (`Get-ComputerInfo`) instead, ensuring consistent output and resolving the issue with missing results.

### DIFF
--- a/providers/os/resources/windows.go
+++ b/providers/os/resources/windows.go
@@ -4,8 +4,11 @@
 package resources
 
 import (
+	"encoding/json"
 	"errors"
+	"io"
 
+	"github.com/rs/zerolog/log"
 	"go.mondoo.com/cnquery/v11/llx"
 	"go.mondoo.com/cnquery/v11/providers-sdk/v1/plugin"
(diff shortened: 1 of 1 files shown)


providers/os/resources/windows.go
```diff
@@ -17,16 +20,35 @@ import (
 func (s *mqlWindows) computerInfo() (map[string]interface{}, error) {
 	conn := s.MqlRuntime.Connection.(shared.Connection)
 
-	cmd := windows.PSGetComputerInfo
```
imilchev, line 20: this function should already execute Get-ComputeInfo in case the other function fails. There must be something else wrong. I am trying to test

imilchev:
this is not addressing the root cause for the issue. It's only a workaround and in the current form will break existing functionality for all clients. We need to come up with a solution specifically for ssh without changing other clients

PR #4708, opened 2024-09-30
🧹 allow os-provider to be builtin via config

arlimus:
Previously the os provider had to be inlined via uncommenting it in the `builtin_dev.go` file manually. This is no longer necessary, you can just add it to the builtin providers in the `providers.yaml` and have `make providers/config` take care of the rest.

```yaml
builtin: [os]
```


vjeffrey (approved):

imilchev:
I thought this always worked. That's how I debugged and tested things... What didn't work before this PR exactly?

PR #4712, opened 2024-10-04
Add Ansible long-form CLI help

misterpantz:
Adds long-form CLI help for the Ansible provider.
From what I could see in the code, there are no subcommands and only the path is required. So this is one super simple. Did I miss anything, @chris-rock? 
Thanks!


tas50 (approved):

tas50:
Tested locally. LGTM

PR #4719, opened 2024-10-07
⭐️ device manager: multi partition support

slntopp:


providers/os/connection/snapshot/blockdevices.go
```diff
@@ -151,31 +151,55 @@ func (blockEntries BlockDevices) FindDevice(name string) (BlockDevice, error) {
 }
 
 // Searches all the partitions in the device and finds one that can be mounted. It must be unmounted, non-boot partition
-// If multiple partitions meet this criteria, the largest one is returned.
-func (device BlockDevice) GetMountablePartition() (*PartitionInfo, error) {
+func (device BlockDevice) GetMountablePartitions(includeBoot bool) ([]*PartitionInfo, error) {
```
preslavgerchev: nit: 
```suggestion
func (device BlockDevice) GetMountablePartitions(includeAll bool) ([]*PartitionInfo, error) {
```

you're not just including the boot, but any arbitrary partitions that we find

providers/os/connection/device/device_connection.go
```diff
@@ -54,73 +56,106 @@ func NewDeviceConnection(connId uint32, conf *inventory.Config, asset *inventory
...
-	}
+		scanDir, err := manager.Mount(block)
+		if err != nil {
+			log.Error().Err(err).Msg("unable to complete mount step")
+			res.Close()
+			return nil, err
+		}
+		res.MountedDirs = append(res.MountedDirs, scanDir)
+
+		if conf.Options == nil {
```
preslavgerchev: we can probably move this out of the for loop

providers/os/connection/device/device_connection.go
```diff
@@ -54,73 +56,106 @@ func NewDeviceConnection(connId uint32, conf *inventory.Config, asset *inventory
...
+			conf.Options = make(map[string]string)
+		}
 
-	res.FileSystemConnection = fsConn
+		// create and initialize fs provider
+		conf.Options["path"] = scanDir
```
preslavgerchev, line 86: how would this work with multiple paths,  i think they will just overwrite one another? we're using th same conf here

providers/os/connection/device/linux/device_manager.go
```diff
@@ -52,11 +53,11 @@ func (d *LinuxDeviceManager) IdentifyMountTargets(opts map[string]string) ([]*sn
 		return []*snapshot.PartitionInfo{pi}, nil
 	}
 
-	pi, err := d.identifyViaDeviceName(opts[DeviceName])
+	partitions, err := d.identifyViaDeviceName(opts[DeviceName], opts[MountAllPartitions] == "true")
```
preslavgerchev, line 56: we should validate that device name is not empty if `MountAllPartitions` is true, we want to be explicit in what device we're targeting

providers/os/connection/snapshot/volumemounter.go
```diff
@@ -94,24 +95,43 @@ func (m *VolumeMounter) mountVolume(fsInfo *PartitionInfo) error {
...
-		return err
+
+	for name, dir := range m.ScanDirs {
+		log.Debug().
+			Str("dir", dir).
+			Str("name", name).
+			Msg("unmount volume")
+		if err := Unmount(dir); err != nil {
+			log.Error().Err(err).Msg("failed to unmount dir")
+			return err
```
preslavgerchev: can we try unmounting all first and then throw the err?

providers/os/connection/snapshot/volumemounter.go
```diff
@@ -94,24 +95,43 @@ func (m *VolumeMounter) mountVolume(fsInfo *PartitionInfo) error {
...
+	for name, dir := range m.ScanDirs {
+		log.Debug().
+			Str("dir", dir).
+			Str("name", name).
+			Msg("remove created dir")
+		if err := os.RemoveAll(dir); err != nil {
+			log.Error().Err(err).
+				Str("dir", dir).
+				Msg("failed to remove dir")
+			return err
```
preslavgerchev: same as the unmount comment

providers/os/connection/device/linux/device_manager.go
slntopp: device name is validated inside `identifyViaDeviceName`

providers/os/connection/device/device_connection.go
slntopp: This option is used only to create FileSystemConnection. I can make the options copy, just that'd be "more costly"

providers/os/connection/device/linux/device_manager.go
preslavgerchev: but it can also return if the name is empty, it then does a best-guess search

if mountall is true we should explicitly validate device name is not empty

preslavgerchev:
can we also verify this doesn't break current behaviour of the device connection?

slntopp:
```shell
[ec2-user@ip-172-31-15-93 ~]$ sudo ./cnquery scan --inventory-file inventory.yaml --verbose --output=yaml 2>/dev/null 1> vmtdish.log
[ec2-user@ip-172-31-15-93 ~]$ sudo cnquery scan --inventory-file inventory.yaml --verbose --output=yaml 2>/dev/null 1> upstream.log
[ec2-user@ip-172-31-15-93 ~]$ diff vmtdish.log upstream.log 
2,3c2,3
<   //explorer.api.mondoo.com/assets/2nA7f5QahsmDZ7THearJtBAFrjb:
<     mrn: //explorer.api.mondoo.com/assets/2nA7f5QahsmDZ7THearJtBAFrjb
---
>   //explorer.api.mondoo.com/assets/2nA7gObNb6y08Wvh6VfdRmfuRN5:
>     mrn: //explorer.api.mondoo.com/assets/2nA7gObNb6y08Wvh6VfdRmfuRN5
6c6
<   //explorer.api.mondoo.com/assets/2nA7f5QahsmDZ7THearJtBAFrjb:
---
>   //explorer.api.mondoo.com/assets/2nA7gObNb6y08Wvh6VfdRmfuRN5:
```

inventory:
```yaml
spec:
  assets:
  - name: vmtd
    platform_ids:
    - //platformid.api.mondoo.app/runtime/google-cloud-vmtd/
    connections:
    - type: device
      options:
        device-name: /dev/sdz
        mount-all-partitions: true
        verbose: true
```

providers/os/connection/device/linux/device_manager.go
slntopp: But if device name is empty, it will return the best-guess, mount-all will be ignored:

https://github.com/mondoohq/cnquery/blob/834ade4ca2ca6413f9225a81a2860670fa142329/providers/os/connection/device/linux/device_manager.go#L133-L142

PR #4720, opened 2024-10-07
Add long-form help for arista provider

misterpantz:


providers/arista/config/config.go
```diff
@@ -19,6 +19,12 @@ var Config = plugin.Provider{
 			Name:      "arista",
 			Use:       "arista user@host",
 			Short:     "an Arista EOS device",
+			Long: `Use the arista provider to query resources on an Arista EOS device
+
+Example:
+  cnquery shell arista <user@host>
+			'
```
tas50: ?

tas50 (approved):

PR #4726, opened 2024-10-08
🐛 Fix regex for matching RPM packages with unorthodox vendor name

### LINKED ISSUE #4714: Linux: os.packages resource doesn't return any data on SUSE 15 OS
Tested on SuSE15

`packages` just returns a list containing the mondoo packages on that system:
```
packages.list: [
  0: package name="cnquery" version="11.24.0-1"
  1: package name="cnspec" version="11.24.0-1"
  2: package name="mondoo" version="11.24.0-1"
]
```

It still works on Debian-type OS.

image

VasilSirakov:
Fixes #4714 

SuSE's vendor name changed for some reason from `SUSE LLC` to `SUSE LLC <https://www.suse.com/>` which no longer matched the regex pattern.

### In this PR:
- Added `/`, `:`, `<` and `>` as allowed characters in the vendor regex pattern.
- Cleaned up vendor name after the fact and removed this tag at the earliest opportunity. It is not particularly useful and was causing trouble with CPEs generation.
    - Used regex to remove any string of characters from the vendor name between a closed pair of angle brackets `<...>`. That way, we are agnostic as to what the contents of this tag are and we do not hardcode some arbitrary string which may change in the future. This assumes that angle brackets are not used in some other way in the vendor name which I think is a fair assumption given that this is an edge case.
- Added unit tests to verify that the packages are correctly parsed and the regex trims out the tag correctly.

Manually tested this on a VM running SuSE 15.6 and `packages.list` produces a correct output:
image

@chris-rock Let me know in case any of my assumptions are invalid and you'd like to handle this differently.

### DIFF
--- a/providers/os/resources/packages/rpm_packages.go
+++ b/providers/os/resources/packages/rpm_packages.go
@@ -30,7 +30,7 @@ const (
 	RpmPkgFormat = "rpm"
 )
 
-var RPM_REGEX = regexp.MustCompile(`^([\w-+]*)\s(\d*|\(none\)):([\w\d-+.:]+)\s([\w\d]*|\(none\))__([\w\d\s,\.]+)__(.*)$`)
+var RPM_REGEX = regexp.MustCompile(`^([\w-+]*)\s(\d*|\(none\)):([\w\d-+.:]+)\s([\w\d]*|\(none\))__([\w\d\s,/<>:\.]+)__(.*)$`)
 
 // ParseRpmPackages parses output from:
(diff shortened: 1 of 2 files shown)


VasilSirakov:
@chris-rock Let me know in case I should review this with someone else.

chris-rock:
@VasilSirakov work with @czunker on this PR

czunker (approved):
Thanks @VasilSirakov

PR #4732, opened 2024-10-09
✨ find device by trailing letters

slntopp:
```
Before Change
BenchmarkFindDevice-8           1000000000               0.0000609 ns/op               0 B/op          0 allocs/op

Naive
BenchmarkFindDevice-8           1000000000               0.0001089 ns/op               0 B/op          0 allocs/op

Optimised
BenchmarkFindDevice-8           1000000000               0.0000595 ns/op               0 B/op          0 allocs/op
```

### DIFF
--- a/providers/os/connection/snapshot/blockdevices.go
+++ b/providers/os/connection/snapshot/blockdevices.go
@@ -129,43 +129,49 @@ func (blockEntries BlockDevices) GetMountablePartitionByDevice(device string) (*
 // Searches for a device by name
 func (blockEntries BlockDevices) FindDevice(requested string) (BlockDevice, error) {
 	log.Debug().Str("device", requested).Msg("searching for device")
-	trailing := requested[len(requested)-1:]
-	trailingMatched := []BlockDevice{}
 
-	// LongestMatchingSuffix returns the length of the longest common suffix of two strings
-	lms := func(s1, s2 string) int {
-		n1 := len(s1)
-		n2 := len(s2)
+	requestedName := strings.TrimPrefix(requested, "/dev/")
+
+	// LongestMatchingSuffix returns the length of the longest common suffix of requested and provided string
+	lmsCache := map[string]int{}
+	lms := func(s string) int {
+		if v, ok := lmsCache[s]; ok {
+			return v
+		}
+
+		n1 := len(requested)
+		n2 := len(s)
 
 		// Start from the end of both strings
 		i := 0
-		for i < int(math.Min(float64(n1), float64(n2))) && s1[n1-i-1] == s2[n2-i-1] {
+		for i < int(math.Min(float64(n1), float64(n2))) && requested[n1-i-1] == s[n2-i-1] {
 			i++
 		}
+
+		lmsCache[s] = i
 		return i
 	}
 
-	for i := range blockEntries.BlockDevices {
-		d := blockEntries.BlockDevices[i]
-		log.Debug().Str("name", d.Name).Interface("children", d.Children).Interface("mountpoint", d.MountPoint).Msg("found block device")
-		fullDeviceName := "/dev/" + d.Name
-		if fullDeviceName == requested {
-			log.Debug().Str("name", d.Name).Msg("found matching device")
-			return d, nil
-		}
+	sorted := false
+	devices := blockEntries.BlockDevices
 
-		if d.Name[len(d.Name)-1:] == trailing {
-			trailingMatched = append(trailingMatched, d)
+	for !sorted {
+		sorted = true
+
+		for i := 0; i < len(devices)-1; i++ {
+			if devices[i].Name == requestedName {
+				return blockEntries.BlockDevices[i], nil
+			}
+
+			if lms(devices[i].Name) < lms(devices[i+1].Name) {
+				devices[i], devices[i+1] = devices[i+1], devices[i]
+				sorted = false
+			}
 		}
 	}
 
-	sort.SliceStable(trailingMatched, func(i, j int) bool {
-		return lms(trailingMatched[i].Name, requested) > lms(trailingMatched[j].Name, requested)
-	})
-
-	if len(trailingMatched) > 0 {
-		log.Debug().Str("name", trailingMatched[0].Name).Msg("found matching device")
-		return trailingMatched[0], nil
+	if lms(devices[0].Name) > 0 {
+		return devices[0], nil
 	}
 
 	return BlockDevice{}, fmt.Errorf("no block device found with name %s", requested)
--- a/providers/os/connection/snapshot/blockdevices.go
+++ b/providers/os/connection/snapshot/blockdevices.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


providers/os/connection/snapshot/blockdevices.go
```diff
@@ -126,28 +127,57 @@ func (blockEntries BlockDevices) GetMountablePartitionByDevice(device string) (*
...
-		end := strings.TrimPrefix(name, "/dev/sd")
-		secondName = "/dev/xvd" + end
+func (blockEntries BlockDevices) FindDevice(requested string) (BlockDevice, error) {
+	log.Debug().Str("device", requested).Msg("searching for device")
+
+	requestedName := strings.TrimPrefix(requested, "/dev/")
+
+	lmsCache := map[string]int{}
+	// LongestMatchingSuffix returns the length of the longest common suffix of requested and provided string
+	lms := func(s string) int {
```
preslavgerchev: can we extract this to a private fn here? it should also be easier to test as then we only need to provide strings

providers/os/connection/snapshot/blockdevices_test.go
```diff
@@ -118,6 +118,73 @@ func TestGetMountablePartitionByDevice(t *testing.T) {
 	})
 }
 
+func BenchmarkFindDevice(b *testing.B) {
```
preslavgerchev: curious: why do we need to benchmark this? we dont expect this to iterate over a lot of partitions

slntopp: well, implementation looks a bit bloated, wanted to remove any concerns before they are raised 😁

providers/os/connection/snapshot/blockdevices.go
```diff
@@ -125,29 +126,59 @@ func (blockEntries BlockDevices) GetMountablePartitionByDevice(device string) (*
...
 
-// Searches for a device by name
-func (blockEntries BlockDevices) FindDevice(name string) (BlockDevice, error) {
-	log.Debug().Str("device", name).Msg("searching for device")
-	var secondName string
-	if strings.HasPrefix(name, "/dev/sd") {
-		// sdh and xvdh are interchangeable
-		end := strings.TrimPrefix(name, "/dev/sd")
-		secondName = "/dev/xvd" + end
+func lms(lmsCache map[string]int, s1, s2 string) int {
```
preslavgerchev: can you add some doc to this function: whats the cache and the return val? also can you extend the name? its not clear to me what `lms` stands for just by looking at it

providers/os/connection/snapshot/blockdevices_test.go
preslavgerchev: i dont think we need the benchmark testing here, we're not running this against a large amount of elements

providers/os/connection/snapshot/blockdevices.go
```diff
@@ -126,7 +126,9 @@ func (blockEntries BlockDevices) GetMountablePartitionByDevice(device string) (*
 	return &PartitionInfo{Name: devFsName, FsType: partitions[0].FsType}, nil
 }
 
-func lms(lmsCache map[string]int, s1, s2 string) int {
+// LongestMatchingSuffix returns the length of the longest common suffix of two strings
+// and caches the result (lengths of the matching suffix) for future calls with the same string
+func LongestMatchingSuffix(lmsCache map[string]int, s1, s2 string) int {
```
preslavgerchev: curious: do we need the caching? whats the use case for trying to fetch this twice for the same string?

slntopp: We compare the lengths of two findings in the loop (because FindDevice actually sorts the devices per lms), and then we check it one more time in the end
LongestMatchingSuffix does iterate over devices names itself
I get it that this function isn't called often and strings aren't that long, but just why make it slower from start

providers/os/connection/snapshot/blockdevices.go
```diff
@@ -53,9 +76,43 @@ func (cmdRunner *LocalCommandRunner) GetBlockDevices() (*BlockDevices, error) {
...
 }
 
+func (blockEntries *BlockDevices) FindAliases() {
+	entries, err := os.ReadDir("/dev")
+	if err != nil {
+		log.Warn().Err(err).Msg("Can't read /dev directory")
+		return
+	}
+
+process_symlinks:
```
preslavgerchev: if we were to read the entries and then pass those to a function that looks like `findAlias(block BlockDevice, entries []fs.Info)`, would that eliminate the need for the labeled loop? i think we should avoid this

providers/os/connection/snapshot/blockdevices_test.go
```diff
@@ -12,6 +12,54 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestBlockDevicesUnamrshal(t *testing.T) {
```
preslavgerchev: nit:
```suggestion
func TestBlockDevicesUnmarshal(t *testing.T) {
```

nice test!

providers/os/connection/snapshot/blockdevices.go
```diff
@@ -125,29 +182,70 @@ func (blockEntries BlockDevices) GetMountablePartitionByDevice(device string) (*
...
+
 // Searches for a device by name
-func (blockEntries BlockDevices) FindDevice(name string) (BlockDevice, error) {
-	log.Debug().Str("device", name).Msg("searching for device")
-	var secondName string
-	if strings.HasPrefix(name, "/dev/sd") {
-		// sdh and xvdh are interchangeable
-		end := strings.TrimPrefix(name, "/dev/sd")
-		secondName = "/dev/xvd" + end
+func (blockEntries BlockDevices) FindDevice(requested string) (BlockDevice, error) {
```
preslavgerchev, line 209: can we update the tests to also include cases for alias matching?

providers/os/connection/snapshot/blockdevices.go
```diff
@@ -125,29 +182,70 @@ func (blockEntries BlockDevices) GetMountablePartitionByDevice(device string) (*
...
-		d := blockEntries.BlockDevices[i]
-		log.Debug().Str("name", d.Name).Interface("children", d.Children).Interface("mountpoint", d.MountPoint).Msg("found block device")
-		fullDeviceName := "/dev/" + d.Name
-		if name != fullDeviceName { // check if the device name matches
-			if secondName == "" || secondName != fullDeviceName {
-				continue
+
+	requestedName := strings.TrimPrefix(requested, "/dev/")
+	lmsCache := map[string]int{}
+	bestMatch := struct {
```
preslavgerchev, line 219: can we simplify this a bit? i feel like there's a way to make this a bit more readable by iterating over all devices and storing their lms and then taking the one with the longest lms at the end, something like:
```
lmsMap := map[string]int{}
	devicesMap := map[string]BlockDevice{}
	for _, d := range devices {
		devicesMap[d.Name] = d
		if d.Name == requestedName {
			return d, nil
		}

		lms := LongestMatchingSuffix(requested, d.Name)
		for _, alias := range d.Aliases {
			// we also try all the aliases and see if those are a better match
			aliasLms := LongestMatchingSuffix(requested, alias)
			if aliasLms > lms {
				lms = aliasLms
			}
		}
		lmsMap[d.Name] = lms
	}

	device := BlockDevice{}
	for name, lms := range lmsMap {
		if lms > 0 && lms > lmsMap[device.Name] {
			device = devicesMap[name]
		}
	}

	// if we have a name, we've found a device
	if device.Name != "" {
		return devicesMap[device.Name], nil
	}
```

that avoids the need for a private struct and the iteration, starting from the 1st index

i have not actually tested this code so it could be wrong

providers/os/connection/snapshot/blockdevices.go
```diff
@@ -125,29 +191,74 @@ func (blockEntries BlockDevices) GetMountablePartitionByDevice(device string) (*
...
+		Device BlockDevice
+		Lms    int
+	}{
+		Device: BlockDevice{},
+		Lms:    0,
+	}
+
+	for i := 0; i < len(devices); i++ {
```
preslavgerchev: nit:
```suggestion
	for _, d := range devices {
```

preslavgerchev (approved):
nice one @slntopp

PR #4733, opened 2024-10-10
Update AWS long-form help

misterpantz:


providers/aws/config/config.go
```diff
@@ -18,9 +18,40 @@ var Config = plugin.Provider{
 	ConnectionTypes: []string{provider.DefaultConnectionType, string(awsec2ebsconn.EBSConnectionType)},
 	Connectors: []plugin.Connector{
 		{
-			Name:    "aws",
-			Use:     "aws",
-			Short:   "an AWS account",
+			Name:  "aws",
+			Use:   "aws",
+			Short: "an AWS account",
+			Long: `Use the aws provider to query the resources in an AWS Organization.
```
tas50: I believe this should be account since we're doing account scans

providers/aws/config/config.go
```diff
@@ -18,9 +18,40 @@ var Config = plugin.Provider{
...
+	cnspec scan aws
+	cnquery scan aws -f mondoo-aws-incident-response.mql.yaml --querypack mondoo-incident-response-aws
+	cnquery shell aws --role <role-arn>
+	cnspec scan aws ec2 instance-connect <user@host>
+	cnspec scan aws ec2 instance-connect <user@host> --identity-file <path>
+	cnspec scan aws ec2 ebs <snapshot-id>
+	cnspec scan aws --filters region=ap-south-1
+
+Notes:
+  If you set the ATLASSIAN_ADMIN_TOKEN environment variable, you can omit the profile flag.
```
tas50: Wrong env var here

misterpantz: DOH!

tas50 (approved):

PR #4742, opened 2024-10-11
✨ Add support for comma-separated values on the same key for EC2 tag filters

### LINKED ISSUE #4739: Add support for multi-value EC2 tag filtering
First and foremost, we should investigate if AWS even supports this type of filtering on multiple values for the same tag.

#### Description:
We currently support include/exclude filtering when querying EC2 instances, however, each tag is restricted to a single value. We should add support for multiple values per tag. An example use-case could be `Name: {value}` where the same key `Name` is used with different values.

The simplest way to implement this would be to specify the values as a comma-separated list that would look something along the lines of `ec2:tag:key1=val1,val2,val3` and `exclude:ec2:tag:key1=val1,val2,val3`. Same is already being done for instance identifiers and regions.

VasilSirakov:
Closes #4739 

This adds the possibility of specifying comma-separated values for the same key in a include/exclude tag filter. Those filters can now be specified as follows:
```
ec2:tag:Name=name1,name2,name3
```
and
```
exclude:ec2:tag:Name=name1,name2,name3
```

We keep treating the filter as a single `key : value` so this is backwards compatible. We just expand the string value right before applying the filters.

#### Note:
There is a quirk with how filters are applied. Comma-separated values on the same key are applied with a logical `OR`. If we specify 2 tag filters on different keys, they are applied with a logical `AND`.

### DIFF
--- a/providers/aws/resources/aws_ec2.go
+++ b/providers/aws/resources/aws_ec2.go
@@ -774,7 +774,7 @@ func (a *mqlAwsEc2) getEc2Instances(ctx context.Context, svc *ec2.Client, filter
 	for k, v := range filters.Tags {
 		params.Filters = append(params.Filters, ec2types.Filter{
 			Name:   aws.String(fmt.Sprintf("tag:%s", k)),
-			Values: []string{v},
+			Values: strings.Split(v, ","),
 		})
 	}
 	if len(filters.InstanceIds) > 0 {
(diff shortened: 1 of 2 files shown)


providers/aws/resources/aws_ec2.go
```diff
@@ -1787,10 +1787,12 @@ func shouldExcludeInstance(instance ec2types.Instance, filters connection.Ec2Dis
 		}
 	}
 	for k, v := range filters.ExcludeTags {
-		for _, iTag := range instance.Tags {
-			if iTag.Key != nil && *iTag.Key == k &&
-				iTag.Value != nil && *iTag.Value == v {
-				return true
+		for _, tagValue := range strings.Split(v, ",") {
```
preslavgerchev, line 1790: can we add a test for this?

providers/aws/resources/aws_ec2.go
```diff
@@ -1787,10 +1787,12 @@ func shouldExcludeInstance(instance ec2types.Instance, filters connection.Ec2Dis
...
+			for _, iTag := range instance.Tags {
+				if iTag.Key != nil && *iTag.Key == k &&
+					iTag.Value != nil && *iTag.Value == tagValue {
+					return true
+				}
```
philipbalinov, line 1795: nit: doesn't this for block essentially check whether `filters.ExcludeTags` and `instance.Tags` intersect? I'm no go expert, but there must be a built-in function for that ❓

VasilSirakov: I feel like in Go the idiomatic way to do most things with maps/slices is just for loops. But even if there was such a function, we are expanding the values of `filters.ExcludeTags` which can be comma-separated so it wouldn't apply here.

preslavgerchev (approved):

PR #4760, opened 2024-10-16
🐛 use short windows computer info command as fallback always

imilchev:
The error message about the command length doesn't need to be in English. It can be in any other language Windows supports, therefore checking explicitly the error text is not a feasible approach to capture all command length errors.

This PR changes the behaviour so we always use the short command as a fallback to any type of error


chris-rock:
fixes https://github.com/mondoohq/server/issues/9345

chris-rock (approved):

PR #4763, opened 2024-10-17
🐛 allow running shodan provider with inventory file

imilchev:
Using an inventory file which provides the Shodan token as a vault secret triggeres an error. This is now fixed and we can run scans with vault secrets for Shodan

### DIFF
--- a/providers/shodan/provider/provider.go
+++ b/providers/shodan/provider/provider.go
@@ -6,9 +6,10 @@ package provider
 import (
 	"context"
 	"errors"
-	"go.mondoo.com/cnquery/v11/providers-sdk/v1/vault"
 	"os"
 	"strings"
+
+	"go.mondoo.com/cnquery/v11/providers-sdk/v1/vault"
 
 	"go.mondoo.com/cnquery/v11/llx"
 	"go.mondoo.com/cnquery/v11/providers-sdk/v1/inventory"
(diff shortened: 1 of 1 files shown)


providers/shodan/provider/provider.go
```diff
@@ -51,10 +52,9 @@ func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error)
 	if token == "" {
 		token = os.Getenv("SHODAN_TOKEN")
 	}
-	if token == "" {
-		return nil, errors.New("no shodan token provided, use --token or SHODAN_TOKEN")
```
imilchev, line 55: This is also validated in the connection. We don't need to check it here. This makes sure inventory files with vault secrets are accepted. If no token is provided, we will error out when the connection is created https://github.com/mondoohq/cnquery/blob/ffa713536aed31ebed30c94ed015ff9d5c7d95ee/providers/shodan/connection/connection.go#L53-L55

imilchev, line 55: This is also validated in the connection. We don't need to check it here. This makes sure inventory files with vault secrets are accepted. If no token is provided, we will error out when the connection is created https://github.com/mondoohq/cnquery/blob/ffa713536aed31ebed30c94ed015ff9d5c7d95ee/providers/shodan/connection/connection.go#L53-L55

afiune (approved):

PR #4767, opened 2024-10-18
:star: Create AWS resource aws.ec2.instances {iamInstanceProfile}

mm-weber:
This adds the following resource:
instanceProfileResource

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -3014,6 +3014,8 @@ private aws.ec2.instance @defaults("instanceId region state instanceType archite
   instanceType string
   // Tags on the instance
   tags map[string]string
+  // Instance profile of the instance.
+  IamInstanceProfile() dict
   // Image that was used for the instance
   image() aws.ec2.image
   // Launch time of the instance
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -799,7 +799,7 @@ private aws.iam.user @defaults("arn name") {
 }
 
 // AWS IAM instanceProfile
-private aws.iam.instanceProfile @defaults("arn instanceProfileName") {
+private aws.iam.instanceProfile @defaults("arn instanceProfileId") {
   // ARN of the IAM instanceProfile
   arn string
   // Time when instanceProfile was created
@@ -3030,7 +3030,7 @@ private aws.ec2.instance @defaults("instanceId region state instanceType archite
   // Tags on the instance
   tags map[string]string
   // Instance profile of the instance.
-  iamInstanceProfile() dict
+  iamInstanceProfile() aws.iam.instanceProfile
   // Image that was used for the instance
   image() aws.ec2.image
   // Launch time of the instance
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -711,6 +711,7 @@ aws.iam {
   virtualMfaDevices() []aws.iam.virtualmfadevice
   // List of server certificates stored in IAM
   serverCertificates() []dict
+  instanceProfiles() []aws.iam.instanceProfile
 }
 
 // Entry in AWS IAM credential report
@@ -795,6 +796,19 @@ private aws.iam.user @defaults("arn name") {
   accessKeys() []dict
   // Login profile for the user
   loginProfile() aws.iam.loginProfile
+}
+
+private aws.iam.instanceProfile @defaults("arn instanceProfileName") {
+  // ARN of the IAM instanceProfile
+  arn string
+  // Time when instanceProfile was created
+  createDate time
+  // ID of the IAM instanceProfile
+  instanceProfileId string
+  // name of the IAM instanceProfile
+  instanceProfileName string
+  // Tags for the IAM user
+  tags map[string]string
 }
 
 // AWS IAM login profile for a user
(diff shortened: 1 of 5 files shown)


providers/aws/resources/aws.lr
```diff
@@ -3014,6 +3014,8 @@ private aws.ec2.instance @defaults("instanceId region state instanceType archite
   instanceType string
   // Tags on the instance
   tags map[string]string
+  // Instance profile of the instance.
+  IamInstanceProfile() dict
```
tas50: Shouldn't this be `iamInstanceProfile` per our naming?

mm-weber: Thanks @tas50 ! 
I updated the resource accordingly.
Works now as expected.

instanceProfileResource

providers/aws/resources/aws.lr
```diff
@@ -3014,6 +3014,8 @@ private aws.ec2.instance @defaults("instanceId region state instanceType archite
   instanceType string
   // Tags on the instance
   tags map[string]string
+  // Instance profile of the instance.
+  iamInstanceProfile() dict
```
vjeffrey: i think it would be a good idea to try to make this a real resource - even if all we show is this information here, the arn and id like we do now -- 
i say this bc we could easily tie together some stuff if we do so, like later on adding aws.iam.instanceprofiles list resource, which could list all instance profiles and have references to the roles attached to those instance profiles, which gives us some good iam role/policy checking that folks could test around for their ec2 instances

all that said, we can also add the dict in now and update later.. but if you have the time, if you could update this to a real resource (and still just return the arn and id like we do now, we dont need to fetch the full instance profile yet), that would set us up for the future, instead of having to later deprecate this one and add another that has the resource reference

vjeffrey (approved):
this works well!

i would prefer if the instance profile was a real resource instead of a dict, i left a longer comment explaining my thinking, but putting my +1 here now in case we need this quickly - it does the job, i just think if we make it a real resource we'll avoid a little bit of rework later

tas50:
+1 to having real resources vs. dicts. It makes for better auto completes and docs. Overall it's a better user experience even if it's more work for us.

providers/aws/resources/aws.lr
```diff
@@ -797,6 +798,20 @@ private aws.iam.user @defaults("arn name") {
   loginProfile() aws.iam.loginProfile
 }
 
+// AWS IAM instanceProfile
+private aws.iam.instanceProfile @defaults("arn instanceProfileId") {
+  // ARN of the IAM instanceProfile
+  arn string
+  // Time when instanceProfile was created
```
```suggestion
  // Time when the instance profile was created
```

providers/aws/resources/aws.lr
```diff
@@ -797,6 +798,20 @@ private aws.iam.user @defaults("arn name") {
...
+  createDate time
+  // ID of the IAM instanceProfile
+  instanceProfileId string
+  // name of the IAM instanceProfile
```
```suggestion
  // Name of the instance profile
```

providers/aws/resources/aws.lr
```diff
@@ -797,6 +798,20 @@ private aws.iam.user @defaults("arn name") {
   loginProfile() aws.iam.loginProfile
 }
 
+// AWS IAM instanceProfile
```
```suggestion
// AWS IAM instance profile
```

providers/aws/resources/aws.lr
```diff
@@ -797,6 +798,20 @@ private aws.iam.user @defaults("arn name") {
...
+private aws.iam.instanceProfile @defaults("arn instanceProfileId") {
+  // ARN of the IAM instanceProfile
```
```suggestion
  // ARN of the instance profile
```

providers/aws/resources/aws.lr
```diff
@@ -797,6 +798,20 @@ private aws.iam.user @defaults("arn name") {
...
+  arn string
+  // Time when instanceProfile was created
+  createDate time
+  // ID of the IAM instanceProfile
+  instanceProfileId string
+  // name of the IAM instanceProfile
+  instanceProfileName string
+  // Tags for the IAM user
```
misterpantz: This isn't tags for the instance profile? Just checking. :)

providers/aws/resources/aws.lr
```diff
@@ -3014,6 +3029,8 @@ private aws.ec2.instance @defaults("instanceId region state instanceType archite
   instanceType string
   // Tags on the instance
   tags map[string]string
+  // Instance profile of the instance.
```
```suggestion
  // Instance profile of the instance
```

misterpantz (commented):
Some docs suggestions for you, @mm-weber!

providers/aws/resources/aws.lr.manifest.yaml
```diff
@@ -2010,6 +2010,8 @@ resources:
     fields:
       arn: {}
       createDate: {}
+      iamRole: {}
```
tas50: This goes away now right?

providers/aws/resources/aws.lr
mm-weber: Good call, thanks @misterpantz !

providers/aws/resources/aws.lr.manifest.yaml
mm-weber: good catch @tas50 !

mm-weber:
Thank you all for your help on this @tas50 @misterpantz and especially @vjeffrey for walking me through her process !

Done some cleanup and fixed a little thing, think it's ready to go now.

PR #4776, opened 2024-10-24
✨ Streamline AWS discovery filters

VasilSirakov:
#### In this PR:
- Made include/exclude region filters general for all assets. Incorporated them in `.Regions()` function so that should be applied everywhere.
- Removed tag filter from the general discovery filters. It did not make sense to have the same tag filter for all discovered assets as that is way too restrictive and often times would conflict, e.g. if you filter by a tag you use for EC2 instances, that automatically ignores every other asset in your AWS account. Furthermore, many assets do not have tags (containers/accounts, etc) so you need to ensure you do not filter those out which makes having global tag filters difficult to implement properly and reason about. Likely in the future we can make the filters more modular and have dedicated filters for different assets.
- Modified the existing filter keys in an attempt to make them easier to use, e.g. `ec2:iid` instead of `ec2:instance-ids`. Also, changed `exclude:ec2:xyz` into `ec2:exclude:xyz` such that the resource type the filter applies to is at the start.
- Removed functions that were no longer applicable after refactor.
- Made some code improvements.
- Fixed bug where `DescribeClusters` and `DescribeTasks` were not retrieving tags from AWS. They have to be specifically requested.

### DIFF
--- a/providers/aws/connection/connection.go
+++ b/providers/aws/connection/connection.go
@@ -155,9 +155,9 @@ func parseOptsToFilters(opts map[string]string) DiscoveryFilters {
 		case k == "ec2:exclude:iid":
 			d.Ec2DiscoveryFilters.ExcludeInstanceIds = append(d.Ec2DiscoveryFilters.ExcludeInstanceIds, strings.Split(v, ",")...)
 		case strings.HasPrefix(k, "ec2:tag:"):
-			d.Ec2DiscoveryFilters.Tags[strings.TrimPrefix(k, "tag:")] = v
+			d.Ec2DiscoveryFilters.Tags[strings.TrimPrefix(k, "ec2:tag:")] = v
 		case strings.HasPrefix(k, "ec2:exclude:tag:"):
-			d.Ec2DiscoveryFilters.ExcludeTags[strings.TrimPrefix(k, "exclude:tag:")] = v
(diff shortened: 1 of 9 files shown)


providers/aws/connection/connection.go
```diff
@@ -125,38 +137,30 @@ func NewAwsConnection(id uint32, asset *inventory.Asset, conf *inventory.Config)
...
-		case k == "exclude:ec2:regions":
-			d.Ec2DiscoveryFilters.ExcludeRegions = append(d.Ec2DiscoveryFilters.ExcludeRegions, strings.Split(v, ",")...)
-		case k == "all:regions", k == "regions":
-			d.GeneralDiscoveryFilters.Regions = append(d.GeneralDiscoveryFilters.Regions, strings.Split(v, ",")...)
-		case k == "ec2:instance-ids":
+		case k == "regions":
+			d.DiscoveryFilters.Regions = append(d.DiscoveryFilters.Regions, strings.Split(v, ",")...)
+		case k == "exclude:regions":
+			d.DiscoveryFilters.ExcludeRegions = append(d.DiscoveryFilters.ExcludeRegions, strings.Split(v, ",")...)
+		case k == "ec2:iid":
```
preslavgerchev: shouldnt' this be `ec2:instance-ids`?

preslavgerchev (approved):

PR #4785, opened 2024-10-28
✨ Add platform PURLs

czunker:
Add the asset platform in package URL format.

### DIFF
--- a/providers/os/resources/purl/platform_purl_test.go
+++ b/providers/os/resources/purl/platform_purl_test.go
@@ -1,0 +1,178 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package purl
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.mondoo.com/cnquery/v11/providers-sdk/v1/inventory"
+)
+
+func TestNewPlatformPurl(t *testing.T) {
+	tests := []struct {
+		name     string
+		platform *inventory.Platform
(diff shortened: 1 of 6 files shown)


providers/os/resources/purl/platform_purl_test.go
```diff
@@ -0,0 +1,178 @@
...
+			want:    "pkg:os/esxi/@7.0.3?arch=x86_64&distro=esxi-7.0.3",
+			wantErr: "",
+		},
+		{
+			name: "kubernetes deployment",
+			platform: &inventory.Platform{
+				Name:    "k8s-deployment",
+				Version: "1.27",
+			},
+			want:    "pkg:os/k8s-deployment/@1.27?distro=k8s-deployment-1.27",
```
czunker: This looks wrong. It is not an os and not a distro.

providers/os/resources/purl/platform_purl_test.go
```diff
@@ -0,0 +1,178 @@
...
+			wantErr: "",
+		},
+		{
+			name: "aws platform",
+			platform: &inventory.Platform{
+				Name: "aws",
+			},
+			want:    "pkg:os/aws/?distro=aws",
```
czunker: Same for cloud providers.

Other terms I could think of:
- platform
- runtime

???

chris-rock: Let's switch to platform

chris-rock (approved):
Nice, lets try it

PR #4795, opened 2024-10-30
✨ Support searching /dev/disk/by-id aliases.

preslavgerchev:
This fixes the GCP snapshot scanning. In GCP devices are aliased under `/dev/disk-by-id/<device-name>`. This PR builds on top of the already supported alias finding by looking into that directory as well


slntopp:
I think it would've made more sense to just search in multiple paths instead of copying the method, since the logic is the same
1. Look for symlinks in the dir
2. Trim to base name
3. Look for matches

imilchev (approved):

PR #4798, opened 2024-10-30
🔨 new `logger.FuncDur()` + no cache expiration

afiune:
A couple of changes here:
    
**🧹 Do not expire the cache for github provider**

    We do not have a long running server where we need to clear our cache
    every once in a while, instead we execute a single scan and we expect it
    to finish (end execution) which ultimately clears the cache. For those
    long running executions, we want to persist the cache until we finish
    scanning.

    This change sets the github provider cache to No expiration.

**🔧 new `tracer.FuncDur()`**

    `FuncDur` must be used in a `defer` statement.

    This new helper function receives the function name and the time
    when a function started and logs out the time it took to run.

 ```go
package mypackage

func MyFunction() {
    defer tracer.FuncDur(time.Now(), "mypackage.MyFunction")

    ...
}
 ```

Signed-off-by: Salim Afiune Maya <afiune@mondoo.com>

### DIFF
--- a/providers-sdk/v1/util/tracer/func.go
+++ b/providers-sdk/v1/util/tracer/func.go
@@ -1,0 +1,28 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package tracer
+
+import (
+	"time"
+
+	"github.com/rs/zerolog/log"
+)
+
+// FuncDur must be used in a `defer` statement. It receives the function name and the time
+// when a function started and logs out the time it took to run.
+//
+// ```go
+//
+//	func MyFunction() {
+//		defer tracer.FuncDur(time.Now(), "mypackage.MyFunction")
+//
+//		...
+//	}
+//
+// ```
+func FuncDur(start time.Time, name string) {
+	log.Trace().Str("func", name).
+		TimeDiff("took", time.Now(), start).
+		Msgf("tracer.FuncDur>")
+}
--- a/providers/github/resources/github.go
+++ b/providers/github/resources/github.go
@@ -29,7 +29,7 @@ func getUser(ctx context.Context, runtime *plugin.Runtime, conn *connection.Gith
 	}
 	g := obj.(*mqlGithub)
 	if g.memoize == nil {
-		g.memoize = memoize.NewMemoizer(30*time.Minute, 1*time.Hour)
+		g.memoize = memoize.NewMemoizer(0, 0)
 	}
 	res, err, _ := g.memoize.Memoize("user-"+user, func() (interface{}, error) {
 		log.Debug().Msgf("fetching user %s", user)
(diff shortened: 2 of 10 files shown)


providers/github/resources/github.go
```diff
@@ -29,7 +29,7 @@ func getUser(ctx context.Context, runtime *plugin.Runtime, conn *connection.Gith
 	}
 	g := obj.(*mqlGithub)
 	if g.memoize == nil {
-		g.memoize = memoize.NewMemoizer(30*time.Minute, 1*time.Hour)
+		g.memoize = memoize.NewMemoizer(0, 0)
```
chris-rock: I am not sure we should go down this route. Ok to make it longer, but users can potentially use github provider with `cnspec serve`

providers-sdk/v1/util/tracer/func.go
```diff
@@ -0,0 +1,28 @@
...
+// ```go
+//
+//	func MyFunction() {
+//		defer tracer.FuncDur(time.Now(), "mypackage.MyFunction")
+//
+//		...
+//	}
+//
+// ```
+func FuncDur(start time.Time, name string) {
```
chris-rock: @jaym what do you think about using opentelemetry instead. This would allow us to use standard tracing?

jaym: I'm not sure how much overhead opentelemetry has, but my guess is that it would not be appropriate if you want to be measuring things in the mql execution. For things like us making remote calls, it could make sense. Would need to solve how you get spans from the providers back to the root process.

There's also https://pkg.go.dev/runtime/trace which I think should be low overhead when not tracing. And the docs say it plays nicely with go's trace tools. Not sure how we would handle forked processes.

Both these probably also require making sure context is properly handled and passed where it needs to be

providers/github/resources/github.go
afiune: Interesting, I wasn't aware of this feature. Thanks for catching this!  

Ideally, we should invalidate the cache on every "scan", but I don't see an easy way to do that,
so I'll just increase the expiration to 2 hours (4x).

providers-sdk/v1/util/tracer/func.go
afiune: This is a minimal implementation to avoid adding much overhead, I do agree with both of you that we should use opentelemetry, can we get this in and release to continue troubleshooting of some issues we have and swap this in the future?

providers/github/resources/github.go
imilchev: Every scan kills the provider (even is serve mode). We don't keep providers running indefinitely l. The moment the scan is done, all providers are killed. For the next scan, we start with a fresh set of providers

afiune:
Split this into:
* https://github.com/mondoohq/cnquery/pull/4799
* https://github.com/mondoohq/cnquery/pull/4800

PR #4800, opened 2024-10-31
🔨 new `logger.FuncDur()`

afiune:
`FuncDur` must be used in a `defer` statement.

This new helper function receives the function name and the time when the
function started and logs out the time it took to run.

Example:
```go
package mypackage

func MyFunction() {
   defer logger.FuncDur(time.Now(), "mypackage.MyFunction")

   ...
}
```


chris-rock (approved):
Thank you @afiune

PR #4804, opened 2024-11-01
✨ Read windows appx packages on a FS scan.

preslavgerchev:
Support looking for appx packages on a windows FS scan. We look for `appxmanifest.xml` files and parse those to determine the package info.


providers/os/resources/packages/windows_packages.go
```diff
@@ -38,20 +41,20 @@ const (
 // https://docs.microsoft.com/en-us/previous-versions/windows/desktop/ff357803(v=vs.85)
 var (
 	wsusClassificationGUID = map[string]WSUSClassification{
-		"5c9376ab-8ce6-464a-b136-22113dd69801 ": Application,
```
chris-rock, line 41: Thank you for the cleanup

providers/os/resources/packages/appx_manifest.go
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package packages
+
+import "encoding/xml"
+
+// AppxManifest represents the structure of an AppxManifest.xml file
+type AppxManifest struct {
```
chris-rock: I would like to see a test here where we parse a manifest file in go test.

providers/os/resources/packages/windows_packages.go
```diff
@@ -268,6 +283,30 @@ func (w *WinPkgManager) getInstalledApps() ([]Package, error) {
...
 }
 
+func (w *WinPkgManager) getAppxPackages() ([]Package, error) {
+	canRunCmd := w.conn.Capabilities().Has(shared.Capability_RunCommand)
+	// we always prefer to use the powershell command to get the appx packages, fallback to filesystem if not possible
+	if !canRunCmd && (w.conn.Type() == shared.Type_FileSystem || w.conn.Type() == shared.Type_Device) {
+		return w.getFsAppxPackages()
+	}
+
+	b, err := windows.Version(w.platform.Version)
```
chris-rock, line 293: I think we want to create a seperate function like `getPwshAppxPackages()`. Then the function is easier to read.

chris-rock (commented):
Great addition @preslavgerchev Lets add some tests and we are good to go.

providers/os/resources/packages/appx_manifest.go
preslavgerchev: done

providers/os/resources/packages/windows_packages.go
preslavgerchev: good point, done

providers/os/resources/packages/appx_manifest.go
preslavgerchev: also added a test for `toPackage` which convers the appxPackage to just a package

PR #4808, opened 2024-11-02
⭐️ support Windows 2025 on arm via SSH

chris-rock:
```
cnquery shell ssh administrator@192.168.64.3 --ask-pass
→ connected to Windows Server 2025 Datacenter

cnquery> asset { platform version build labels arch  }
asset: {
  build: "5000"
  version: "26311"
  arch: "ARM64"
  labels: {
    windows.mondoo.com/display-version: "24H2"
    windows.mondoo.com/product-type: "3"
  }
  platform: "windows"
}
cnquery> 
```

### DIFF
--- a/providers/os/detector/detector_win.go
+++ b/providers/os/detector/detector_win.go
@@ -13,7 +13,41 @@ import (
 	"go.mondoo.com/cnquery/v11/providers/os/registry"
 )
 
+// runtimeWindowsDetector uses powershell to gather information about the windows system
 func runtimeWindowsDetector(pf *inventory.Platform, conn shared.Connection) (bool, error) {
+	// most systems support wmi, but windows on arm does not ship with wmic, therefore we are trying to use windows
+	// builds from registry key first. If that fails, we try to use wmi
+	// see https://techcommunity.microsoft.com/t5/windows-it-pro-blog/wmi-command-line-wmic-utility-deprecation-next-steps/ba-p/4039242
+
+	if pf.Labels == nil {
+		pf.Labels = map[string]string{}
+	}
+
+	//  try to get build + ubr number (win 10+, 2019+)
+	current, err := win.GetWindowsOSBuild(conn)
+	if err == nil && current.UBR > 0 {
+		pf.Name = "windows"
+		pf.Title = current.ProductName
+		pf.Version = current.CurrentBuild
+		pf.Build = strconv.Itoa(current.UBR)
+		pf.Arch = current.Architecture
+
+		var productType string
+		switch current.ProductType {
+		case "WinNT":
+			productType = "1" // Workstation
+		case "ServerNT":
+			productType = "3" // Server
+		case "LanmanNT":
+			productType = "2" // Domain Controller
+		}
+
+		pf.Labels["windows.mondoo.com/product-type"] = productType
+		pf.Labels["windows.mondoo.com/display-version"] = current.DisplayVersion
+		return true, nil
+	}
+
+	// fallback to wmi if the registry key is not available
 	data, err := win.GetWmiInformation(conn)
 	if err != nil {
 		log.Debug().Err(err).Msg("could not gather wmi information")
@@ -29,19 +63,7 @@ func runtimeWindowsDetector(pf *inventory.Platform, conn shared.Connection) (boo
 
 	// FIXME: we need to ask wmic cpu get architecture
 	pf.Arch = data.OSArchitecture
-
-	if pf.Labels == nil {
-		pf.Labels = map[string]string{}
-	}
 	pf.Labels["windows.mondoo.com/product-type"] = data.ProductType
-
-	// optional: try to get the ubr number (win 10 + 2019)
-	current, err := win.GetWindowsOSBuild(conn)
-	if err != nil {
-		log.Debug().Err(err).Msg("could not parse windows current version")
-	} else if current.UBR > 0 {
-		pf.Build = strconv.Itoa(current.UBR)
-	}
 
 	return true, nil
 }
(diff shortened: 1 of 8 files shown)


providers/os/detector/testdata/detect-windows2019.toml
```diff
@@ -3,11 +3,16 @@ stdout = """Node,BootDevice,BuildNumber,BuildType,Caption,CodeSet,CountryCode,Cr
 VAGRANT,\\Device\\HarddiskVolume1,17763,Multiprocessor Free,Microsoft Windows Server 2019 Datacenter Evaluation,1252,1,Win32_OperatingSystem,Win32_ComputerSystem,,VAGRANT,-420,TRUE,TRUE,TRUE,3,FALSE,,FALSE,256,2,721716,979372,1922780,20190906065515.000000-420,,20190908011749.580533-420,20190908042731.608000-420,0409,Microsoft Corporation,4294967295,137438953344,{en-US},Microsoft Windows Server 2019 Datacenter Evaluation|C:\\Windows|\\Device\\Harddisk0\\Partition2,0,69,1,80,Vagrant,64-bit,1033,400,18,,,,,FALSE,TRUE,3,,00431-20000-00000-AA838,0,0,1179648,OK,400,\\Device\\HarddiskVolume2,C:\\Windows\\system32,C:,,3276340,2096692,10.0.17763,C:\\Windows
 """
 
-[commands."powershell -c \"Get-ItemProperty -Path 'HKLM:\\SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion' -Name CurrentBuild, UBR, EditionID | ConvertTo-Json\""]
+[commands."55dbc0e9b838caa11145eed07f6e73644bda27bf65a0c58a52291f9a18384481"]
```
philipbalinov, line 6: why are we replacing the verbose subsection name with a sha?

chris-rock: Because this is a longer script now, we use the script hash.

providers/os/detector/detector_win.go
```diff
@@ -13,7 +13,41 @@ import (
...
+		pf.Name = "windows"
+		pf.Title = current.ProductName
+		pf.Version = current.CurrentBuild
+		pf.Build = strconv.Itoa(current.UBR)
+		pf.Arch = current.Architecture
+
+		var productType string
+		switch current.ProductType {
+		case "WinNT":
+			productType = "1" // Workstation
```
preslavgerchev, line 38: is there any reason why we use the digits here and not `Workstation`, `Server`, etc?

providers/os/detector/windows/build_version.go
```diff
@@ -82,8 +87,15 @@ func ParseWinRegistryCurrentVersion(r io.Reader) (*WindowsCurrentVersion, error)
 
 // powershellGetWindowsOSBuild runs a powershell script to retrieve the current version from windows
 func powershellGetWindowsOSBuild(conn shared.Connection) (*WindowsCurrentVersion, error) {
-	pscommand := "Get-ItemProperty -Path 'HKLM:\\SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion' -Name CurrentBuild, UBR, EditionID | ConvertTo-Json"
-	cmd, err := conn.RunCommand(powershell.Wrap(pscommand))
+	pscommand := `
+$info = Get-ItemProperty -Path 'HKLM:\SYSTEM\CurrentControlSet\Control\ProductOptions' -Name ProductType
```
preslavgerchev, line 91: nit: why not use our existing way of fetching registry keys to build this? the only thing thats extra here seems to be the processor architecture which can be a separate smaller pwsh cmd to fetch

preslavgerchev (approved):

providers/os/detector/detector_win.go
chris-rock: I think it makes sense to change this, but this way it is compatible to the current implementation. We need to check and adjust policies like https://github.com/mondoohq/cnspec-policies/blob/main/core/mondoo-windows-11-compatibility.mql.yaml#L22-L23 to use a new label based approach.

providers/os/detector/detector_win.go
```diff
@@ -13,7 +13,41 @@ import (
...
+	// builds from registry key first. If that fails, we try to use wmi
+	// see https://techcommunity.microsoft.com/t5/windows-it-pro-blog/wmi-command-line-wmic-utility-deprecation-next-steps/ba-p/4039242
+
+	if pf.Labels == nil {
+		pf.Labels = map[string]string{}
+	}
+
+	//  try to get build + ubr number (win 10+, 2019+)
+	current, err := win.GetWindowsOSBuild(conn)
+	if err == nil && current.UBR > 0 {
```
afiune, line 28: Would be nice to log this error even though we are falling back to wmi

providers/os/detector/windows/build_version.go
chris-rock: I was just adjusting the current implementation with minimal change.  I agree, that we should use the new registry key implementation for this. Since I have no time to do this, I recommend we do this in a follow up PR.

providers/os/detector/detector_win.go
afiune: Same question, specially since PS command returns a string as a product type and here we are using a number as a string? 🤔

providers/os/detector/detector_win.go
chris-rock: Feel free to add this to the PR.

providers/os/detector/detector_win.go
afiune: Let's do this with https://github.com/mondoohq/cnquery/issues/4815

providers/os/detector/windows/build_version.go
afiune: Let's do this with https://github.com/mondoohq/cnquery/issues/4815

PR #4811, opened 2024-11-04
✨ make `components.Select()` generic

afiune:
This change is adding a new `Selector` interface that the function `components.Select()`
accepts so that we can select any type of slice.

Additionally, I improved the `components.List()` function to use the `tabwriter` when
running things without a TTY.

### Added Simple Examples

Screenshot 2024-11-04 at 11 10 21 AM


cli/components/_examples/selector/main.go
imilchev, line 1: why not have this added to a readme or some other development doc?

cli/components/list_raw_object.go
```diff
@@ -0,0 +1,65 @@
...
+	"fmt"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/rs/zerolog/log"
+	"go.mondoo.com/cnquery/v11/cli/theme"
+)
+
+// Object is the interface that a list need to implement so we can display its objects.
+type Object interface {
```
imilchev: can we have a more specific name for this interface? I find this one too generic making it unclear what it's purpose is. Perhaps call it `CliListableItem` or something similar?

cli/components/list_raw_object.go
```diff
@@ -0,0 +1,65 @@
...
+//		list := components.List(theme.OperatingSystemTheme, "string", customStrings)
+//		fmt.Printf(list)
+//	}
+//
+// ```
+func List[O Object](theme *theme.Theme, listType string, list []O) string {
+	b := &strings.Builder{}
+	w := tabwriter.NewWriter(b, 1, 1, 1, ' ', tabwriter.TabIndent)
+
+	log.Debug().Msgf("discovered %d %s(s)", len(list), listType)
```
imilchev: i think this message makes no sense here. The discovery mechanism is related to asset discovery when starting a scan. You are building a generic function that can be used for any type of object (not necessarily an asset).

cli/components/selector.go
```diff
@@ -0,0 +1,64 @@
...
+	"fmt"
+	"os"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/rs/zerolog/log"
+)
+
+// Selector is the interface that items need to implement so that we can select them.
+type Selector interface {
+	HumanName() string
```
imilchev: nit: weird name. Perhaps call it `FriendlyName` or similar.
Would be nice to add documentation what this function is used for

cli/components/selector.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"fmt"
+	"os"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/rs/zerolog/log"
+)
+
+// Selector is the interface that items need to implement so that we can select them.
+type Selector interface {
```
imilchev: this one also seems to generic. `CliSelectableItem` maybe?

cli/components/list_raw_object.go
```diff
@@ -0,0 +1,65 @@
...
+
+	for i := range list {
+		assetObj := list[i]
+
+		for i, key := range assetObj.PrintableKeys() {
```
imilchev: I am not sure I am in favour of this construction. We are building a generic way of displaying lists of items and selecting an item from that list. I think the way an item from the list is displayed, should be defined outside of this function. The current implementation seems like it provides some "customization" options but it is actually very limited.

If you don't need the keys and values individually, why not just make the interface for `Object` have a function `Print() string` or something similar, and just call this here. Then it is up to the `Object` to specify how it is printed on the CLI. We also remove the `PrintableKeys` and `PrintableValue`.

cli/components/_examples/selector/main.go
afiune: good idea! I can add some docs.

This example is useful to test this interactive code, you can run it with
```
$ go run cli/components/_examples/selector/main.go
```

I'll add it to the docs. 👍🏽

cli/components/list_raw_object.go
afiune: Will change. 👍🏽

cli/components/list_raw_object.go
afiune: I disagree.

Leaving the formatting to the implementation (caller) is not an option, if we want that, we should not have this function at all.

The whole point of these helpers is to help and standardize how we print/interact with "stuff".

cli/components/selector.go
afiune: I didn't want to re-implement the function for the assets, since it already exists and, it is the only place we use it (for now).

If we are ok with a new function, then I'll go with `Display()`

cli/components/selector.go
afiune: Sure thing.

imilchev (approved):

cli/components/list_raw_object.go
afiune: Great catch!

PR #4820, opened 2024-11-06
✨device manager: extended partitions info

slntopp:


providers/os/connection/device/device_connection.go
```diff
@@ -90,6 +95,8 @@ func NewDeviceConnection(connId uint32, conf *inventory.Config, asset *inventory
 		}
 	}
 
+	log.Info().Any("partitions", res.partitions).Msg("device connection> == PARTITIONS ==")
```
preslavgerchev: leftover

preslavgerchev (approved):

providers/os/connection/device/device_connection.go
slntopp: good catch, thx

PR #4823, opened 2024-11-07
✨ fstab parser

slntopp:
```
cnquery run -c "fstab.entries { * }"
fstab.entries: [
  0: {
    fsck: 1
    options: "defaults,noatime"
    device: "UUID=7c4e7e0e-ce36-42f9-b456-16f78f3a1eb1"
    fstype: "xfs"
    mountpoint: "/"
    dump: 1
  }
  1: {
    fsck: 2
    options: "defaults,noatime,uid=0,gid=0,umask=0077,shortname=winnt,x-systemd.automount"
    device: "UUID=C155-24D2"
    fstype: "vfat"
    mountpoint: "/boot/efi"
    dump: 0
  }
]
```

#### Device connection example
```
cnquery run device -c "fstab.entries { * }" --device-name sdby
fstab.entries: [
  0: {
    fsck: 1
    options: "discard,commit=30,errors=remount-ro"
    device: "LABEL=cloudimg-rootfs"
    fstype: "ext4"
    mountpoint: "/"
    dump: 0
  }
  1: {
    fsck: 2
    options: "defaults"
    device: "LABEL=BOOT"
    fstype: "ext4"
    mountpoint: "/boot"
    dump: 0
  }
  2: {
    fsck: 1
    options: "umask=0077"
    device: "LABEL=UEFI"
    fstype: "vfat"
    mountpoint: "/boot/efi"
    dump: 0
  }
]
```

### DIFF
--- a/providers/os/resources/fstab.go
+++ b/providers/os/resources/fstab.go
@@ -15,7 +15,13 @@ import (
 
 func (f *mqlFstab) entries() ([]any, error) {
 	conn := f.MqlRuntime.Connection.(shared.Connection)
-	fstabFile, err := conn.FileSystem().Open("/etc/fstab")
+
+	fs := conn.FileSystem()
+	if fs == nil {
+		return nil, errors.New("filesystem not available")
+	}
+
+	fstabFile, err := fs.Open("/etc/fstab")
 	if err != nil {
 		return nil, err
 	}
--- a/providers/os/resources/fstab.go
+++ b/providers/os/resources/fstab.go
@@ -36,7 +36,10 @@ func initFstab(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[strin
 }
 
 func (f *mqlFstab) entries() ([]any, error) {
-	conn := f.MqlRuntime.Connection.(shared.Connection)
+	conn, ok := f.MqlRuntime.Connection.(shared.Connection)
+	if !ok {
+		return nil, errors.New("wrong connection type")
+	}
 
 	fs := conn.FileSystem()
 	if fs == nil {
--- a/providers/os/resources/fstab.go
+++ b/providers/os/resources/fstab.go
@@ -10,8 +10,30 @@ import (
 	"strings"
 
 	"go.mondoo.com/cnquery/v11/llx"
+	"go.mondoo.com/cnquery/v11/providers-sdk/v1/plugin"
 	"go.mondoo.com/cnquery/v11/providers/os/connection/shared"
 )
+
+func initFstab(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error) {
+	if x, ok := args["path"]; ok {
+		path, ok := x.Value.(string)
+		if !ok || path == "" {
+			path = "/etc/fstab"
+		}
+
+		f, err := CreateResource(runtime, "fstab", map[string]*llx.RawData{
+			"path": llx.StringData(path),
+		})
+		if err != nil {
+			return nil, nil, err
+		}
+		args["path"] = llx.StringData(path)
+		return args, f, nil
+	}
+
+	args["path"] = llx.StringData("/etc/fstab")
+	return args, nil, nil
+}
 
 func (f *mqlFstab) entries() ([]any, error) {
 	conn := f.MqlRuntime.Connection.(shared.Connection)
@@ -21,7 +43,7 @@ func (f *mqlFstab) entries() ([]any, error) {
 		return nil, errors.New("filesystem not available")
 	}
 
-	fstabFile, err := fs.Open("/etc/fstab")
+	fstabFile, err := fs.Open(f.GetPath().Data)
 	if err != nil {
 		return nil, err
 	}
--- a/providers/os/resources/fstab.go
+++ b/providers/os/resources/fstab.go
@@ -1,0 +1,109 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+package resources
+
+import (
+	"bufio"
+	"errors"
+	"io"
+	"strconv"
+	"strings"
+
+	"go.mondoo.com/cnquery/v11/llx"
+	"go.mondoo.com/cnquery/v11/providers/os/connection/shared"
+)
+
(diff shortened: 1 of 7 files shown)


providers/os/resources/fstab.go
```diff
@@ -0,0 +1,109 @@
...
+	"strconv"
+	"strings"
+
+	"go.mondoo.com/cnquery/v11/llx"
+	"go.mondoo.com/cnquery/v11/providers/os/connection/shared"
+)
+
+func (f *mqlFstab) entries() ([]any, error) {
+	conn := f.MqlRuntime.Connection.(shared.Connection)
+	fstabFile, err := conn.FileSystem().Open("/etc/fstab")
```
preslavgerchev: i thihnk we should check if `conn.FileSystem()` returns an actual value. maybe some connections return `nil` here

providers/os/resources/fstab.go
```diff
@@ -0,0 +1,109 @@
...
+	"io"
+	"strconv"
+	"strings"
+
+	"go.mondoo.com/cnquery/v11/llx"
+	"go.mondoo.com/cnquery/v11/providers/os/connection/shared"
+)
+
+func (f *mqlFstab) entries() ([]any, error) {
+	conn := f.MqlRuntime.Connection.(shared.Connection)
```
preslavgerchev: we should check if its castable

slntopp: Do you mean generally or just in this particular case? I "copied" this over from iptables

preslavgerchev: best case: generally, but just fixing this case should do it 😄

providers/os/resources/fstab.go
```diff
@@ -0,0 +1,115 @@
...
+
+	fs := conn.FileSystem()
+	if fs == nil {
+		return nil, errors.New("filesystem not available")
+	}
+
+	fstabFile, err := fs.Open("/etc/fstab")
```
preslavgerchev: can we make this a param of the actual MQL resource? that will let us read fstabs on snapshots too

slntopp: It does read fstab on snapshots already(example in PR description), since mounted dir is handled by FileSystem connecion.
I will add the init() for the edge case where /etc is mounted separately, it's probably going to be an issue eitherway

providers/os/resources/os.lr
```diff
@@ -929,6 +931,28 @@ iptables.entry {
...
+
+  entries() []fstab.entry
+}
+
+private fstab.entry @defaults("device mountpoint") {
+  // Device referenced in the fstab, e.g. LABEL=rootfs
+  device string
+  // Mountpoint, e.g. '/'
+  mountpoint string
+  // Filesystem type, e.g. ext4
```
misterpantz:
```suggestion
  // File system type, e.g., ext4
```
I try to follow the Google dev style guide as much as we can. https://developers.google.com/style/word-list#file-system

providers/os/resources/os.lr
```diff
@@ -929,6 +931,28 @@ iptables.entry {
...
 
+fstab @defaults("path") {
+  init(path? string)
+  path string
+
+  entries() []fstab.entry
+}
+
+private fstab.entry @defaults("device mountpoint") {
+  // Device referenced in the fstab, e.g. LABEL=rootfs
```
misterpantz:
```suggestion
  // Device referenced in the fstab, e.g., LABEL=rootfs
```

providers/os/resources/os.lr
```diff
@@ -929,6 +931,28 @@ iptables.entry {
...
+  device string
+  // Mountpoint, e.g. '/'
```
misterpantz:
```suggestion
  // Mount point, e.g., '/'
```

providers/os/resources/os.lr
```diff
@@ -929,6 +931,28 @@ iptables.entry {
...
+  mountpoint string
+  // Filesystem type, e.g. ext4
+  fstype string
+  // Mount options, e.g. defaults
```
misterpantz:
```suggestion
  // Mount options, e.g., defaults
```

providers/os/resources/os.lr
```diff
@@ -929,6 +931,28 @@ iptables.entry {
...
+  options string
+  // Dump frequency, e.g. 0
```
misterpantz:
```suggestion
  // Dump frequency (0 for full backup or an integer above 0, incremental backup, copies all files new or modified since the last dump of a lower level)
```

providers/os/resources/os.lr
```diff
@@ -929,6 +931,28 @@ iptables.entry {
...
+  dump int
+  // Filesystem check order, e.g. 1
```
misterpantz:
```suggestion
  // File system check order, e.g., 1
```

providers/os/resources/os.lr
misterpantz: In what units?

providers/os/resources/os.lr
misterpantz: Could we list all the options? Or point our users somewhere?

slntopp: [The list is quite big](https://man.archlinux.org/man/ext4.5#Mount_options_for_ext4)
Also it's bit tough with the references, since it's linux, there are just too many of them.

providers/os/resources/os.lr
slntopp: doesn't have any
> The dump level (any integer). A level 0, full backup, specified by -0 guarantees the entire file system is copied (but see also the -h option below). A level number above 0, incremental backup, tells dump to copy all files new or modified since the last dump of a lower level. The default level is 0. Historically only levels 0 to 9 were usable in dump, this version is able to understand any integer as a dump level.

providers/os/resources/os.lr
misterpantz: No problem, we just link to that list!

providers/os/resources/os.lr
misterpantz: Thank you. I wrote a brief version.

providers/os/resources/os.lr
misterpantz: Per our discussion, closing. 😄

preslavgerchev (approved):

PR #4840, opened 2024-11-11
🐛 Consolidate discovery options for all providers and remove duplicate all/auto from CLI help info.

VasilSirakov:
Fixes https://github.com/mondoohq/cnspec/issues/1473

#### In this PR:
- Keep (all/auto) as default for all providers.
- Ensure every provider specifies its supported discovery options in `config.go` (except all/auto).
- Removed some unused code.


VasilSirakov:
As per @imilchev remarks, we should keep `all` and `auto` for all providers by default and simply remove in `config.go` the occurrences where individual providers specify `DiscoveryAll` and `DiscoveryAuto` to avoid having duplicates.

imilchev (approved):

PR #4847, opened 2024-11-13
🐛 Load provider flags from environment variables

afiune:
This change allows `cnquery` to load all provider flags via environment variables. For instance, if a provider exposes a flag named `foo`, the user can export the environment variable `MONDOO_FOO` to provide a value.

The trick here was to switch loading the flag values from `cobra` to `viper`, since we already ran `viper.BindPFlag()` inside the command `PreRun`, we can now use `viper.Get()` directly.

### DIFF
--- a/cli/providers/providers.go
+++ b/cli/providers/providers.go
@@ -5,8 +5,6 @@ package providers
 
 import (
 	"encoding/json"
-	"go.mondoo.com/cnquery/v11/utils/piped"
-	"go.mondoo.com/ranger-rpc/status"
 	"os"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


cli/providers/providers.go
```diff
@@ -318,35 +318,22 @@ func attachFlags(flagset *pflag.FlagSet, flags []plugin.Flag) {
 	}
 }
 
-func getFlagValue(flag plugin.Flag, cmd *cobra.Command) *llx.Primitive {
+func getFlagValue(flag plugin.Flag) *llx.Primitive {
 	switch flag.Type {
 	case plugin.FlagType_Bool:
-		v, err := cmd.Flags().GetBool(flag.Long)
```
chris-rock, line 324: Do we need a bind equivalent somewhere? Like in https://github.com/mondoohq/cnspec/blob/main/apps/cnspec/cmd/scan.go#L105

afiune: We already have it for all flags inside a provider. Though you're pointing to cnspec code so I'm assuming you want to know if this will work there. I can test, but this is working for cnquery.

afiune: https://github.com/mondoohq/cnquery/blob/main/cli/providers/providers.go#L377

chris-rock: Perfecto. Then lets do it.

PR #4848, opened 2024-11-13
🧹 Adding Country location to NamedLocation under conditional access - MS365

HRouhani:
Screenshot from 2024-11-13 16-00-33

### DIFF
--- a/providers/ms365/resources/conditional-access.go
+++ b/providers/ms365/resources/conditional-access.go
@@ -20,6 +20,9 @@ func (a *mqlMicrosoftConditionalAccess) namedLocations() ([]interface{}, error) 
 
 	ctx := context.Background()
 	namedLocations, err := graphClient.Identity().ConditionalAccess().NamedLocations().Get(ctx, nil)
+	if err != nil {
+		return nil, transformError(err)
+	}
 
 	var locationDetails []interface{}
 	for _, location := range namedLocations.GetValue() {
@@ -66,6 +69,9 @@ func (a *mqlMicrosoftConditionalAccessNamedLocations) countryLocations() ([]inte
 
 	ctx := context.Background()
 	namedLocations, err := graphClient.Identity().ConditionalAccess().NamedLocations().Get(ctx, nil)
+	if err != nil {
+		return nil, transformError(err)
+	}
 
 	var locationDetails []interface{}
 	for _, location := range namedLocations.GetValue() {
--- a/providers/ms365/resources/conditional-access.go
+++ b/providers/ms365/resources/conditional-access.go
@@ -11,7 +11,11 @@ import (
 	"go.mondoo.com/cnquery/v11/providers/ms365/connection"
 )
 
-func (a *mqlMicrosoftConditionalAccess) namedLocations() ([]interface{}, error) {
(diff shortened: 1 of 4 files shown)


providers/ms365/resources/ms365.lr
```diff
@@ -60,6 +60,8 @@ microsoft.tenant @defaults("name") {
 microsoft.conditionalAccess {
   // IP named location
   namedLocations() []microsoft.conditionalAccess.ipNamedLocation
+  // Country-based named location
+  countryLocations() []microsoft.conditionalAccess.countryNamedLocation
```
czunker, line 70: From reading the docs, I think, this should be called `namedLocations()`. It can be a country or an IP.

providers/ms365/resources/ms365.lr
```diff
@@ -70,6 +72,14 @@ microsoft.conditionalAccess.ipNamedLocation @defaults("name trusted") {
   trusted bool
 }
 
+// Microsoft Conditional Access Country named location
+microsoft.conditionalAccess.countryNamedLocation @defaults("name lookupMethod") {
```
czunker, line 82: Same here:
`namedLocation` instead of `country...`

czunker (commented):
Code wise LGTM.
But I think we should rename the resource.

providers/ms365/resources/ms365.lr
HRouhani: Good points. I changed the structure totally to make it both be under the umbrella of namedLocation

providers/ms365/resources/ms365.lr
HRouhani: Thanks I did

HRouhani:
After the merging I need to change all the policies get effected with this changes!

providers/ms365/resources/conditional-access.go
```diff
@@ -55,3 +52,50 @@ func (a *mqlMicrosoftConditionalAccess) namedLocations() ([]interface{}, error)
...
+
+func (a *mqlMicrosoftConditionalAccessNamedLocations) countryLocations() ([]interface{}, error) {
+	conn := a.MqlRuntime.Connection.(*connection.Ms365Connection)
+	graphClient, err := conn.GraphClient()
+	if err != nil {
+		return nil, err
+	}
+
+	ctx := context.Background()
+	namedLocations, err := graphClient.Identity().ConditionalAccess().NamedLocations().Get(ctx, nil)
```
czunker, line 75: You should check the error here.

providers/ms365/resources/conditional-access.go
```diff
@@ -20,9 +20,6 @@ func (a *mqlMicrosoftConditionalAccess) namedLocations() ([]interface{}, error)
 
 	ctx := context.Background()
 	namedLocations, err := graphClient.Identity().ConditionalAccess().NamedLocations().Get(ctx, nil)
-	if err != nil {
```
czunker: I think, you should keep the err check.

czunker (commented):
Nice refactor, just error checks are missing.

providers/ms365/resources/conditional-access.go
HRouhani: thanks, done

HRouhani: done

PR #4851, opened 2024-11-13
🧹 Replace unmaintained mapstructure

czunker:
See https://github.com/99designs/gqlgen/pull/3343 for more details.


go.mod
```diff
@@ -64,7 +65,6 @@ require (
 	github.com/miekg/dns v1.1.62
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/mitchellh/hashstructure/v2 v2.0.2
-	github.com/mitchellh/mapstructure v1.5.0
```
chris-rock, line 67: Can we also add the config to prevent the addition going forward?

chris-rock (approved):
Thank you @czunker

go.mod
czunker: Good call. Done.

PR #4859, opened 2024-11-14
✨ add aliases to partitions info

slntopp:


providers/os/connection/snapshot/partition.go
```diff
@@ -11,6 +11,7 @@ type PartitionInfo struct {
 	Name   string
 	FsType string
 
+	Aliases []string
```
preslavgerchev, line 17: can you write a comment above on what those aliases are?

slntopp: updated them all

preslavgerchev: nice!

preslavgerchev (approved):

PR #4863, opened 2024-11-15
🐛 bring back loading provider flags from env vars

afiune:
The problem from my original PR https://github.com/mondoohq/cnquery/pull/4847  to load provider flags from environment
variables was that I didn't consider the case where a provider defines a flag with
`ConfigEntry = "-"`. For those cases, we do NOT bind the flag to the `viper` config
and therefore, we couldn't load these flags.

This PR fixes that issue and, for those flags that are not bind to the config, we will
continue to fetch the value directly from the flag itself, that is, from `cobra`.

### DIFF
--- a/test/providers/os_test.go
+++ b/test/providers/os_test.go
@@ -226,7 +226,7 @@ func TestProvidersEnvVarsLoading(t *testing.T) {
 		}
 	})
 
-	t.Run("command with	flags set to not bind to config (ConfigEntry=\"-\")", func(t *testing.T) {
+	t.Run("command with flags set to not bind to config (ConfigEntry=\"-\")", func(t *testing.T) {
 		t.Run("should work via direct flag", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


test/providers/os_test.go
```diff
@@ -185,3 +186,70 @@ func TestOsProviderSharedTests(t *testing.T) {
...
+
+		// Should have packages
+		if assert.NotEmpty(t, c) {
+			x := c[0]
+			assert.NotNil(t, x.Packages)
+			assert.True(t, len(x.Packages) > 0)
+		}
+	})
+
+	t.Run("command with	flags set to not bind to config (ConfigEntry=\"-\")", func(t *testing.T) {
```
jaym: some weird spacing hin here

jaym (approved):

test/providers/os_test.go
afiune:
```suggestion
	t.Run("command with flags set to not bind to config (ConfigEntry=\"-\")", func(t *testing.T) {
```

PR #4895, opened 2024-11-20
Add curl timeout for cloud metadata lookup at unix systems on Azure

tomtrix:
Dear Mondoo team,

this change should fix the issue https://github.com/mondoohq/cnspec/issues/1486

Kind regards
Tom

### DIFF
--- a/providers/os/id/azcompute/azcompute.go
+++ b/providers/os/id/azcompute/azcompute.go
@@ -17,7 +17,7 @@ import (
 
 const (
 	identityUrl                   = "http://169.254.169.254/metadata/instance?api-version=2021-02-01"
-	metadataIdentityScriptWindows = `Invoke-RestMethod -TimeoutSec 5 -Headers @{"Metadata"="true"} -Method GET -URI http://169.254.169.254/metadata/instance?api-version=2021-02-01 -UseBasicParsing | ConvertTo-Json`
+	metadataIdentityScriptWindows = `Invoke-RestMethod -TimeoutSec 15 -Headers @{"Metadata"="true"} -Method GET -URI http://169.254.169.254/metadata/instance?api-version=2021-02-01 -UseBasicParsing | ConvertTo-Json`
 )
 
 func MondooAzureInstanceID(instanceID string) string {
--- a/providers/os/id/azcompute/azcompute.go
+++ b/providers/os/id/azcompute/azcompute.go
@@ -57,7 +57,7 @@ func (m *commandInstanceMetadata) Identify() (Identity, error) {
 	var instanceDocument string
 	switch {
 	case m.platform.IsFamily(inventory.FAMILY_UNIX):
-		cmd, err := m.conn.RunCommand("curl --retry 5 --retry-delay 1 --max-time 5 --noproxy '*' -H Metadata:true " + identityUrl)
+		cmd, err := m.conn.RunCommand("curl --retry 5 --retry-delay 1 --connect-timeout 1 --retry-max-time 5 --max-time 10 --noproxy '*' -H Metadata:true " + identityUrl)
 		if err != nil {
 			return Identity{}, err
 		}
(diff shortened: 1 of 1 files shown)


providers/os/id/azcompute/azcompute.go
```diff
@@ -57,7 +57,7 @@ func (m *commandInstanceMetadata) Identify() (Identity, error) {
 	var instanceDocument string
 	switch {
 	case m.platform.IsFamily(inventory.FAMILY_UNIX):
-		cmd, err := m.conn.RunCommand("curl --noproxy '*' -H Metadata:true " + identityUrl)
+		cmd, err := m.conn.RunCommand("curl --max-time 1 --noproxy '*' -H Metadata:true " + identityUrl)
```
chris-rock: @tomtrix Thank you raising this issue. I think we definitely need to add a max-time setting. 1 second seems to be quite rigid. To make this even more reliable, what do you think about:

```
--retry 3 --retry-delay 1 --max-time 5
```

tomtrix: @chris-rock seems fine to me too! I've set the timeout to 1 sec. because you did so at the PowerShell version of the metadata lookup.

https://github.com/mondoohq/cnquery/blob/9e1b2eb0c17e2045617c9350bccf1c102e4c672b/providers/os/id/azcompute/azcompute.go#L20

Shall we change both commands to the same values?

chris-rock: Great idea. Lets align this.

tomtrix: @chris-rock which version of PowerShell does cnspec require? Options for "retry" and "delay" have been implement in PowerShell 7, most Windows systems still use PowerShell 5 at the moment.

If we wanna be backward compatible, we can increase timeout to 5 seconds without retries only.

PowerShell 5.1: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-restmethod?view=powershell-5.1#-timeoutsec

PowerShell 7.4 (LTS): https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-restmethod?view=powershell-7.4#-maximumretrycount

tomtrix: @chris-rock I think we should go with PowerShell 5 to ensure backward compatibility.

I've adjusted PowerShell call to ensure a **connection timeout of max. 15 seconds**.

`curl` instead got four parameters:

* `--retry 3`
* `--retry-delay 1`
* [`--connect-timeout 5`](https://curl.se/docs/manpage.html#--connect-timeout) to ensure a **connection timeout of max. 5 seconds** per retry
* [`--max-time 30`](https://curl.se/docs/manpage.html#-m) to "prevent hanging" when the API response takes to long

chris-rock: Lets stick with Powershell 5 for now. I think we should stick with 5 seconds for Pwsh, otherwise we hang on so long. In normal cases the metadata api should always respond much faster. So I think we can use shorter times just specifically for the metadata api. I still tend to stick with --max-time 5 since the payload from the metadata api is so small. WDYT?

tomtrix: @chris-rock for me timeouts are relevent for the "worst case" - aren't they?

If the Metadata API responds faster, the command goes on right after getting a response.

The timeout is relevant in edge cases only - i.e. Linux VMs on Hyper-V.

For a bit better performance we could adjust **connection timeout** to 5 seconds for sure - I will take care of it.

tomtrix: @chris-rock I've adjusted the timeouts to 5 seconds and removed the `--connect-timeout` from `curl`. Feel free to merge. :)

tomtrix: @chris-rock after testing out several `curl` parameters, I think we should go with `curl --retry 5 --retry-delay 1 --connect-timeout 1 --retry-max-time 5 --max-time 10 ...`. This ensure a maximum **connection timeout of 5 seonds** and a **maximum tranfer timeout of 10 seconds**.

image

jaym (approved):

providers/os/id/azcompute/azcompute.go
chris-rock: Awesome. Thank you @tomtrix

PR #4896, opened 2024-11-20
🧹 cache windows smbios response 

imilchev:
Retrieving the smbios info for windows can be expensive. This PR makes sure the resposne is cached in the manager and that we use a single manager during detection. It will make sure we will run the powershell script once instead of 3 times


chris-rock (approved):
Looks great. Thank you @imilchev

PR #4899, opened 2024-11-20
🧹 use CIM instead of WMI for smbios

imilchev:
WMI is deprecated and can be slow, we migrate to its successor

I compared the results for `machine.bios`, `machine.chassis`, `machine.baseboard`, `machine.system` and they seem to return the same info


chris-rock (approved):
Thank you @imilchev

PR #4903, opened 2024-11-21
🛠️ new `util/zerologadapter` for retryable requests

afiune:
We use the `retryablehttp` package to retry http requests and we always want the logs to go to `debug` level so, we need an adapter.

We have implemented this adapter many times, this change removes all the duplicated code and creates a util package that all providers can use.

Example:

```go
retryClient := retryablehttp.NewClient()
retryClient.Logger = zerologadapter.New(log.Logger)
```

### DIFF
--- a/providers-sdk/v1/util/zerologadapter/adapter.go
+++ b/providers-sdk/v1/util/zerologadapter/adapter.go
@@ -1,0 +1,57 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package zerologadapter
+
+import "github.com/rs/zerolog"
+
+// New returns a new adapter for the zerolog logger to the LeveledLogger interface. This struct is
+// mainly used in conjunction with a retryable http client to convert all retry logs to debug logs.
+//
+// NOTE that all messages will go to debug level.
+//
+// e.g.
+// ```go
+// retryClient := retryablehttp.NewClient()
+// retryClient.Logger = zerologadapter.New(log.Logger)
+// ```
+func New(logger zerolog.Logger) *Adapter {
+	return &Adapter{logger}
+}
+
+type Adapter struct {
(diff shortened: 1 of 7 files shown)


preslavgerchev (approved):
nice simplification @afiune

providers-sdk/v1/util/zerologadapter/adapter.go
```diff
@@ -0,0 +1,57 @@
...
+func (z *Adapter) Warn(msg string, keysAndValues ...interface{}) {
+	z.logger.Debug().Fields(convertToFields(keysAndValues...)).Msg(msg)
+}
+
+func convertToFields(keysAndValues ...interface{}) map[string]interface{} {
+	fields := make(map[string]interface{})
+	for i := 0; i < len(keysAndValues); i += 2 {
+		if i+1 < len(keysAndValues) {
+			keyString, ok := keysAndValues[i].(string)
+			if ok { // safelty first, eventhough we always expect a string
```
preslavgerchev:
```suggestion
			if ok { // safety first, eventhough we always expect a string
```

czunker:
Just rebase on main.
https://github.com/mondoohq/cnquery/pull/4905 should have fixed the failing tests.

providers-sdk/v1/util/zerologadapter/adapter.go
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package zerologadapter
```
chris-rock: I think this package should be in https://github.com/mondoohq/cnquery/tree/main/logger

afiune: I put it inside the `providers-sdk/v1` to make it clear that providers can and should use it, Ill move it though.

PR #4932, opened 2024-11-27
🧹 Pull in the latest version of GitHub package

tas50:
Tested pulling a repo data and it seems to work as expected


czunker (approved):
Release notes mention some breaking changes: https://github.com/google/go-github/releases

But they all read to me, like they should already break compilation.

PR #4934, opened 2024-11-29
🐛 consistent processes executable for unix and linux

### LINKED ISSUE #4889: Mac processes executable has incorrect value
**Describe the bug**
The `executable` field for a MacOS process has incorrect value. For linux the value of `executable` holds just the executable name. For MacOS processes, the same field holds the full path to the executable.

MacOS:
Image

Linux:
Image

**To Reproduce**

1. Open `cnspec shell`
2. Run `processes{executable}`
3. Note the mistake

**Expected behavior**
The value of `executable` should behave the same as for Linux. It should contain just the executable name, not the full path

imilchev:
Fixes #4889

### DIFF
--- a/providers/os/resources/processes/unixps.go
+++ b/providers/os/resources/processes/unixps.go
@@ -54,16 +54,17 @@ type ProcessEntry struct {
 }
 
 func (p ProcessEntry) ToOSProcess() *OSProcess {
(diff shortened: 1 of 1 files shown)


czunker (approved):
LGTM
Thanks @imilchev

PR #4970, opened 2024-12-11
⚡ fetch org repositories in parallel 

afiune:
Introducing an internal go package called `workerpool` that we can use to send parallel
requests when needed.

:zap: For this change, I am making the fetching of repositories for an organization faster.

Tested this code with an **organization that has around 3k repositories**

### Before (~2 Minutes)
```
TRC logger.FuncDur> func=provider.github.repositories took=102803.621667
```
### After (~5 seconds)
```
TRC logger.FuncDur> func=provider.github.repositories took=4567.576542
```


providers/github/resources/github_org.go
```diff
@@ -262,26 +265,66 @@ func (g *mqlGithubOrganization) repositories() ([]interface{}, error) {
...
 	for {
-		repos, resp, err := conn.Client().Repositories.ListByOrg(conn.Context(), orgLogin, listOpts)
-		if err != nil {
+
+		// exit as soon as we collect all repositories
+		if len(allRepos) >= int(repoCount) {
+			break
+		}
+
+		// send as many request as workers we have
```
jaym: Why do we limit by number of workers? Doesn't this artificially throttle things

afiune: Good point. Let me check that.

afiune: @jaym I updated the PR

internal/workerpool/collector.go
```diff
@@ -0,0 +1,34 @@
...
+				c.errors = append(c.errors, err)
+			}
+
+			c.requestsRead++
+		}
+	}()
+}
+
+func (c *collector[R]) RequestsRead() int64 {
+	return c.requestsRead
```
jaym: i think there might be a data race here on requestsRead

internal/workerpool/pool.go
```diff
@@ -0,0 +1,117 @@
...
+		if !p.Processing() {
+			return
+		}
+		<-ticker.C
+	}
+}
+
+// PendingRequests returns the number of pending requests.
+func (p *Pool[R]) PendingRequests() int64 {
+	return p.requestsSent - p.collector.RequestsRead()
```
jaym: i'm not really familiar with the atomics api in go, but i would expect some sort of synchronized read on requestsSent

internal/workerpool/collector.go
afiune: Fixed! Thanks for pointing it out, Jay.

internal/workerpool/pool.go
afiune: atomic helps to do thread safe operations, yes, I missed the read calls, they are now there! 🙌🏽

afiune:
Used the [data race detector](https://go.dev/doc/articles/race_detector) from go to validate.
```
$ go test -race go.mondoo.com/cnquery/v11/internal/workerpool -v
=== RUN   TestPoolSubmitAndRetrieveResult
--- PASS: TestPoolSubmitAndRetrieveResult (0.01s)
=== RUN   TestPoolHandleErrors
--- PASS: TestPoolHandleErrors (0.01s)
=== RUN   TestPoolMultipleTasksWithErrors
--- PASS: TestPoolMultipleTasksWithErrors (0.01s)
=== RUN   TestPoolHandlesNilTasks
--- PASS: TestPoolHandlesNilTasks (0.00s)
=== RUN   TestPoolProcessing
--- PASS: TestPoolProcessing (0.06s)
=== RUN   TestPoolClosesGracefully
--- PASS: TestPoolClosesGracefully (0.11s)
=== RUN   TestPoolWithManyTasks
--- PASS: TestPoolWithManyTasks (15.03s)
PASS
ok  	go.mondoo.com/cnquery/v11/internal/workerpool	16.715s
```

jaym (approved):

PR #4971, opened 2024-12-11
✨ device: linux lvm support

slntopp:
#### What has changed

After mount targets (partitions) are discovered, we'd attempt to either find `fstab` file or detect `ostree` partition.

##### `fstab`

If `fstab` configuration is discovered, provider will parse it and mount all discovered partitions using "opts" configured in the file. This ensures all required features and flags are used, for instance if partition has `btrfs` or `xfs` sub-volumes.
Additionally, device manager would add same partition twice or more if it's mounted multiple times with different opts.

Additionally, device manager will attempt to mount these partitions as `/tmp/cnspec-scanxxxx` + `<fstab mountpoint>`

By the time we get to attempt detect asset, partitions are already mounted and the scan dir has the structure closest to that in the original VM.  

##### `ostree`

`ostree` filesystem together with edge case distros like Fedora CoreOS have a very different file structure as well as missing lot of usual service files (like `hostname`, `machine-id`, etc)
First of all, device manager attempts to `untangle` the pointers and locate the actual and latest root filesystem.
Additionally this PR introduces `bind` option to the volume mounter, which doesn't mount partition again with `--bind` as one may think, though but adjusts the fs-connection used for asset detection and can be used anywhere else, where the files search relies on standard linux tree structure. For malware detection and similar operations `Bind` can be ignored to get the full overview of the system.

##### moby/mount instead of sys/unix

Since `fstab` options include both flags and "data" we need a reliable way to split those, so mount doesn't crash out with `invalid argument` error. moby mount provides that, so we simply don't have to reinvent a bicycle here.

##### inject platform ids fallback

As mentioned, some distros don't have usual means of ID-ing the OS like hostname, machine-id or ssh-host-keys. We can however inject custom platform-ids in this cases. With this PR we don't error out from asset detection if no platform ids were found and there are injected ones.


vjeffrey:
@slntopp can you add some words about how to best test this new functionality?

slntopp:
1. Attach volume(s) with Linux on it
2. Run `cnspec scan device --device-name /dev/<your-device>`
3. Enjoy successful scan

Difference is that before if root partition was `btrfs`, `xfs` with `ostree`, has other complex mounting configuration - scanner wouldn't be able to detect asset by failing to find related files (os-release and analogues)

go.mod
```diff
@@ -287,7 +287,10 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
 
-require github.com/moby/buildkit v0.18.1
+require (
+	github.com/moby/buildkit v0.18.1
```
preslavgerchev, line 291: nit: why not put those in with the other direct deps

providers/os/config/config.go
```diff
@@ -300,6 +300,12 @@ var Config = plugin.Provider{
 					Desc:   "Mount all partitions of the block device",
 					Option: plugin.FlagOption_Hidden,
 				},
+				{
+					Long:   "skip-attempt-expand-partitions",
```
preslavgerchev, line 304: what would be the use case for that?

providers/os/connection/device/device_connection.go
```diff
@@ -83,12 +96,29 @@ func NewDeviceConnection(connId uint32, conf *inventory.Config, asset *inventory
...
+
+		if block.MountPoint == "" {
+			scanDir, err := manager.Mount(block)
+			if err != nil {
+				log.Error().Err(err).Msg("unable to complete mount step")
+				continue
+			}
+			block.MountPoint = scanDir
+		}
+		res.MountedDirs = append(res.MountedDirs, block.MountPoint)
```
preslavgerchev, line 112: should we only append this if the block was mounted by us? is that then used for unmounting?

providers/os/connection/device/linux/device_manager.go
```diff
@@ -75,11 +87,221 @@ func (d *LinuxDeviceManager) IdentifyMountTargets(opts map[string]string) ([]*sn
...
+	if err != nil {
+		log.Error().Err(err).Msg("unable to mount partitions with fstab")
+		d.UnmountAndClose()
+		return partitions, nil
+	}
+
+	return partitions, nil
+}
+
+func (d *LinuxDeviceManager) HintFSTypes(partitions []*snapshot.PartitionInfo) ([]resources.FstabEntry, error) {
```
preslavgerchev: can this be a private fn?

providers/os/connection/device/linux/device_manager.go
```diff
@@ -75,11 +87,221 @@ func (d *LinuxDeviceManager) IdentifyMountTargets(opts map[string]string) ([]*sn
...
+			return nil, nil
+		}
+
+	}
+
+	return nil, errors.New("fstab not found")
+}
+
+func (d *LinuxDeviceManager) AttemptFindFstab(dir string) ([]resources.FstabEntry, error) {
+	cmd := exec.Command("find", dir, "-type", "f", "-wholename", `*/etc/fstab`)
```
preslavgerchev: can we use the cmd runner here? its part of the vol mounter. we could inject it explicitly in the device manager

providers/os/connection/device/linux/device_manager.go
```diff
@@ -75,11 +87,221 @@ func (d *LinuxDeviceManager) IdentifyMountTargets(opts map[string]string) ([]*sn
...
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		log.Error().Err(err).Msg("error searching for fstab")
+		return nil, nil
+	}
+
+	if l := len(strings.Split(string(out), "\n")); l > 2 {
```
preslavgerchev, line 164: can you privde a sample of what the expected output here is such that it requires a `l > 2` check?

preslavgerchev: also, maybe best to extract that in a different func which can also be tested (just the p[arsing of the fstab output)

providers/os/connection/device/linux/device_manager.go
```diff
@@ -75,11 +87,221 @@ func (d *LinuxDeviceManager) IdentifyMountTargets(opts map[string]string) ([]*sn
...
+		return nil, nil
+	}
+
+	return resources.ParseFstab(bytes.NewReader(fstabFile))
+}
+
+func (d *LinuxDeviceManager) AttemptFindOSTree(dir string, partition *snapshot.PartitionInfo) bool {
+	info, err := os.Stat(path.Join(dir, "ostree"))
+	if err != nil {
+		if os.IsNotExist(err) {
```
preslavgerchev, line 187: nit: && the 2 checks

providers/os/connection/snapshot/blockdevices.go
```diff
@@ -225,6 +234,16 @@ func (device BlockDevice) GetPartitions(includeBoot bool, includeMounted bool) (
 	return partitions, nil
 }
 
+func mapLVM2Partitions(part BlockDevice, partitions *[]*PartitionInfo) {
```
preslavgerchev: can we get this to reutnr the lv2m parts (children) and then append those to `partitions` in the caller of `mapLVM..`, would make for better readability

providers/os/connection/snapshot/partition.go
```diff
@@ -21,6 +21,17 @@ type PartitionInfo struct {
 	Uuid string
 	// (optional) MountPoint is the partition mount point
 	MountPoint string
+
+	// (optional) MountOptions are the mount options
+	MountOptions []string
+	// (optional) Bind adjusts the root for FS connection
+	Bind string
```
preslavgerchev: If i understand correctly, this is for OS trees where the root is not at the mountpoint but somewhere else? Maybe we call this `RootDir` or something? Should we make this always filled in and for:
* os tree partitions its probably not equal to `Mountpoint`
* for other parts, its usually equal to `Mountpoint`

but we always use this property to initiate FS scans from, so that code is consistent and its just the `RootDir` logic that varies

providers/os/connection/snapshot/volumemounter.go
```diff
@@ -62,17 +97,32 @@ func (m *VolumeMounter) createScanDir() (string, error) {
...
 	opts = stringx.DedupStringArray(opts)
-	scanDir := m.ScanDirs[fsInfo.Name]
+	opts = saninizeOptions(opts)
+
+	scanDir := m.ScanDirs[fsInfo.key()]
 	log.Debug().Str("fstype", fsInfo.FsType).Str("device", fsInfo.Name).Str("scandir", scanDir).Str("opts", strings.Join(opts, ",")).Msg("mount volume to scan dir")
 	return Mount(fsInfo.Name, scanDir, fsInfo.FsType, opts)
 }
 
+func saninizeOptions(options []string) []string {
```
preslavgerchev:
```suggestion
func sanitizeOptions(options []string) []string {
```

providers/os/connection/snapshot/blockdevices.go
preslavgerchev: can we also add tests for lvm2 parts?

go.mod
slntopp: auto generated by `go mod tidy`

providers/os/config/config.go
slntopp: skip mounting-remounting cycle, for example if we want just specific single partition maybe?

providers/os/connection/device/device_connection.go
slntopp: This is used by other providers that use device connection indirectly, i.e. without fs connection.

providers/os/connection/device/linux/device_manager.go
slntopp: it's just a single line followed by `\n`, so split by `\n` results into 2 chunks, output is smth like:
```
/mnt/etc/fstab\n
```

so we can confidently say that we've found fstab if there is a single result in the output.

providers/os/connection/device/linux/device_manager.go
slntopp: two checks are there, so if ostree simply doesn't exist this function fails silently, but if it's something else we get to log the error

providers/os/connection/snapshot/partition.go
slntopp: Thing is that MountPoint might be changing, but Bind shouldn't
Also, there are use cases where we aren't interested in root but whole fs
I propose we can add it as a method, so it's consistent

providers/os/connection/snapshot/partition.go
```diff
@@ -21,6 +22,17 @@ type PartitionInfo struct {
...
```
preslavgerchev: can we make this private now that you've hidden it in a func?

providers/os/connection/snapshot/partition.go
```diff
@@ -21,6 +22,17 @@ type PartitionInfo struct {
...
+}
+
+type MountPartitionDto struct {
```
preslavgerchev, line 33: can you add some docs on this?

providers/os/connection/snapshot/partition.go
```diff
@@ -21,6 +22,17 @@ type PartitionInfo struct {
...
+	*PartitionInfo
+
+	ScanDir *string
```
preslavgerchev, line 37: is there a chance this is nil? can we do with representing the absence as an empty string>

providers/os/connection/snapshot/volumemounter.go
```diff
@@ -62,17 +97,32 @@ func (m *VolumeMounter) createScanDir() (string, error) {
...
-	scanDir := m.ScanDirs[fsInfo.Name]
+	opts = sanitizeOptions(opts)
+
+	scanDir := m.ScanDirs[fsInfo.key()]
 	log.Debug().Str("fstype", fsInfo.FsType).Str("device", fsInfo.Name).Str("scandir", scanDir).Str("opts", strings.Join(opts, ",")).Msg("mount volume to scan dir")
 	return Mount(fsInfo.Name, scanDir, fsInfo.FsType, opts)
 }
 
+func sanitizeOptions(options []string) []string {
+	var sanitized []string
```
preslavgerchev: nit: init the array here

providers/os/connection/snapshot/blockdevices.go
```diff
@@ -225,6 +234,18 @@ func (device BlockDevice) GetPartitions(includeBoot bool, includeMounted bool) (
 	return partitions, nil
 }
 
+func mapLVM2Partitions(part BlockDevice) (partitions []*PartitionInfo) {
+	for _, p := range part.Children {
+		devFsName := "/dev/mapper/" + p.Name
+		partitions = append(partitions, &PartitionInfo{
```
preslavgerchev: nit: can we make a blockdevice -> partitioninfo func? i think we use this in more places than here

preslavgerchev (approved):

providers/os/connection/snapshot/partition.go
slntopp: It'll be nil in most cases. Normally volume mounter handles scandir creation. We need this so we can mount sub dir under existing scan dir, e.g. /tmp/cnspec-scanX/mnt

slntopp: Also, nil check is in place at the MountP side

providers/os/connection/snapshot/blockdevices.go
slntopp: added a method to BlockDevice

preslavgerchev (approved):

PR #4974, opened 2024-12-12
🐛 fix find files for tar

### LINKED ISSUE #4894: Find is failing on a debian based container that has find
**Describe the bug**

Find is failing on the `nginx` container which is Debian based and has find.

```
docker run -it nginx /bin/bash
root@2c952940aee8:/# find . -name 'issue'
./etc/issue
```

Image

To repro this you can enable the container secrets policy and then run `cnspec scan docker image nginx`

imilchev:
Fixes #4894

### DIFF
--- a/providers/os/connection/container/image_connection_test.go
+++ b/providers/os/connection/container/image_connection_test.go
@@ -231,11 +231,11 @@ func TestImageConnections(t *testing.T) {
 				fSearch := fs.(*tar.FS)
 
 				if test.testfile == "/etc/alpine-release" {
-					infos, err := fSearch.Find("/", regexp.MustCompile(`alpine-release`), "file")
+					infos, err := fSearch.Find("/", regexp.MustCompile(`alpine-release`), "file", nil, nil)
 					require.NoError(t, err)
 					assert.Equal(t, 1, len(infos))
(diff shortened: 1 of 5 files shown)


czunker:
I don't get the error anymore, but it does not find files:
```
cnquery run container image nginx:1-alpine -c "files.find(from: '/etc', name: 'os-release')"
...
files.find.list: []
```
vs.
```
docker run -ti nginx:1-alpine find /etc -name os-release
/etc/os-release
```

imilchev:
That seems like a different bug. Name filtering never worked for fs find. I fixed it too.

czunker (approved):
Both work now:
```
cnquery run container image nginx:1-alpine -c "files.find(from: '/etc', name: 'os-release')" 

files.find.list: [
  0: file path="/etc/os-release" size=188 permissions.string="-rw-r--r--"
]
```
```
cnquery run container image nginx:1-alpine -c "files.find(from: '/etc', regex: '.*release')"
files.find.list: [
  0: file path="/etc/alpine-release" size=7 permissions.string="-rw-r--r--"
  1: file path="/etc/os-release" size=188 permissions.string="-rw-r--r--"
]
```

Thanks @imilchev

PR #4976, opened 2024-12-12
⭐️ plugin cache + verify github conn once

afiune:
## Testing ...

Follow up of https://github.com/mondoohq/cnquery/pull/4973


imilchev (commented):
I think this is a more elegant way of achieving the same result. https://github.com/mondoohq/cnquery/blob/main/providers/github/resources/github.go#L18-L49

We have this memoizer which can cache responses for certain calls. It caches them for a configurable amount of time. We already use this in the github provider for caching users, for example. I think you can use a memoizer for the validation too

providers-sdk/v1/plugin/service.go
```diff
@@ -24,11 +25,14 @@ type Service struct {
 
 	lastHeartbeat int64
 	heartbeatLock sync.Mutex
+
+	Cache store.KeyValue
```
preslavgerchev, line 29: is this thread-safe? iirc w can have multiple conns using 1 provider?

preslavgerchev: ~~also, if we only need this for github, why not make it a github-connection specific thing?~~

was thinking out loud here, this is provider/plugin level

afiune:
@imilchev if we want that approach, lets review this PR instead https://github.com/mondoohq/cnquery/pull/4980

I still think having a key value cache out-of-the-box for providers is a nice thing to have. But that is a discussion for later.

PR #4978, opened 2024-12-13
Fix the vmware technology URLs

tas50:
The top level here should really be vmware so we don't end up with vsphere under vsphere

Paging @arlimus on this one


chris-rock (approved):
Thank you @tas50

arlimus (approved):
kudos!

PR #4980, opened 2024-12-14
⚡️ verify github connection only once

afiune:
Supersedes https://github.com/mondoohq/cnquery/pull/4976

When a user scans a single GH repository, this was not a problem, but for users that scan an entire organization with thousands of repositories, we were creating a connection for each asset and validating it thousands of times.

This changes the behavior to verify the connection only once which will save us thousands of GH API requests.


providers/github/connection/connection.go
```diff
@@ -38,25 +40,75 @@ type GithubConnection struct {
 	asset  *inventory.Asset
 	client *github.Client
 	ctx    context.Context
+
+	// Used to avoid verifying a client with the same options more than once
+	Hash uint64
```
imilchev, line 45: nit: can we call this optionsHash or something similar to make its purpose more clear?

imilchev (approved):

providers/github/connection/connection.go
afiune: I'll rename it in a diff PR

afiune: https://github.com/mondoohq/cnquery/issues/4995

PR #4996, opened 2024-12-16
⭐️ Add Purl to Macos and Windows systems

afiune:
Fixes https://github.com/mondoohq/cnquery/issues/4957


providers/os/resources/purl/purl_types.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func init() {
+	// merge packageurl.KnownTypes and the extension types
+	for t := range packageurl.KnownTypes {
+		KnownTypes[t] = struct{}{}
+	}
+}
+
+func ValidType(t string) bool {
```
czunker: I don't see this being used. Did I miss something?

czunker: When we use this somewhere else, could we add some tests to it?
Some types we use and some we do no longer want to use like `windows_app`, `windows-app`.

czunker (approved):

providers/os/resources/purl/purl.go
```diff
@@ -18,6 +18,23 @@ const (
 	QualifierEpoch  = "epoch"
 )
 
+// PackageURL is a helper struct that renters a package url based of an inventory
```
czunker, line 21: `renders` instead of `renters`

providers/os/resources/purl/purl.go
```diff
@@ -43,44 +60,92 @@ func NewQualifiers(qualifier map[string]string) packageurl.Qualifiers {
...
-// see https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst for more information
-func NewPackageUrl(pf *inventory.Platform, name string, version string, arch string, epoch string, purlType string) string {
-	qualifiers := map[string]string{}
-	if arch != "" {
-		qualifiers[QualifierArch] = arch
+// NewPackageURL creates a new package url for a given platform, name, version, and type.
+//
+// For more information, see:
+// https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst
+func NewPackageURL(pf *inventory.Platform, t Type, name, version string, modifiers ...Modifier) *PackageURL {
```
czunker, line 67: That made me stumble.
First thought was, that we could derive the type from the platform. But no:
- Windows: windows, appx, nuget, ...
- Ubuntu: deb, snap, brew, ...
- ...

:face_with_diagonal_mouth:

czunker (approved):
Thanks @afiune

providers/os/resources/purl/purl.go
afiune: yeah, we really can't 😭

PR #5034, opened 2024-12-23
Improve printing of discovery options in help

tas50:
Current:

```
--discover strings             Enable the discovery of nested assets. Supports: all,auto,repos,users,organization,terraform,k8s-manifests
```

With this PR

```
--discover strings             Enable the discovery of nested assets. Supports: all, auto, repos, users, organization, terraform, k8s-manifests
```


tas50:
If someone wants to make this last join be ", or" this would be even better

vjeffrey (approved):

PR #5037, opened 2024-12-23
:arrow_up: upgrade google/go-github to `v69`

tas50:
This is required for some future work. There are a number of breaking changes introduced in this release. I was able to fetch some basic data from an org and repo w/o issue, but we should probably dig a bit deeper there.


afiune (approved):
Checked the breaking changes from `v67` ➡️  `v68` and any issue would have been caught at compile time.

We should go all the way to the new version `v69` which has WAY more breaking changes:
* Some also would have been caught at compile time
* We do not use any `Ruleset` that got most changes
* Some error messages changed, luckily we do not rely on error text in general 🙌🏽 

We should be good to merge, but first I will cut a release with the latest fixes.

afiune:
https://github.com/mondoohq/cnquery/pull/5174

PR #5043, opened 2024-12-26
Improve the MQL resource pack docs generation

tas50:
- Rename the Mondoo pack to be Mondoo Platform so it's more clear what this does
- Avoid `Mondoo Mondoo Platform` by renaming the pack headers


misterpantz (approved):
super!

PR #5046, opened 2024-12-28
⭐️ human readable impact value

chris-rock:
Previously we only supported int values for query impacts, see https://github.com/mondoohq/cnspec-policies/blob/main/core/mondoo-linux-security.mql.yaml#L344

```yaml
  - uid: mondoo-linux-security-address-space-layout-randomization-aslr-is-enabled
    title: Ensure address space layout randomization (ASLR) is enabled
    impact: 90
    filters: |
      asset.kind != "container-image"
      asset.runtime != "docker-container"
    mql: |
      kernel.parameters["kernel.randomize_va_space"] == 2
```

With this change, we can also use typical ratings which makes it easier for users to read the values:

```yaml
  - uid: mondoo-linux-security-address-space-layout-randomization-aslr-is-enabled
    title: Ensure address space layout randomization (ASLR) is enabled
    impact: "critical"
    filters: |
      asset.kind != "container-image"
      asset.runtime != "docker-container"
    mql: |
      kernel.parameters["kernel.randomize_va_space"] == 2
```

### DIFF
--- a/explorer/impact.go
+++ b/explorer/impact.go
@@ -14,10 +14,10 @@ import (
 // Impact represents severity rating scale when impact is provided as human-readable string value
 var impactMapping = map[string]int32{
 	"none":     0,
-	"low":      10,
-	"medium":   40,
-	"high":     70,
-	"critical": 100,
+	"low":      20,
+	"medium":   55,
+	"high":     80,
+	"critical": 95,
 }
 
 func (v *Impact) HumanReadable() string {
--- a/explorer/impact.go
+++ b/explorer/impact.go
@@ -11,6 +11,15 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+// Impact represents severity rating scale when impact is provided as human-readable string value
+var impactMapping = map[string]int32{
+	"none":     0,
+	"low":      10,
+	"medium":   40,
+	"high":     70,
+	"critical": 100,
+}
+
 func (v *Impact) HumanReadable() string {
 	if v.Value == nil {
 		return "unknown"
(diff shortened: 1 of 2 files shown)


tas50:
Is the long term goal here also to rename that field so we don't cause confusion about impact vs. risk rating?

explorer/impact.go
```diff
@@ -11,6 +11,15 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+// Impact represents severity rating scale when impact is provided as human-readable string value
+var impactMapping = map[string]int32{
+	"none":     0,
+	"low":      10,
+	"medium":   40,
+	"high":     70,
+	"critical": 100,
```
arlimus: I'd recommend to place the numbers in the middle of each of the ranges. Currently critical is on the top end of the range, while the others are on the low end. I think it would help better balance things. (That said, I could see the argument to leave critical the way you have it)

chris-rock: Right now, I used always the lowest number except for critical because it felt strange to have 90 for critical. I see both options: lowest or middle values working. I would appreciate if you could propose the middle values so I can just adjust to those.

arlimus: The middle would be: 95 critical, 80 high, 55 medium, 20 low, 0 none

arlimus: The problem with having the lower values on everything is that it skews the calculation on all evaluations this way. I.e. it's very easy to have something that is high (low end 70) have a much too lower influence on the overall calculation (of an asset or a policy) because it is at the very bottom end of the scale. That's why I'd recommend against the lower end of the scale. Btw same reason applies to the top end of the scale. Third reason to pick the middle is to give us some wiggle room to adjust severities in policies within each band to prioritize things higher/lower

arlimus:
Superseded and merged in b3b21e8bb007a78c0db87ef3ec5e03e481911ff0

PR #5083, opened 2025-01-09
⭐️ Add tags to objectref

jaym:
We can use the tags to provide additional metadata for variants like what icon to use and what title to show

An example:
```
queries:
  - uid: mondoo-edr-policy-ensure-edr-agent-is-installed
    title: Ensure EDR Agent is installed
    variants:
        - uid: mondoo-edr-policy-ensure-edr-agent-is-installed-macos
          tags:
            mondoo.com/filter-title: macos
        - uid: mondoo-edr-policy-ensure-edr-agent-is-installed-linux
          tags:
            mondoo.com/filter-title: linux
        - uid: mondoo-edr-policy-ensure-edr-agent-is-installed-windows
          tags:
            mondoo.com/filter-title: windows
```


czunker (approved):
LGTM

Could you please add the example from the PR description to the code, when we already have some example variant somewhere? I'm not sure we have something like this already.

PR #5093, opened 2025-01-14
Add featureflag to force shell completions for windows

jaym:
It can be run with

```
MONDOO_FEATURES=ForceShellCompletions cnquery shell local
```

edit:
I guess on windows that would be
```
$env:MONDOO_FEATURES = "ForceShellCompletions"
cnquery shell local
```

screenshot

### DIFF
--- a/cli/shell/completer.go
+++ b/cli/shell/completer.go
@@ -4,6 +4,8 @@
 package shell
 
 import (
+	"runtime"
+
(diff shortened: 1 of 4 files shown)


jaym:
fyi, i tested this in terminal, powershell, and the command shell and they all work on windows server 2022

vjeffrey (approved):

PR #5094, opened 2025-01-15
✨ add ability to fetch ms365 application by id

vjeffrey:
allow users to fetch by name or id
```
 microsoft.application(id: "2efd0330-112c-4971-ab20-eaa54c.....") { * }
```

```
microsoft.application(name: "nametest") { * }

```

### DIFF
--- a/providers/ms365/resources/applications.go
+++ b/providers/ms365/resources/applications.go
@@ -176,17 +176,53 @@ func newMqlMicrosoftApplication(runtime *plugin.Runtime, app models.Applicationa
 }
 
 func initMicrosoftApplication(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error) {
-	// we only look up the application if we have been supplied by its name and nothing else
-	raw, ok := args["name"]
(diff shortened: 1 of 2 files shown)


providers/ms365/resources/applications.go
```diff
@@ -176,18 +176,50 @@ func newMqlMicrosoftApplication(runtime *plugin.Runtime, app models.Applicationa
...
+		name := raw.Value.(string)
+		id, err := fetchApplicationIdByName(runtime, name)
+		if err != nil {
+			return nil, nil, err
+		}
+		appId = *id
 	}
-	name := raw.Value.(string)
+	if appId == "" {
+		raw, ok := args["id"]
```
jaym: should id take precedence if both are specified

jaym:
should there be changes to the resource definition file?

providers/ms365/resources/applications.go
vjeffrey: right now we're assuming that only one will be provided/are rejecting if more than one provided. but ya, if we got both it would be better it would be better to use id

jaym (approved):

PR #5106, opened 2025-01-21
⭐ epoch support in version

arlimus:
Add support for epochs in MQL's `semver` type.

```coffee
> version("1:1.2.3") < version("2:1.0.0")
true
```

This works for both deb/rpm-style epochs ("NN:version") and python-style epochs ([PEP440](https://peps.python.org/pep-0440/)).

Additionally you can access the epoch:

```coffee
# deb/rpm-style epoch
> version("5:1.2.3").epoch
5

# python-style epoch
> version("3!2.1").epoch
3
```

You can compare versions with and without epochs as well:

```coffee
> version("1:1.0") > version("2009.12.03")
true
```

Internal note: This required a new wrapper around the version, which can now be extended with more metadata. Currently we have the epoch stored.


arlimus:
Update after discussion with @chris-rock:

```coffee
> version("1.2.3")

> version("1.2.3", type="semver")
```

chris-rock (approved):
Thank you @arlimus Great improvement.

PR #5111, opened 2025-01-22
:sparkles: Add additional field to GKE: releaseChannel

mm-weber:


providers/gcp/resources/gcp.lr
```diff
@@ -1490,6 +1490,8 @@ private gcp.project.gkeService.cluster @defaults("name description zone status c
   identityServiceConfig dict
   // Configuration options for the NetworkPolicy feature
   networkPolicyConfig dict
+  // Configuration for Release Channel feature
+  releaseChannel dict
```
preslavgerchev: do we plan to extend this or does this only hold a string? if so, its probably best to just make this a `releaseChannel string` in that case.

if it's a dict we should probably rename to something like `releaseConfig` maybe?

providers/gcp/resources/gcp.lr
```diff
@@ -1490,6 +1490,8 @@ private gcp.project.gkeService.cluster @defaults("name description zone status c
   identityServiceConfig dict
   // Configuration options for the NetworkPolicy feature
   networkPolicyConfig dict
+  // The release channel that the cluster is subscribed to.
```
misterpantz:
```suggestion
  // The release channel that the cluster is subscribed to
```

mm-weber: Thanks @misterpantz !

mm-weber:
Converted to draft adding more fields.

preslavgerchev (approved):

PR #5121, opened 2025-01-26
🐛 fix handling of aws filters (ensure filters are copied in cfg clone)

vjeffrey:
fixes https://github.com/mondoohq/cnspec/issues/1531

### DIFF
--- a/providers-sdk/v1/inventory/inventory.go
+++ b/providers-sdk/v1/inventory/inventory.go
@@ -389,11 +389,21 @@ func (p *Platform) PrettyTitle() string {
 type cloneSettings struct {
 	noDiscovery        bool
 	parentConnectionId *uint32
+	withFilters        bool
 }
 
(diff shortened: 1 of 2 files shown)


providers-sdk/v1/inventory/inventory.go
```diff
@@ -433,6 +443,9 @@ func (cfg *Config) Clone(opts ...CloneOption) *Config {
 	if cloneSettings.parentConnectionId != nil {
 		clonedObject.ParentConnectionId = *cloneSettings.parentConnectionId
 	}
+	if cloneSettings.withFilters {
+		clonedObject.Discover.Filter = cfg.Discover.Filter
```
VasilSirakov: Since this is a function that clones a struct, wouldn't it be better as a precaution to create a new map and iteratively copy all elements onto it?

VasilSirakov (approved):

PR #5131, opened 2025-01-29
🧹 add labels to gcp discovered assets

vjeffrey:
this will add the labels to the actual assets that are discovered


tas50:
Do we need to refactor the asset overview policy after this ships?

vjeffrey:
> Do we need to refactor the asset overview policy after this ships?
what's the concern? oh, is it that we may end up with a doubled up list of tags in that view?

jaym (approved):

PR #5132, opened 2025-01-29
🐛 Return empty python packages if dir does not exist.

preslavgerchev:


providers/os/resources/python.go
```diff
@@ -210,8 +208,9 @@ func collectPythonPackages(runtime *plugin.Runtime, fs afero.Fs, path string) ([
 
 	fileList, err := afs.ReadDir(path)
 	if err != nil {
-		if !os.IsNotExist(err) {
+		if os.IsNotExist(err) {
```
preslavgerchev, line 211: the ! felt wrong to me, i think this should be just `if os.IsNotExist` instead of `if !os...`

czunker (approved):
LGTM
Thanks @preslavgerchev

PR #5138, opened 2025-01-29
🐛 fix `contains` on `dict` type

arlimus:
We are experiencing issues where two `dict` types being compared don't come up with the right results. Here is an example I ran to produce this:

```coffee
> region = terraform.plan.variables.where( name == "gcp_region" ).first.value;
  terraform.plan.variables.where(name == "service_vpc_connector").all(v: v.value.contains(region) )
[failed] [].all()
  actual:   [
    0: terraform.plan.variable value="projects/prj-acc7/locations/us-central1/connectors/preprod-gen-central1-01" name="service_vpc_connector"
  ]
```

As you can see above we are looking at the `value` of the varaible, which clearly is set to the region but doesn't come up. If we print out the variables individually, it all comes out correctly.

The problem is that the string comparison doesn't work when we work with `dict` types because in that example we have to create code that has to with with arbitrary content and work with filtering on another `dict`.

This PR fixes the issue.

```coffee
> region = terraform.plan.variables.where( name == "gcp_region" ).first.value;
  terraform.plan.variables.where(name == "service_vpc_connector").all(v: v.value.contains(region) )
[ok] value: true
```

CC @scottford-io


chris-rock (approved):
Great improvement @arlimus

PR #5144, opened 2025-01-31
🐛  github: fix precedence flag vs env-var

afiune:
Github provider has two authentication methods.

1. Application credentials
2. Personal access token

We give precedence to the former and, if both auth methods are provided,
we will output a warning.

Closes https://github.com/mondoohq/cnquery/issues/5036

### DIFF
--- a/providers/github/provider/provider.go
+++ b/providers/github/provider/provider.go
@@ -62,6 +62,13 @@ func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error)
 		conf.Options[connection.OPTION_ENTERPRISE_URL] = string(x.Value)
 	}
 
+	// Github provide has two authentication methods.
+	//
+	// 1. Application credentials
+	// 2. Personal access token
+	//
+	// We give precedence to the former and, if both auth methods are provided,
+	// we will output a warning.
 	isAppAuth := false
 	appId, ok := flags[connection.OPTION_APP_ID]
 	if ok && len(appId.Value) > 0 {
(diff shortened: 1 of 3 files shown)


providers/github/provider/provider.go
```diff
@@ -73,19 +80,25 @@ func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error)
...
+	if token == "" && len(os.Getenv("GITHUB_TOKEN")) != 0 {
 		token = os.Getenv("GITHUB_TOKEN")
+		log.Debug().Msg("loaded token from GITHUB_TOKEN env variable")
 	}
 	if token == "" && !isAppAuth {
 		return nil, errors.New("a valid GitHub authentication is required, pass --token '<yourtoken>', set GITHUB_TOKEN environment variable or provide GitHub App credentials")
 	}
 	if token != "" {
+		if isAppAuth {
+			log.Warn().Msg("both authentication methods provided, using application credentials")
```
vjeffrey, line 101: it looks like here we warn that we're going to use the application credentials but then set the token as the credential? is that correct?

afiune: That is correct. We can also NOT set it, it is a noop at this time so, more like a preference.

vjeffrey: ok. i do find it confusing to set it if we're not using it, but things are working 👍

vjeffrey (approved):

providers/github/provider/provider.go
afiune: Will not set it to avoid confusion. 👍🏽

PR #5147, opened 2025-01-31
Replace repo-level token with GH App

philipbalinov:
RELEASR_ACTION_TOKEN seems to have expired: `Error: Bad credentials`

### DIFF
--- a/.github/workflows/providers.yaml
+++ b/.github/workflows/providers.yaml
@@ -188,10 +188,18 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ ! inputs.skip_publish }}
     steps:
+      - name: Generate token
+        id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/providers.yaml
```diff
@@ -188,10 +188,18 @@ jobs:
...
     if: ${{ ! inputs.skip_publish }}
     steps:
+      - name: Generate token
+        id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.MONDOO_CHECKOUT_APP_ID }}
+          private-key: ${{ secrets.MONDOO_CHECKOUT_APP_PRIVATE_KEY }}
+          owner: mondoohq
+          repositories: |
```
imilchev, line 198: is this correct? aren't we missing repos or something?

philipbalinov: yep, should say `releasr`

imilchev (approved):

PR #5149, opened 2025-01-31
🐛 fix fetching of service accounts on gke cluster nodepool config

vjeffrey:
fixes https://github.com/mondoohq/cnquery/issues/5133

### DIFF
--- a/providers/gcp/resources/iam.go
+++ b/providers/gcp/resources/iam.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/rs/zerolog/log"
 	"go.mondoo.com/cnquery/v11/llx"
 	"go.mondoo.com/cnquery/v11/providers-sdk/v1/plugin"
 	"go.mondoo.com/cnquery/v11/providers/gcp/connection"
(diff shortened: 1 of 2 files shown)


providers/gcp/resources/iam.go
```diff
@@ -74,7 +73,15 @@ func initGcpProjectIamServiceServiceAccount(runtime *plugin.Runtime, args map[st
 			return args, sa, nil
 		}
 	}
-	return nil, nil, errors.New("service account not found")
+
+	args["name"] = llx.NilData
```
imilchev, line 79: why is this change needed? I think the error should be there... If we reach this point in the code, then we tried to find something that doesn't exist. Setting everything to nil will acts as if there is an actual service account with all nil fields

vjeffrey: hmm ya you're right, let me see if there's another way to do this - basically you can have references to external service accounts, which means they wont be found, and that's ok, but we still need to be able to reference it

mm-weber:
Fixes: https://github.com/mondoohq/cnquery/issues/5164

providers/gcp/resources/iam.go
vjeffrey: so i've tried playing around with this, but if we return the error then we block loading the rest of the information, but it's expected to have references to external service accounts

```
go run apps/cnquery/cnquery.go shell gcp project (prod-account) --discover gke-clusters
```

with the error returned, we block the rest of the output:
```
nquery> gcloud.project.gke.clusters { nodePools { config {*} }}
1 error occurred:
	* 1 error occurred:
	* 1 error occurred:
	* service account not found
gcloud.project.gke.clusters: [
  0: {
    nodePools: 1 error occurred:
	* 1 error occurred:
	* service account not found

```

vs.

```
cnquery> gcloud.project.gke.clusters { nodePools { config {*} }}
gcloud.project.gke.clusters: [
  0: {
    nodePools: [
      0: {
        config: {
          gcfsConfig: null
          machineType: "e2-standard-4"
          diskSizeGb: 50
          sandboxConfig: null
          ....
         ```

jaym (approved):

PR #5156, opened 2025-02-03
✨ override builtin functions

afiune:
Today, there is no official way to override builtin functions like `length`, this change tries to do the most minimal change in MQL to allow a unique pattern for providers to override builtin functions.

The core change is an additional check when executing a bound function. We check if the resource defines the builtin function like `length`, and if so, we prioritize it. If the provider doesn't define it, we execute the function as we did before, by loading first the builtin function and if not existing, then we run the resource function.

Here is an example of this pattern to override the `length` builtin function.

Having a resource that exposes a function that loads big amounts of resources:

```
cloud {
  resources() []resources
}
```

The provider implementation would look something like this:
```go
func (c *mqlCloud) resources() ([]interface, error) {
	// API call to fetch all resources
}
```

When running the MQL query `cloud.resources.length`, we first load all the resources, and then we count them.

Since we do not need any information about the resources themselves, this could delay policies if we do things like:
```
cloud.resources.length < 5 && cloud.resources.length > 1
```

An alternative to improve these resources is to override the builtin function with a more performant implementation.

```
cloud {
  // update function with a new custom list resource
  resources() customResources
}

// definition of a custom list resource
customResources {
  []resource

  // overrides builtin function
  length() int
}
```

The new implementation moves the logic that loads the resources into a `list()` function and exposes a new function that overrides the builtin function:
would look like:
```go
func (c *mqlCloud) resources() (*mqlCustomResources, error) {
	// Only initializes the custom list resource
}
func (c *mqlCustomResources) list() ([]interface, error) {
	// API call to fetch all resources
}
// length() overrides the builtin function,
func (c *mqlCustomResources) length() (int64, error) {
        // This should be a more performant way to count the "resources"
}
```

Additionally, this change moves the `findField()` function from the `compiler` to the resource `Schema`.


chris-rock:
Converted to draft as it requires more discussion with @arlimus

afiune:
Talked with @arlimus and we are ready to review this PR.

mql/mql_test.go
```diff
@@ -350,3 +350,25 @@ func TestDictMethods(t *testing.T) {
...
+		{
+			Code:        "mos.groups.length",
+			ResultIndex: 0, Expectation: int64(5),
+		},
+		// This calls the native builtin `length` function
+		{
+			Code:        "mos.groups.list.length",
+			ResultIndex: 0, Expectation: int64(7),
+		},
+		// Same here, builtint `length` function
```
arlimus: nit: builtin

afiune:
```suggestion
		// Same here, builtin `length` function
```

arlimus (approved):
Thank you for contributing the overrides @afiune ! 

For anyone who may find this PR in the future:

We introduce this builtin override pattern as an effective short-term pattern to take care of large-scale MQL. It gives us a ton of flexibility to add optimized solutions for some of these super long-running calls.

In parallel we are working on a separate RFC to tackle this problem long-term. The advantages of this override-pattern is the ability to do very efficient calls, which are just not possible today. The reasons we will take a different path mid-term is to avoid changes to the underlying expected user behavior of these builtin functions (eg: other types of return values, other behaviors despite having the same name). (from my past experience in eg ruby and scala: it's a fun feature, very powerful, but is harder to read and easier to make mistakes)

PR #5173, opened 2025-02-07
🐛multi vol followup

slntopp:


providers/os/connection/device/linux/device_manager.go
```diff
@@ -91,15 +92,15 @@ func (d *LinuxDeviceManager) IdentifyMountTargets(opts map[string]string) ([]*sn
 		partitions = append(partitions, partitionsForDevice...)
 	}
 
-	if err := errors.Join(errs...); err != nil {
-		return partitions, err
-	}
-
 	if opts[SkipAttemptExpandPartitions] == "true" {
+		log.Warn().Msg("!!! skipping attempt to expand partitions")
```
preslavgerchev: debug log?

providers/os/connection/device/linux/device_manager.go
```diff
@@ -91,15 +92,15 @@ func (d *LinuxDeviceManager) IdentifyMountTargets(opts map[string]string) ([]*sn
...
 		return partitions, nil
 	}
 
-	return d.attemptExpandPartitions(partitions)
+	partitions, err = d.attemptExpandPartitions(partitions)
```
preslavgerchev, line 104: since we're not returning on err anymore, does `attemptExpandPartitions` protect against possible nil arrays being passed in?

slntopp: it's range loops all the way, but i've added a safety check anyway

providers/os/connection/device/linux/device_manager.go
slntopp: oops

preslavgerchev (approved):

PR #5183, opened 2025-02-08
🐛 azure: fail gracefully if unable to get security contacts

afiune:
Closes https://github.com/mondoohq/cnquery/issues/4997

### DIFF
--- a/providers/azure/resources/cloud_defender.go
+++ b/providers/azure/resources/cloud_defender.go
@@ -17,7 +17,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/security/armsecurity"
-	security "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/security/armsecurity"
+	"github.com/rs/zerolog/log"
 )
 
 const (
@@ -61,7 +61,7 @@ func (a *mqlAzureSubscriptionCloudDefenderService) defenderForServers() (interfa
 	if err != nil {
 		return nil, err
 	}
-	vmPricing, err := clientFactory.NewPricingsClient().Get(ctx, fmt.Sprintf("subscriptions/%s", subId), "VirtualMachines", &security.PricingsClientGetOptions{})
+	vmPricing, err := clientFactory.NewPricingsClient().Get(ctx, fmt.Sprintf("subscriptions/%s", subId), "VirtualMachines", &armsecurity.PricingsClientGetOptions{})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


afiune:
Failing because of https://github.com/mondoohq/cnquery/pull/5184

afiune:
## Test
```
cnquery> azure.subscription.cloudDefender.securityContacts {alertNotifications }
! fail gracefully error="failed to fetch security contacts from https://management.azure.com/subscriptions/1234567-abc6-1234-xyz3-3943298239/providers/Microsoft.Security/securityContacts: 500 Internal Server Error"
azure.subscription.cloudDefender.securityContacts: []
```

preslavgerchev (approved):

providers/azure/resources/cloud_defender.go
```diff
@@ -454,19 +454,21 @@ func (a *mqlAzureSubscriptionCloudDefenderService) securityContacts() ([]interfa
 	if err != nil {
 		return nil, err
 	}
+	res := []interface{}{}
 	list, err := getSecurityContacts(ctx, armConn)
 	if err != nil {
-		return nil, err
+		// https: //github.com/mondoohq/cnquery/issues/4997
+		log.Warn().Err(err).Msg("fail gracefully")
```
chris-rock: Why is that a warn?

providers/azure/resources/cloud_defender.go
```diff
@@ -454,19 +454,21 @@ func (a *mqlAzureSubscriptionCloudDefenderService) securityContacts() ([]interfa
...
+		return res, nil
 	}
-	res := []interface{}{}
 	for _, contact := range list {
 		alertNotifications, err := convert.JsonToDict(contact.Properties.AlertNotifications)
```
chris-rock: do we need to check if contact or contact.Properties are set?

providers/azure/resources/cloud_defender.go
```diff
@@ -454,19 +454,21 @@ func (a *mqlAzureSubscriptionCloudDefenderService) securityContacts() ([]interfa
...
 		if err != nil {
-			return nil, err
+			log.Debug().Err(err).Msg("unable to convert armsecurity.Contact.Properties.AlertNotifications to dict")
 		}
 		notificationsByRole, err := convert.JsonToDict(contact.Properties.NotificationsByRole)
```
chris-rock: do we need to check if contact or contact.Properties are set?

providers/azure/resources/cloud_defender.go
```diff
@@ -454,19 +454,21 @@ func (a *mqlAzureSubscriptionCloudDefenderService) securityContacts() ([]interfa
...
+		// https: //github.com/mondoohq/cnquery/issues/4997
+		log.Warn().Err(err).Msg("fail gracefully")
+		return res, nil
 	}
-	res := []interface{}{}
 	for _, contact := range list {
 		alertNotifications, err := convert.JsonToDict(contact.Properties.AlertNotifications)
 		if err != nil {
-			return nil, err
+			log.Debug().Err(err).Msg("unable to convert armsecurity.Contact.Properties.AlertNotifications to dict")
```
chris-rock: in which cases can this happen? We should continue to error out and handle the case properly

providers/azure/resources/cloud_defender.go
afiune: I was debating between this an a simple debug log message, I fee that is a better option, would you agree?

providers/azure/resources/cloud_defender.go
afiune: yeah, we probably should have done that, I'll add  the check.

providers/azure/resources/cloud_defender.go
afiune: yeah, we probably should have done that, I'll add  the check.

providers/azure/resources/cloud_defender.go
afiune: We haven't seen this but, the issue says that, any error trying to get the security contacts should fail gracefully. So I checked other places that we could end up in an undesirable error.

afiune:
@chris-rock updated with your comments, thank you!

Also, I much rather go this way https://github.com/mondoohq/cnquery/pull/5185 - Thoughts?

PR #5198, opened 2025-02-09
🧹 Cleanup extra port group fields/resources in the vsphere provider

tas50:
These are not there in code


vjeffrey:
https://github.com/mondoohq/cnquery/pull/4813 arent some of those in there?

vjeffrey:
oh nevermind, i see the field changed to camelcase

vjeffrey (approved):

PR #5199, opened 2025-02-10
🐛 Fix namespace for Red Hat PURLs

czunker:


czunker:
This is how they are formatted in OSV: https://security.access.redhat.com/data/osv/RHBA-2024:6680.json

providers/os/resources/purl/purl.go
```diff
@@ -80,8 +80,15 @@ func NewPackageURL(pf *inventory.Platform, t Type, name, version string, modifie
 		// and if the distro is set via labels, set it as a namespace
 		if pf.Labels != nil && pf.Labels[detector.LabelDistroID] != "" {
 			purl.Namespace = pf.Labels[detector.LabelDistroID]
+		} else if pf.Name != "" {
+			// Fallback to the platform name if the distro-id is not set
+			purl.Namespace = pf.Name
 		}
 	}
+	// The distro might be rhel, but the namespace for the PURL needs to be redhat.
+	if purl.Namespace == "rhel" {
```
chris-rock: that feels strange, we should not set it to rhel in the first place.

providers/os/resources/purl/purl.go
```diff
@@ -80,8 +80,15 @@ func NewPackageURL(pf *inventory.Platform, t Type, name, version string, modifie
 		// and if the distro is set via labels, set it as a namespace
 		if pf.Labels != nil && pf.Labels[detector.LabelDistroID] != "" {
 			purl.Namespace = pf.Labels[detector.LabelDistroID]
+		} else if pf.Name != "" {
+			// Fallback to the platform name if the distro-id is not set
+			purl.Namespace = pf.Name
```
chris-rock, line 81: why do we need that?

providers/os/resources/purl/purl.go
czunker: Removed.

providers/os/resources/purl/purl.go
czunker: I changed this, so that we only rely on the platform name for the namespace.

chris-rock (approved):

PR #5202, opened 2025-02-10
✨ Do not error out on SBOM generation, use the sbom ErrorMessage instead

preslavgerchev:
This ensures that if the SBOM query pack errors out, we still get an SBOM pack with a status of Failed and the correct error message. If one of the queries error out the whole SBOM generation will fail. For example, this report will fail without the current changes:
```json
// this is just the sbom packages piece, the rest is omitted for brevity
"//local.cnquery.io/run/local-execution/queries/mondoo-sbom-packages": {
                    "content": {
                        "packages.list": {
                            "error": "error fetching sbom packages for some reason"
                        }
                    }
                }
```

### DIFF
--- a/sbom/generator/generator.go
+++ b/sbom/generator/generator.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/pkg/errors"
 	"go.mondoo.com/cnquery/v11"
 	"go.mondoo.com/cnquery/v11/cli/reporter"
 	"go.mondoo.com/cnquery/v11/mrn"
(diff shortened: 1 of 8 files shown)


sbom/generator/generator.go
```diff
@@ -22,7 +23,13 @@ func NewBom(report *reporter.Report) ([]*sbom.Sbom, error) {
 	return GenerateBom(report)
 }
 
+func GenerateBomV2(r *reporter.Report) []*sbom.Sbom {
```
chris-rock: why do we need that? Can we not just use GenerateBom and return nil as error?

preslavgerchev: I wanted to avoid forcing any upstream repos that use this to migrate since the signature changes. We could also change `GenerateBom`'s signature to never return an error too, I think this cleans up the usage for callers here as they don't have to handle an error

VasilSirakov: But you already change the behavior of `GenerateBOM` to never return an error so the underlying usages would have a redundant err check if they don't migrate to V2. Fixing the compilation error after changing the signature would be trivial and actually an improvement.

preslavgerchev: Agreed, I will change this

chris-rock (approved):
Thank you @preslavgerchev

PR #5204, opened 2025-02-11
🐛 Do not stop sbom generation on the first encountered unmarshal error.

VasilSirakov:
#### In this PR:
- removed the short-circuiting on the first encountered unmarshal error.
- added test cases with more detailed assertions ensuring that the unaffected packages in the sbom are still filled in regardless if an error is encountered.


sbom/generator/generator.go
```diff
@@ -201,3 +205,12 @@ func enrichPlatformIds(ids []string) []string {
 	}
 	return platformIds
 }
+
+func getSortedMapKeys[T any](m map[string]T) []string {
```
preslavgerchev: you can also use an iterator for this since go 1.23:
```
for k := range maps.Keys(m) {
        // ...
    }
```

VasilSirakov: Nice, didn't know that.

preslavgerchev (approved):

PR #5208, opened 2025-02-12
✨ version(.., type: "semver")

arlimus:
```
cnquery> version("1.2.3", type: 'semver')
version: 1.2.3

cnquery> version("1:1.2.3", type: 'semver')
version '1:1.2.3' is not a semantic version
version: 1:1.2.3

cnquery> version("1:1.2.3", type: 'debian')
version: 1:1.2.3

cnquery> version("1!1.2.3", type: 'python')
version: 1!1.2.3

cnquery> version("1:1.2.3")
version: 1:1.2.3
```

Also adds support for parsing named arguments on builtin type conversion functions (like `version`).

Part 1 of https://github.com/mondoohq/cnquery/pull/5106#issuecomment-2606678743

### DIFF
--- a/llx/builtin_version.go
+++ b/llx/builtin_version.go
@@ -37,6 +37,10 @@ func NewVersion(s string) Version {
 	}
 }
 
+func (v *Version) IsSemver() bool {
+	return v.Version != nil && v.epoch == 0
+}
+
 func parseEpoch(v string) (*semver.Version, int) {
 	prefix := reEpoch.FindString(v)
 	if prefix == "" {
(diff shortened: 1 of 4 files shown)


llx/builtin_version.go
```diff
@@ -37,6 +37,10 @@ func NewVersion(s string) Version {
 	}
 }
 
+func (v *Version) IsSemver() bool {
+	return v.Version != nil && v.epoch == 0
```
chris-rock: I am not sure I understand why epoch 0 is an indicator of semver? Should we not try to parse it an return false when semver.NewVersion returns an error?

chris-rock (approved):
Thank you @arlimus

PR #5212, opened 2025-02-13
🐛 Align photon PURL namespace with ecosystem

czunker:
Align the namespace with the osv ecosystem.

'pkg:rpm/photon/...' -> 'pkg:rpm/photon%20os/...'

### DIFF
--- a/providers/os/resources/purl/purl.go
+++ b/providers/os/resources/purl/purl.go
@@ -77,9 +77,14 @@ func NewPackageURL(pf *inventory.Platform, t Type, name, version string, modifie
 		// use the platform architecture for the package
 		purl.Arch = pf.Arch
 
-		if pf.Name != "" {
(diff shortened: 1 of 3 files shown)


providers/os/resources/purl/purl.go
```diff
@@ -77,9 +77,14 @@ func NewPackageURL(pf *inventory.Platform, t Type, name, version string, modifie
 		// use the platform architecture for the package
 		purl.Arch = pf.Arch
 
-		if pf.Name != "" {
-			purl.Namespace = pf.Name
+		platformName := pf.Name
```
chris-rock: I recommend we rename it to `purlNamespace := pf.Name` to make clear that we do not change the platform name but instead determine the correct purl namespace.

chris-rock (approved):
Thank you @czunker

PR #5216, opened 2025-02-14
🧹 Wire up the remainder of aws.iam.role.createdAt

tas50:
It was in the code but not the lr which caused failures

```
[failed] [].all()
  error: [aws] cannot set 'createdAt' in resource 'aws.iam.group', field not found
```


chris-rock:
The resource is already initialized with the value https://github.com/mondoohq/cnquery/blob/main/providers/aws/resources/aws_iam.go#L1428

chris-rock (approved):
Thank you @tas50

PR #5218, opened 2025-02-14
🐛  fix aws s3 bucket public()

afiune:
We now check Public Access Block settings first, then Bucket PolicyStatus
which is the native way that AWS computes if a bucket is public or not, but
if that doesn't work, we still parse the bucket policy to see if it public.

### DIFF
--- a/providers/aws/resources/aws_s3.go
+++ b/providers/aws/resources/aws_s3.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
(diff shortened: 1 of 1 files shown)


vjeffrey:
this looks to be evaluating public correctly, thank you!
```
  30: {
    public: true
    policy: {
      version: ""
      document: "{}"
      statements: []
      name: ""
      id: ""
    }
    name: "test-vj-public"
    location: "us-east-2"
  }
  31: {
    public: false
    policy: {
      version: ""
      document: "{}"
      statements: []
      name: ""
      id: ""
    }
    name: "testint2"
    location: "us-east-2"
  }
  ```
  any idea why the policy is empty for all of these?

afiune:
@vjeffrey Yes. The problem is that `aws.s3.bucket.policy` (`mqlAwsS3BucketPolicy`) doesn't implement IDs properly.

vjeffrey:
> @vjeffrey Yes. The problem is that aws.s3.bucket.policy (mqlAwsS3BucketPolicy) doesn't implement IDs properly.

kk, would b great to fix in for that, since itll be hard for folks to evalluate their s3 bucket policy if we dont return the value for it

vjeffrey (approved):

PR #5227, opened 2025-02-14
🐛 fix core `asset.labels` + ✨  new `asset.platformMetadata`

afiune:
The labels of an asset come from the asset itself, not the platform, but as some point
we made a mistake to use the platform labels, this PR is the first step to fix this issue.

In this PR we are:
* Introducing a new field to the proto message `asset.Platform.Metadata` with the goal
to give more clarity and avoid confusions with `asset.Labels`
* We are deprecating `asset.Platform.Labels` in favor of `asset.Platform.Metadata` (this field
will be remove in `v12`)
* We are introducing a new `asset.platformMetadata` to the `asset` MQL schema, this
new field will have the `asset.Platform.Metadata`
* We are fixing the `asset.labels` MQL resource to have the actual asset labels but, we are
going to merge this field with the platform labels for backwards compatibility

After merging this PR, we will start migrating the use of `asset.Platform.Labels` to
`asset.Platform.Metadata`, so that in `v12` we can remove `asset.Platform.Labels.


arlimus:
Why did we use Platform.Labels before and what change does it make to switch to Asset.Labels? Are there different code paths that set these across different providers? I'm worried that with both fields existing we might run into similar problems again.

chris-rock:
I do some extra testing but this is what I discussed with @afiune as a solution. It stuck me that we use asset.Annotations but not asset.Labels. I am 99% sure the labels from platform are actually incorrect.

arlimus:
@chris-rock I asked @afiune to investigate the solution, so we make sure it doesn't break things. We shouldn't have both structures because it can lead to bugs like this one :)

afiune:
Thank you both, I agree with @arlimus, if we don't use one, we should get rid of it.

I will do some digging and report back any findings.

afiune:
@chris-rock @arlimus I found a few places where we are using `asset.platform.labels`, I can update the tests but I want to check with you both if this change is heading in the right direction, or not.

cli/reporter/cnquery_report.pb.go
```diff
@@ -4,7 +4,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.35.2
-// 	protoc        v5.29.2
+// 	protoc        v5.29.3
```
chris-rock: Those updates should not be part of this PR since they distract the review.

providers-sdk/v1/upstream/mvd/mvd.go
```diff
@@ -26,18 +26,18 @@ func (r *VulnReport) Cves() []*CVE {
 	return cveList
 }
 
-// MvdPlatform converts the inventory.Platform to the
+// MvdPlatform converts the inventory.Asset that contains the platform to the
 // platform object we use for MVD
-func NewMvdPlatform(pf *inventory.Platform) *Platform {
-	if pf == nil {
+func NewMvdPlatform(asset *inventory.Asset) *Platform {
```
chris-rock: That is not correct. Mvd Platform is just a conversion of the platform object.

providers/os/resources/packages/apk_packages.go
```diff
@@ -137,8 +137,8 @@ func ParseApkUpdates(input io.Reader) (map[string]PackageUpdate, error) {
 
 // Arch, Manjaro
 type AlpinePkgManager struct {
-	conn     shared.Connection
-	platform *inventory.Platform
+	conn  shared.Connection
+	asset *inventory.Asset
```
chris-rock: pkg managers do not need to know the asset information

providers-sdk/v1/inventory/inventory.proto
```diff
@@ -304,7 +304,12 @@ message Platform {
   // ^^
 
   string runtime = 21;
+  // FIXME: DEPRECATED, remove in v12 (or later) vv
   map<string, string> labels = 22;
```
chris-rock, line 308: we to set this id as locked? like reserved 22

providers/core/resources/core.lr
```diff
@@ -44,6 +44,8 @@ asset @defaults("name platform version") {
   fqdn string
   // Build version of the platform (optional)
   build string
+  // Platform Metadata (e.g. key values from /etc/os/release)
+  platformMetadata map[string]string
```
chris-rock, line 48: +1

providers/core/provider/provider.go
```diff
@@ -62,20 +63,29 @@ func (s *Service) Connect(req *plugin.ConnectReq, callback plugin.ProviderCallba
 	}
 
 	asset := req.Asset
+	// FIXME: remove in v12 (or later) vv
+	// we merge `asset.Labels` and `asset.Platform.Labels` for backwards compatibility
+	assetLabelsMergedV11Compat := mapx.Merge(asset.Labels, asset.Platform.Labels)
```
chris-rock: +1

chris-rock: platform labels has precedence over asset label

chris-rock (approved):
Approval pending the discussed changes

PR #5228, opened 2025-02-15
🐛 fix `aws.s3.bucket.policy` resource

afiune:
This change allows running `aws.s3.bucket.policy` when in discovery mode.

```
cnspec shell aws --discover s3-buckets
```

Also, fixes the issue discovered at https://github.com/mondoohq/cnquery/pull/5218
where we noticed that the policy resource doesn't implement the `id()` func properly.

Closes https://github.com/mondoohq/cnquery/issues/5169

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -1822,9 +1822,11 @@ private aws.s3.bucket.corsrule @defaults("name") {
 }
 
 // Amazon S3 bucket policy
-private aws.s3.bucket.policy @defaults("bucketName exists version") {
+private aws.s3.bucket.policy @defaults("bucketName version") {
   // Unique ID for the policy
   id string
+  // Deprecated
+  name string
   // Bucket name that this policy belongs
   bucketName string
   // Document for the policy
@@ -1833,10 +1835,7 @@ private aws.s3.bucket.policy @defaults("bucketName exists version") {
   version() string
   // List of statements for the policy
   statements() []dict
-  // Whether the bucket policy exists
-  exists bool
 }
-
 
 // AWS Application Auto Scaling
 aws.applicationAutoscaling @defaults("namespace") {
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -1825,7 +1825,7 @@ private aws.s3.bucket.corsrule @defaults("name") {
 private aws.s3.bucket.policy @defaults("bucketName version") {
   // Unique ID for the policy
   id string
-  // Deprecated
+  // Deprecated. Use `bucketName` instead
   name string
   // Bucket name that this policy belongs
   bucketName string
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -1822,17 +1822,19 @@ private aws.s3.bucket.corsrule @defaults("name") {
 }
 
 // Amazon S3 bucket policy
-private aws.s3.bucket.policy @defaults("name version") {
+private aws.s3.bucket.policy @defaults("bucketName exists version") {
   // Unique ID for the policy
   id string
-  // Name for the policy
-  name string
+  // Bucket name that this policy belongs
+  bucketName string
   // Document for the policy
   document string
   // Version of the policy
   version() string
   // List of statements for the policy
   statements() []dict
+  // Whether the bucket policy exists
+  exists bool
 }
 
 
--- a/providers/aws/resources/aws_s3.go
+++ b/providers/aws/resources/aws_s3.go
@@ -191,21 +191,6 @@ func (a *mqlAwsS3Bucket) id() (string, error) {
 	return a.Arn.Data, nil
 }
 
-func (a *mqlAwsS3Bucket) emptyAwsS3BucketPolicy() (*mqlAwsS3BucketPolicy, error) {
-	res, err := CreateResource(a.MqlRuntime, "aws.s3.bucket.policy", map[string]*llx.RawData{
-		"bucketName": llx.StringData(a.Name.Data),
-		"document":   llx.StringData("{}"),
-		"version":    llx.StringData(""),
-		"id":         llx.StringData(""),
-		"exists":     llx.BoolData(false),
-		"statements": llx.ArrayData([]interface{}{}, types.Dict),
-	})
-	if err != nil {
-		return nil, err
-	}
-	return res.(*mqlAwsS3BucketPolicy), nil
-}
-
 func (a *mqlAwsS3Bucket) policy() (*mqlAwsS3BucketPolicy, error) {
 	conn := a.MqlRuntime.Connection.(*connection.AwsConnection)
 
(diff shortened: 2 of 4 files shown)


providers/aws/resources/aws.lr
```diff
@@ -1822,17 +1822,19 @@ private aws.s3.bucket.corsrule @defaults("name") {
...
 
 // Amazon S3 bucket policy
-private aws.s3.bucket.policy @defaults("name version") {
+private aws.s3.bucket.policy @defaults("bucketName exists version") {
   // Unique ID for the policy
   id string
-  // Name for the policy
-  name string
+  // Bucket name that this policy belongs
+  bucketName string
```
afiune, line 1831: This might be a breaking change, but a bucket policy doesn't have a name. It belongs to a bucket and, to calculate the bucket policy ID we need the bucket name, so I re purpose this field.

If this is not acceptable, I am ok creating an internal struct to pass the bucket name but, we need to do something about this `name`field that says it is the name of the policy, but it's not.

providers/aws/resources/aws.lr
```diff
@@ -1822,17 +1822,19 @@ private aws.s3.bucket.corsrule @defaults("name") {
...
   // Document for the policy
   document string
   // Version of the policy
   version() string
   // List of statements for the policy
   statements() []dict
+  // Whether the bucket policy exists
+  exists bool
```
afiune: Buckets can have a policy, but not all of them has it. This is very helpful, I think.

providers/aws/resources/aws.lr
imilchev: we don't know whether this is used by someone or not. Perhaps keep both for a while and add a comment to `name` that it's deprecated. We can remove it in next major release

providers/aws/resources/aws.lr
imilchev: I don't understand why this is needed. If there is no bucket policy then that resource shouldn't be created, right? Do we have some specific requirement to have the MQL object even when such resource in AWS doesn't exist?

providers/aws/resources/aws.lr
afiune: If someone is using it, they are using it incorrectly but I agree, I'll put it back with a deprecation.

providers/aws/resources/aws.lr
vjeffrey: we should be able to use null comparisons if we need to check that a policy exists or does not exist on a bucket. like aws.s3.bucket.policy != null -- does that not work?

afiune: It sounds it is possible! 🙌🏽 Thank you both for the correction 💯

vjeffrey:
i'm hitting a panic when i try to test this
```
cnquery> aws.s3.buckets { policy != null }
[signal SIGSEGV: segmentation violation code=0x2 addr=0x8 pc=0x1071ee7d0]

goroutine 176 [running]:
go.mondoo.com/cnquery/v11/providers/aws/resources.(*mqlAwsS3BucketPolicy).MqlID(0x10cd66390?)
	/Users/vj/go/src/go.mondoo.io/cnquery/providers/aws/resources/aws.lr.go:20888
```

providers/aws/resources/aws_s3.go
```diff
@@ -205,28 +202,33 @@ func (a *mqlAwsS3Bucket) policy() (*mqlAwsS3BucketPolicy, error) {
 	policy, err := svc.GetBucketPolicy(ctx, &s3.GetBucketPolicyInput{
 		Bucket: &bucketname,
 	})
-	if err != nil {
-		if isNotFoundForS3(err) {
-			return emptyAwsS3BucketPolicy(a.MqlRuntime)
-		}
+	if err != nil && !isNotFoundForS3(err) {
```
vjeffrey: perhaps it would be useful to set the plugin state here? `plugin.StateIsNull | plugin.StateIsSet`

afiune: Nice, I learned something new today. Thanks @vjeffrey 👍🏽

vjeffrey:
hmm. the other panic is gone:
```
cnquery> aws.s3.buckets { policy != null }
aws.s3.buckets: [
  0: {
    policy != null: false
  }
```

but i'm seeing this:
```
cnquery> aws.s3.buckets { public }
[signal SIGSEGV: segmentation violation code=0x2 addr=0x98 pc=0x10bb6fbb8]

goroutine 83 [running]:
go.mondoo.com/cnquery/v11/providers/aws/resources.(*mqlAwsS3BucketPolicy).parsePolicyDocument(0x14000faa008?)
	/Users/vj/go/src/go.mondoo.io/cnquery/providers/aws/resources/aws_s3.go:714 +0x18
go.mondoo.com/cnquery/v11/providers/aws/resources.(*mqlAwsS3Bucket).public(0x14000faa008)
	/Users/vj/go/src/go.mondoo.io/cnquery/providers/aws/resources/aws_s3.go:462 +0x23c
```

mm-weber:
@afiune

Tested this PR on the proposed solution, something seems to be off, still:
image

This does throw an error, too:
image

Tested this against via `cnspec shell aws --discover s3-buckets` using a s3 bucket with those properties:
image via `aws.s3.bucket {*}` .

afiune:
@vjeffrey @mm-weber ugh, I think that there are a few moving parts that I missed, specially since we merged https://github.com/mondoohq/cnquery/pull/5218 and that is where it's throwing a panic.



Will have something to review soon.

providers/aws/resources/aws_s3.go
```diff
@@ -133,6 +158,9 @@ func initAwsS3Bucket(runtime *plugin.Runtime, args map[string]*llx.RawData) (map
 	var arn string
 	if args["arn"] != nil {
 		arn = args["arn"].Value.(string)
+		if !strings.HasPrefix(arn, "arn:aws:s3:") {
+			return nil, nil, errors.Newf("not a valid bucket ARN '%s'", arn)
```
afiune, line 162: During testing I detected that, if you are NOT in discovery and, you try to access `aws.s3.bucket` then we try to "create a bucket resource" of the AWS account, which is incorrect.

Now, we will exit with helpful error message:
```
cnquery> aws.s3.bucket
not a valid bucket ARN 'arn:aws:sts::123456789012'
aws.s3.bucket: no data available
```

providers/aws/resources/aws.lr
```diff
@@ -1822,11 +1822,13 @@ private aws.s3.bucket.corsrule @defaults("name") {
 }
 
 // Amazon S3 bucket policy
-private aws.s3.bucket.policy @defaults("name version") {
+private aws.s3.bucket.policy @defaults("bucketName version") {
   // Unique ID for the policy
   id string
-  // Name for the policy
+  // Deprecated
```
vjeffrey: nit: can you add "use bucketName instead" so that the deprecation reason is clear?

afiune:
```suggestion
  // Deprecated. Use `bucketName` instead
```

afiune: Done

vjeffrey (approved):

PR #5236, opened 2025-02-18
🐛 parse malformed `package.json` field `engines`

afiune:
Closes https://github.com/mondoohq/cnquery/issues/5235


chris-rock (approved):
Thank you @afiune

PR #5238, opened 2025-02-18
🐛 Detect windows 11 name

jaym:
Windows 11 returns `Windows 10` for `ProductName` in `HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion`. We use that as the platform title. This corrects the platform title if its a build that falls into with windows 11 build numbers


chris-rock (approved):
Thank you @jaym

PR #5245, opened 2025-02-19
🐛 k8s: don't fail during api resource discovery

afiune:
Closes https://github.com/mondoohq/cnquery/issues/5165

A sand-boxed cluster doesn't have metrics server enabled:
https://cloud.google.com/kubernetes-engine/docs/concepts/sandbox-pods#limitations-incompatible

This code fails just because that single API resource, out of 40+ resources, fails.

Screenshot 2025-02-18 at 6 30 06 PM

Now we will just display a warning that "some" api resources aren't available.

### DIFF
--- a/providers/k8s/connection/shared/resources/discovery.go
+++ b/providers/k8s/connection/shared/resources/discovery.go
@@ -73,9 +73,8 @@ func (d *Discovery) SupportedResourceTypes() (*ApiResourceIndex, error) {
 	log.Debug().Msg("query api resource types")
 	resList, err := d.discoveryClient.ServerPreferredResources()
 	if err != nil {
-		var groupDiscoveryErr *discovery.ErrGroupDiscoveryFailed
-		if errors.As(err, groupDiscoveryErr) {
-			log.Warn().Err(err).Msg("one or more kubernetes API groups fail to load")
+		if discovery.IsGroupDiscoveryFailedError(err) {
+			log.Debug().Err(err).Msg("one or more kubernetes API groups fail to load")
 		} else {
 			return nil, errors.Wrap(err, "failed to fetch api resource types from kubernetes")
 		}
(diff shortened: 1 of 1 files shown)


providers/k8s/connection/shared/resources/discovery.go
```diff
@@ -73,7 +73,7 @@ func (d *Discovery) SupportedResourceTypes() (*ApiResourceIndex, error) {
 	log.Debug().Msg("query api resource types")
 	resList, err := d.discoveryClient.ServerPreferredResources()
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to fetch api resource types from kubernetes")
+		log.Warn().Err(err).Msg("failed to fetch api resource types from kubernetes")
```
chris-rock: It does not feel right to not throw the error here.

afiune:
Talked with @imilchev and we agreed that this is the right place but, we should not ignore
all errors, but instead only the `ErrGroupDiscoveryFail`.

providers/k8s/connection/shared/resources/discovery.go
```diff
@@ -10,7 +10,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/pkg/errors"
```
afiune, line 13: We should not use this package, it is archived.

providers/k8s/connection/shared/resources/discovery.go
```diff
@@ -73,7 +73,12 @@ func (d *Discovery) SupportedResourceTypes() (*ApiResourceIndex, error) {
 	log.Debug().Msg("query api resource types")
 	resList, err := d.discoveryClient.ServerPreferredResources()
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to fetch api resource types from kubernetes")
+		var groupDiscoveryErr *discovery.ErrGroupDiscoveryFailed
+		if errors.As(err, groupDiscoveryErr) {
+			log.Warn().Err(err).Msg("one or more kubernetes API groups fail to load")
```
imilchev: nit: I think this one is not necessarily a warning. I would log it as a debug message. 

Also you can do:
```go
if discovery.IsGroupDiscoveryFailedError(err) {
  ...
}
```

imilchev (approved):

PR #5247, opened 2025-02-20
⭐ show assessments inside of blocks

### LINKED ISSUE #5246: Improve Readability of .containsNone Function Output in Terraform Checks
The `.containsNone` function in MQL is useful for ensuring that specific strings do not appear within an array. However, when used in Terraform checks, the output it returns upon failure is difficult to interpret due to excessive verbosity. Instead of providing a concise summary of which values failed the check, the function outputs the entire resource change structure, making it challenging to identify the offending values.

For example, when checking if a specific GCP API (`parallelstore.googleapis.com`) is not enabled in a Terraform plan file, the following query is used:

```
terraform.plan.resourceChanges.where( name == "enable_project_apis").all( change.after.triggers.apis.split(" ").containsNone(["parallelstore.googleapis.com"]))
```

If this check fails, the current output is a large and difficult-to-parse structure that includes all APIs enabled in the plan, rather than clearly highlighting the presence of the restricted API.

**Expected Behavior**

- The output should clearly indicate which value(s) caused the failure.
- Provide a concise and readable summary of the violation, rather than the entire resource change object.
- The output should clearly indicate which value(s) caused the failure.
- Provide a concise and readable summary of the violation, rather than the entire resource change object.

Example:

```
[failed] containsNone check failed: Found restricted API(s): ["parallelstore.googleapis.com"]
```

**Actual Behavior**

- The function outputs an extensive list of all APIs included in the Terraform plan.
- Parsing through the response to find the violating API is cumbersome.
- Unnecessary duplication of data makes debugging difficult.

**Suggested Enhancement**

- Modify `.containsNone` to return only the violating elements instead of the entire list.
- Provide a human-readable summary of the failure.
- Consider adding an option for structured output (e.g., JSON) for better parsing in automation workflows.

**Use Case:**

This improvement will help security teams and infrastructure engineers quickly identify compliance violations without needing to sift through lengthy Terraform plan outputs.

**References**

Example failure output (current behavior) can be found in the original issue description above.

Impact: Improves usability and debugging efficiency when working with .containsNone in MQL queries.

arlimus:
This has been a very long time coming. Despite being an important step forward, it's also only just the first step in terms of the code changes that will be necessary to clean this up long-term.

So, ...

## What is this?

Currently if we run this:

```coffee
> [1,2,3].containsOnly([1,2])
[failed] [].containsOnly()
  expected: == []
  actual:   [
    0: 3
  ]
```

It shows us what's up. 

But what if we put it inside a block?

```coffee
> mondoo { [1,2,3].containsOnly([1,2]) }
mondoo: {
  [].containsOnly(): false
}
```

Darn, all the useful output just disappeared.

This PR make it so that instead we get:

```coffee
> mondoo { [1,2,3].containsOnly([1,2]) }
mondoo: {
  [failed] [].containsOnly()
    expected: == []
    actual:   [
      0: 3
    ]
}
```

This example is a bit constructed, but works just as well with naturally occurring blocks like:

**TODO**

Fixes https://github.com/mondoohq/cnquery/issues/5246

## How does it work?

This is where things get a bit tricky, because this PR is only a first step to solve the problem, but ultimately needs follow-ups to increase clarity.

**1: Expand Code Bundles**

By default, code bundles don't need to translate from CodeIDs into refs. This only ever becomes useful when we print things and deal with CodeIDs and try to go back to the compiled code (CodeBundle) and figure out how best to render it. We add a type that is only used for printing so we don't need to extend the CodeBundle for this task (no foreseeable future where we'd ever want to transmit this data, ie not part of proto).

**2: Start to building assessments inside the data printer**

Printing is currently split up in an assessment and a data-printing phase. If something is handled in the assessment phase it will be printed as an assessment and that's it. If not, it will be printed as data. 

When we receive an outer block, it switches into data mode. However, the block may contain assertions, that are then ignored and never collected nor printed. This PR changes this behavior and starts to look for assertions inside of blocks.

**3. Start to collect datapoints inside of blocks**

An old piece of code had previously removed this:

```
	// BUG (jaym): collectRefDatapoints prevents us from collecting datapoints.
	// Collecting datapoints for blocks didn't work correctly until 6.7.0.
	// See https://gitlab.com/mondoolabs/mondoo/-/merge_requests/2639
	// We can fix this after some time has passed. If we fix it too soon
	// people will start having their queries fail if a falsy datapoint
	// is collected.
```

Looks like we are far past that point, so let's start collecting.

### DIFF
--- a/apps/cnquery/cmd/plugin.go
+++ b/apps/cnquery/cmd/plugin.go
@@ -81,7 +81,7 @@ func (c *cnqueryPlugin) RunQuery(conf *run.RunQueryConfig, runtime *providers.Ru
 			return errors.Wrap(err, "failed to compile command")
 		}
 
-		out.WriteString(logger.PrettyJSON((b)) + "\n" + printer.DefaultPrinter.CodeBundle(b))
+		out.WriteString(logger.PrettyJSON((b)) + "\n" + printer.DefaultPrinter.CodeBundle(b.Expand()))
 		return nil
 	}
 
--- a/cli/printer/mql.go
+++ b/cli/printer/mql.go
@@ -17,7 +17,7 @@ import (
 	"golang.org/x/exp/slices"
 )
 
-func (print *Printer) Datas(bundle *llx.CodeBundle, results map[string]*llx.RawResult) string {
+func (print *Printer) Datas(bundle *llx.ExpandedCodeBundle, results map[string]*llx.RawResult) string {
 	var res strings.Builder
 	i := 0
 	for _, v := range results {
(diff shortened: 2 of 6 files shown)


arlimus:
Superseded by https://github.com/mondoohq/mql/pull/10631

PR #5267, opened 2025-02-26
✨ Set `asset.kind` to `virtualmachine` when in cloud

afiune:
This change consolidates `asset.kind` for `virtualmachine` and `baremetal`.

I am proposing having a single file where we define these asset kinds so that we
can later use them in the backend (follow up PR).

We are also changing the kind of an asset if:
* We are running an OS scan
* The scan is happening in a cloud instance

## How to test this change

* Build cnquery binary
* Check that the `asset.kind` comes back as `virtualmachine`

Alternatively, build the binary locally and connect to an `aws` instance via:
```
cnquery shell aws --discover instances
```
Select and asset and check that `asset.kind` comes back as `virtualmachine`


providers/os/id/clouddetect/clouddetect.go
```diff
@@ -29,16 +29,25 @@ var detectors = []detectorFunc{
 	gcp.Detect,
 }
 
+const AssetKind = "virtualmachine"
```
chris-rock: Lets find a better name for this

chris-rock: The shared kinds probably belong into the provider sdk

chris-rock: lets move it into platform.go

providers/os/id/clouddetect/clouddetect.go
```diff
@@ -29,16 +29,25 @@ var detectors = []detectorFunc{
...
+
 type detectResult struct {
 	platformId         string
 	platformName       string
 	relatedPlatformIds []string
 }
 
-func Detect(conn shared.Connection, p *inventory.Platform) (PlatformID, PlatformName, []RelatedPlatformID) {
+type PlatformInfo struct {
```
chris-rock, line 39: nit: add comments

providers/os/id/clouddetect/clouddetect.go
```diff
@@ -29,16 +29,25 @@ var detectors = []detectorFunc{
...
+	Name               string
+	Kind               string
+	RelatedPlatformIDs []string
+}
+
+func Detect(conn shared.Connection, p *inventory.Platform) PlatformInfo {
 	mgr, err := smbios.ResolveManager(conn, p)
 	if err != nil {
-		return "", "", nil
+		return PlatformInfo{"", "", "", nil}
```
chris-rock: Should this return nil? But we need to check the error handling on the other side.

providers/os/id/clouddetect/clouddetect.go
afiune: I can't put it there since we will have a cyclical dependency, but I think it would be a great idea to move these constants (for the near future) into the actual `inventory` package.

I think we will benefit from using them in the backend.

providers/aws/connection/platform.go
```diff
@@ -14,7 +16,7 @@ func GetPlatformForObject(platformName string, accountId string) *inventory.Plat
 		return &inventory.Platform{
 			Name:                  platformName,
 			Title:                 getTitleForPlatformName(platformName),
-			Kind:                  "aws-object",
+			Kind:                  getPlatformKind(platformName),
```
afiune: This needs to be reverted. We only want to label os scans.

PR #5284, opened 2025-03-03
✨ os cloud resource

afiune:
New `os` resource to expose cloud configuration for an asset running in the cloud.

~Depends on https://github.com/mondoohq/cnquery/pull/5267~ 🟢 merged!

## How to test this change

* Build the `os` provider (how? docs [here](https://github.com/mondoohq/cnquery/blob/main/docs/development.md#debug-providers))
* Build the `cnquery` binary
* Inspect the new `cloud` resource, try `cloud.provider` and `cloud.instance {*}`

Alternatively, build the provider and binary locally and connect to an `aws` instance via:
```
cnquery shell aws --discover instances
```
Select and asset and inspect the new `cloud` resource, try `cloud.provider` and `cloud.instance {*}`

## AWS
Example output of the new `cloud.instance` resource on an EC2 instance.

```
cnquery> cloud.instance: {
  publicIpv4: [
    0: ipv4Address ip=54.90.132.8
  ]
  privateIPv4: [
    0: ipv4Address ip=172.31.24.71
  ]
  publicHostname: "ec2-54-90-132-8.compute-1.amazonaws.com"
  privateHostname: "ip-172-31-24-71.ec2.internal"
  metadata: {
    ami-id: "ami-05b10e08d247fb927"
    ami-launch-index: 0.000000
    ami-manifest-path: "(unknown)"
    block-device-mapping: {
      ami: "xvda"
      root: "/dev/xvda"
    }
    events: {
      maintenance: {
        history: []
        scheduled: []
      }
    }
    hostname: "ip-172-31-24-71.ec2.internal"
    iam: {
      info: {
        Code: "Success"
        InstanceProfileArn: "arn:aws:iam::247094954579:instance-profile/SSM-ROLE"
        InstanceProfileId: "AIPATTCABZJJTBIG66FMP"
        LastUpdated: "2025-03-01T16:52:00Z"
      }
      security-credentials: "SSM-ROLE"
    }
    identity-credentials: {
      ec2: {
        info: {
          AccountId: "247094954579"
          Code: "Success"
          LastUpdated: "2025-03-01T16:51:29Z"
        }
        security-credentials: "ec2-instance"
      }
    }
    instance-action: "none"
    instance-id: "i-018ed722754cc4057"
    instance-life-cycle: "on-demand"
    instance-type: "t2.micro"
    local-hostname: "ip-172-31-24-71.ec2.internal"
    local-ipv4: "172.31.24.71"
    mac: "0a:ff:de:6b:e3:19"
    managed-ssh-keys: {
      active-keys: "ec2-user"
      signer-cert: "-----CERTIFICATE GOES HERE-----"
      signer-ocsp: {
        06B25927C42A721631C1EFD9431E648FA62E1E39: "qoI6ZRR"
        15E336D89683E51B6A134048088EBE2423EE0B4D: "bs7NKs9Sg"
        2AD974A775F73CBDBBD8F5AC3A49255FA8FB1F8C: "p2EoDx6qA=="
        9E99A48A9960B14926BB7F3B02E22DA2B0AB7280: "N4uhhaGlhQ=="
      }
    }
    metrics: "vhostmd"
    network: {
      interfaces: {
        macs: {
          0a:ff:de:6b:e3:19: {
            device-number: 0.000000
            interface-id: "eni-087a980ef8e9331bc"
            ipv4-associations: "54.90.132.8"
            local-hostname: "ip-172-31-24-71.ec2.internal"
            local-ipv4s: "172.31.24.71"
            mac: "0a:ff:de:6b:e3:19"
            owner-id: 247094954579.000000
            public-hostname: "ec2-54-90-132-8.compute-1.amazonaws.com"
            public-ipv4s: "54.90.132.8"
            security-group-ids: "sg-023de70d2bca17fbb"
            security-groups: "launch-wizard-5"
            subnet-id: "subnet-0d8e638520470e7de"
            subnet-ipv4-cidr-block: "172.31.16.0/20"
            vpc-id: "vpc-0f51ccdf5ed9d9816"
            vpc-ipv4-cidr-block: "172.31.0.0/16"
            vpc-ipv4-cidr-blocks: "172.31.0.0/16"
          }
        }
      }
    }
    placement: {
      availability-zone: "us-east-1c"
      availability-zone-id: "use1-az4"
      region: "us-east-1"
    }
    profile: "default-hvm"
    public-hostname: "ec2-54-90-132-8.compute-1.amazonaws.com"
    public-ipv4: "54.90.132.8"
    public-keys: "0=afiune"
    reservation-id: "r-077561058b9a4c23a"
    security-groups: "launch-wizard-5"
    services: {
      domain: "amazonaws.com"
      partition: "aws"
    }
    system: "xen-on-nitro"
  }
}
cnquery> 
```

The public and private list of IP addresses comes like this:
```
cnquery> cloud.instance.publicIpv4 {*}
cloud.instance.publicIpv4: [
  0: {
    ip: 54.161.92.232
    subnet: ""
    gateway: ""
    cidr: ""
    broadcast: ""
  }
]
cnquery> cloud.instance.privateIpv4 {*}
cloud.instance.privateIpv4: [
  0: {
    subnet: 172.31.16.0/20
    ip: 172.31.24.71
    broadcast: "172.31.31.255"
    cidr: 172.31.24.71/20
    gateway: ""
  }
]
```
Works on Windows too 💯 

Screenshot 2025-03-03 at 3 17 27 PM

## GCP

https://github.com/mondoohq/cnquery/pull/5286

## Azure
TBA


providers/os/resources/os.lr.manifest.yaml
```diff
@@ -355,6 +375,22 @@ resources:
...
     min_mondoo_version: 5.15.0
+  ipv4Address:
+    fields:
+      broadcast: {}
+      cidr: {}
+      gateway: {}
+      ip: {}
+      subnet: {}
+    min_mondoo_version: 9.0.0
+  ipv4Addresss:
```
tas50: This looks like a typo

providers/os/resources/os.lr.manifest.yaml
```diff
@@ -83,6 +83,26 @@ resources:
...
+    fields:
+      metadata: {}
+      privateHostname: {}
+      privateIPv4: {}
+      privateIpv4: {}
+      publicHostname: {}
+      publicIpv4: {}
+    is_private: true
+    min_mondoo_version: 9.0.0
+  cloud.metadata:
```
tas50: Is this meant to be here?

providers/os/resources/os.lr.manifest.yaml
afiune: interesting, I did make a typo once, but I fixed it, I guess the code generation doesn't remove this on its own? 🤔 

Good catch!

providers/os/resources/os.lr.manifest.yaml
afiune: same

providers/os/resources/os.lr.manifest.yaml
tas50: Yep they only add. never remove

tas50:
@afiune While testing this I'm hitting a lot of performance issues related to how many calls we need to do. If the user is just trying to get `cloud.instance.` they end up with over 1,000 curl requests when remotely running. It seems like we're always fetching the entire metadata even though we only need the IPs for the default fields there.

I ran:

```
cnquery shell aws --discover instances --verbose
```

then selected a single instance and ran:

```
cloud.instance
```

afiune:
@tas50 you are hitting different issues here:
* When you run that command, the `aws` provider discovers all instances, even though you just want to connect to a single asset
* The discovery for each instance is the place where we detect and fill all `asset` information, that's why we use now `clouddetect` which ultimately hits the instance metadata service
* We are NOT fetching the entire metadata, that only happens when you connect to the asset and run `cloud.XYZ`, what you see is a single call to the IMDS to fetch the instance identity

Remember that, `--discover instances` is meant to do a holistic discovery, I think there is a feature/flag missing here for the `cnquery shell` command to allow connecting to a single asset without doing the entire discovery.

tas50:
I'll send you over the output. It has some sensitive stuff in it

vjeffrey:
i built the os and aws providers, ran local `cnquery shell aws --discover instances` with `DEBUG=1`
i selected an ec2 instance
at that point, i should just be querying that ec2 instance

i ran `asset { * }`, which worked normally like always
then i ran `cloud.instance`
that was a couple minutes ago and it appears my shell is looping calling the apis over and over again

Screenshot 2025-03-06 at 19 42 15

it just loops forever for me rn -- did i miss something in the setup?

afiune:
@tas50 @vjeffrey Thank you for your feedback, there was a bug in the crawler that is not fixed. It takes ~30 seconds to fetch all the metadata information, I think that is reasonable but if you think it is not, we could make improvements to reduce the time.

tas50:
Retested and things are pretty performant at this point. I'm ok with merging this PR as-is, but let's set aside some time for further improvements in a different PR to see what can be done to optimize the calls when the default fields are used.

Thanks for digging in @afiune

vjeffrey:
@afiune can you open an issue to ensure we address the performance? im hesitant to merge this without knowing we're tracking that somewhere. it's slightly unreasonable that we could call all those apis when only requesting a single field like the ip address. that said, i know you want to get the other clouds done and then address this, so let's move forward

vjeffrey (approved):

tas50:
https://github.com/mondoohq/cnquery/issues/5295 5-10 second goal

PR #5288, opened 2025-03-05
🐛 warn users not to use standalone filters

arlimus:
The most common examples look like this:

```coffee
> packages.where("noname")

or

> packages.contains("something")
```

These will always be true, because they don't test the condition!

The correct use is:

```coffee
> packages.where(name == "noname")
```

This will be followed up by the linter (to catch this) and v12 which will make this a breaking change going forward.

### DIFF
--- a/mqlc/builtin_resource.go
+++ b/mqlc/builtin_resource.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"strconv"
 
+	"github.com/rs/zerolog/log"
 	"go.mondoo.com/cnquery/v11/llx"
 	"go.mondoo.com/cnquery/v11/mqlc/parser"
 	"go.mondoo.com/cnquery/v11/providers-sdk/v1/resources"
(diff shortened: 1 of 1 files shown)


mqlc/builtin_resource.go
```diff
@@ -153,6 +154,9 @@ func compileResourceWhere(c *compiler, typ types.Type, ref uint64, id string, ca
 	if err != nil {
 		return types.Nil, err
 	}
+	if refs.isStandalone {
+		log.Warn().Msg("Detected a standalone function in a where-clause! Please make sure you use filter the list, as this is most likely unintended")
```
misterpantz: "use filter the list" doesn't make sense to me. :) Just "filter the list" or "use a filter on the list"?

misterpantz:
```suggestion
		log.Warn().Msg("Detected a standalone function in a WHERE clause! This is probably unintended. Be sure to filter the list.")
```

tas50 (approved):

arlimus:
Thank you both for review and feedback!!

PR #5290, opened 2025-03-06
✨ integer-based IP addresses

arlimus:
As suggested by @frozen425 add support for integer-based IP addresses:

```coffee
> ip(2885681153)
ip: "172.0.0.1"
```

Currently limited to IPv4 only, I'm not sure yet what the approach would be for 128bit addresses (IPv6) with only 64-bit integers...

### DIFF
--- a/llx/builtin_ip.go
+++ b/llx/builtin_ip.go
@@ -59,6 +59,23 @@ func NewIP(s string) IP {
 
 var bitmasks = []byte{0x00, 0x80, 0xc0, 0xe0, 0xf0, 0xf8, 0xfc, 0xfe, 0xff}
 
+func int2ip(i int64) string {
+	cur := i
+
+	d := byte(cur % 256)
+	cur = cur >> 8
+
+	c := byte(cur % 256)
+	cur = cur >> 8
+
+	b := byte(cur % 256)
+	cur = cur >> 8
+
+	a := byte(cur % 256)
+
+	return net.IPv4(a, b, c, d).String()
+}
+
 func countMaskBits(b []byte) int {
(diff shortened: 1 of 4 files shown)


llx/builtin_ip.go
```diff
@@ -59,6 +59,23 @@ func NewIP(s string) IP {
 
 var bitmasks = []byte{0x00, 0x80, 0xc0, 0xe0, 0xf0, 0xf8, 0xfc, 0xfe, 0xff}
 
+func int2ip(i int64) string {
```
afiune, line 62: could we use `uint32` instead for this function? we only need 32 bits and, ipv4 addresses are naturally unsigned.

llx/builtin_ip.go
```diff
@@ -59,6 +59,23 @@ func NewIP(s string) IP {
...
+	cur := i
+
+	d := byte(cur % 256)
```
afiune: Interested to know why you used modulus (`% 256`) instead of bitwise and (`& 0xFF`)

afiune (approved):
One nitpick to save memory, but out of that, looks good!

llx/builtin_ip.go
arlimus: I have a follow-up that changes this in a few moments, goes into basically any int type

llx/builtin_ip.go
arlimus: it aligns a bit more with the descriptive approach of calculating the IP I found, but bitwise op would work just as well, I wonder if it's possibly faster in the compiler :thinking:

PR #5291, opened 2025-03-06
[Issue - 10441] Improved k8s cluster discovery

Chichichkin:
# 🚀 Overview
Addressing the issue 👉 [#10441](https://github.com/mondoohq/server/issues/10441)

## 🎯 User-facing change?

✅ The new way of building the platform IDs relies solely on the namespaces now.

## ⚠️ Risk Assessment: what can go wrong?

🟢 n/a

### DIFF
--- a/providers/k8s/connection/manifest/connection_test.go
+++ b/providers/k8s/connection/manifest/connection_test.go
@@ -97,7 +97,7 @@ func TestManifestDiscovery(t *testing.T) {
 	}
 	inv, err := resources.Discover(pluginRuntime, cnquery.Features{})
 	require.NoError(t, err)
-	require.Len(t, inv.Spec.Assets, 2)
+	require.Len(t, inv.Spec.Assets, 3)
 
 	conn.InventoryConfig().Discover.Targets = []string{"all"}
 	pluginRuntime = &plugin.Runtime{
@@ -108,7 +108,7 @@ func TestManifestDiscovery(t *testing.T) {
 	}
 	inv, err = resources.Discover(pluginRuntime, cnquery.Features{})
 	require.NoError(t, err)
-	require.Len(t, inv.Spec.Assets, 2)
+	require.Len(t, inv.Spec.Assets, 3)
 
 	conn.InventoryConfig().Discover.Targets = []string{"deployments"}
 	pluginRuntime = &plugin.Runtime{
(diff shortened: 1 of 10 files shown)


imilchev:
we need to hold off merging this PR. I believe this change conflicts with how we build platformids when scanning manifests. I need to take a look how to set this up such that it works for both manifests and clusters

imilchev:
This PR makes sure we have more stable platform IDs for k8s assets. With the previous implementation different platform IDs were generating for all workloads depending whether the scan would discover the cluster asset or not.

With this change the platform ids should always be the same:

When scanning a cluster:
- cluster asset has it's own platform ID, which is based on the `kube-system` namespace UID `//platformid.api.mondoo.app/runtime/k8s/uid/1bda2ae0-27fe-49ed-90fd-39d3bb8da955`
- workload IDs have platform IDs based on the parent namespace's UID `//platformid.api.mondoo.app/runtime/k8s/uid/ee1df06e-55c1-4ce3-b0a7-820efed0417a/namespace/default/pods/name/nginx`

When scanning manifests:
- manifest asset has it's own platform ID based on the path hash for the manifest/directory `//platformid.api.mondoo.app/runtime/k8s/uid/ed37af0ce97030c84a07fb9d3311841f5f180a77573c664da6fbe3d8dc481896`
- workload IDs have platform IDs based on the manifest/directory path `//platformid.api.mondoo.app/runtime/k8s/uid/ed37af0ce97030c84a07fb9d3311841f5f180a77573c664da6fbe3d8dc481896/namespace/mondoo-operator/deployments/name/mondoo-operator-controller-manager`

providers/k8s/connection/manifest/connection_test.go
```diff
@@ -97,7 +97,7 @@ func TestManifestDiscovery(t *testing.T) {
 	}
 	inv, err := resources.Discover(pluginRuntime, cnquery.Features{})
 	require.NoError(t, err)
-	require.Len(t, inv.Spec.Assets, 2)
+	require.Len(t, inv.Spec.Assets, 3)
```
czunker, line 100: According to the comments in this PR, the manifest part didn't change. Why did this increase?

czunker (commented):
I can currently only think of ~~two~~ one case~~s~~ where this might not work:
- The user has no privileges to access kube-system. Are we still able to get the uuid from the namespace list?

Did I read this correctly, that we do not discover the cluster when a namespace filter is set?

providers/k8s/connection/manifest/connection_test.go
imilchev: we discover the namespaces now. Before we skipped over them which was actually a bug. We turned on namespace discovery for api scans, but not for manifests

Chichichkin:

PR #5292, opened 2025-03-07
✨ ip.cidr, ip.address and raw IP storage

arlimus:
Adds the `address` field, which only returns the address itself:

```coffee
# obscure IPv4, printed nicely
> ip("192.168.0.0/24").address
"192.168.0.0"

> ip("2001:db8:3c4d:15::1a2f:1a2b/48").address
"2001:db8:3c4d:15::1a2f:1a2b"
```

Adds the `cidr` field, which returns the [CIDR](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing) notation of the IP address. We fill in the prefix length for IPv4 class A/B/C networks and IPv6 automatically:

```coffee
# obscure IPv4, printed nicely
> ip(2885681153).cidr
"172.0.0.1/16"

> ip("2001:db8:3c4d:15::1a2f:1a2b").cidr
"2001:db8:3c4d:15::1a2f:1a2b/64"
```

**internal:**

The bigger thing this PR does is overhaul the internal data structure for IP so we don't have to constantly parse it again. This increases efficiency but also means we have to take care of serialization/deserialization.

- `Mask` was renamed to `PrefixLength` because that's what it actually is
- `NewIP` is really a `ParseIP` which also separates it from `ParseIntIP` (which is demonstrated above)

Why proto-based marshaling?
Not an easy decision, but one that is more maintainable while the interface still shakes out. A byte-optimized marshaling approach will be added later as an alternative.


llx/llx.proto
```diff
@@ -155,3 +155,9 @@ message Assessment {
   bool is_assertion = 3;
   repeated AssessmentItem results = 4;
 }
+
+message IP {
+  bytes address = 1;
+  bool has_prefix = 2;
+  int32 prefix_length = 3;
+}
```
afiune, line 163: Nice!

llx/builtin_ip.go
```diff
@@ -49,17 +62,10 @@ func NewIP(s string) IP {
...
-		Version: version,
-		Mask:    mask,
-	}
+	return version, mask
 }
 
-var bitmasks = []byte{0x00, 0x80, 0xc0, 0xe0, 0xf0, 0xf8, 0xfc, 0xfe, 0xff}
-
-func int2ip(i int64) string {
+func int2ip[T int | int64 | int32 | uint | uint64 | uint32](i T) net.IP {
```
afiune: beautiful ✨

llx/builtin_ip.go
```diff
@@ -264,82 +280,177 @@ func (i IP) Cmp(other IP) int {
...
+	return StringData(v.CIDR()), 0, nil
+}
+
+func any2ip(v any) (RawIP, bool) {
+	switch t := v.(type) {
+	case RawIP:
+		return t, true
+	case string:
+		return ParseIP(t), true
+	case int64:
```
afiune, line 436: are we missing some types here? like `int` and others? 🤔

afiune: I'm guessing no, but if so, we won't ever use the other types, I think.

arlimus: Not at this position, because the internal values can only be these 3, ie for int it can only be an int64. Much of the backend is built around this.

llx/builtin_ip.go
afiune: Just a question since I see in MQL we use `int64` everywhere. It looks awesome but I'm curious if we are even using all of those types? 🤔

afiune (approved):
Pretty nice, just a question about the `int2ip()` func.

llx/builtin_ip.go
arlimus: yeah tbh I just went over-eager here :grin: If I'm honest, we only need int and int64, just like the other method

Issue #5261, opened 2025-02-24
Ignore packages with emtpy name on Windows


czunker:
We see a lot of errors, like this one:
```
  "jsonPayload": {
    "level": "error",
    "error": "purl is missing name",
    "purl": "pkg:windows/?arch=x86_64",
    "message": "could not parse purl"
  },
```

This looks like a Windows package without a name.
We should figure out whether the package should have a name or otherwise ignore it.

czunker:
![Image](https://github.com/user-attachments/assets/9654d8fc-0228-43ca-a1ff-2fabc1fbbfed)

PR #5294, opened 2025-03-07
Ignore packages with empty names

claytoncasey01:
If the package has an empty name skip it. It can cause errors downstream later due to creating invalid PUrls that can't be parsed.

Fixes: #5261.

### DIFF
--- a/providers/os/resources/packages/windows_packages.go
+++ b/providers/os/resources/packages/windows_packages.go
@@ -452,6 +452,10 @@ func getPackageFromRegistryKeyItems(children []registry.RegistryKeyItem, platfor
 		return nil
 	}
 
+	if displayName == "" {
+		return nil
+	}
+
 	pkg := &Package{
 		Name:    displayName,
 		Version: displayVersion,
@@ -471,6 +475,7 @@ func getPackageFromRegistryKeyItems(children []registry.RegistryKeyItem, platfor
 		}
 	} else {
 		log.Debug().Msg("ignored package since information is missing")
+		return nil
 	}
 	return pkg
 }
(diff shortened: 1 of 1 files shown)


providers/os/resources/packages/windows_packages.go
```diff
@@ -452,6 +452,12 @@ func getPackageFromRegistryKeyItems(children []registry.RegistryKeyItem, platfor
 		return nil
 	}
 
+	if displayName == "" || displayVersion == "" {
```
chris-rock: why do we have empty packages in the first place?

claytoncasey01: From a bit of quick research it seems to be somewhat common (for windows): partial installs, bad uninstalls, or even some software just not putting a DisplayName in the registry during the setup/install process. 

I can debug the registry code where we pull display value and then actually check the path that's failing to confirm this.

chris-rock: I just want to find out if we need to fix something in cnquery. We should mark this as TODO, since generally we should not have empty package names in the first place.

chris-rock (approved):

providers/os/resources/packages/windows_packages.go
```diff
@@ -129,7 +129,15 @@ func (p winAppxPackages) toPackage(platform *inventory.Platform) Package {
...
 		Format:  "windows/appx",
 		Vendor:  p.Publisher,
-		PUrl:    purl.NewPackageURL(platform, purl.TypeAppx, p.Name, p.Version).String(),
+	}
+
+	// TODO: We need to figure out why we have empty displayNames.
+	// this is common in windows but we need to verify it is a windows
+	// issue and not a cnquery issue.
+	if p.Name != "" {
+		pkg.PUrl = purl.NewPackageURL(platform, purl.TypeAppx, p.Name, p.Version).String()
```
czunker: The package will remain on the list, only without a PURL.

The other part of the PR skips the package completely. I think we should do the same with this.

czunker (approved):
LGTM
Thanks  @claytoncasey01

PR #5304, opened 2025-03-10
🐛 log error when getting latest cnquery version for status cli cmd

kkereziev:
Desirable outcome:
- Log warning message that fetching the latest version failed, instead of throwing an error.

Follow up actions:
- Open PR with the new patch version in cnspec.

### DIFF
--- a/apps/cnquery/cmd/status.go
+++ b/apps/cnquery/cmd/status.go
@@ -105,7 +105,7 @@ func checkStatus() (Status, error) {
 
 	latestVersion, err := cnquery.GetLatestVersion(httpClient)
 	if err != nil {
-		return s, cli_errors.NewCommandError(errors.Wrap(err, "failed to get latest version"), 1)
+		log.Warn().Err(err).Msg("Failed to get latest version")
 	}
 
 	s.Client.LatestVersion = latestVersion


kkereziev:

kkereziev:
recheck

apps/cnquery/cmd/status.go
```diff
@@ -105,7 +105,7 @@ func checkStatus() (Status, error) {
 
 	latestVersion, err := cnquery.GetLatestVersion(httpClient)
 	if err != nil {
-		return s, cli_errors.NewCommandError(errors.Wrap(err, "failed to get latest version"), 1)
+		log.Warn().Err(err).Msg("Failed to get latest cnquery version")
```
imilchev: I think the same code is re-used in `cnspec` so if your message says `Failed to get latest cnquery version`, it would show the same thing when running this from `cnspec`. I think you should keep the original message so it's applicable for both cnquery and cnspec.

apps/cnquery/cmd/status.go
```diff
@@ -105,7 +105,7 @@ func checkStatus() (Status, error) {
...
 	}
 
 	s.Client.LatestVersion = latestVersion
```
imilchev, line 111: this is rendered in the CLI and it will show as empty string in case we had an error retrieving the latest version. Maybe in case of an error, you should assign `UNKNOWN` or just print the same message here.

kkereziev: There is a guard against empty string - https://github.com/mondoohq/cnquery/blob/main/apps/cnquery/cmd/status.go#L183

PR #5305, opened 2025-03-10
⭐️ os cloud resource for azure

afiune:
Part 3 of https://github.com/mondoohq/cnquery/pull/5284

This PR enables the new os resource `cloud` in Azure VMs.

## How to test this change

* Build the `os` provider (how? docs [here](https://github.com/mondoohq/cnquery/blob/main/docs/development.md#debug-providers))
* Build the `cnquery` binary
* Inspect the new `cloud` resource, try `cloud.provider` and `cloud.instance {*}`


cli/printer/mql.go
```diff
@@ -787,6 +787,9 @@ func (print *Printer) Data(typ types.Type, data interface{}, codeID string, bund
 		return print.Secondary(data.(string))
 
 	case types.IP:
+		if data == nil {
+			return print.Secondary("null")
+		}
```
afiune, line 792: This is fixed separately here https://github.com/mondoohq/cnquery/pull/5307

providers/os/id/azcompute/azcompute.go
```diff
@@ -16,8 +16,11 @@ import (
 )
 
 const (
-	identityUrl                   = "http://169.254.169.254/metadata/instance?api-version=2021-02-01"
+	instanceMetadataScriptUnix    = "curl --retry 5 --retry-delay 1 --connect-timeout 1 --retry-max-time 5 --max-time 10 --noproxy '*' -H Metadata:true http://169.254.169.254/metadata/instance?api-version=2021-02-01"
 	metadataIdentityScriptWindows = `Invoke-RestMethod -TimeoutSec 5 -Headers @{"Metadata"="true"} -Method GET -URI http://169.254.169.254/metadata/instance?api-version=2021-02-01 -UseBasicParsing | ConvertTo-Json`
+
+	loadbalancerMetadataScriptUnix    = "curl --retry 5 --retry-delay 1 --connect-timeout 1 --retry-max-time 5 --max-time 10 --noproxy '*' -H Metadata:true http://169.254.169.254/metadata/loadbalancer?api-version=2021-02-01"
```
preslavgerchev: can we use a newer api-version? 
https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=windows#supported-api-versions

afiune: Updated it ✅

providers/os/resources/cloud/instance_metadata.go
```diff
@@ -52,3 +57,82 @@ func (m InstanceMetadata) PrivateIP() string {
...
+	m.PublicIpv4[index] = merged
+}
+func (m *InstanceMetadata) findPublicIP(ip string) int {
+	return slices.IndexFunc(m.PublicIpv4, func(address Ipv4Address) bool {
+		return address.IP == ip
+	})
+}
+
+// AddOrUpdatePrivateIP adds or updates one or many Ipv4Addresses
+func (m *InstanceMetadata) AddOrUpdatePrivateIP(ips ...Ipv4Address) {
```
preslavgerchev, line 92: whats the use case in which we get the same ip twice so it needs to be updated? i'd expect the metadata to give us the IP addresses just once

afiune: After doing extensive research, I discovered that the public IP address may come from two different endpoints within the IMDS, either the `instance/` metadata or the `loadbalander/` one.

Docs:
https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=windows#endpoint-categories

So we don't know where will it come from, but we check both places looking. The question is, what do we do if it is present in both places? I say we merge the information in the best possible way, which is this implementation, but I am open to suggestions.

preslavgerchev: if the instance is behind a LB, will the instance endpoint still give us back the IP info?
Maybe it's best to just define a precedence order here and use whichever one has higher prio?

This code is also fine, I just want to make sure there isnt an easier way to handle this case

afiune: The precedence comes from who calls first this function, we prioritize the information from the load balancer, if it exits. Then look at the instance metadata:

https://github.com/mondoohq/cnquery/blob/8b43a91b47fdcff961398a27d76deedb6524563b/providers/os/resources/cloud/azure.go#L53-L115

I can move the logic from here to there 👆🏽  if that makes things easier.

preslavgerchev (approved):

PR #5309, opened 2025-03-11
🐛 omit storing `force` flag in config

afiune:
Additionally, we have a way to add more fields we want to omit from storing
in the config.

### DIFF
--- a/apps/cnquery/cmd/login.go
+++ b/apps/cnquery/cmd/login.go
@@ -200,9 +200,6 @@ func register(token string, annotations map[string]string, timer int, splay int,
 		}
 
 		log.Debug().Msg("store configuration")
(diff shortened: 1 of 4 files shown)


apps/cnquery/cmd/login.go
```diff
@@ -200,9 +200,6 @@ func register(token string, annotations map[string]string, timer int, splay int,
 		}
 
 		log.Debug().Msg("store configuration")
-		// overwrite force, otherwise it will be stored
-		viper.Set("force", false)
```
afiune, line 204: This never worked. And it was trying to unset a config setting in a single place. Moved it to the storing of the config instead.

vjeffrey (approved):

PR #5314, opened 2025-03-13
⭐️ cloud detect + metadata for VMware

afiune:


providers/os/id/vmware/vmtoolsd/vmtoolsd.go
```diff
@@ -0,0 +1,170 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	return strings.TrimSpace(string(data)), nil
+}
+
+// vmtoolsdGuestInfo runs vmtoolsd to fetch guest info with the provided key
+func (m *CommandInstanceMetadata) vmtoolsdGuestInfo(key string) (string, error) {
+	return m.RunCommand(fmt.Sprintf("%s --cmd \"info-get guestinfo.%s", m.vmtoolsd(), key))
```
tas50:
```suggestion
	return m.RunCommand(fmt.Sprintf("%s --cmd \"info-get guestinfo.%s\"", m.vmtoolsd(), key))
```

tas50: This fixed my local failures

providers/os/id/vmware/vmtoolsd/vmtoolsd.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func (m *CommandInstanceMetadata) Identify() (Identity, error) {
+	identt := Identity{}
+
+	hostname, err := m.Hostname()
+	if err != nil {
+		return identt, err
+	}
+	identt.Hostname = "//platformid.api.mondoo.app/runtime/vmware/instance/" + hostname
```
jaym: why is the hostname a platform id

jaym:
will this change the platform ids for all existing assets causing them to show up as duplicates?

afiune:
@jaym this is ready for review again, thanks for pointing out the ID problem, now we are using the `UUID` as the identifier and we are NOT using the `hostname` at all for platform ids since we will detect that via the "hostname detector".

There is a place where we can get related platform IDs for the vSphere and vCenter, I added those since they are related to where this VM is running on.

PR #5324, opened 2025-03-17
🐛 Fix missing package arch in windows and macos

claytoncasey01:
- Fixes an issue with missing arch in packages for windows and macos


czunker:
Works on Windows:
image

czunker (approved):
Thanks @claytoncasey01

PR #5331, opened 2025-03-18
✨ Authenticate against AKS cluster by mimicking kubelogin managed identity credential token flow.

VasilSirakov:
#### In this PR:
Added a `--kubelogin` boolean flag to the Kubernetes provider which gives the option to authenticate against a remote AKS cluster using a JWT bearer token. This flow mimics the flow `kubelogin` ([docs](https://learn.microsoft.com/en-us/azure/aks/kubelogin-authentication)) uses without needing the `kubelogin` CLI to be present locally.

Azure provides us a kubeconfig file with the following exec:
```bash
kubelogin get-token --environment AzurePublicCloud --server-id {value} --client-id {value} --tenant-id {value} --login devicecode
```
this would then get modified to:
```bash
 kubelogin get-token --login msi --server-id {value}
 ```
which is the flow implemented in this PR (uses the inferred managed identity).

Running cnquery with:
```bash
cnquery shell k8s --kubelogin
```
yields
image

Also compiled this into cnspec and ran it against a policy:
image

 #### Notes:
We initially wanted to pull in the entire [kubelogin](https://github.com/Azure/kubelogin) as a dependency but decided to hold off for now due to #2883 to be on the safe side and only used the relevant code from `azidentity`. We can always come back to this at a later point and include it since we're talking about 4-5 lines of code here.

### DIFF
--- a/providers/k8s/config/config.go
+++ b/providers/k8s/config/config.go
@@ -77,7 +77,7 @@ Examples:
 					Long:    "kubelogin",
 					Type:    plugin.FlagType_Bool,
 					Default: "false",
-					Desc:    "Authenticate against a remote Azure AD enabled Kubernetes cluster using a default user or system-assigned Azure identity.",
+					Desc:    "Authenticate against a remote Azure AD enabled Kubernetes cluster using an Azure identity.",
 				},
 			},
 		},
--- a/providers/k8s/config/config.go
+++ b/providers/k8s/config/config.go
@@ -73,6 +73,12 @@ Examples:
 					Default: "",
 					Desc:    "HTTP proxy to use for container pulls",
 				},
+				{
+					Long:    "kubelogin",
+					Type:    plugin.FlagType_Bool,
+					Default: "false",
+					Desc:    "Use kubelogin auth flow to authenticate against an AKS cluster using a managed identity credential.",
+				},
 			},
 		},
 	},
--- a/providers/k8s/config/config.go
+++ b/providers/k8s/config/config.go
@@ -77,7 +77,7 @@ Examples:
 					Long:    "kubelogin",
 					Type:    plugin.FlagType_Bool,
 					Default: "false",
-					Desc:    "Use kubelogin auth flow to authenticate against an AKS cluster using a managed identity credential.",
+					Desc:    "Authenticate against a remote Azure AD enabled Kubernetes cluster using a default user or system-assigned Azure identity.",
 				},
 			},
 		},
--- a/providers/k8s/connection/api/auth.go
+++ b/providers/k8s/connection/api/auth.go
@@ -1,0 +1,70 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package api
+
+import (
+	"context"
+	"strconv"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/rs/zerolog/log"
+	"go.mondoo.com/cnquery/v11/providers-sdk/v1/inventory"
+	"go.mondoo.com/cnquery/v11/providers-sdk/v1/util/azauth"
+	"go.mondoo.com/cnquery/v11/providers/k8s/connection/shared"
+	"k8s.io/client-go/rest"
+)
+
+const (
+	// AKS Microsoft Entra server application ID used by AKS for user authentication. The bearer token needs to be issued
+	// for this application (aud claim in JWT).
+	// https://learn.microsoft.com/en-us/azure/aks/kubelogin-authentication#how-to-use-kubelogin-with-aks
+	serverAppId  = "6dae42f8-4368-4678-94ff-3960e28e3630"
+	defaultScope = "/.default"
+)
+
+// attempt to get a bearer token using the kubelogin flow and attach it to the rest config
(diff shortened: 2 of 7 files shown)


providers/k8s/connection/api/auth.go
```diff
@@ -0,0 +1,71 @@
...
+	"go.mondoo.com/cnquery/v11/providers/k8s/connection/shared"
+	"k8s.io/client-go/rest"
+)
+
+const (
+	// AKS Microsoft Entra server application ID used by AKS for user authentication. The bearer token needs to be issued
+	// for this application (aud claim in JWT).
+	// https://learn.microsoft.com/en-us/azure/aks/kubelogin-authentication#how-to-use-kubelogin-with-aks
+	serverAppId  = "6dae42f8-4368-4678-94ff-3960e28e3630"
+	defaultScope = "/.default"
```
czunker: Is this correct?
What I found in the docs, I'm not sure what the correct value should be: https://learn.microsoft.com/en-us/entra/identity-platform/scopes-oidc#the-default-scope

providers/k8s/provider/provider.go
```diff
@@ -77,6 +78,10 @@ func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error)
 		conf.Options[shared.OPTION_NAMESPACE_EXCLUDE] = string(ns.Value)
 	}
 
+	if val, ok := req.Flags[shared.OPTION_KUBELOGIN]; ok {
+		conf.Options[shared.OPTION_KUBELOGIN] = strconv.FormatBool(val.RawData().Value.(bool))
```
czunker, line 82: You first cast it to a bool and then convert it back to a string.
Isn't there an easier way to achieve this?

providers/k8s/connection/api/auth.go
```diff
@@ -0,0 +1,71 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	if !kubeloginAuth {
+		return nil
+	}
+
+	log.Debug().Msg("attempting to get a bearer token using the kubelogin flow")
```
czunker: As this is the general k8s provider, we might want to check whether we try to connect against AKS. Can we figure this out?

Or from another perspective:
How would the errors look like, when someone tries to connect to GKE with this auth flow. Would the user recognize from the messages, that this is only applicable to AKS?
I saw the help text, but I'd assume some users do not recognize the AKS part.

providers/k8s/connection/api/auth.go
VasilSirakov: The `.default` scope is used by kubelogin for this flow [here](https://github.com/Azure/kubelogin/blob/main/pkg/internal/token/execCredentialPlugin.go#L92).

providers/k8s/provider/provider.go
VasilSirakov: Technically, this bool to string conversion here is not necessary since we can just not assign this option when the flag is not specified (i.e. `false`) and the behavior would bo equivalent.

However, for the inventory case I wanted to have this `kubelogin: true|false` format and in order to unify the code paths, I also did the conversion here. If you prefer, I can remove this particular conversion.

czunker: Sounds legit. Please add a comment so that this is not removed during another code change.

providers/k8s/config/config.go
```diff
@@ -73,6 +73,12 @@ Examples:
 					Default: "",
 					Desc:    "HTTP proxy to use for container pulls",
 				},
+				{
+					Long:    "kubelogin",
+					Type:    plugin.FlagType_Bool,
+					Default: "false",
+					Desc:    "Use kubelogin auth flow to authenticate against an AKS cluster using a managed identity credential.",
```
preslavgerchev: nit: does it have to be managed identity? i presume i can also use this with my (cli) creds

providers/k8s/connection/api/auth.go
VasilSirakov: You're right, the current help text can be misinterpreted and we should make the documentation clearer. This auth flow should never be used for any other scenario and any other cloud.

I'll take a look at the errors, I suppose wrapping them with a some additional context wouldn't hurt.

providers/k8s/config/config.go
VasilSirakov: So, this replicates the MSI (Managed Service Identity) flow [here](https://azure.github.io/kubelogin/concepts/login-modes/msi.html). Internally it uses a managed identity token credential which means it has to be a system-assigned or user-assigned managed identity.

Currently it infers a default identity to authenticate with, however, in a scenario where there are multiple identities, it does not know how to choose between them and errors. We can consider adding an extra flag/opt for disambiguating it with a specific resource identifier if necessary or we could do that in a follow-up.

I also tested this now from my local CLI and it authenticated successfully but I presume that's because my account has the `Owner` role over the entire subscription despite not being a managed identity.

providers/k8s/provider/provider.go
VasilSirakov: Actually, now I saw that this is common throughout cnquery for such boolean flags. I think there's not much value in adding a comment in this case.

providers/k8s/config/config.go
preslavgerchev: i'd imagine this works with just any AZ identity as long as you have the right permissions, be that managed or a local (CLI) one as token fetching is behind an interface

VasilSirakov:
@preslavgerchev @czunker feedback has been processed or answered. you can take another look.

czunker (approved):
LGTM
Thanks @VasilSirakov

providers/k8s/connection/api/auth.go
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package api
```
preslavgerchev: nit: can we rename this file to `aks_auth.go`

PR #5332, opened 2025-03-18
🐛 lsblk: empty mountpoints

slntopp:
Depending on a version of lsblk json output might differ slightly
Has `MOUNTPOINT` only: https://man7.org/linux/man-pages/man8/lsblk.8.html
Has both `MOUNTPOINT` and `MOUNTPOINTS`  https://man.archlinux.org/man/lsblk.8.en#:~:text=The%20relationship%20between,with%20the%20device.

Bonus: handle `null` entries

### DIFF
--- a/providers/os/resources/lsblk.go
+++ b/providers/os/resources/lsblk.go
@@ -6,6 +6,7 @@ package resources
 import (
 	"encoding/json"
 	"errors"
+	"slices"
 
 	"go.mondoo.com/cnquery/v11/llx"
 	"go.mondoo.com/cnquery/v11/types"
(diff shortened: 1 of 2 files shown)


providers/os/resources/lsblk.go
```diff
@@ -34,7 +35,19 @@ func (l *mqlLsblk) list() ([]interface{}, error) {
 		d := blockEntries.Blockdevices[i]
 		for i := range d.Children {
 			entry := d.Children[i]
-			entry.Mountpoints = append(entry.Mountpoints, entry.Mountpoint)
+			// Some versions of the lsblk return [null] instead of empty array
+			entry.Mountpoints = slices.Collect(func(yield func(interface{}) bool) {
```
afiune: Should this happen inside the parsing of the mount block entries?

afiune (approved):
Left a question. If we do move the code, we can add test data and include tests.

Code looks good.

providers/os/resources/lsblk.go
slntopp: Good point, will update

PR #5335, opened 2025-03-19
✨ parse scsi devices form `/sys/`

slntopp:


providers/os/connection/device/linux/lun_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestParseLsscsiOutput(t *testing.T) {
+func TestParseLsscsiCLIOutput(t *testing.T) {
```
preslavgerchev, line 12: can we remove this + the `parseLsscsiOutput` func?

preslavgerchev: ah, i see its used as a fallback

preslavgerchev (approved):

providers/os/connection/device/linux/lun.go
```diff
@@ -22,6 +25,49 @@ type scsiDeviceInfo struct {
...
+		}
+
+		entry, err := os.Readlink(path.Join("/sys/block/", block.Name()))
+		if err != nil {
+			log.Warn().Err(err).Str("block", block.Name()).Msg("failed to readlink")
+			continue
+		}
+		entry, err = filepath.Abs(path.Join("/sys/block/", entry))
+		if err != nil {
+			log.Warn().Err(err).Str("block", block.Name()).Msg("failed to get absolute path")
```
afiune, line 55: do we need these logs to be warnings?

afiune (commented):
I see a number of warning level logs, those are displayed to users and might be too noisy.

providers/os/connection/device/linux/lun.go
slntopp: Something has to be very broken for these to fail and produce a log, it'd be very helpful to have that log.

PR #5336, opened 2025-03-19
🐛 fetching VPC endpoint services

afiune:
This change uses the API `DescribeVpcEndpoints()` to fetch information about endpoint services and delays loading of additional data for when it is accessed.

Fixes https://github.com/mondoohq/cnquery/issues/5318


vjeffrey (approved):
thank you!

PR #5344, opened 2025-03-21
✨ use generics for pointer convertion

afiune:

### DIFF
--- a/providers/aws/resources/aws_codebuild.go
+++ b/providers/aws/resources/aws_codebuild.go
@@ -126,8 +126,8 @@ func initAwsCodebuildProject(runtime *plugin.Runtime, args map[string]*llx.RawDa
 	if err != nil {
 		return nil, nil, err
 	}
-	args["arn"] = llx.StringData(convert.ToString(project.Arn))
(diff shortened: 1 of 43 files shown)


providers/aws/resources/aws_codebuild.go
```diff
@@ -126,8 +126,8 @@ func initAwsCodebuildProject(runtime *plugin.Runtime, args map[string]*llx.RawDa
 	if err != nil {
 		return nil, nil, err
 	}
-	args["arn"] = llx.StringData(convert.ToString(project.Arn))
-	args["description"] = llx.StringData(convert.ToString(project.Description))
+	args["arn"] = llx.StringDataPtr(project.Arn)
+	args["description"] = llx.StringDataPtr(project.Description)
```
afiune, line 130: There are cases like this one where, we could use `StringDataPtr()` instead of `StringData()` but, I do not know the implications of that change since one SETs the field to an empty string and the other one doesn't set it, but returns a `Nil` data type.

I am on board to leave these as is and not use `StringDataPtr()` but I am open to suggestions.

jaym (approved):

vjeffrey (approved):

PR #5370, opened 2025-03-25
🧹 lift type restrictions for version resource

chris-rock:
For the version resource we have been very restrictive. Checking for the type string prevented us from using Ref and Dict for `version`.


arlimus (approved):
Thank you for fixing my oversight!!

PR #5375, opened 2025-03-26
🧹 Remove dependabot go mod bumping

tas50:
We have a single PR generation for all the providers vs. individual PRs that break things


.github/dependabot.yml
```diff
@@ -1,286 +0,0 @@
----
-version: 2
-updates:
-  - package-ecosystem: github-actions
```
chris-rock: Lets keep the github actions. I think that would be useful.

tas50: I updated it to just remove the got bits

PR #5377, opened 2025-03-27
🐛 os: pin `cloud.instance` to version `>=11.45.0`

afiune:


chris-rock:
Why is that required? 9.x was the introduction of the plugin mechanism.

jaym:
Before we introduced plugins, that field told clients if they were allowed to try to get a field from a resource. I do not know if it still even works

jaym (approved):

afiune:
@jaym @chris-rock and I talked, this should help, but we will see.

PR #5390, opened 2025-03-28
⭐️ os network interfaces

afiune:
This change introduces a new resource to the `os` provider called `network`, this
resource today gathers all the network interfaces from the system and additional
information about the interface like ip addresses (v4 and v6), subnet, gateway, the
status and flags of the interface, as well as the MAC address.

# Data Collection Methods

The new `networki` package has an entry point function called `Interfaces()` that
returns the list of interfaces in the system. The way this function work is by detecting
the operating system we are connected to, and based of that information, it runs multiple
detector functions to collect as much information from different sources:
```go
detectors := []func() ([]Interface, error){
	...
}
```

In the future, we can add more of these to enhance the data we collect.

# Testing

Ran tests agains all three operating systems we need to support.

## `macOS` 🍎 

Screenshot 2025-03-28 at 1 50 36 PM

## `Windows` 🪟 

Screenshot 2025-04-01 at 3 54 59 PM

### `amazonlinux`

Screenshot 2025-04-01 at 4 11 07 PM

### `Debian`

Screenshot 2025-04-01 at 4 11 48 PM

### DIFF
--- a/providers/os/id/networki/linux_n.go
+++ b/providers/os/id/networki/linux_n.go
@@ -19,16 +19,34 @@ import (
 
 // detectLinuxInterfaces detects network interfaces on Linux.
 func (n *neti) detectLinuxInterfaces() ([]Interface, error) {
+	var errs []error
+	interfaces := []Interface{}
+
+	// List of detectors that collect network interfaces, we stop executing them as
+	// soon as one of them succeeds collecting all the information
 	detectors := []func() ([]Interface, error){
 		n.getLinuxCmdInterfaces,
 		n.getLinuxSysfsInterfaces,
+	}
+	for _, detectFn := range detectors {
+		detectedInterfaces, err := detectFn()
+		if err == nil && len(detectedInterfaces) != 0 {
+			interfaces = AddOrUpdateInterfaces(interfaces, detectedInterfaces)
+			break
+		}
+		log.Debug().Err(err).
+			Msg("os.network.interface> unable to detect network interfaces")
+		errs = append(errs, err)
+	}
+
+	// List of enrichment functions that collect additional information that we
+	// couldn't gather in the detectors.
+	enrichments := []func() ([]Interface, error){
 		n.getLinuxIPv4GatewayDetails,
 		n.getLinuxIPv6GatewayDetails,
 	}
 
-	var errs []error
-	interfaces := []Interface{}
-	for _, detectFn := range detectors {
+	for _, detectFn := range enrichments {
 		detectedInterfaces, err := detectFn()
 		if err != nil {
 			log.Debug().Err(err).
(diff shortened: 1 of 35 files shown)


atomic111:
@afiune thank you

providers/os/id/networki/linux_n.go
```diff
@@ -0,0 +1,280 @@
...
+			if matches := macRegex.FindStringSubmatch(line); matches != nil {
+				// Match MAC address
+				currentInterface.SetMAC(matches[1])
+			} else if matches := ipRegex.FindStringSubmatch(line); matches != nil {
+				// Match IPv4 address
+				ip := NewIPv4WithPrefixLength(
+					matches[1],
+					parseInt(matches[2]),
+				)
+				// if ip.Broadcast != matches[3] {
```
jaym: should this be removed?

providers/os/id/networki/windows_n.go
```diff
@@ -0,0 +1,242 @@
...
+	"strings"
+
+	"github.com/cockroachdb/errors"
+	"github.com/rs/zerolog/log"
+)
+
+// detectWindowsInterfaces detects network interfaces on Windows.
+func (n *neti) detectWindowsInterfaces() ([]Interface, error) {
+	detectors := []func() ([]Interface, error){
+		n.getWindowsIpconfigCmdInterfaces,
```
jaym: do we actually need both of these? is something missing from get-netip

providers/os/id/networki/linux_n.go
```diff
@@ -0,0 +1,280 @@
...
+
+	"github.com/cockroachdb/errors"
+	"github.com/rs/zerolog/log"
+	"github.com/spf13/afero"
+)
+
+// detectLinuxInterfaces detects network interfaces on Linux.
+func (n *neti) detectLinuxInterfaces() ([]Interface, error) {
+	detectors := []func() ([]Interface, error){
+		n.getLinuxCmdInterfaces,
```
jaym, line 28: do we need both ip add and sysfs?

providers/os/id/networki/linux_n.go
afiune: Will delete. 💯

providers/os/id/networki/windows_n.go
afiune: For some reason, I wouldn't get all the information from one place. 

* get-netip was missing the gateway
* ipconfig was missing the MTU and virtualization

I could look at it again and see if there is another command I can use to get the gateway on the get-netip detector and, avoid running the the next one if one of the succeeds.

providers/os/id/networki/linux_n.go
afiune: We don't, sysfs is just a backup, I will split them and run either or, good point. 

In the future, it we do need statistics, we might need both. But not today.

jaym (approved):

PR #5399, opened 2025-04-02
✨ Use debian package list as evidence instead of its contents.

VasilSirakov:


providers/os/resources/packages/dpkg_packages.go
```diff
@@ -215,39 +215,11 @@ func (dpm *DebPkgManager) Available() (map[string]PackageUpdate, error) {
 }
 
 func (dpm *DebPkgManager) Files(name string, version string, arch string) ([]FileRecord, error) {
-	fs := dpm.conn.FileSystem()
+	path := "/var/lib/dpkg/info/" + name + ".list"
```
preslavgerchev: do we need to check if either of the paths exist?

VasilSirakov: good point, I assumed that since we already know the package is installed, the inferred `.list` file would be present but perhaps it wouldn't hurt to add a check for that @chris-rock?

preslavgerchev: the loop before had a check (`fs.Stat`), if it does not exist, no entries were added. I assume we still want to do the same

VasilSirakov: that's a fair point, better to have this check instead of risking reporting a non-existent file

PR #5409, opened 2025-04-03
Add status to sbom package

claytoncasey01:
Adds the status field to the sbom proto so we can use for determining efix in server.


chris-rock:
We do not need this for now.

PR #5411, opened 2025-04-03
✨ add docs to builtin types, expose their schema, and operators

arlimus:
This only gets the ball rolling, but it starts to expose these builtin types and starts to add docs to them. A lot more work needs to be done to give docs to all of them, but it gets us started at least.

We now also get programmatic access to the `mqlc.BuiltinDocs`, which is a `Schema` containing all our types and their fields and functions.

### DIFF
--- a/mqlc/builtin.go
+++ b/mqlc/builtin.go
@@ -411,8 +411,11 @@ func availableFields(c *compiler, typ types.Type) map[string]llx.Documentation {
 	return res
 }
 
+// BuiltinSchema captures all internal types and their metadata.
 // We could have just as well used the `resources.Schema` here.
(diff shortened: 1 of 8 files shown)


mqlc/builtin.go
```diff
@@ -344,3 +410,39 @@ func availableFields(c *compiler, typ types.Type) map[string]llx.Documentation {
 
 	return res
 }
+
+// We could have just as well used the `resources.Schema` here.
+// However, semantically t
```
chris-rock: there is something missing with the comment

arlimus: ouch, who did that :joy: :see_no_evil:

PR #5417, opened 2025-04-04
✨ Safe manipulation of `inventory.Platform`

afiune:
One of the core data structures of Mondoo’s plugin system is the `inventory.Asset` struct that contains details about an asset, including platform information that is stored in `Asset.Platform`.

The `inventory.Platform` struct stores important information used by the Mondoo platform (`server`) like the runtime, the platform kind, and additional metadata.

When we run `cnquery` or `cnspec` most of the times the first thing we do is an initial discovery to detect the asset and platform we are going to scan, then we pass these core data structures to our providers where, sometimes, we do additional discovery and enrichment of data. An example of this is the discovery of AWS instances (`aws` provider) that later gets enriched with additional information from the `os` provider.

# Change

We are introducing two accessors. One at the Asset level and the other one directly into the Platform struct. These new accessor should be used through out the code base to safely manipulate the platform of an asset.

### DIFF
--- a/providers-sdk/v1/inventory/asset.go
+++ b/providers-sdk/v1/inventory/asset.go
@@ -178,3 +178,12 @@ func (a *Asset) AddMondooLabels(root *Asset) {
 		}
 	}
 }
+
+// SetPlatform is an accessor to correctly set the platform of the asset.
+func (a *Asset) SetPlatform(pf *Platform) {
+	if a.Platform == nil {
+		a.Platform = pf
+	} else {
+		a.Platform.Merge(pf)
+	}
+}
--- a/providers-sdk/v1/inventory/asset.go
+++ b/providers-sdk/v1/inventory/asset.go
@@ -179,8 +179,9 @@ func (a *Asset) AddMondooLabels(root *Asset) {
 	}
 }
 
-// SetPlatform is an accessor to correctly set the platform of the asset.
-func (a *Asset) SetPlatform(pf *Platform) {
+// MergePlatform merges the provided platform into the asset's platform.
+// If the asset's platform is `nil`, the provided platform is set.
+func (a *Asset) MergePlatform(pf *Platform) {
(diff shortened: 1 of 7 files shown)


providers-sdk/v1/inventory/asset.go
```diff
@@ -178,3 +178,12 @@ func (a *Asset) AddMondooLabels(root *Asset) {
 		}
 	}
 }
+
+// SetPlatform is an accessor to correctly set the platform of the asset.
+func (a *Asset) SetPlatform(pf *Platform) {
```
chris-rock: I think semantically, SetPlatform should override the data. If the caller wants to merge, the caller should call MergePlatform().

providers-sdk/v1/inventory/inventory.go
```diff
@@ -318,13 +320,20 @@ var (
...
 )
 
-func (p *Platform) IsFamily(family string) bool {
-	for i := range p.Family {
-		if p.Family[i] == family {
-			return true
+// Merge performs a deep merge of the provided platform.
+func (p *Platform) Merge(pf *Platform) {
+	if pf != nil {
+		if err := mergo.Merge(p, pf, mergo.WithOverride); err != nil {
```
chris-rock, line 326: How do we ensure it does the right merge? Do we handle precedence?

afiune: Yes we do. There are two scenarios.
1. Default - When the new platform struct HAS values and the origin does NOT have, we set them
2. WithOverride - When the new platform struct HAS values, we set it

For both options, there has to bee non-zero values to take action.

I propose the override since, if any detector actually detects something, we most likely want that value inside the platform struct without loosing anything else we have collected.

Without the override, we are saying that the first thing that gets discovered is what has precedence, if we want that, we can remove the `WithOverride`.

vjeffrey (approved):

PR #5423, opened 2025-04-06
⭐ parse.xml

arlimus:
Add support for parsing XML files, either by providing the file or by providing the XML contents directly.

Using inline data:

```coffee
> parse.xml(content: '<root />').params
parse.xml.params: {
  root: {}
}
```

Via file `example.xml`:

```xml
<root>
  <box>
    <hello a="1"/>
  </box>
  <box>
    <world b="2">
      <c>3</c>
      4
    </world>
  </box>
  <box>🌎</box>
</root>
```

Running this in MQL:

```coffee
> parse.json('example.xml').params
parse.xml.params: {
  root: {
    box: [
      0: {
        hello: {
          @a: "1"
        }
      }
...
```

Since the conversion from XML to a flat (JSON-like) structure isn't standardized, we took an approach very similar to other formatters (e.g. [xq](https://github.com/sibprogrammer/xq), [jsonformatter](https://jsonformatter.org/xml-to-json)). This means:

- single items are added as a flat child element, e.g. the first `<box/>` element has a flat field `hello` above
- once there are multiple elements, they are turned into a list, e.g. `root.box` is an array above
- attributes are added as `@attribute` fields in the element
- if an element only has text contents, the value is just set to the text, e.g. `<root>val</root>` is turned into `{"root": "val"}`
- if an element has a mix of child elements and text contents, all text contents are added to the `__text` child field

Future consideration are to add access to the entire XML structure, which is more complex to traverse, but may be required for some use-cases. We will also handle streaming data separately (for all of these formats)

### DIFF
--- a/providers/os/resources/parse.go
+++ b/providers/os/resources/parse.go
@@ -165,7 +165,9 @@ func (x *xmlElem) addAttr(a []xml.Attr) {
 	if len(a) == 0 {
 		return
 	}
-	x.attributes = map[string]string{}
+	if x.attributes == nil {
+		x.attributes = map[string]string{}
+	}
 	for ai := range a {
 		attr := a[ai]
 		x.attributes[attrName(attr.Name)] = attr.Value
--- a/providers/os/resources/parse.go
+++ b/providers/os/resources/parse.go
@@ -204,10 +204,12 @@ func (x *xmlElem) UnmarshalXML(decoder *xml.Decoder, start xml.StartElement) err
 		case xml.CharData:
 			cur := path[len(path)-1]
 			v := strings.TrimSpace(string(elem))
-			cur.children = append(cur.children, &xmlElem{
-				data:      v,
-				isElement: false,
-			})
+			if v != "" {
+				cur.children = append(cur.children, &xmlElem{
+					data:      v,
+					isElement: false,
+				})
+			}
 		}
 	}
 }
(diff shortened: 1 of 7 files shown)


providers/os/resources/parse.go
```diff
@@ -130,6 +132,190 @@ func (s *mqlParseJson) params(content string) (interface{}, error) {
...
+	children   []*xmlElem
+	data       string
+	isElement  bool
+}
+
+func (x *xmlElem) addAttr(a []xml.Attr) {
+	if len(a) == 0 {
+		return
+	}
+	x.attributes = map[string]string{}
```
afiune: This always re-initializes the map.
```suggestion
	if x.attributes == nil {
		x.attributes = map[string]string{}
	}
```

providers/os/resources/parse.go
```diff
@@ -130,6 +132,190 @@ func (s *mqlParseJson) params(content string) (interface{}, error) {
...
+	res := map[string]any{}
+	for k, v := range x.attributes {
+		res["@"+k] = v
+	}
+
+	for i := range x.children {
+		child := x.children[i]
+		data, isElem, params := child._params()
+
+		// text data is added flat
```
afiune, line 232: nit I wonder if we should separate the text content into a slice to simplify this code, maybe adding the slice to the `xmlElem` and append it in `UnmarshalXML()` for the `xml.CharData` case.

afiune (approved):
One nit and one comment that needs to get in before merging (map re-initialization), besides that, this looks good to me!

providers/os/resources/parse.go
chris-rock: Thank you @afiune

PR #5439, opened 2025-04-08
✨ Add WIF auth support

claytoncasey01:
This PR includes several changes to enhance the authentication process. The most important changes are for adding support for Workload Identity Federation (WIF). 

- Integrate WIF json config into the existing `config.go` flow.
- New fields in `upstream.proto` to support `ExchangeExternalToken` endpoint
- Added an implementation of `ExchangeExternalToken` in `sts.go`, along with some helper functions to get a json token for the different providers (GCP, Azure, and Github Actions).
- Updates to `GetServiceCredential` function in `config.go` to handle a new authentication method `wif`, this uses the new `ExchangeExternalToken` and attempts to get the user a valid Service Account.

#### Testing
1. Build cnquery for linux `make cnquery/build/linux`
2. Make an auth binding for google in the UI and save it to a `.json` file
3. Upload the cnquery binary and wif config file to GCP cloud shell
4. Run `cnquery status --config path/to/config.json`


cli/config/config.go
```diff
@@ -117,15 +119,73 @@ func InitViperConfig() {
 		_, err := AppFs.Stat(Path)
 		if err == nil {
 			log.Debug().Str("configfile", viper.ConfigFileUsed()).Msg("try to load local config file")
-			if err := viper.ReadInConfig(); err == nil {
-				LoadedConfig = true
+
+			// Check if this is a WIF config file
+			isWif, wifErr := IsWifConfigFormat(Path)
```
jaym: is it not possible to make wif config a subset of our normal config file?

providers-sdk/v1/upstream/sts.go
```diff
@@ -53,6 +60,194 @@ func ExchangeSSHKey(apiEndpoint string, identityMrn string, resourceMrn string)
...
+	if issuerURI == "" {
+		switch provider {
+		case "gcp":
+			issuerURI = "https://accounts.google.com"
+		case "azure":
+			issuerURI = "https://sts.windows.net/"
+		case "github":
+			issuerURI = "https://token.actions.githubusercontent.com"
+		default:
+			issuerURI = "https://accounts.google.com"
```
jaym: probably just want to error here?

cli/config/config.go
claytoncasey01: Maybe, I will look into that a bit.

providers-sdk/v1/upstream/sts.go
claytoncasey01: Thinking about it, yeah, since we are now using the config the UI gives after creating the binding, this whole switch can go away since it should always have an issuerURI and if it doesn't then that's an error.

vjeffrey:
✨ im gonna set aside some time to review this later today, thank you!

vjeffrey (approved):
thank you, this is working great!

cli/config/config.go
afiune: Can we get rid of the error in this function and only return a boolean? 99% of the times we won't hit it and when we do, it won't be useful to print an error that says that we couldn't read the config file since we will later try to read it.

cli/config/config.go
```diff
@@ -117,13 +119,71 @@ func InitViperConfig() {
...
+			viper.Set("audience", audience)
+		}
+
+		// Set the issuer URI if available
+		if issuerUri := viper.GetString("issuerUri"); issuerUri != "" {
+			viper.Set("issuer_uri", issuerUri)
+		}
+
+		// Set the API endpoint from universeDomain if available
+		if universeDomain := viper.GetString("universeDomain"); universeDomain != "" {
```
afiune, line 196: Since all of these `viper.Get` function calls are happening before `viper.AutomaticEnv`, we won't have support for loading these options from environment variables.

If we want support, I would recommend moving them down below.

cli/config/config.go
```diff
@@ -245,14 +309,56 @@ func (c *CommonOpts) GetFeatures() cnquery.Features {
...
+
+			// Validate required fields for WIF
+			if c.Audience == "" {
+				log.Error().Msg("missing required field 'audience' for WIF authentication")
+				return nil
+			}
+
+			// If apiEndpoint is not set, check for universeDomain in viper
+			if c.APIEndpoint == "" {
+				universeDomain := viper.GetString("universeDomain")
```
afiune: How can we end up here? If we have `c.Authentication.Method == "wif"` I think it means we loaded the WIF config during `InitViperConfig` so this feels redundant and we should get rid of it unless we are covering some edge case that I am not seeing. 🤔

cli/config/config.go
```diff
@@ -295,6 +401,11 @@ func (c *CommonOpts) GetParentMrn() string {
 }
 
 func (c *CommonOpts) UpstreamApiEndpoint() string {
+	// First check for environment variable override
+	if envEndpoint := os.Getenv("MONDOO_API_ENDPOINT"); envEndpoint != "" {
```
afiune: The loading of environment variables like this one happens automatically, why do we need this?

cli/config/config.go
claytoncasey01: Yeah this is definitely redundant, removed.

cli/config/config.go
claytoncasey01: Left overs from when I was passing this stuff as flags in cnspec, removed.

cli/config/config.go
claytoncasey01: Nice catch! Moved the `viper.Get` calls below `viper.AutomaticEnv` with my latest push.

PR #5450, opened 2025-04-14
🧹 Improve our pretty platform values

tas50:
- Consistent capitalization
- Match what we're doing in asset overview now


chris-rock (approved):
Thank you @tas50

PR #5453, opened 2025-04-15
🧹 add more npm standard locations

chris-rock:


preslavgerchev (approved):

providers/os/resources/npm.go
```diff
@@ -32,8 +32,14 @@ var (
 		"/home/*/.npm-global/lib",
 		// Windows
 		"C:\\Users\\*\\AppData\\Roaming\\npm",
+		"C:\\Program Files\\nodejs\\node_modules\\npm",
+		"C:\\Users\\*\\node_modules",
 		// macOS
 		"/Users/*/.npm-global/lib",
```
tas50, line 38: @chris-rock Do we also want to support NVM on the mac: with paths like `/Users/tsmith/.nvm/versions/node/v22.13.1/lib/node_modules`

PR #5457, opened 2025-04-15
✨ add new exception types

kkereziev:


explorer/cnquery_explorer.pb.go
```diff
@@ -49,14 +52,20 @@ var (
 		3: "ACTIVATE",
 		4: "IGNORE",
```
tas50, line 60: Is this the old snooze name before it became snooze? If so we can reuse this for Risk Accepted. We're just renaming Snooze -> Risk Accepted

kkereziev: Ya, this pr wasn't finished, sorry, will transition it as a draft.

kkereziev: Unfortunately we can't fully get rid of `IGNORE` because it is used elsewhere - https://github.com/mondoohq/cnspec/blob/d8782404d422eb4340dd4299a7299d363f526775/policy/cnspec_policy.proto#L71, there are multiple examples in our internal services too. This enum is used in multiple contexts. `IGNORE` and `RISK_ACCEPTED` will have to co-exist for now.

PR #5458, opened 2025-04-15
🐛 Fetch primary group members from /etc/passwd

jaym:
The primary group for a user isn't in /etc/group

https://askubuntu.com/questions/1373865/etc-group-why-the-username-does-not-appear-in-its-own-primary-group


vjeffrey (approved):
✨ thank u!

PR #5461, opened 2025-04-16
✨ cache aws connection and upstream client

afiune:
This is an attempt to reduce the load of the AWS provider. We are going to cache the upstream client which uses ranger under the hood and creates goroutines to fulfill the requests. This is not a problem for a few runtimes, but when we do thousands of them, we really don't need to create that many clients.


mql/mql_test.go
```diff
@@ -92,6 +92,7 @@ func TestMqlSimple(t *testing.T) {
 			assert.Equal(t, one.assertion, res.Value)
 		})
 	}
+	goleak.VerifyNone(t)
```
afiune:
```suggestion
```

providers/aws/connection/connection.go
```diff
@@ -118,32 +129,58 @@ func NewAwsConnection(id uint32, asset *inventory.Asset, conf *inventory.Config)
...
+	c.opts.options = asset.Options
+	if conf.Discover != nil {
+		c.Filters = parseOptsToFilters(conf.Discover.Filter)
+	}
+	return c, nil
+}
+
+func (c *AwsConnection) Hash() uint64 {
+	// generate hash of the config options used to generate this connection
+	// used to avoid verifying a client with the same options more than once
```
afiune:
```suggestion
	// generate hash of the config options used to to initialize this connection,
	// we use this to avoid verifying a client with the same options more than once
```

providers/aws/connection/connection.go
```diff
@@ -118,32 +129,58 @@ func NewAwsConnection(id uint32, asset *inventory.Asset, conf *inventory.Config)
...
+}
+
+func (c *AwsConnection) Verify() (string, error) {
+	cfgCopy := c.cfg.Copy()
+	identity, err := CheckIam(c.cfg.Copy())
 	if err != nil {
 		log.Debug().Err(err).Msg("could not gather details of AWS account")
+		err1 := err
 		// try with govcloud region
 		cfgCopy.Region = "us-gov-west-1"
```
afiune:
```suggestion
	identity, err := CheckIam(c.cfg.Copy())
	if err != nil {
		log.Debug().Err(err).Msg("could not gather details of AWS account")
		// try with govcloud region, store error to return it if this last option does not work
		err1 := err
		cfgCopy := c.cfg.Copy()
		cfgCopy.Region = "us-gov-west-1"
```

PR #5464, opened 2025-04-16
✨ Add Memoizer to `plugin.Service`

afiune:
We have identified a pattern where, our plugins will need to have a way to cache stuff when creating runtimes, connections and clients, by adding a `Memoizer` interface that allows us to do that, and embedding it into the `plugin.Service` struct, all providers will have it out-of-the-box.


explorer/scan/local_scanner.go
```diff
@@ -205,7 +205,7 @@ func CreateProgressBar(discoveredAssets *DiscoveredAssets, disableProgressBar bo
 	return multiprogress, nil
 }
 
-func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *upstream.UpstreamConfig) (*explorer.ReportCollection, error) {
+func (s *LocalScanner) distributeJob(ctx context.Context, job *Job, upstream *upstream.UpstreamConfig) (*explorer.ReportCollection, error) {
```
afiune, line 208: This was my editor telling me that `ctx` should always be the first argument of a func, it is an internal function so it shouldn't be a problem, happy to remove this if we don't like it here.

providers-sdk/v1/plugin/service.go
```diff
@@ -132,6 +141,7 @@ func (s *Service) Disconnect(req *DisconnectReq) (*DisconnectRes, error) {
 	s.runtimesLock.Lock()
 	defer s.runtimesLock.Unlock()
 	s.doDisconnect(req.Connection)
+	s.Flush() // flush our Memoizer
```
jaym: something here doesn't feel quite right. You remove everything from the cache when any runtime disconnects. But there could still be other runtimes that are still running

providers-sdk/v1/util/memoize/memoize.go
```diff
@@ -7,48 +7,97 @@
...
+// Flush will remove all items from the cache, set `storage` to `nil` and
+// run the garbage collector to release long running goroutines.
+func (m *goCache) Flush() {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	if m.storage != nil {
+		m.storage.Flush()
+		m.storage = nil
+		go runtime.GC()
```
jaym: not convinced we need to start a new goroutine for runtime.GC

providers-sdk/v1/util/memoize/memoize.go
```diff
@@ -7,48 +7,97 @@
...
 		}
 
 		return data, innerErr
 	})
-	return value, err, false
+	return value, false, err
+}
+
+// Flush will remove all items from the cache, set `storage` to `nil` and
+// run the garbage collector to release long running goroutines.
```
jaym, line 83: can you clarify this comment. I'm under the impression that running the garbage collector cleans up memory, not goroutines

providers-sdk/v1/plugin/service.go
afiune: Good point. Should we check if there are still runtimes, and we only flush once we have no runtime?

providers-sdk/v1/util/memoize/memoize.go
afiune: GC might take some time, if we don't do it in a goroutine, it will block until GC finishes.

providers-sdk/v1/util/memoize/memoize.go
afiune: It does both. It cleans goroutines because of the underlying implementation (`go-cache`) that uses [runtime finalizers](https://github.com/patrickmn/go-cache/blob/master/cache.go#L1123).

More info here https://github.com/patrickmn/go-cache/issues/166#issuecomment-1494191038

I think we could try to switch to https://github.com/go-pkgz/expirable-cache on a follow up PR.

jaym (approved):

mql/mql_test.go
```diff
@@ -75,6 +92,7 @@ func TestMqlSimple(t *testing.T) {
 			assert.Equal(t, one.assertion, res.Value)
 		})
 	}
+	goleak.VerifyNone(t)
```
afiune: redundant
```suggestion
```

PR #5465, opened 2025-04-16
Send Slack notification on provider build failure

benr:
Addresses https://github.com/mondoohq/server/issues/10891


vjeffrey (approved):
nice! thank u!

PR #5480, opened 2025-04-25
🐛 OS Provider Segfault Fix

claytoncasey01:
The `os` provider has some dependency on the `network` provider. This PR adds a dependencies array to providers that be be updated in the providers configuration. It will search through these dependencies and install any that are missing. 

This is to fix: https://github.com/mondoohq/server/issues/10794

### DIFF
--- a/providers/defaults.go
+++ b/providers/defaults.go
@@ -11,19 +11,21 @@ import "go.mondoo.com/cnquery/v11/providers-sdk/v1/plugin"
 // to tell users what providers are used for common connections, when there
 // is no other way to find out.
 var DefaultProviders Providers = map[string]*Provider{
+
 	"ansible": {
 		Provider: &plugin.Provider{
 			Name:            "ansible",
 			ID:              "go.mondoo.com/cnquery/v11/providers/ansible",
 			ConnectionTypes: []string{"ansible"},
 			Connectors: []plugin.Connector{
+
 				{
 					Name:  "ansible",
 					Use:   "ansible PATH",
 					Short: "an Ansible playbook",
 				},
 			},
-			Dependencies: []string{},
+			Dependencies: []string(nil),
 		},
 	},
 
@@ -33,13 +35,14 @@ var DefaultProviders Providers = map[string]*Provider{
 			ID:              "go.mondoo.com/cnquery/v9/providers/arista",
 			ConnectionTypes: []string{"arista"},
 			Connectors: []plugin.Connector{
+
 				{
 					Name:  "arista",
 					Use:   "arista user@host",
 					Short: "an Arista EOS device",
 				},
 			},
-			Dependencies: []string{},
+			Dependencies: []string(nil),
 		},
 	},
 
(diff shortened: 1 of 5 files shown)


afiune (commented):
As we chatted, I don't think that installing the provider is the right thing to do here, instead look at the `lr` CLI and see if you can render the missing resources from the import that the os provider has at https://github.com/mondoohq/cnquery/blob/main/providers/os/resources/os.lr#L5 - If you look at the `os.lr.manifest.yaml` those resources are missing.

claytoncasey01:
> As we chatted, I don't think that installing the provider is the right thing to do here, instead look at the `lr` CLI and see if you can render the missing resources from the import that the os provider has at https://github.com/mondoohq/cnquery/blob/main/providers/os/resources/os.lr#L5 - If you look at the `os.lr.manifest.yaml` those resources are missing.

@afiune @vjeffrey and I discussed this morning and decided for now just to go with a simple dependencies array approach. We can revisit this approach if dependencies become a normal thing.

providers/defaults.go
```diff
@@ -11,18 +11,21 @@ import "go.mondoo.com/cnquery/v11/providers-sdk/v1/plugin"
...
 			ConnectionTypes: []string{"ansible"},
 			Connectors: []plugin.Connector{
+
 				{
 					Name:  "ansible",
 					Use:   "ansible PATH",
 					Short: "an Ansible playbook",
 				},
 			},
+			Dependencies: []string(nil),
```
afiune, line 28: do we need all of these initializations? by default, dependencies should be empty. 🤔 what am I missing?

claytoncasey01: That entire file is auto generated, so it appears to just default to []string(nil) when it doesn't have any dependencies slice in the config for that provider. Having this still allows us to range over this without explicit nil checks. We can try to force it to remove this but it seems to be the default behavior when this file is being generated. 

We could add a check to only include it in the generated `defaults.go` if it actually has data but we would then also have to add a nil check before we attempt to loop through it.

claytoncasey01: All of that said, I don't really have a preference one way or another.

providers/providers.go
```diff
@@ -462,6 +470,29 @@ func installVersion(name string, version string) (*Provider, error) {
...
+func installDependencies(provider *Provider, existing Providers) {
+	for _, depName := range provider.Dependencies {
+		// Check if dependency is already installed
+		depProvider := existing.Lookup(ProviderLookup{ProviderName: depName})
+		if depProvider == nil {
+			upstreamDep := DefaultProviders.Lookup(ProviderLookup{ProviderName: depName})
+			if upstreamDep != nil {
+				depProvider, err := Install(upstreamDep.Name, "")
+				if err != nil {
+					log.Warn().Err(err).
```
afiune: Why installing a dependency doesn't return an error? 🤔

claytoncasey01: Just a miss on my part, added.

afiune (approved):
Lets get this fixed.

PR #5481, opened 2025-04-25
🐛 Fix crash in processes resource on AIX

jaym:
This code was calling log.Fatal when it failed to parse a line. The line was something like
```
18875522                                    -      <idle>
```
Looks like some sort of kernel process

Regardless, this change allows idle and kproc. And it removes uses of fatal and just returns an error


claytoncasey01 (approved):
LGTM 🚢

PR #5489, opened 2025-04-29
🐛 Fix process handling for Linux kernel threads

### LINKED ISSUE #5422: OS Provider: `processes {flags, command} ` cannot display kernel threads
**Describe the bug**
Running the command `processes {flags, command}` runs into an error for Kernel processes.
```
cnspec> processes { pid command flags }
1 error occurred:
	* no command provided
processes.list: [
  0: {
    flags: {
      deserialize: "31"
      switched-root: ""
      system: ""
    }
    command: "/usr/lib/systemd/systemd --switched-root --system --deserialize 31 "
    pid: 1
  }
  1: {
    flags: no command provided
    command: ""
    pid: 2
  }
```

**To Reproduce**
See above

**Expected behavior**
Internally we're using the `/proc/<uid>/cmdline` info to get the flags and the command.
For kernel processes on Linux this file is empty.

Would it be possible in that case to use the `/proc/<uid>/comm` file to get the  Kernel thread name and add empty flags?

```
# cat `/proc/<uid>/comm`
kthreadd
```

 os 11.3.41

glower:
this change fixes issue where kernel threads weren't properly detected due to empty /proc/<pid>/cmdline
and adds fallback to read process name from /proc/<pid>/comm when cmdline is empty

resolves #5422

### DIFF
--- a/providers/os/resources/processes/linuxproc.go
+++ b/providers/os/resources/processes/linuxproc.go
@@ -85,6 +85,20 @@ func (lpm *LinuxProcManager) Process(pid int64) (*OSProcess, error) {
 		return nil, err
 	}
 
+	// If cmdline is empty (likely a kernel thread), read from /proc/<pid>/comm instead
+	if cmdline == "" {
+		commf, err := lpm.conn.FileSystem().Open(filepath.Join(pidPath, "comm"))
+		if err != nil {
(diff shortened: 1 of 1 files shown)


chris-rock (approved):
Thank you @glower

PR #5495, opened 2025-05-01
⭐️ `lr`: add provider dependencies to schema

afiune:
Resources within a provider can depend on resources from another provider,
for example; The `os` provider depends on `network` provider.

The `lr` parser already reads these imports but it wasn't passing it to the
schema. This change adds the list of dependencies so that we can install
them at runtime.


claytoncasey01 (approved):
Nice, thanks @afiune

jaym (approved):

PR #5504, opened 2025-05-06
🐛 Correctly handle large UIDs/GIDs as float64

glower:
resolves https://github.com/mondoohq/cnquery/issues/5451

On AIX and potentially other platforms using filesystems like JFS/NFSv4/ZFS, users and groups like 'nobody' can be assigned very large unsigned 32-bit UIDs/GIDs, such as 4294967294.

This fix will handle UIDs/GIDs as float64 values internally to properly process numeric data


providers/os/resources/group.go
```diff
@@ -135,7 +135,7 @@ func (x *mqlGroups) list() ([]interface{}, error) {
 		group := groups[i]
 		nu, err := CreateResource(x.MqlRuntime, "group", map[string]*llx.RawData{
 			"name": llx.StringData(group.Name),
-			"gid":  llx.IntData(group.Gid),
+			"gid":  llx.FloatData(group.Gid),
```
chris-rock, line 138: I think we would need to update the schema for the group

providers/os/resources/users/testdata/debian.toml
```diff
@@ -11,7 +11,7 @@ mail:x:8:12:mail:/var/spool/mail:/sbin/nologin
 operator:x:11:0:operator:/root:/sbin/nologin
 games:x:12:100:games:/usr/games:/sbin/nologin
 ftp:x:14:50:FTP User:/var/ftp:/sbin/nologin
-nobody:x:99:99:Nobody:/:/sbin/nologin"""
+nobody:x:4294967294:4294967294:Nobody:/:/sbin/nologin"""
```
chris-rock, line 14: Instead of changing debian, I think we should add a AIX related test, or can this happen on linux too?

glower: JFS/NFSv4 can be also used on linux, but not by default, ZFS not sure

tas50: ZFS is not in Linux by default but can be added. I use it with my Debian NAS

providers/os/resources/group.go
glower: I changes both, `gid` and `uid` to be floats and the change in the code escalated a bit :)
I hope I did not miss any place

PR #5505, opened 2025-05-06
🐛 Pass in correct auto-update flag to EnsureProvider.

preslavgerchev:
Putting this in draft as that seems to break specific tests


afiune:
Superseded by https://github.com/mondoohq/cnquery/pull/5510

Issue #4879, opened 2024-11-18
Discover applications installed via snap on Linux distributions


czunker:
**Is your feature request related to a problem? Please describe.**
We do not discover applications installed via snap on Linux systems

**Describe the solution you'd like**
Query snap for installed applications and report them as packages.

**Describe alternatives you've considered**
Ignore them. Not an option.

**Additional context**
-

chris-rock:
Best is we parse `cat /var/lib/snapd/state.json` (requires root) file. If that is not possible we need to fallback to 

```
snap list --all --color=never
Name              Version   Rev    Tracking         Publisher          Notes
core22            20250315  1908   latest/stable    canonical✓         base
google-cloud-cli  520.0.0   328    latest/stable/…  google-cloud-sdk✓  classic
snapd             2.67.1    23771  latest/stable    canonical✓         snapd
```

PR #5519, opened 2025-05-08
✨ Collect snap packages

czunker:
This collects data about the present snap packages. It currently appllies to:
- Ubuntu
- Debian
- Fedora

Fixes: https://github.com/mondoohq/cnquery/issues/4879


czunker:
This is slow over ssh connections:
```
time cnspec run vagrant default -c "packages.where(format == 'snap'){ name version purl }" 
packages.where.list: [
  0: {
    version: "1.0"
    purl: "pkg:snap/ubuntu/bare@1.0?distro=ubuntu-22.04"
    name: "bare"
  }
...
  11: {
    version: "1.18.5"
    purl: "pkg:snap/ubuntu/vault@1.18.5?arch=amd64&distro=ubuntu-22.04"
    name: "vault"
  }
]
cnspec run vagrant default -c   16,69s user 20,81s system 36% cpu 1:41,40 total
```

Alternatives described in #4879 would require running as root or executing a command.

providers/os/resources/packages/snap_packages.go
```diff
@@ -0,0 +1,129 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	pkgList := []Package{}
+	for _, file := range files {
+		manifest, err := afs.Open(file)
+		if err != nil {
+			log.Error().Err(err).Str("file", file).Msg("could not open manifest file")
```
glower, line 70: should we not add `continue` here? otherwise manifest would be `nil`

providers/os/resources/packages/snap_packages.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+		pkg, err := spm.parseSnapManifest(manifest)
+		if err != nil {
+			log.Error().Err(err).Str("file", file).Msg("could not parse manifest file")
```
glower, line 75: same here but with `pkg`

providers/os/resources/packages/detect.go
```diff
@@ -13,27 +13,35 @@ import (
...
+		// TODO: do we really need to make this a blocking call, we could update available updates async
+		// we try to retrieve the available updates
+		available, err := pm.Available()
+		if err != nil {
+			log.Debug().Err(err).Msg("mql[packages]> could not retrieve available updates")
+			available = map[string]PackageUpdate{}
+		}
+		log.Debug().Int("updates", len(available)).Msg("mql[packages]> available updates")
+		for k, v := range available {
+			availableList[k] = v
```
glower, line 42: could it be that if two different package managers (e.g., `apt` and `snap`) both provide an update for the same package name (e.g., "firefox"), the entry would be overwritten?
Don't know if this could be a problem or not

providers/os/resources/packages/snap_packages.go
czunker: Good catch. Thank you.

providers/os/resources/packages/snap_packages.go
czunker: Good catch. Thank you.

providers/os/resources/packages/snap_packages.go
czunker: I opted not to defer but to explicitly close because of the loop.

providers/os/resources/packages/detect.go
czunker: Currently not, because snap does not provide available at the moment.

czunker:
> fun fact, I was working in the team responsible for snaps at Canonical :)

We need to talk :smile:

PR #5523, opened 2025-05-09
🐛 Fix npm package ID to be uniq across directories

czunker:
When the same package is installed in different locations, generate unique IDs to show correct information for each package.

Before:
```
cnspec run vagrant default -c "npm.packages.where(name == /yosay/){ id name version files }" 
npm.packages.where.list: [
  0: {
    name: "yosay"
    id: "yosay/usr/local/lib/node_modules/yo/package.json"
    version: "^2.0.2"
    files: [
      0: pkgFileInfo path="/usr/local/lib/node_modules/yo/package.json"
    ]
  }
  1: {
    name: "yosay"
    id: "yosay/usr/local/lib/node_modules/generator-code/package.json"
    version: "^3.0.0"
    files: [
      0: pkgFileInfo path="/usr/local/lib/node_modules/generator-code/package.json"
    ]
  }
]
```

With this PR:
```
npm.packages.where.list: [
  0: {
    name: "yosay"
    id: "yosay"
    version: "^2.0.2"
    files: [
      0: pkgFileInfo path="/usr/local/lib/node_modules/yo/package.json"
    ]
  }
  1: {
    name: "yosay"
    id: "yosay"
    version: "^2.0.2"
    files: [
      0: pkgFileInfo path="/usr/local/lib/node_modules/yo/package.json"
    ]
  }
]

```

To reproduce:
```
npm install -g yo generator-code
```

### DIFF
--- a/providers/os/resources/npm.go
+++ b/providers/os/resources/npm.go
@@ -23,25 +23,23 @@ import (
 	"go.mondoo.com/cnquery/v11/types"
 )
 
-var (
-	defaultNpmPaths = []string{
-		// Linux
-		"/usr/local/lib",
-		"/opt/homebrew/lib",
-		"/usr/lib",
-		"/home/*/.npm-global/lib",
-		// Windows
-		"C:\\Users\\*\\AppData\\Roaming\\npm",
-		"C:\\Program Files\\nodejs\\node_modules\\npm",
-		"C:\\Users\\*\\node_modules",
-		// macOS
-		"/Users/*/.npm-global/lib",
-		// Container app paths
-		"/app",
-		"/home/node/app",
-		"/usr/src/app",
-	}
-)
+var defaultNpmPaths = []string{
+	// Linux
+	"/usr/local/lib",
+	"/opt/homebrew/lib",
+	"/usr/lib",
+	"/home/*/.npm-global/lib",
+	// Windows
+	"C:\\Users\\*\\AppData\\Roaming\\npm",
+	"C:\\Program Files\\nodejs\\node_modules\\npm",
+	"C:\\Users\\*\\node_modules",
+	// macOS
+	"/Users/*/.npm-global/lib",
+	// Container app paths
+	"/app",
+	"/home/node/app",
+	"/usr/src/app",
+}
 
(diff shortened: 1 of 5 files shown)


providers/os/resources/npm_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// To get the correct data, we need distinct IDs
+	require.NotEqual(t, pkg2.MqlID(), pkg3.MqlID())
+	require.Equal(t, "yosay/usr/local/lib/node_modules/yo/package.json", pkg2.MqlID())
+	require.Equal(t, "yosay/usr/local/lib/node_modules/generator-code/package.json", pkg3.MqlID())
+}
+
+// Mock callbacks for testing
+// These are needed during calls to CreateSharedResource
+type providerCallbacks struct{}
```
preslavgerchev, line 80: can we extract this in a test util file so it can be used elsewhere too?

providers/os/resources/npm.go
```diff
@@ -322,8 +320,14 @@ func newNpmPackage(runtime *plugin.Runtime, pkg *languages.Package) (*mqlNpmPack
 		mqlFiles = append(mqlFiles, lf)
 	}
 
+	path := ""
+	if len(mqlFiles) > 0 {
```
preslavgerchev, line 324: can we make `mqlFiles` to be a `[]*mqlPkgFileInfo{}` list? that way we avoid the casting. not sure if this causes a problem later in the code

providers/os/resources/npm_test.go
czunker: I gave it a try.
When we extract this to the testutils, it results in an import cycle.
When we move it into the mockprovider, the k8s tests fail.

Seem like they are to specific to the tests.

providers/os/resources/npm.go
czunker: No,
L335:
```
"files":   llx.ArrayData(mqlFiles, types.Resource("pkgFileInfo")),
```
This requires `[]interfac{}`.

preslavgerchev (approved):

PR #5525, opened 2025-05-10
Add a resource for parsing journald

tas50:
Replaces a query like this:

```
["/etc/systemd/journald.conf"].where(file(_).exists) {
  file(_) {
    parse.ini("/etc/systemd/journald.conf").sections["Journal"]["ForwardToSyslog"].downcase == "yes"
  }
}
```

with this

```
journald.config.params.ForwardToSyslog == "yes"
```


arlimus (approved):
Glad you found another use for the auditd pattern!

PR #5526, opened 2025-05-12
🐛 Do not stop package collection if snap dir is not present.

preslavgerchev:


czunker (approved):
Thanks @preslavgerchev

PR #5529, opened 2025-05-12
✨ support inRange on strings

arlimus:
The operator `inRange` currently only works on numbers. This makes it work with number-like strings as well, by parsing the number in the string and then comparing them.

```coffee
> "2".inRange(1,3)
true
```


chris-rock (approved):
Awesome. Thank you @arlimus

PR #5536, opened 2025-05-13
:broom: Fix logindefs resource, allow for trailing comments

mm-weber:
logindefs resources didn't pick up on this line because of the comment:
```
PASS_MIN_DAYS 1  # Updated by CIS Build Kit
```

Updated resource to ignore comments after the setting, added testdata and test


czunker (approved):
LGTM
Thanks @mm-weber

PR #5538, opened 2025-05-15
✨ add string.notIn([..])

arlimus:
Add the counterpart to the `in` call:

```coffee
> 'term'.notIn(['a','b','c'])
true

> 'term'.notIn(['term'])
false
```


chris-rock (approved):
Very helpful. Thank you @arlimus

PR #5546, opened 2025-05-16
⭐ auditd.rules resource

arlimus:


mm-weber:
Thanks for working on this @arlimus !

I've played around with it a little trying to re-write this check:
```
  - uid: cis-rocky-linux-9--6.3.3.15
    title: Ensure successful and unsuccessful attempts to use the chcon command are collected
    impact: 60
    props:
      - uid: auditFiles
        title: Return the content from all /etc/audit/rules.d and /etc/audit/audit.rules
        mql: |
          auditFiles = files.find(from: "/etc/audit/rules.d",regex:'.*\.rules$' , type: "file").list.map(path) + ["/etc/audit/audit.rules"]
          return auditFiles.map(file(_).content.lines.where( _ == /^[^#]/ ))
    variants:
      - uid: linux-successful-and-unsuccessful-attempts-chcon-command-collected-static
      - uid: linux-successful-and-unsuccessful-attempts-chcon-command-collected-live-system
  - uid: linux-successful-and-unsuccessful-attempts-chcon-command-collected-static
    filters: mondoo.capabilities.contains("run-command") != true
    mql: |
      props.auditFiles.flat.where(_ == /chcon/).unique != empty
      props.auditFiles.flat.where(_ == /chcon/).unique.any(
        split("-").containsAll(["a always,exit ","F path=/usr/bin/chcon ","F perm=x "])
          && split("=").any(split(" ").any(_ == logindefs.params.UID_MIN))
          && split("=").any(split(" ").any(_ == "auid>")
          && split("-").containsAll(["F auid!=","1 ","F key=perm_chng"])
            || split("-").containsAll(["F auid!=unset ","k perm_chng"])
        )
      )
```

## Problem, still need to use `split` to get values properly:

Let's return one possible ways what the rule could look like with the new resource:
```
cnspec> auditd.rules.syscalls.where(fieldEntries.any(_ == "path=/usr/bin/chacl") ) {*}
auditd.rules.syscalls.where: [
  0: {
    list: "exit"
    syscalls: []
    keyname: "perm_chng"
    action: "always"
    fieldEntries: [
      0: "path=/usr/bin/chacl"
      1: "perm=x"
      2: "auid>=1000"
      3: "auid!=unset"
    ]
  }
]
```

1) Unfortunately it's still impossible to work without weird usage of `split` in the current version:
```
auditd.rules.syscalls.where(fieldEntries.any(_ == "path=/usr/bin/chacl") ) {fieldEntries.where(_.split("auid>=")[1]) {_.split('auid>=')} }
auditd.rules.syscalls.where: [
  0: {
    fieldEntries.where: [
      0: {
        split: [
          0: ""
          1: "1000"
        ]
      }
    ]
  }
]
```

I'm trying to achieve this here:
```
cnspec> auditd.rules.syscalls.where(fieldEntries.any(_ == "path=/usr/bin/chacl") ) {fieldEntries.where(_.split("auid>=")[1]).any(_.split('auid>=')[1] == logindefs.params.UID_MIN) }
auditd.rules.syscalls.where: [
  0: {
    [].any(): true
  }
]
```

## Possible solution:

Would it be possible to hard-code the possible to make the list `fieldEntries[]` into a hashmap relying on the possible options ? (we can start with a limited set of possible options)

Note: You cannot write `"perm= x"`, as this will break the audit.rules syntax, and lead to errors loading the rules. This makes things easier.

```
auditd.rules.syscalls.where: [
  0: {
    list: "exit"
    syscalls: []
    keyname: "perm_chng"
    action: "always"
    fieldEntries: {
      path=: "/usr/bin/chacl"
      perm=: "x"
      auid>=: "1000"
      auid!=: "unset"
    }
  }
]
```

This looks a bit weird, so we could just remove the "=" string, so it looks like that:
```
auditd.rules.syscalls.where: [
  0: {
    list: "exit"
    syscalls: []
    keyname: "perm_chng"
    action: "always"
    fieldEntries: {
      path: "/usr/bin/chacl"
      perm: "x"
      auid>: "1000"
      auid!: "unset"
    }
  }
]
```

This would allow to write the check in question like:
```
auditd.rules.syscalls.where(fieldEntries.any(_ == "path=/usr/bin/chacl")) {fieldEntries['auid>'] == logindefs.params.UID_MIN}
```

mm-weber:
## List of hard-coded keys we need at the moment:
```
fieldEntries: {
      arch:
      auid>: 
      auid!: 
      exit:
      path: 
      perm: 
      
}
```

mm-weber:
> Thanks for working on this @arlimus !
> 
> I've played around with it a little trying to re-write this check:
> 
> ```
>   - uid: cis-rocky-linux-9--6.3.3.15
>     title: Ensure successful and unsuccessful attempts to use the chcon command are collected
>     impact: 60
>     props:
>       - uid: auditFiles
>         title: Return the content from all /etc/audit/rules.d and /etc/audit/audit.rules
>         mql: |
>           auditFiles = files.find(from: "/etc/audit/rules.d",regex:'.*\.rules$' , type: "file").list.map(path) + ["/etc/audit/audit.rules"]
>           return auditFiles.map(file(_).content.lines.where( _ == /^[^#]/ ))
>     variants:
>       - uid: linux-successful-and-unsuccessful-attempts-chcon-command-collected-static
>       - uid: linux-successful-and-unsuccessful-attempts-chcon-command-collected-live-system
>   - uid: linux-successful-and-unsuccessful-attempts-chcon-command-collected-static
>     filters: mondoo.capabilities.contains("run-command") != true
>     mql: |
>       props.auditFiles.flat.where(_ == /chcon/).unique != empty
>       props.auditFiles.flat.where(_ == /chcon/).unique.any(
>         split("-").containsAll(["a always,exit ","F path=/usr/bin/chcon ","F perm=x "])
>           && split("=").any(split(" ").any(_ == logindefs.params.UID_MIN))
>           && split("=").any(split(" ").any(_ == "auid>")
>           && split("-").containsAll(["F auid!=","1 ","F key=perm_chng"])
>             || split("-").containsAll(["F auid!=unset ","k perm_chng"])
>         )
>       )
> ```
> 
> ## Problem, still need to use `split` to get values properly:
> Let's return one possible ways what the rule could look like with the new resource:
> 
> ```
> cnspec> auditd.rules.syscalls.where(fieldEntries.any(_ == "path=/usr/bin/chacl") ) {*}
> auditd.rules.syscalls.where: [
>   0: {
>     list: "exit"
>     syscalls: []
>     keyname: "perm_chng"
>     action: "always"
>     fieldEntries: [
>       0: "path=/usr/bin/chacl"
>       1: "perm=x"
>       2: "auid>=1000"
>       3: "auid!=unset"
>     ]
>   }
> ]
> ```
> 
> 1. Unfortunately it's still impossible to work without weird usage of `split` in the current version:
> 
> ```
> auditd.rules.syscalls.where(fieldEntries.any(_ == "path=/usr/bin/chacl") ) {fieldEntries.where(_.split("auid>=")[1]) {_.split('auid>=')} }
> auditd.rules.syscalls.where: [
>   0: {
>     fieldEntries.where: [
>       0: {
>         split: [
>           0: ""
>           1: "1000"
>         ]
>       }
>     ]
>   }
> ]
> ```
> 
> I'm trying to achieve this here:
> 
> ```
> cnspec> auditd.rules.syscalls.where(fieldEntries.any(_ == "path=/usr/bin/chacl") ) {fieldEntries.where(_.split("auid>=")[1]).any(_.split('auid>=')[1] == logindefs.params.UID_MIN) }
> auditd.rules.syscalls.where: [
>   0: {
>     [].any(): true
>   }
> ]
> ```
> 
> ## Possible solution:
> Would it be possible to hard-code the possible to make the list `fieldEntries[]` into a hashmap relying on the possible options ? (we can start with a limited set of possible options)
> 
> Note: You cannot write `"perm= x"`, as this will break the audit.rules syntax, and lead to errors loading the rules. This makes things easier.
> 
> ```
> auditd.rules.syscalls.where: [
>   0: {
>     list: "exit"
>     syscalls: []
>     keyname: "perm_chng"
>     action: "always"
>     fieldEntries: {
>       path=: "/usr/bin/chacl"
>       perm=: "x"
>       auid>=: "1000"
>       auid!=: "unset"
>     }
>   }
> ]
> ```
> 
> This looks a bit weird, so we could just remove the "=" string, so it looks like that:
> 
> ```
> auditd.rules.syscalls.where: [
>   0: {
>     list: "exit"
>     syscalls: []
>     keyname: "perm_chng"
>     action: "always"
>     fieldEntries: {
>       path: "/usr/bin/chacl"
>       perm: "x"
>       auid>: "1000"
>       auid!: "unset"
>     }
>   }
> ]
> ```
> 
> This would allow to write the check in question like:
> 
> ```
> auditd.rules.syscalls.where(fieldEntries.any(_ == "path=/usr/bin/chacl")) {fieldEntries['auid>'] == logindefs.params.UID_MIN}
> ```

arlimus:
@mm-weber can you describe to me in words what you are trying to check in this auditd rule? Ignore MQL for a second, just in words, what should the line contain?

arlimus:
Ok, I'll help clarify a few things in the meantime:

1. The fields in syscall monitoring aren't just split by `=`. I can see the mixup, since the manpage clearly says `-F field=value`. However, just reading further down, you see: "The audit system considers uids to be unsigned numbers." and then a bunch of comparison operations. That means `auid>=1000` is not an assignment, i.e. the field is still `auid` not `auid>`. The same is true for `!=` and others.
2. I'm going to assume you are trying to see if `-F auid>=X` is set with X at a minimum 1000. Which means you are really looking for two things: you are looking for a field that has both the X value at min 1k and the operation set to > or >=.

I pulled the original check from the rocky linux policy: "The operating system must generate audit records for successful/unsuccessful uses of the chcon command." 

The config they suggest is:

```
-a always,exit -F path=/usr/bin/chcon -F perm=x -F auid>=${UID_MIN} -F auid!=unset -k perm_chng
```

That means we are looking for a rule on chcon, with perm=x, whose auid >= MIN. The `-k` should be optional afaik (it's just for search).

### 1: explicit, expanded

Ok If I were to translate that ideally, let's assume the explicit form first:

```coffee
rules.contains(rule:
  rule.action == "always",
  rule.list == "exit",
  rule.fields.contains( key == "path" && operation == "=" && value == "/usr/bin/chcon" ) &&
  rule.fields.contains( key == "perm" && operation == "=" && value == "x" ) &&
  rule.fields.contains( key == "auid" && operation == ">=" && value == UID_MIN ) &&
  rule.fields.contains( key == "auid" && operation == "!=" && value == "unset" )
)
```

This will work. It's just very verbose. But can work if I read the fields into `{key, operation, value}`.

mm-weber:
@arlimus Thanks for cutting to the meat of the issue.

The solution 1) looks like the one we should pick. I do not see the verbosity as a drawback when it comes to such complex rules. Maybe one more suggestion on top of this, since you are already kinda  proposing it with your example. How about this?

```
rules.contains(rule:
  rule.action == "always",
  rule.list == "exit",
  rule.field( key == "path" && operation == "=" && value == "/usr/bin/chcon" ),
  rule.field( key == "perm" && operation == "=" && value == "x" ),
  rule.field( key == "auid" && operation == ">=" && value == UID_MIN ),
  rule.field( key == "auid" && operation == "!=" && value.in(["unset",-1,4294967295]) )
)
```

Can we allow for 1 or more rule.field entries and allow for that syntax?

The issue with rules being altered downstream seem out-of-scope for building the resource to me.
If you look at this example:

```
# 1. Delete all rules
-D

# 2. Set buffer
-b 8192

# 3. Audit execution of chcon by non-system users
-a always,exit -F path=/usr/bin/chcon -F perm=x -F auid>=${UID_MIN} -F auid!=unset -k perm_chng

# 4. BUT, never audit chcon execution from certain location
-a never,exit -F path=/usr/bin/chcon -F dir=/tmp -F perm=x -k ignore_tmp_chcon

# 5. Audit failed attempts to use chcon on specific sensitive file
-a always,exit -F path=/usr/bin/chcon -F success=0 -F obj=/etc/secret.txt -k secret_chcon_fail

# 6. Enable auditing
-e 1
```

You can create extra checks for that in MQL if necessary.
That being said, you certainly CAN completely overwrite rules downstream:
```
-a always,exit -F path=/usr/bin/chcon -F perm=x -F auid>=${UID_MIN} -F auid!=unset -k perm_chng

-a never,exit -F path=/usr/bin/chcon -F perm=x -F auid>=${UID_MIN} -F auid!=unset -k perm_chng
```

arlimus:
Updated this PR and the example in the first comment to include the parsed syscall fields. These allow testing on individual field values now.

Overall flow testing of these rules will follow later on

providers-sdk/v1/testutils/testdata/arch.json
```diff
@@ -155,15 +157,15 @@
         },
         {
           "Resource": "file",
-          "ID": "/dummy.xml",
+          "ID": "/dummy.null.json",
```
arlimus, line 160: note: the automatic ingestion of new resources fixed the sorting of these file resources, i.e. this is not changing it from dummy.xml to dummy.null.json but merely changing the order in which these resources appear in the file

chris-rock (approved):
Thank you @arlimus

PR #5550, opened 2025-05-19
⭐ Add asset discovery for aws-eks-cluster assets

tas50:
Expose these as an asset so we can improve check output


vjeffrey (approved):
tested locally, working great!

PR #5554, opened 2025-05-19
⭐️ enable cloud resource for EBS volume scanning

afiune:
## 💁‍♂️ Problem/Context
We recently shipped cloud and cloud IP detection in Mondoo for agent-based scans on cloud hosts. This feature has been very useful in properly presenting hosts as running in the cloud and has also enabled us to expose internet connected hosts via a new risk factor.

Unfortunately if the user is scanning their infrastructure using the EBS snapshot scanning capabilities in the AWS Serverless integration, they will not get this data back. While we cannot fetch the metadata endpoint during the EBS volume scan we can populate much of the cloud resource data including:

* Asset.kind is virtualmachine
* Provider (obviously this is going to be AWS)
* Local IP / Public IP (via further inspection of the AWS instance via API calls)

## 📖 User Story
As a user I want to understand the configuration data of all the assets in my business, so that I can better respond to Mondoo findings.


providers/os/id/awsebs/windows_ebs.go
```diff
@@ -0,0 +1,105 @@
...
+	} else {
+		// we are looking for something similar to:
+		//
+		// Starting services...
+		// Hostname: WIN-1234ABCD
+		// Configuration done.
+		//
+		launchLogs, err := afero.ReadFile(m.conn.FileSystem(), `\ProgramData\Amazon\EC2Launch\log\agent.log`)
+		if err == nil {
+			if match := winHostnameRE.FindString(string(launchLogs)); match != "" {
```
glower: I think if you want to capture group you should use `FindStringSubmatch`

providers/os/id/aws/aws.go
```diff
@@ -24,9 +25,24 @@ func readValue(conn shared.Connection, fPath string) string {
 	return string(content)
 }
 
-func Detect(conn shared.Connection, p *inventory.Platform, smbiosMgr smbios.SmBiosManager) (string, string, []string) {
+func Detect(conn shared.Connection, p *inventory.Platform) (string, string, []string) {
 	var values []string
-	if p.IsFamily(inventory.FAMILY_LINUX) {
+	if conn.Type() == shared.Type_FileSystem {
```
preslavgerchev, line 30: you should double check if thats the type we give to the device connection too

providers/os/id/awsebs/unix_ebs.go
```diff
@@ -0,0 +1,148 @@
...
+
+	cloudInitLog, err := afero.ReadFile(m.conn.FileSystem(), "/var/log/cloud-init.log")
+	if err != nil {
+		log.Debug().Err(err).Msg("unable to read cloud-init log")
+	} else {
+		// Detect public IP from the cloud-init.log file - we are searching for something similar to this line:
+		//
+		// 'http://169.254.169.254:80/2021-03-23/meta-data/network/interfaces/macs/0a:ff:ed:34:f3:6d/ipv4-associations/54.146.163.122'
+		//
+		ipv4PublicRE := regexp.MustCompile(`meta-data/network/interfaces/macs/([0-9a-f:]+)/ipv4-associations/([0-9.]+)`)
```
preslavgerchev, line 70: do we need this for snapshots? this information might be outdated/stale by the time we're reading the snapshot

preslavgerchev: we're also relying on parsing log files, which makes it prone to breaking

providers/os/id/awsebs/awsebs.go
```diff
@@ -0,0 +1,276 @@
...
+		connections := asset.GetConnections()
+		index := slices.IndexFunc(connections, func(c *inventory.Config) bool {
+			if c != nil && c.Type == shared.Type_Device.String() {
+				return true
+			}
+			return false
+		})
+		// @afiune we can't use `device.PlatformIdInject` because of a cyclic dep
+		// TODO move it to a shared package
+		platformInjected, ok := connections[index].Options["inject-platform-ids"]
```
preslavgerchev, line 158: connections[index] will blow up here if it's -1 (not found)

preslavgerchev: also it feels a bit weird to have the flag reading here as that's connection specific

i think we want to move away from this and just inject asset.PlatformIds directly instead of this option

providers/os/id/aws/aws.go
afiune: At first, I thought the type was `shared.Type_Device` but after lots of debugging (difficult because you have to test on a "runner") I figured out this is the type we land on.

I think it is because we nest two types inside the device connection struct https://github.com/mondoohq/cnquery/blob/main/providers/os/connection/device/device_connection.go#L34-L35 
 and we have config and asset there and in the nested fs struct https://github.com/mondoohq/cnquery/blob/main/providers/os/connection/fs/filesystem.go#L55-L58

Somehow we endup having the nested struct instead of the main device connection struct.

providers/os/id/awsebs/unix_ebs.go
afiune: RE snapshot. Maybe 🤷🏽  - @tas50 can you chime in?

RE outdated. Yes, but there is no other way I can find to detect this information, though remember this is a fallback and is "best effort" kind of collection. Open to not do this cc @tas50

providers/os/id/awsebs/awsebs.go
afiune: @preslavgerchev I'd love to hear you idea about injecting the platform IDs directly, I'll ping you to chat about it.

providers/os/id/awsebs/unix_ebs.go
tas50: We take the snapshot ourselves so by the time we scan it there should be pretty minimal delay right?

providers/os/id/awsebs/awsebs.go
afiune: Preslav and I synced and we decided to do this cleanup in a different change. That is, remove the `inject-platform-ids` from both, `cnquery` and `serverless` projects.

preslavgerchev (approved):

PR #5558, opened 2025-05-20
🐛 Prevent trailing space in ParseProcessCmdline output

glower:
The ParseProcessCmdline function was producing output with an
erroneous trailing space when processing /proc/[pid]/cmdline files.
This is because the cmdline format uses null bytes as argument
separators and includes a final null terminator for the entire
argument list.


afiune (approved):
After CI is green. Which is broken in `main` atm.

PR #5561, opened 2025-05-20
Add `machine.baseboard` support for AIX

jaym:
```
cnquery> machine.baseboard { * }
machine.baseboard: {
  version: "IBM,FW950.D0 (VL950_175)"
  manufacturer: "IBM"
  serial: "7835450"
  assetTag: ""
  product: "IBM,9009-22A"
}
```

Fixes https://github.com/mondoohq/cnquery/issues/5542

### DIFF
--- a/providers/os/resources/smbios/aix.go
+++ b/providers/os/resources/smbios/aix.go
@@ -1,0 +1,87 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package smbios
+
+import (
+	"bufio"
+	"errors"
+	"io"
+	"strings"
+
+	"go.mondoo.com/cnquery/v11/providers/os/connection/shared"
+)
+
+// AIXSmbiosManager gets system information from AIX systems.
+// smbios is not a thing on AIX, but we implement this interface
+type AIXSmbiosManager struct {
+	provider shared.Connection
+}
+
+func (s *AIXSmbiosManager) Name() string {
+	return "AIX Smbios Manager"
+}
+
+func (s *AIXSmbiosManager) Info() (*SmBiosInfo, error) {
+
+	cmd, err := s.provider.RunCommand("prtconf")
+	if err != nil {
+		return nil, err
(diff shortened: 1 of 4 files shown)


syrull:
Looks good, tests are failing however?

providers/os/resources/smbios/aix.go
```diff
@@ -0,0 +1,79 @@
...
+		line := scanner.Text()
+		// Process the line to extract relevant information
+		if strings.Contains(line, "System Model") {
+			baseBoardInfo.Model = strings.TrimSpace(strings.Split(line, ":")[1])
+			sysInfo.Model = baseBoardInfo.Model
+		} else if strings.Contains(line, "Machine Serial Number") {
+			baseBoardInfo.SerialNumber = strings.TrimSpace(strings.Split(line, ":")[1])
+			sysInfo.SerialNumber = baseBoardInfo.SerialNumber
+		} else if strings.Contains(line, "Firmware Version") {
+			baseBoardInfo.Version = strings.TrimSpace(strings.Split(line, ":")[1])
```
afiune: All of these splits assume that there will always be a `:`, which most likely will be there but if not, we will panic.

Could we create a simple helper function that does this splitting and trimming safely?

providers/os/resources/smbios/aix.go
```diff
@@ -0,0 +1,79 @@
...
+			sysInfo.Version = baseBoardInfo.Version
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		return baseBoardInfo, sysInfo, err
+	}
+	return baseBoardInfo, sysInfo, nil
```
afiune, line 86: nit
```suggestion
	return baseBoardInfo, sysInfo, scanner.Err()
```

afiune (approved):
A couple of comments.

providers/os/resources/smbios/aix.go
jaym: i prefer not to do things this way. it makes using a debugger more difficult

PR #5565, opened 2025-05-20
🧹 update tests

chris-rock:


explorer/scan/discovery_test.go
```diff
@@ -151,7 +151,7 @@ func TestDiscoverAssets(t *testing.T) {
 		inv := getInventory()
 		discoveredAssets, err := DiscoverAssets(context.Background(), inv, nil, recording.Null{})
 		require.NoError(t, err)
-		assert.Len(t, discoveredAssets.Assets, 3)
+		assert.Len(t, discoveredAssets.Assets, 4)
```
chris-rock, line 154: I am not 100% clear why this changed.

imilchev: I think this happened because of my change that merged here https://github.com/mondoohq/cnquery/commit/15161230099466995a27fe0a9d0c434710ca07b9

We now discover k8s namespaces always. I guess it took effect yesterday because yesterday the providers go released. This test relies on the released providers and doesn't use the builtin.

imilchev (approved):

PR #5566, opened 2025-05-21
🧹 Remove version pin from docker dep

czunker:
The new release of https://github.com/google/go-containerregistry/releases/tag/v0.20.4 allows us to remove the pin


providers/os/resources/auditpol.go
```diff
@@ -22,7 +22,7 @@ func (p *mqlAuditpol) list() ([]interface{}, error) {
 	cmd := o.(*mqlPowershell)
 	out := cmd.GetStdout()
 	if out.Error != nil {
-		return nil, fmt.Errorf("could not run auditpol: " + out.Error.Error())
+		return nil, fmt.Errorf("could not run auditpol: %s", out.Error.Error())
```
glower: nit: `fmt.Errorf("could not run auditpol: %w", out.Error)`

providers/os/resources/secpol.go
```diff
@@ -33,7 +33,7 @@ func (s *mqlSecpol) policy() (*windows.Secpol, error) {
 	cmd := o.(*mqlCommand)
 	out := cmd.GetStdout()
 	if out.Error != nil {
-		return nil, fmt.Errorf("could not run auditpol: " + out.Error.Error())
+		return nil, fmt.Errorf("could not run auditpol: %s", out.Error.Error())
```
glower: same nit here :)

providers/os/resources/auditpol.go
czunker: Nice, thanks. I changed it.

providers/os/resources/secpol.go
czunker: Nice, thanks. I changed it.

PR #5568, opened 2025-05-21
🐛 detect OpenSSH 9.9 and force Key Exchange Algorithms

afiune:
OpenSSH 9.9 introduces hybrid post-quantum algorithms, and when users try to connect via SSH, they get a key signature compatibility issue, most likely due to the go package https://pkg.go.dev/golang.org/x/crypto/ssh not supporting it yet or misinterpreting it.

Note that, there is also a reported bug that recommends upgrading to OpenSSH 9.9p2 https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1103392 so we recommend users to upgrade their servers to that version.

Regardless, this change introduces an extra step during an SSH connection that detects if the server's advertised key exchange methods include unsupported hybrid ones, and if so, we will force the key exchange algorithm to a compatible one.


jaym:
i tested this and it does work, even when openssh client does not on my computer

providers/os/connection/ssh/ssh.go
```diff
@@ -16,6 +15,7 @@ import (
 	"time"
 
 	awsconf "github.com/aws/aws-sdk-go-v2/config"
+	"github.com/cockroachdb/errors"
```
jaym, line 18: probably should just use "errors"

jaym (approved):

PR #5577, opened 2025-05-27
Configure Coordinator from ENV in init(), prepare for explicit init

glower:
Modifies Coordinator initialization to read auto-update settings
from MONDOO_ prefixed environment variables directly within the package
init() for immediate effect. Adds mutex for safe future explicit
re-initialization by host applications.

### DIFF
--- a/providers/coordinator.go
+++ b/providers/coordinator.go
@@ -47,7 +47,7 @@ var BuiltinCoreID = coreconf.Config.ID
 var Coordinator ProvidersCoordinator
 
 func newCoordinator(globalAutoUpdateCfg UpdateProvidersConfig) *coordinator {
-	log.Info().
+	log.Debug().
 		Bool("global auto-update is enabled", globalAutoUpdateCfg.Enabled).
 		Msg("providers.newCoordinator() is called")
 
@@ -249,7 +249,7 @@ func (c *coordinator) RemoveRuntime(runtime *Runtime) {
 }
 
 func (c *coordinator) GetRunningProvider(id string, update UpdateProvidersConfig) (*RunningProvider, error) {
-	log.Info().
+	log.Debug().
 		Str("id", id).
 		Bool("update is enabled", update.Enabled).
 		Msg("coordinator.GetRunningProvider() is called")
@@ -270,10 +270,6 @@ func (c *coordinator) GetRunningProvider(id string, update UpdateProvidersConfig
 // unsafeStartProvider will start a provider and add it to the list of running providers. Must be called
 // with a mutex lock around it.
 func (c *coordinator) unsafeStartProvider(id string, update UpdateProvidersConfig) (*RunningProvider, error) {
-	log.Info().
-		Str("id", id).
-		Bool("update is enabled", update.Enabled).Msg(">>> unsafeStartProvider()")
-
 	if x, ok := builtinProviders[id]; ok {
 		// We don't warn for core providers, which are the only providers
 		// built into the binary (for now).
--- a/providers/runtime.go
+++ b/providers/runtime.go
@@ -141,7 +141,7 @@ func (r *Runtime) setProviderConnection(c *plugin.ConnectRes, err error) {
 }
 
 func (r *Runtime) addProvider(id string) (*ConnectedProvider, error) {
-	log.Info().
+	log.Debug().
 		Str("id", id).
 		Bool("update is enabled", r.AutoUpdate.Enabled).
 		Msg("Runtime.addProvider() is called")


providers/coordinator.go
```diff
@@ -296,14 +296,23 @@ func (c *coordinator) unsafeStartProvider(id string, update UpdateProvidersConfi
...
 		// things are still functional. We want to consider failure, possibly
 		// with a config entry in the future.
-		updated, err := TryProviderUpdate(provider, update)
-		if err != nil {
-			log.Error().
-				Err(err).
+
+		// Only skip k8s provider updates when running in Kubernetes (docker container)
+		// unless explicitly enabled via MONDOO_AUTO_UPDATE=true in operator code
+		if provider.Name == "k8s" && os.Getenv("MONDOO_AUTO_UPDATE") == "false" {
```
preslavgerchev: not fully familiar with how the k8s scans are ran but there's a flag to disable auto updates when running a scan, i think it's `--auto-update=false`.

maybe that's a better solution instead of putting k8s-specific logic in here

imilchev: I think this code is not needed at all. There is already `if update.Enabled` condition on line 295. The problem is that that value is incorrect when the flag is set. We don't need another condition here but need to make sure the original one works as expected

glower: This is probably always `true` because we introduced this change here: https://github.com/mondoohq/cnquery/pull/5510/files 
```		
AutoUpdate: UpdateProvidersConfig{
   Enabled: true,
},
```

glower: > not fully familiar with how the k8s scans are ran but there's a flag to disable auto updates when running a scan, i think it's `--auto-update=false`.

setting `MONDOO_AUTO_UPDATE`=false should behave exactly in the same way as `--auto-update=false`, it is done somewhere in our Viper setup.

> maybe that's a better solution instead of putting k8s-specific logic in here

But it's kinda is only the problem in k8s. I can test if adding `--auto-update=false` to the command in the container will help or not

imilchev: @glower you are correct. The CLI flag and the env var are binding to the same setting. Doesn't matter which one you use, it will behave the same way.

I think that code that hard-codes the update to `true` is the issue. It should always read from the setting. If you implement this, then we should be good to go. I think it looks like it's just a k8s problem but it isn't... perhaps it's just easiest to trigger that case for k8s. It really depends when a new runtime gets created and picks up that hardcoded flag. My proposal is to just make sure that we always use the setting and we don't have hard-coded values for this, then it should all work correctly

providers/coordinator.go
```diff
@@ -103,13 +105,16 @@ func (c *coordinator) NewRuntime() *Runtime {
 }
 
 func (c *coordinator) newRuntime() *Runtime {
+	config.InitViperConfig()
```
imilchev: I think this shouldn't depend on the cli package. You would somehow need to pipe the `autoUpdate` value from outside the coordinator. I think the best would probably be to have the autoupdate settings defined on the coordinator itself since that setting is defined once per cnquery/cnspec instance. 
You would need to figure out from where to set the value though

glower: For now, I've made it so the Coordinator immediately respects MONDOO_AUTO_UPDATE  environment variables when `cnquery` initializes (even inside `cnspec`). So, it will work inside of k8s provider running the the docker.

The main goal here is to pave the way for a new InitGlobalCoordinator(config) function. I will make a change so `cnspec`  will call this directly at startup, passing whatever auto-update settings they've decided on from their own flags or configs. 

The current mutex bits are just to ensure this future change in `cnspec` can safely override the temporary ENV-based setup without issues.

This keeps`cnspec` working as is while letting us use ENVs for control immediately. Next step: update cnspec to use the new function.

providers/providers.go
```diff
@@ -61,9 +65,37 @@ func init() {
 
 	LastProviderInstall = time.Now().Unix()
 
-	// Initialize the global coordinator instance
-	coordinator := newCoordinator()
-	Coordinator = coordinator
+	initMutex.Lock()
+	defer initMutex.Unlock()
```
afiune: Since this code is inside the `init()` function, I don't think we need a mutex.

providers/providers.go
```diff
@@ -61,9 +65,37 @@ func init() {
...
+
+		legacyCoordinatorCfg := UpdateProvidersConfig{
+			Enabled:         autoUpdateEnabled,
+			RefreshInterval: 60 * 60,
+		}
+		Coordinator = newCoordinator(legacyCoordinatorCfg)
+	}
+}
+
+func InitGlobalCoordinator(cfg UpdateProvidersConfig) {
```
afiune: what is this for?

providers/runtime.go
```diff
@@ -56,6 +56,12 @@ func (c *coordinator) RuntimeWithShutdownTimeout(timeout time.Duration) *Runtime
 	return runtime
 }
 
+func (c *coordinator) RuntimeWithUpdateProvidersConfig(update UpdateProvidersConfig) *Runtime {
```
afiune: Not sure where will this be use. But if we need it, let's add a test.

I also feel that instead of creating multiple functions to initialize a runtime like `RuntimeWith...()` we should come up with options to pass to a single init function. Something like `Runtime(opts ...RuntimeFunc)`.

providers/providers.go
glower: i wanted to remove the config part for the auto reload from init and use InitGlobalCoordinator in the cnspec and pass the confing from there.

afiune: The global `Coordinator` should act as a singleton. We initialize it once and use it as the default coordinator. If callers (like `cnspec`) doesn't want the default, they should be able to create their own coordinator, but with the modifications you are doing in the init function I think cnspec will use the default without issues.

I would recommend removing this function, the mutex, and the `coordinatorInitializedByHost` flag.

providers/providers.go
```diff
@@ -61,9 +65,37 @@ func init() {
...
+	defer initMutex.Unlock()
+
+	if !coordinatorInitializedByHost { // Check if host already initialized
+		log.Warn().Msg("DEPRECATED: cnquery.providers.Coordinator is being implicitly initialized. " +
+			"The host application should call InitGlobalCoordinator for proper configuration.")
+
+		viper.SetEnvPrefix("mondoo")
+		viper.AutomaticEnv()
+
+		autoUpdateEnabled := viper.GetBool("auto_update")
```
afiune: We should probably do this instead:
```suggestion
		autoUpdateEnabled := true
		config.InitViperConfig()
		if viper.IsSet("auto_update") {
			autoUpdateEnabled = viper.GetBool("auto_update")
		}
```

Just like we do here https://github.com/mondoohq/cnquery/blob/main/cli/providers/providers.go#L57-L62

providers/providers.go
```diff
@@ -61,9 +65,37 @@ func init() {
...
+
+		log.Info().
+			Bool("autoUpdateEnabled", autoUpdateEnabled).
+			Msg("cnquery.providers.init(): Implicitly initializing global Coordinator from ENV (MONDOO_AUTO_UPDATE).")
+
+		legacyCoordinatorCfg := UpdateProvidersConfig{
```
afiune: out of curiosity: why is this called legacy?

providers/providers.go
glower: I think it's more or less the same, but let me change it for consistency reasons

providers/providers.go
glower: I wanted to remove this part from the init function and replace it with a explicit call (and mutex), 
but you convinced me that init() is actually doing exactly that

providers/runtime.go
glower: I forgot to remove this part, I did several iterations of the fix

providers/coordinator.go
```diff
@@ -46,11 +46,16 @@ var BuiltinCoreID = coreconf.Config.ID
 
 var Coordinator ProvidersCoordinator
 
-func newCoordinator() *coordinator {
+func newCoordinator(globalAutoUpdateCfg UpdateProvidersConfig) *coordinator {
+	log.Info().
```
afiune: Info level logs are always displayed to the user. Can we switch these new log messages to Debug level?

afiune (approved):
Code looks good. Just need to update the log level. Thanks @glower !

providers/coordinator.go
glower: done

providers/providers.go
```diff
@@ -61,9 +62,24 @@ func init() {
...
+	viper.SetEnvPrefix("mondoo")
+	viper.AutomaticEnv()
+
+	autoUpdateEnabled := true
+	config.InitViperConfig()
+	if viper.IsSet("auto_update") {
+		autoUpdateEnabled = viper.GetBool("auto_update")
+	}
+
+	log.Info().
```
chris-rock, line 74: This should not be Info()

providers/providers.go
```diff
@@ -61,9 +62,24 @@ func init() {
...
 	LastProviderInstall = time.Now().Unix()
 
-	// Initialize the global coordinator instance
-	coordinator := newCoordinator()
-	Coordinator = coordinator
+	viper.SetEnvPrefix("mondoo")
+	viper.AutomaticEnv()
+
+	autoUpdateEnabled := true
+	config.InitViperConfig()
```
chris-rock, line 69: viper should not be even imported here. This should not be used outside of the cmd package

PR #5590, opened 2025-05-28
✨ aws codedeploy resource

afiune:
Screenshot 2025-05-28 at 3 07 11 PM


providers/aws/resources/aws_codedeploy.go
```diff
@@ -0,0 +1,508 @@
...
+			batchGetInput := &codedeploy.BatchGetDeploymentsInput{DeploymentIds: listOutput.Deployments}
+			batchGetOutput, err := svc.BatchGetDeployments(ctx, batchGetInput)
+			if err != nil {
+				return nil, errors.Wrapf(err, "could not batch get deployment details")
+			}
+
+			for _, depInfo := range batchGetOutput.DeploymentsInfo {
+				// Construct a synthetic ARN or use deploymentId for uniqueness
+				syntheticArn := fmt.Sprintf(
+					"arn:aws:codedeploy:%s:%s:deployment/%s/%s/%s",
```
vjeffrey, line 407: this doesnt seem to match the arn references for code deploy in https://docs.aws.amazon.com/codedeploy/latest/userguide/security_iam_service-with-iam.html ?

providers/aws/resources/aws.lr
```diff
@@ -3648,3 +3648,113 @@ private aws.timestream.liveanalytics.table @defaults("name region") {
...
+  // ARN of the application
+  arn string
+  // ID of the application
+  applicationId string
+  // Name of the application
+  applicationName string
+  // Compute platform of the application (Server, Lambda, or ECS)
+  computePlatform string
+  // Time the application was created
+  createDate time
```
tas50: createdAt is what we're trying to standardize on for AWS

providers/aws/resources/aws_codedeploy.go
afiune: There is no ARN for deployments. That's why it is synthetic. Happy to remove it if it is confusing.

vjeffrey: ah ok, makes sense

providers/aws/resources/aws.lr.manifest.yaml
```diff
@@ -649,6 +649,105 @@ resources:
...
+    platform:
+      name:
+      - aws
+  aws.codedeploy.application:
+    fields:
+      applicationId: {}
+      applicationName: {}
+      arn: {}
+      computePlatform: {}
+      createDate: {}
```
```suggestion
```

vjeffrey (approved):

PR #5604, opened 2025-05-30
🧹 avoid reporting panics from dev environments

afiune:
When we are doing development, we may cause a panic on purpose, or by mistake.
On those cases, we don't want to report back to the platform.

Only production releases will.

### DIFF
--- a/.github/workflows/test-report.yml
+++ b/.github/workflows/test-report.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Download and Extract Artifacts
-      uses: dawidd6/action-download-artifact@v9
+      uses: dawidd6/action-download-artifact@v10
       with:
         run_id: ${{ github.event.workflow_run.id }}
         path: artifacts
--- a/providers/ansible/config/config.go
+++ b/providers/ansible/config/config.go
@@ -12,7 +12,7 @@ import (
 var Config = plugin.Provider{
 	Name:            "ansible",
 	ID:              "go.mondoo.com/cnquery/v11/providers/ansible",
-	Version:         "11.0.68",
+	Version:         "11.0.71",
 	ConnectionTypes: []string{provider.DefaultConnectionType},
 	Connectors: []plugin.Connector{
 		{
--- a/providers/arista/config/config.go
+++ b/providers/arista/config/config.go
@@ -12,7 +12,7 @@ import (
 var Config = plugin.Provider{
 	Name:            "arista",
 	ID:              "go.mondoo.com/cnquery/v9/providers/arista",
-	Version:         "11.0.81",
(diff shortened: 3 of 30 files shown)


providers-sdk/v1/upstream/health/errors.go
```diff
@@ -20,6 +20,10 @@ import (
 type PanicReportFn func(product, version, build string, r any, stacktrace []byte)
 
 func ReportPanic(product, version, build string, reporters ...PanicReportFn) {
+	if build != "dev" {
```
vjeffrey: should this be build == "dev" ?

afiune: Yup. I moved it and forgot to change the logic. 🤦

afiune:
```suggestion
	if build == "dev" {
```

Makefile
```diff
@@ -10,10 +10,11 @@ ifndef MANIFEST_VERSION
...
 endif
 
-ifndef TAG
-# echo "read TAG from git"
-TAG=$(shell git log --pretty=format:'%h' -n 1)
-endif
+# The build in the Makefile is hardcoded to be "dev" for development
+# purposes like avoid sending alerts from our development environemnts.
+#
+# The real build tag is being set by our release pipelines.
```
czunker: I'm not sure what you are referring to.
Looking at the latest release, there is no TAG or BUILD set: https://github.com/mondoohq/cnquery/actions/runs/15436676015/job/43444557238

Makefile
```diff
@@ -33,8 +34,8 @@ ifeq ($(TARGETOS),windows)
 	BIN_SUFFIX=".exe"
 endif
 
-LDFLAGS=-ldflags "-s -w -X go.mondoo.com/cnquery/v11.Version=${VERSION} -X go.mondoo.com/cnquery/v11.Build=${TAG}" # -linkmode external -extldflags=-static
-LDFLAGSDIST=-tags production -ldflags "-s -w -X go.mondoo.com/cnquery/v11.Version=${LATEST_VERSION_TAG} -X go.mondoo.com/cnquery/v11.Build=${TAG} -s -w"
+LDFLAGS=-ldflags "-s -w -X go.mondoo.com/cnquery/v11.Version=${VERSION} -X go.mondoo.com/cnquery/v11.Build=${BUILD}" # -linkmode external -extldflags=-static
```
czunker: When you change the env var here, are you sure this still works for an actual release?

afiune: Yes, I'm sure. `Build` is getting set directly here https://github.com/mondoohq/cnquery/blob/main/.goreleaser.yml#L45

czunker: Ah, thank you. I missed this one.

czunker (approved):
Thanks @afiune

PR #5605, opened 2025-05-31
Add city/country fields to Shodan hosts

tas50:
It's one of the more interesting features

### DIFF
--- a/providers/shodan/resources/host.go
+++ b/providers/shodan/resources/host.go
@@ -6,8 +6,9 @@ package resources
 import (
 	"context"
 	"errors"
-	"go.mondoo.com/cnquery/v11/llx"
 	"strings"
+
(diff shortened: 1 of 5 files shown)


providers/shodan/resources/host.go
```diff
@@ -6,9 +6,10 @@ package resources
 import (
 	"context"
 	"errors"
-	"go.mondoo.com/cnquery/v11/llx"
 	"strings"
 
+	"go.mondoo.com/cnquery/v11/llx"
+
```
chris-rock: nit: empty newline

chris-rock (approved):
Thank you @tas50

providers/shodan/resources/host.go
tas50: Fixed

PR #5636, opened 2025-06-05
⭐️ sts: allow passing JWT Token

afiune:
### Summary

Currently, our WIF integration does not support supplying a JSON Web Token (JWT) as part of the credential exchange process. This change adds support for using a JWT as a credential source during the WIF authentication flow. It allows users to pass in a JWT token via environment variable or configuration file during the WIF exchange process.

### Motivation

In modern cloud-native environments, particularly when integrating with cloud providers (e.g., AWS, GCP, Azure), it is common to perform secure federated identity exchanges using JWTs issued by an OIDC-compliant identity provider.

Supporting JWT input will:
- Improve compatibility with existing Identity Providers (IdPs)
- Enable seamless authentication for CI/CD workflows and short-lived credentials
- Reduce reliance on service account keys and static credentials

#### Example Usage (Env + Config)

Use new environment variable:
```env
JWT_TOKEN=<your-jwt-token>
```

Plus the WIF configuration (already existing) similar to:
```json
{
    "universeDomain": "https://api.edge.mondoo.com",
    "scopes": [
        "//iam.api.mondoo.app/roles/agent"
    ],
    "type": "external_account",
    "audience": "//captain.api.mondoo.app/spaces/epic-mclaren-654267",
    "subjectTokenType": "urn:ietf:params:oauth:token-type:jwt",
    "issuerUri": "https://accounts.google.com"
}
```

To test this you have to have a workload identity provider, in this case, for Google.

https://mondoo.com/docs/platform/maintain/access/non-human/wif/#which-wif-providers-does-mondoo-support

#### Example only config

If users are generating the WIF configuration on the fly, they can already inject the JWT token in it
rather than using the environment variable. This example shows how to generate a token and build
the configuration all at once.

```bash
echo '{
    "universeDomain": "https://api.edge.mondoo.com",
    "scopes": [
        "//iam.api.mondoo.app/roles/agent"
    ],
    "type": "external_account",
    "audience": "//captain.api.mondoo.app/spaces/epic-mclaren-654267",
    "subjectTokenType": "urn:ietf:params:oauth:token-type:jwt",
    "issuerUri": "https://accounts.google.com",
    "jwtToken": "'$(gcloud auth print-identity-token)'"
}' | jq . > /tmp/wif.json
```
**Note** that `jq` is used only for formatting purposes.

### Acceptance Criteria

* JWT token can be passed via environment variable or config
* Token is used in the WIF exchange flow in place of static credentials
~* Validation errors are raised if the JWT is malformed or expired~

### Additional Context

Adding support for JWT aligns with best practices recommended by cloud providers and will help us maintain a secure, scalable identity federation architecture.

### DIFF
--- a/providers-sdk/v1/upstream/sts.go
+++ b/providers-sdk/v1/upstream/sts.go
@@ -60,11 +60,23 @@ func ExchangeSSHKey(apiEndpoint string, identityMrn string, resourceMrn string) 
 	}, nil
 }
 
-func ExchangeExternalToken(apiEndpoint string, audience string, issuerURI string) (*ServiceAccountCredentials, error) {
-	// Fetch the identity token from the cloud provider
-	jsonToken, err := fetchIdentityToken(audience)
-	if err != nil {
-		return nil, err
+func ExchangeExternalToken(apiEndpoint, audience, issuerURI, jwtToken string) (*ServiceAccountCredentials, error) {
+	if jwtToken == "" {
+		// Try to fetch the token from an environment variable
+		jwtToken = os.Getenv("JWT_TOKEN")
+	}
+
+	if jwtToken == "" {
+		// Try to fetch the identity token from the cloud provider
+		jsonToken, err := fetchIdentityToken(audience)
+		if err != nil {
+			return nil, err
+		}
+		jwtToken = jsonToken
+	}
+
+	if jwtToken == "" {
+		return nil, fmt.Errorf("no identity token to use for an external exchange")
 	}
 
 	stsClient, err := NewSecureTokenServiceClient(apiEndpoint, ranger.DefaultHttpClient())
@@ -75,7 +87,7 @@ func ExchangeExternalToken(apiEndpoint string, audience string, issuerURI string
 	request := &ExchangeExternalTokenRequest{
 		Audience:  audience,
 		IssuerUri: issuerURI,
-		JwtToken:  jsonToken,
+		JwtToken:  jwtToken,
 	}
 	resp, err := stsClient.ExchangeExternalToken(context.Background(), request)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


providers-sdk/v1/upstream/sts.go
```diff
@@ -60,11 +60,23 @@ func ExchangeSSHKey(apiEndpoint string, identityMrn string, resourceMrn string)
...
 
-func ExchangeExternalToken(apiEndpoint string, audience string, issuerURI string) (*ServiceAccountCredentials, error) {
-	// Fetch the identity token from the cloud provider
-	jsonToken, err := fetchIdentityToken(audience)
-	if err != nil {
-		return nil, err
+func ExchangeExternalToken(apiEndpoint, audience, issuerURI, jwtToken string) (*ServiceAccountCredentials, error) {
+	if jwtToken == "" {
+		// Try to fetch the token from an environment variable
+		jwtToken = os.Getenv("JWT_TOKEN")
```
czunker, line 66: Our other env vars are prefixed with `MONDOO_`. Shouldn't we also use the pattern here?

czunker (commented):
The PR description mentions:
> Validation errors are raised if the JWT is malformed or expired
Where does this happen? Does the stsClient raise these errors?

czunker:
I gave this a try, but I get an error:
```
→ using wif authentication method, generate temporary credentials
x could not exchange external (wif) token error="failed to fetch identity token from any supported cloud provider"
```
I adjusted the above example wif config to another space and stored the gcloud output into the env var.
Am I missing something in the space config?  Do we have something to test this with?

afiune:
@czunker You do have to configure a workload identity provider, in this case, for Google.

https://mondoo.com/docs/platform/maintain/access/non-human/wif/#which-wif-providers-does-mondoo-support

afiune:
@czunker the validation is not in the client but in the server, I forgot to remove that acceptance criteria from the original GH issue that describes the improvement.

Good catch!

PR #5648, opened 2025-06-10
⭐ Extend microsoft.users to contain the data for assignedLicenses

glower:
This should work like this:

```
microsoft.users {assignedLicenses}
microsoft.users.list: [
  0: {
    assignedLicenses: [
      0: microsoft.user.assignedLicense skuId="c42b9cae-ea4f-4ab7-9717-81576235ccac"
    ]
  }
  1: {
    assignedLicenses: []
  }
  2: {
    assignedLicenses: [
      0: microsoft.user.assignedLicense skuId="c42b9cae-ea4f-4ab7-9717-81576235ccac"
    ]
  }
```

### DIFF
--- a/providers/ms365/resources/ms365.lr
+++ b/providers/ms365/resources/ms365.lr
@@ -81,6 +81,14 @@ microsoft.users {
   search string
 }
 
+// This type represents a single license assigned to a user.
+private microsoft.user.assignedLicense @defaults("skuId") {
+  // A collection of the unique identifiers for plans that have been disabled.
+  disabledPlans []string
+  // The unique identifier for the SKU.
+  skuId string
+}
+
 // Microsoft Conditional Access Policies
 microsoft.conditionalAccess {
   // Named locations container
@@ -367,6 +375,8 @@ private microsoft.user @defaults("id displayName userPrincipalName") {
   identities []microsoft.user.identity
   // The user's audit log
   auditlog() microsoft.user.auditlog
+  // The licenses that are assigned to the user, including inherited (group-based) licenses
+  assignedLicenses []microsoft.user.assignedLicense
 }
 
 // Microsoft user audit log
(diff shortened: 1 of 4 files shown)


providers/ms365/resources/ms365.lr
```diff
@@ -81,6 +81,16 @@ microsoft.users {
   search string
 }
 
+// This type represents a single license assigned to a user.
+private microsoft.user.assignedLicense @defaults("skuId") {
+  // Internal ID
+  id string
```
afiune: Let's remove this id that shows already three times in the resource. 
1. The actual internal id which is `__id`
2. The `skuId`
3. and this `id`

providers/ms365/resources/ms365.lr.manifest.yaml
```diff
@@ -522,6 +523,13 @@ resources:
       userType: {}
     is_private: true
     min_mondoo_version: 9.0.0
+  microsoft.user.assignedLicense:
+    fields:
+      disabledPlans: {}
+      id: {}
```
afiune: Once you remove the `id` from the resource, you are going to have to manually remove this line, the generator only adds new things, doesn't remove stuff.

glower: removed

providers/ms365/resources/ms365.lr.manifest.yaml
```diff
@@ -564,10 +571,22 @@ resources:
...
     fields:
+      assignedLicenses: {}
       filter: {}
       list: {}
       search: {}
     min_mondoo_version: 9.0.0
+  microsoft.users.assignedLicense:
+    fields:
+      disabledPlans: {}
+      id: {}
```
afiune: this one too needs to be removed, I think 🤔 cc @tas50

tas50: The automation will only add so if you rename things during development you can end up with duplicates. This should get cleaned up

glower: got it, will remove

PR #5650, opened 2025-06-10
⭐ Add depguard tp the linter rule to restrict viper usage

glower:
Introduces a new `depguard` rule to prevent the `github.com/spf13/viper`
package from being imported outside of designated CLI packages.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -33,6 +33,16 @@ linters-settings:
 
   depguard:
     rules:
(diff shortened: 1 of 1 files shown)


chris-rock:
Why are the test not failing for the scanner then? I think it is ok to have an exception for this one use case in local_scanner.go

chris-rock (approved):

PR #5656, opened 2025-06-11
✨ cnquery run `--exit-1-on-failure`

afiune:
This change adds a new flag named `--exit-1-on-failure` which will make
`cnquery` exit with code `1` if any query result failed.

Simple Example: **This should fail.**
```
cnquery run -c "1==2" --exit-1-on-failure
```

Closes https://github.com/mondoohq/cnspec/issues/1671

### DIFF
--- a/apps/cnquery/cmd/plugin.go
+++ b/apps/cnquery/cmd/plugin.go
@@ -123,6 +123,17 @@ func (c *cnqueryPlugin) RunQuery(conf *run.RunQueryConfig, runtime *providers.Ru
 		out.WriteString("[")
 	}
 
+	// anyResultFailed is a flag that will be switched on if any query result failed,
+	// if the flag `exit-1-on-failure` is provided and anyResultFailed is true, we
+	// will exit the program with the exit code `1`
+	var anyResultFailed = false
+	// we defer this check since we want it to be the last thing to be evaluated
+	defer func() {
+		if conf.GetExit_1OnFailure() && anyResultFailed {
+			os.Exit(1)
+		}
+	}()
+
 	for i := range discoveredAssets.Assets {
 		asset := discoveredAssets.Assets[i]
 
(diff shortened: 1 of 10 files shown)


test/commands/cli_test.go
```diff
@@ -35,13 +66,15 @@ func TestCompare(t *testing.T) {
...
+
 	ts.DisableLogging = true
-	ts.Commands["cnquery"] = cmdtest.InProcessProgram("cnquery", func() int {
-		err := cnqueryCmd.Execute()
-		if err != nil {
-			return 1
-		}
-		return 0
-	})
+	ts.Commands["cnquery"] = cmdtest.Program("cnquery")
```
afiune, line 78: The only way the exit code worked was by updating the tests to run the CLI as a binary and not as a in-process execution.

https://github.com/google/go-cmdtest?tab=readme-ov-file#invoking-your-cli-out-of-process

apps/cnquery/cmd/plugin.go
```diff
@@ -123,6 +123,17 @@ func (c *cnqueryPlugin) RunQuery(conf *run.RunQueryConfig, runtime *providers.Ru
...
 	}
 
+	// anyResultFailed is a flag that will be switched on if any query result failed,
+	// if the flag `exit-1-on-failure` is provided and anyResultFailed is true, we
+	// will exit the program with the exit code `1`
+	var anyResultFailed = false
+	// we defer this check since we want it to be the last thing to be evaluated
+	defer func() {
+		if conf.GetExit_1OnFailure() && anyResultFailed {
+			os.Exit(1)
```
preslavgerchev, line 133: if we exit here, we might not run anything that follows `RunQuery`, I think we should just return an err here and let the caller above handle this

afiune: There is nothing really that runs after `RunQuery` https://github.com/mondoohq/cnquery/blob/main/apps/cnquery/cmd/run.go#L86-L89

Would you still like me to change it? 

Further more, shell is another command that does that in a couple of places.
* https://github.com/mondoohq/cnquery/blob/main/apps/cnquery/cmd/shell.go#L142
* https://github.com/mondoohq/cnquery/blob/main/apps/cnquery/cmd/shell.go#L149

preslavgerchev: I think ideally we only call `os.Exit` in the cmds themselves to ensure that anything underneath gets executed correctly, but this is not blocking for me

PR #5665, opened 2025-06-16
🐛 Convert to durations for JSON

jaym:
They were being put in as a timestamp

Also, fixes an issue where we were saying there are 24 mins in an hour instead of 60

### DIFF
--- a/llx/rawdata_json.go
+++ b/llx/rawdata_json.go
@@ -379,9 +379,19 @@ func rawDataJSON(typ types.Type, data interface{}, codeID string, bundle *CodeBu
 			return nil
 		}
 
-		b, err := time.MarshalJSON()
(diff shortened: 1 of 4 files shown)


llx/rawdata_json.go
```diff
@@ -379,9 +379,19 @@ func rawDataJSON(typ types.Type, data interface{}, codeID string, bundle *CodeBu
 			return nil
 		}
 
-		b, err := time.MarshalJSON()
-		buf.Write(b)
-		return err
+		if time.Unix() > 0 {
+			b, err := time.MarshalJSON()
+			buf.Write(b)
```
czunker: nit: I guess `b` is empty in case of an error, but somehow writing before checking the error reads strange to me.

PR #5666, opened 2025-06-16
✨ Detect activated SLES modules/extensions

czunker:
This adds the actiavted SLES modules to the platform metadata.

```
cnquery run container image registry.suse.com/suse/sle15 -c "asset.platformMetadata"
asset.platformMetadata: {
  activated_modules: "Basesystem, Python 3, Server Applications"
  distro-id: "sles"
}
```

### DIFF
--- a/providers/os/detector/detector_all.go
+++ b/providers/os/detector/detector_all.go
@@ -422,7 +422,7 @@ var sles = &PlatformResolver{
 		if pf.Name == "sles" {
 			// SLES can have various modules/repos activated, identify them via filesystem
 			modules := getActivatedSlesModules(conn)
-			pf.Metadata["activated_modules"] = strings.Join(modules, ", ")
+			pf.Metadata["suse/modules"] = strings.Join(modules, ",")
 
 			return true, nil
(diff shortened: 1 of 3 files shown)


providers/os/detector/detector_all.go
```diff
@@ -420,6 +420,10 @@ var sles = &PlatformResolver{
 	IsFamily: false,
 	Detect: func(r *PlatformResolver, pf *inventory.Platform, conn shared.Connection) (bool, error) {
 		if pf.Name == "sles" {
+			// SLES can have various modules/repos activated, identify them via filesystem
+			modules := getActivatedSlesModules(conn)
+			pf.Metadata["activated_modules"] = strings.Join(modules, ", ")
```
chris-rock: i would join this without space here

chris-rock (approved):
Thank you @czunker

providers/os/detector/detector_all.go
chris-rock: maybe we call this `suse/modules`

PR #5701, opened 2025-06-24
⭐️ ms365: Extended microsoft.tenant with formsSettings

### LINKED ISSUE #5584: MS365: Extend microsoft.organization with `formsSettings.isInOrgFormsPhishingScanEnabled`
We need https://learn.microsoft.com/en-us/graph/api/resources/formssettings?view=graph-rest-beta this.

glower:
added support for global formsSettings for ms365 to `microsoft.tenant`

```grapQL
cnquery> microsoft.tenant { formsSettings {*} }
microsoft.tenant: {
  formsSettings: {
    isBingImageSearchEnabled: true
    isExternalShareCollaborationEnabled: true
    id: "d9abc6fc-fd88-4480-a931-2f7939adbac2-forms-settings"
    isRecordIdentityByDefaultEnabled: true
    isExternalShareResultEnabled: true
    isExternalShareTemplateEnabled: true
    isExternalSendFormEnabled: true
    isInOrgFormsPhishingScanEnabled: true
  }

```

resolves #5584

### DIFF
--- a/providers/ms365/resources/ms365.lr
+++ b/providers/ms365/resources/ms365.lr
@@ -78,8 +78,6 @@ microsoft.tenant @defaults("name") {
 
 // Company-wide configuration for apps and services.
 private microsoft.tenant.settings @defaults("isAppAndServicesTrialEnabled isOfficeStoreEnabled") {
-  // Unique synthetic ID based on the parent tenant.
-  id string
   // Controls whether users can start trial subscriptions for apps and services in your organization.
   isAppAndServicesTrialEnabled bool
   // Controls whether users can access the Microsoft Store
@@ -87,9 +85,7 @@ private microsoft.tenant.settings @defaults("isAppAndServicesTrialEnabled isOffi
 }
 
 // Company-wide settings for Microsoft Forms
-private microsoft.tenant.formsSettings @defaults("id isExternalSendFormEnabled isBingImageSearchEnabled") {
-  // Unique synthetic ID based on the parent tenant settings
-  id string
+private microsoft.tenant.formsSettings @defaults("isExternalSendFormEnabled isBingImageSearchEnabled") {
   // Controls whether users can send a link to a form to an external user
   isExternalSendFormEnabled bool
   // Controls whether users can collaborate on a form layout and structure with an external user
--- a/providers/ms365/resources/tenant.go
+++ b/providers/ms365/resources/tenant.go
@@ -161,10 +161,6 @@ func (a *mqlMicrosoft) tenantDomainName() (string, error) {
 	return tenantDomainName, nil
 }
 
-func (m *mqlMicrosoftTenantSettings) id() (string, error) {
-	return m.Id.Data, nil
-}
-
 func (a *mqlMicrosoftTenant) settings() (*mqlMicrosoftTenantSettings, error) {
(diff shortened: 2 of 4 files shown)


.github/actions/spelling/expect.txt
```diff
@@ -127,4 +127,4 @@ manageddevice
 meid
 udid
 Passwordless
-Sspr
```
glower, line 130: added new line, linter was complaining

providers/ms365/resources/ms365.lr
```diff
@@ -84,6 +86,26 @@ private microsoft.tenant.settings @defaults("isAppAndServicesTrialEnabled isOffi
   isOfficeStoreEnabled bool
 }
 
+// Company-wide settings for Microsoft Forms
+private microsoft.tenant.formsSettings @defaults("id isExternalSendFormEnabled isBingImageSearchEnabled") {
```
imilchev: please don't add the `id` to the defaults. That should (almost) never be in the defaults

providers/ms365/resources/tenant.go
```diff
@@ -204,3 +204,49 @@ func (a *mqlMicrosoftTenant) settings() (*mqlMicrosoftTenantSettings, error) {
...
+	settings := formsSetting.GetSettings()
+	if settings == nil {
+		return nil, nil
+	}
+
+	formsSettingId := fmt.Sprintf("%s-forms-settings", a.Id.Data)
+
+	formSetting, err := CreateResource(a.MqlRuntime, "microsoft.tenant.formsSettings",
+		map[string]*llx.RawData{
+			"id":                                  llx.StringData(formsSettingId),
```
imilchev: if you don't need the `id` to be visible to the user, which seems to be the case since that's artificially created, i think you can just define `__id` and remove the `id()` function. That is equivalent to have an id with a function attached. the `__id` is a "private" field to store an id and auto-retrieve it.

imilchev: please check your other PRs for this as well

providers/ms365/resources/ms365.lr
glower: removed

providers/ms365/resources/tenant.go
glower: removed `id` and only `__id`  in this PR and a previews one.

imilchev (approved):

providers/ms365/resources/ms365.lr
```diff
@@ -72,18 +72,36 @@ microsoft.tenant @defaults("name") {
...
-  // Unique synthetic ID based on the parent tenant.
-  id string
   // Controls whether users can start trial subscriptions for apps and services in your organization.
   isAppAndServicesTrialEnabled bool
   // Controls whether users can access the Microsoft Store
   isOfficeStoreEnabled bool
 }
 
+// Company-wide settings for Microsoft Forms
+private microsoft.tenant.formsSettings @defaults("isExternalSendFormEnabled isBingImageSearchEnabled") {
```
chris-rock, line 88: Can we please stop doing using the same resource name twice, once as field and once as resource. There should be no conflict. I recommend renaming the private resource to `microsoft.tenantFormsSettings`

PR #5709, opened 2025-06-24
🧹 network: update default dialer timeout to 2s

afiune:
We are seeing some slow queries that try to access `tls.params`, this is because we have
a timeout of 1m per TLS version, a total of 5 versions so this query could end up running
for 5 minutes.

Since the `tlsshake` package has already a 2s timeout, re-use it to avoid slow queries.


czunker (approved):
LGTM

PR #5720, opened 2025-06-26
✨ Detect activated RHEL modules

czunker:
Implements: https://github.com/mondoohq/cnquery/issues/5695

Example RHEL 9 GCP VM:
```
cnquery run ssh -i ~/.ssh/google_compute_engine XXX -c "asset.platformMetadata" 
asset.platformMetadata: {
  distro-id: "rhel"
  redhat/modules: ""
}
```

With enabled maven module:
```
asset.platformMetadata: {
  distro-id: "rhel"
  rhel/modules: "maven"
}
```

RHEL 8 GCP VM:
```
asset.platformMetadata: {
  distro-id: "rhel"
  redhat/modules: "python36,virt"
}
```

### DIFF
--- a/providers/os/detector/detector_rhel.go
+++ b/providers/os/detector/detector_rhel.go
@@ -6,6 +6,7 @@ package detector
 import (
 	"bufio"
 	"bytes"
+	"path/filepath"
 	"strings"
 
 	"github.com/spf13/afero"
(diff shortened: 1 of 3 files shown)


providers/os/detector/detector_rhel.go
```diff
@@ -0,0 +1,72 @@
...
+		return []string{}
+	}
+
+	modules := []string{}
+	for _, file := range files {
+		if !strings.HasSuffix(file.Name(), ".module") {
+			continue
+		}
+
+		content, err := afs.ReadFile(modulesDir + "/" + file.Name())
```
glower: nit: let us use `filepath.Join(modulesDir, file.Name())`

czunker: Good catch.

Issue #5728, opened 2025-06-27
MS365: `microsoft.identityAndAccess` returns a list of `microsoft.identityAndAccess.policy`


syrull:
**Describe the bug**
`microsoft.identityAndAccess` returns a list of `microsoft.identityAndAccess.policy`

This is rather a category having bunch of other sub categories, `policies` is one of it.

A better way to handle this would be: `microsoft.identityAndAccess.privilegedIdentityManagement.policies` -> returns a list of `microsoft.identityAndAccess.privilegedIdentityManagement.policy`

Issue #5727, opened 2025-06-27
Microsoft Entra: Extend `identityAndAccess` with `organization`


syrull:
Expose `organization` in `identityAndAccess` 

Graph API: https://learn.microsoft.com/en-us/graph/api/organization-get?view=graph-rest-1.0&tabs=http

Cmdlet: https://learn.microsoft.com/en-us/powershell/module/microsoft.graph.identity.directorymanagement/get-mgorganization?view=graph-powershell-1.0

tas50:
## Already available

`organization` is already exposed on `microsoft.identityAndAccess` (it maps to `microsoft.tenant`, backed by the Graph `GET /organization` endpoint you linked). Verified live just now:

```mql
microsoft.identityAndAccess.organization { id name countryLetterCode createdAt }
```

```
microsoft.identityAndAccess.organization: {
  id:                "2919c383-55f1-4d5b-9d3e-352998c3fd2b"
  name:              "Mondoo, Inc"
  countryLetterCode: "US"
  createdAt:         2023-03-24 12:13:45 -0700
}
```

The full `microsoft.tenant` shape (verified domains, assigned/provisioned plans, privacy profile, technical/security notification contacts, on-prem sync, tenant settings, etc.) is reachable from there. Closing as already-implemented — reopen if you need a specific field that isn't on `microsoft.tenant` yet.

PR #5734, opened 2025-06-28
Restructure `microsoft.identityAndAccess` and add `organization`

syrull:
Issue: https://github.com/mondoohq/cnquery/issues/5728


syrull:
Accidentally added `organization` to this branch, which resolves https://github.com/mondoohq/cnquery/issues/5727

This should be a new branch but I didn't realise that I committed into this one, apologies.

providers/ms365/resources/ms365.lr
```diff
@@ -119,16 +119,29 @@ microsoft.users {
   search string
 }
 
-// A container resource for identity and access policies that can be filtered
+// A container resource for identity and access management
 microsoft.identityAndAccess {
-  []microsoft.identityAndAccess.policy
+  // Get the instances of role eligibilities
```
vjeffrey, line 241: this pr needs a rebase - `  roleEligibilityScheduleInstances() []microsoft.identityAndAccess.roleEligibilityScheduleInstance` already exists on main

providers/ms365/resources/ms365.lr
```diff
@@ -172,19 +185,19 @@ private microsoft.identityAndAccess.policy @defaults("id displayName scopeType")
   // The identity of the user who last modified the policy
   lastModifiedBy dict
   // The collection of rules that are part of the policy
-  rules() []microsoft.identityAndAccess.policy.rule
+  rules() []microsoft.identityAndAccess.privilegedIdentityManagement.policy.rule
 }
 
 // A rule defined for a PIM role management policy
-private microsoft.identityAndAccess.policy.rule @defaults("id") {
+private microsoft.identityAndAccess.privilegedIdentityManagement.policy.rule @defaults("id") {
```
vjeffrey, line 311: why is this change to the resource being made?

tas50:
@syrull rebased

providers/ms365/resources/ms365.lr
syrull: This is intentional for https://github.com/mondoohq/mql/issues/5728: `microsoft.identityAndAccess` is being reshaped into a category, and the role-management policy list is moving under `microsoft.identityAndAccess.privilegedIdentityManagement.policies` so PIM policies live under the PIM namespace instead of directly under the broader identity/access category.

providers/ms365/resources/ms365.lr
syrull: Rebased onto current main in 0a5e895b. The `roleEligibilityScheduleInstances()` field is now coming from main and is no longer being introduced as a duplicate by this PR.

philipbalinov:
## Code Review

### Overview
Reshapes `microsoft.identityAndAccess` from a "policies-with-list" resource into a category container. The PIM role-management policies move from `microsoft.identityAndAccess.policy*` to `microsoft.identityAndAccess.privilegedIdentityManagement.policies/.policy*`. Also adds `microsoft.identityAndAccess.organization` returning a `microsoft.tenant`. Net: +418/-195, mostly generated code from `.lr` changes.

### 🛑 Breaking schema changes (most important)
The following public schema entries are **renamed/removed**, not aliased:
- `microsoft.identityAndAccess.policy` → `microsoft.identityAndAccess.privilegedIdentityManagement.policy`
- `microsoft.identityAndAccess.policy.rule` → `…privilegedIdentityManagement.policy.rule`
- `microsoft.identityAndAccess.policy.ruleTarget` → `…privilegedIdentityManagement.policy.rule.target` (note: also `ruleTarget` → `rule.target`)
- `microsoft.identityAndAccess.filter` and `microsoft.identityAndAccess.list` — removed from `.lr.versions`

Existing MQL queries, cnspec policies, and any bundled content referencing `microsoft.identityAndAccess { … }` (iterable) or `microsoft.identityAndAccess.policy` will break at parse/compile time. The repo convention is *"Existing entries are never overwritten"* in `.lr.versions`. Two options:
1. Use `alias` (already supported — see `alias microsoft.organization = microsoft.tenant` at `ms365.lr:7`) to keep old paths working as deprecated aliases for one release.
2. If the team accepts the break, the commit should be prefixed 🛑 per conventions, and a CHANGELOG/migration note should accompany it.

### ⚠ Code duplication: `organization()` reimplements `initMicrosoftTenant`
The new `organization()` method in `identity_and_access.go` is a verbatim copy of `initMicrosoftTenant` in `tenant.go`: same `graphClient.Organization().ByOrganizationId(conn.TenantId()).Get(...)`, same `tenantFields` Select, same `newMicrosoftTenant` call. Prefer delegating:

```go
func (a *mqlMicrosoftIdentityAndAccess) organization() (*mqlMicrosoftTenant, error) {
    res, err := NewResource(a.MqlRuntime, "microsoft.tenant", nil)
    if err != nil {
        return nil, err
    }
    return res.(*mqlMicrosoftTenant), nil
}
```

This lets the existing `initMicrosoftTenant` own the fetch, gives free `__id`-based caching (a single Graph call serves both `microsoft.tenant` and `microsoft.identityAndAccess.organization`), and avoids drift between two copies of the same field list.

### Minor
- **Naming / scoping**: putting `organization` under `identityAndAccess` is a stretch — organization/tenant info isn't really IAM. The PR author noted in a comment that this was accidentally committed to this branch (originally for issue #5727). Worth deciding whether it belongs here at all given `microsoft.tenant` and the `microsoft.organization` alias already exist, and ideally splitting it into a separate PR scoped to #5727.
- **Version jump**: `ms365` config bumps `13.0.6 → 13.0.9` (3 patches) and new `.lr.versions` entries use `13.0.10`. The `13.0.10` is consistent with the bumped version, but the `13.0.6 → 13.0.9` jump is unexplained — convention is a single patch bump (`13.0.6 → 13.0.7`).
- **Dead init pass-through**: `initMicrosoftIdentityAndAccessPrivilegedIdentityManagementPolicies` carries over the pre-existing
  ```go
  if filter, ok := args["filter"]; ok { args["filter"] = filter }
  return args, nil, nil
  ```
  which is a no-op (read & reassign). Not introduced here, but since the function is being moved/renamed anyway, dropping it (and letting the default init run) would clean up.
- **Container `__id`**: new `privilegedIdentityManagement` and `policies` containers are created with `nil` args (no explicit `__id`). Consistent with other singleton containers like `identityAndSignIn`, so OK.
- **No tests**: typical for the codebase. Worth confirming interactively:
  ```
  mql shell ms365
  microsoft.identityAndAccess.privilegedIdentityManagement.policies.list
  microsoft.identityAndAccess.organization { id name verifiedDomains }
  ```

### Verdict
The restructure itself is reasonable and the code-gen is clean, but the **silent breaking change** to the public schema is the headline issue — please decide whether to provide alias paths or label/announce this as a breaking change. The `organization()` duplication should be replaced with a delegation to `microsoft.tenant`. The bundled `organization` addition probably belongs in a separate PR scoped to #5727.

PR #5738, opened 2025-06-29
Add fallback support for kernel.modules when lsmod is not available

### LINKED ISSUE #5641: `kernel.modules` is not resolving the modules if `lsmod` is not installed
**Describe the bug**

`kernel.modules` is not resolving the modules if `lsmod` is not installed, we should rely on `lsmod` as a fallback, we can use `find /sys/module -name "initstate" -exec grep -l "live" {} \;` otherwise.

syrull:
Fixes #5641

This commit implements a fallback mechanism for the kernel.modules resource when the lsmod command is not available on Linux systems. The implementation follows a three-tier approach:

1. Primary: Try /sbin/lsmod command (existing behavior)
2. Fallback 1: Read from /proc/modules if lsmod fails
3. Fallback 2: Parse /sys/module directory structure if /proc/modules is unavailable

Key changes:
- Added ParseLinuxSysModule() function to parse kernel modules from /sys/module
- Updated LinuxKernelManager.Modules() with comprehensive fallback logic
- Added extensive test coverage for all fallback scenarios
- Maintains backward compatibility with existing functionality

The /sys/module fallback reads module directories, checks initstate files for 'live' modules, and extracts size/reference count information from coresize and refcnt files respectively.

Testing:
- All existing tests continue to pass
- New tests verify fallback behavior and /sys/module parsing
- Mock test data simulates systems without lsmod or /proc/modules

### DIFF
--- a/providers/os/resources/kernel/modules.go
+++ b/providers/os/resources/kernel/modules.go
@@ -6,9 +6,14 @@ package kernel
 import (
 	"bufio"
 	"io"
+	"os"
 	"path"
+	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
+
+	"github.com/spf13/afero"
 )
 
 func ParseLsmod(r io.Reader) []*KernelModule {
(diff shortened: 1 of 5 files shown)


providers/os/resources/kernel/modules.go
```diff
@@ -132,3 +137,92 @@ func ParseGenkex(stdout io.Reader) ([]*KernelModule, error) {
...
+			return nil
+		}
+
+		moduleName := info.Name()
+
+		// Check if module is live by reading initstate file
+		initstatePath := filepath.Join(path, "initstate")
+		initstateFile, err := fs.Open(initstatePath)
+		if err != nil {
+			// If initstate doesn't exist, skip this module
```
philipbalinov, line 180: 👍 Note: /sys/module/ will list all modules, both builtin and external/loadable.

missing initstate file means that the module is builtin, so we are skipping those in order to replicate the behaviour of lsmod

philipbalinov (approved):
LGTM @syrull

syrull:
Not merging this as I failed to gather more information on the issue, probably my instances were buggy/restricted hence the missing command.

PR #5742, opened 2025-06-30
Extended microsoft.policies to contain the activityBasedTimeoutPolicy

glower:
resolves ##5582

### DIFF
--- a/providers/ms365/resources/policies.go
+++ b/providers/ms365/resources/policies.go
@@ -124,6 +124,38 @@ func (a *mqlMicrosoftPolicies) authenticationMethodsPolicy() (*mqlMicrosoftPolic
 	return newAuthenticationMethodsPolicy(a.MqlRuntime, resp)
 }
 
+func (a *mqlMicrosoftPolicies) activityBasedTimeoutPolicies() ([]interface{}, error) {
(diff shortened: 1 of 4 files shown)


providers/ms365/resources/policies.go
```diff
@@ -124,6 +124,46 @@ func (a *mqlMicrosoftPolicies) authenticationMethodsPolicy() (*mqlMicrosoftPolic
...
+			"__id":                  llx.StringDataPtr(policy.GetId()),
+			"id":                    llx.StringDataPtr(policy.GetId()),
+			"definition":            llx.ArrayData(convert.SliceAnyToInterface(policy.GetDefinition()), types.String),
+			"displayName":           llx.StringDataPtr(policy.GetDisplayName()),
+			"isOrganizationDefault": llx.BoolDataPtr(policy.GetIsOrganizationDefault()),
+		})
+	if err != nil {
+		return nil, err
+	}
+	return mqlPolicy.(*mqlMicrosoftPoliciesActivityBasedTimeoutPolicy), nil
```
imilchev: nit: no need to do the type cast here since you use this already as `[]interface` in the upper function. You could also just not define a function but directly create the resource above (it doesn't seem like it's re-used).

PR #5743, opened 2025-06-30
✨ Expand scanning granularity in GCP

afiune:
Fine grained asset scanning \(scanning more cloud resources as assets\) makes it easier to explore, understand, and triage security findings\. When individual resources are discovered as assets then users can view asset overview information, create exceptions, and create tickets for individual findings\.

<p></p>

Let's improve our GCP experience by scanning more cloud resources as assets, so that we provide improved visibility into GCP security:

- Cloud SQL PostgreSQL `gcp.project.sql.instances.where(databaseInstalledVersion == /^POSTGRES_/)`

- Cloud SQL MySQL `gcp.project.sql.instances.where(databaseInstalledVersion == /^MYSQL_/)`

- Cloud SQL SQL Server  `gcp.project.sql.instances.where(databaseInstalledVersion == /^SQL_/)`

- Cloud DNS zone `gcp.project.dnsService.managedzone`

- KMS keyrings `gcp.project.kmsService.keyring`

## Test

Ran a scan and the platform showed up the asset. Now we need to write policies.

Screenshot 2025-06-30 at 2 21 15 PM

### DIFF
--- a/providers/gcp/config/config.go
+++ b/providers/gcp/config/config.go
@@ -51,7 +51,7 @@ Examples with the GCP project configured:
 				resources.DiscoveryStorageBuckets,
 				resources.DiscoveryBigQueryDatasets,
 				resources.DiscoverCloudSQLMySQL,
-				resources.DiscoverCloudSQLPostgres,
+				resources.DiscoverCloudSQLPostgreSQL,
 				resources.DiscoverCloudSQLSQLServer,
 				resources.DiscoverCloudDNSZones,
 				resources.DiscoverCloudKMSKeyrings,
--- a/providers/gcp/connection/platform.go
+++ b/providers/gcp/connection/platform.go
@@ -6,6 +6,7 @@ package connection
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"go.mondoo.com/cnquery/v11/providers-sdk/v1/inventory"
 )
@@ -52,7 +53,7 @@ func (c *GcpConnection) PlatformInfo() (*inventory.Platform, error) {
 	if c.opts.platformOverride != "" && c.opts.platformOverride != "gcp" {
 		return &inventory.Platform{
 			Name:    c.opts.platformOverride,
-			Title:   getTitleForPlatformName(c.opts.platformOverride),
+			Title:   GetTitleForPlatformName(c.opts.platformOverride),
 			Family:  []string{"google"},
 			Kind:    "gcp-object",
 			Runtime: "gcp",
@@ -89,7 +90,7 @@ func (c *GcpConnection) PlatformInfo() (*inventory.Platform, error) {
 	return nil, errors.New("unsupported resource type")
 }
 
-func getTitleForPlatformName(name string) string {
+func GetTitleForPlatformName(name string) string {
 	switch name {
 	case "gcp-organization":
 		return "GCP Organization"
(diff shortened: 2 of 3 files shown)


providers/gcp/connection/platform.go
```diff
@@ -111,6 +111,12 @@ func getTitleForPlatformName(name string) string {
 		return "GCP Storage Bucket"
 	case "gcp-bigquery-dataset":
 		return "GCP BigQuery Dataset"
+	case "gcp-sql-instance":
```
tas50: Can we expand this into a platform per SQL type in GCP please

tas50 (changes requested):

providers/gcp/connection/platform.go
afiune: Will do. Do you also want to have different discover flags? e.g. `--discover clou-sql-mysql` ...

afiune: done

providers/gcp/config/config.go
```diff
@@ -50,6 +50,7 @@ Examples with the GCP project configured:
 				resources.DiscoveryGkeClusters,
 				resources.DiscoveryStorageBuckets,
 				resources.DiscoveryBigQueryDatasets,
+				resources.DiscoverCloudSQLs,
```
vjeffrey: i would have expected to see 5 new things here? one for each of the items you listed in the description?

afiune: Yup. Will update this too.

afiune: done

afiune:
Screenshot 2025-06-30 at 2 21 15 PM

providers/gcp/connection/platform.go
```diff
@@ -111,6 +111,16 @@ func getTitleForPlatformName(name string) string {
 		return "GCP Storage Bucket"
 	case "gcp-bigquery-dataset":
 		return "GCP BigQuery Dataset"
+	case "gcp-sql-mysql":
+		return "GCP Cloud SQL MySQL"
```
tas50:
```suggestion
		return "GCP Cloud SQL for MySQL"
```

providers/gcp/connection/platform.go
```diff
@@ -111,6 +111,16 @@ func getTitleForPlatformName(name string) string {
...
+	case "gcp-sql-postgres":
+		return "GCP Cloud SQL Postgres"
```
tas50:
```suggestion
		return "GCP Cloud SQL for PostgreSQL"
```

providers/gcp/connection/platform.go
```diff
@@ -111,6 +111,16 @@ func getTitleForPlatformName(name string) string {
...
+	case "gcp-sql-sqlserver":
+		return "GCP Cloud SQL SQL Server"
```
tas50:
```suggestion
		return "GCP Cloud SQL for SQL Server"
```

providers/gcp/resources/discovery.go
```diff
@@ -35,8 +38,16 @@ const (
 	DiscoveryGkeClusters        = "gke-clusters"
 	DiscoveryStorageBuckets     = "storage-buckets"
 	DiscoveryBigQueryDatasets   = "bigquery-datasets"
+	DiscoverCloudSQLMySQL       = "cloud-sql-mysql"
+	DiscoverCloudSQLPostgres    = "cloud-sql-postgres"
```
tas50: Let's use the full name. They've used that name for at least a decade. 

```suggestion
	DiscoverCloudSQLPostgreSQL    = "cloud-sql-postgresql"
```

afiune: This is not what we get back from Google's API  
* https://github.com/googleapis/google-api-go-client/blob/main/sqladmin/v1beta4/sqladmin-gen.go#L1472
* https://github.com/googleapis/google-api-go-client/blob/main/sqladmin/v1beta4/sqladmin-gen.go#L534-L542

So we will have to do some string replacement.

PR #5750, opened 2025-07-02
⭐️ Improve cnquery sbom command

chris-rock:
This PR improves the `cnquery sbom` experience:

- ⭐️ exclude CPEs from SBOM export
- ⭐️ sbom asset title
- 🧹 users need to explicitly enable the exclusion of the evidence
- ⭐️ add basic protobom support to allow more flexible sbom conversions in the future

You can see the output via:

```
cnquery sbom docker alpine:3.19 --output json --output-target ./alpine-3.19.json
```

It changes the default, so that no evidence and cpes are included. To enable both run:

```
cnquery sbom docker alpine:3.19 --output json --output-target ./alpine-3.19.json --with-evidence --with-cpes
```


sbom/cnquery_sbom.pb.go
```diff
@@ -4,8 +4,8 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.36.6
-// 	protoc        v6.31.0
-// source: sbom.proto
+// 	protoc        v6.31.1
+// source: cnquery_sbom.proto
```
chris-rock, line 8: Needed to rename the proto file to avoid proto conflicts with protobom which also calls their proto sbom.proto

apps/cnquery/cmd/sbom.go
```diff
@@ -27,7 +27,8 @@ func init() {
 	sbomCmd.Flags().StringToString("annotation", nil, "Add an annotation to the asset") // user-added, editable
 	sbomCmd.Flags().StringP("output", "o", "list", "Set output format: "+sbom.AllFormats())
 	sbomCmd.Flags().String("output-target", "", "Set output target to which the SBOM report will be written")
-	sbomCmd.Flags().Bool("with-evidence", false, "Display evidence for each component")
+	sbomCmd.Flags().Bool("with-evidence", false, "Include evidence for each component")
+	sbomCmd.Flags().Bool("with-cpes", false, "Disable the generation of CPEs for each component")
```
czunker: I think the text should be called "Enable ..."

sbom/protobom.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	for _, node := range doc.GetNodeList().GetNodes() {
+		pkg := &Package{
+			Name:    node.Name,
+			Version: node.Version,
+		}
+
+		for key, identifier := range node.GetIdentifiers() {
+			if key == int32(protobom_sbom.SoftwareIdentifierType_PURL) {
```
czunker, line 60: Do we need to generate a PURL when we don't find one?

sbom/protobom.go
```diff
@@ -0,0 +1,95 @@
...
+		purposes := node.GetPrimaryPurpose()
+		if len(purposes) > 0 {
+			switch purposes[0] {
+			case protobom_sbom.Purpose_OPERATING_SYSTEM:
+				bom.Asset.Platform = &Platform{
+					Name:    pkg.Name,
+					Version: pkg.Version,
+					Title:   pkg.Description,
+				}
+				bom.Asset.Platform.Family = familyMap[pkg.Name]
```
czunker, line 86: Is this correct?
The package name is used for the platform?

apps/cnquery/cmd/sbom.go
chris-rock: Yes, I switched it from negative to positive and forgot to adjust it

sbom/protobom.go
chris-rock: Not at the moment

sbom/protobom.go
chris-rock: Yes, because the purpose defines if this is an OS or an application package

apps/cnquery/cmd/sbom.go
chris-rock: Fixed

czunker (approved):
Thanks @chris-rock

PR #5758, opened 2025-07-03
🧹 use vtproto

imilchev:
- Use vtproto when cloning, marshalling and unmarshalling across the code base.
- Add a gRPC codec to use vtproto functions for marshalling and unmarshalling when possible


providers-sdk/v1/plugin/codec.go
imilchev, line 1: this file was taken from https://github.com/vitessio/vitess/blob/main/go/vt/servenv/grpc_codec.go That was referenced in the vtprotobuf docs when talking about grpc bindings

chris-rock (approved):
Thank you @imilchev

PR #5770, opened 2025-07-07
✨ Add InstallLocation to files for Windows packages

czunker:
Screenshot from 2025-07-09 11-20-52

### DIFF
--- a/providers/os/resources/packages/testdata/windows_2019.toml
+++ b/providers/os/resources/packages/testdata/windows_2019.toml
@@ -1,5 +1,5 @@
 # Packages
-[commands."powershell -c \"Get-AppxPackage -AllUsers | Select Name, PackageFullName, Architecture, Version, Publisher  | ConvertTo-Json\""]
+[commands."powershell -c \"Get-AppxPackage -AllUsers | Select Name, PackageFullName, Architecture, Version, Publisher, InstallLocation | ConvertTo-Json\""]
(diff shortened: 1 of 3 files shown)


providers/os/resources/os.lr
```diff
@@ -659,6 +659,9 @@ package @defaults("name version") {
 
   // Package vendor
   vendor string
+
+  // Package installation path
+  installLocation string
```
chris-rock: Can we not use files() for this?

czunker: I first thought so too, but files is about the files which are inside the package and extracted to disk. E.g. for a dpkg package, it is about the files stored in `/usr`, the man pages, ...

This is about the installer itself, as I understood the Windows data.

chris-rock: I think we should keep this consistent with macOS:

```
packages.list[0]: {
  name: "App Store"
  epoch: ""
  purl: "pkg:macos/macos/App%20Store@3.0?arch=arm64"
  arch: "arm64"
  outdated: false
  files: [
    0: pkgFileInfo path="/System/Applications/App Store.app"
  ]
  cpes: []
  format: "macos"
  version: "3.0"
  available: ""
  description: ""
  installed: true
  origin: ""
  vendor: ""
  status: ""
}
```

chris-rock (approved):
Thank you @czunker

PR #5785, opened 2025-07-15
🐛 Prevent CVSS panic

czunker:


providers-sdk/v1/upstream/mvd/cvss/cvss.go
```diff
@@ -407,6 +407,11 @@ func MaxScore(cvsslist []*Cvss) (*Cvss, error) {
 
 	for i := 1; i < len(cvsslist); i++ {
 		entry := cvsslist[i]
+		if entry == nil {
```
preslavgerchev, line 410: what is the use case for getting a nil cvss in that list? I think we should not put nils in the list

preslavgerchev (approved):

PR #5788, opened 2025-07-16
🧪 Fail query compilation if it has no entrypoints or datapoints.

preslavgerchev:
Draft since I am not sure if this is the right place to put this check (also needs tests)


mqlc/mqlc.go
```diff
@@ -2067,6 +2067,17 @@ func (c *compiler) CompileParsed(ast *parser.AST) error {
...
 	c.updateLabels()
 
+	return c.failIfNoEntrypoints()
+}
+
+func (c *compiler) failIfNoEntrypoints() error {
+	if c.Features.IsActive(cnquery.FailIfNoEntryPoints) {
+		for _, b := range c.Result.CodeV2.Blocks {
+			if len(b.Datapoints) == 0 && len(b.Entrypoints) == 0 {
+				return errors.New("failed to compile: received an empty code structure")
```
jaym: we should probably indicate this is a bug in the error message so the user doesn't think they did something wrong

PR #5797, opened 2025-07-17
Removed Index path as it breaks the html of the releases website

username-is-already-taken2:
When this runs the releases website only lists the providers package

image

I've check the times of the workflow and there is no speed advantage for only reindexing a certain path due to the way it evaluates the bucket contents.

Remote Call - https://github.com/mondoohq/releasr/actions/runs/16296722224

Manual Trigger - https://github.com/mondoohq/releasr/actions/runs/16343633311

I'm proposing here to remove that path and just leave the job to run with the defaults.  Which is what I have to run to fix it :)

### DIFF
--- a/.github/workflows/providers.yaml
+++ b/.github/workflows/providers.yaml
@@ -216,4 +216,3 @@ jobs:
           token: ${{ steps.generate-token.outputs.token }}
           repository: "mondoohq/releasr"
           event-type: reindex
-          client-payload: '{ "reindex-path": "providers" }'


philipbalinov:
@username-is-already-taken2 we added the path in there not as a speed optimization, but because there was a rogue releasr running during releases and corrupting the HTML/checksum generation for cnspec.

Come to think of it, we are possibly reintroducing this by removing the path specification here. What we should do is stop `releasr` from generating the root /index.html page if the `--app` parameter was passed.

PR #5808, opened 2025-07-25
🧹 Fix some compiler warnings and conditionals in AWS provider.

VasilSirakov:
### In this PR:
- Fixes some conditionals where the wrong variable was checked.
- Removes redundant err checks.
- Modifies loops to value-only instead of index-only where you need to index `rawResources.Data[i].(mqlType)`.

### DIFF
--- a/providers/aws/resources/aws_iam.go
+++ b/providers/aws/resources/aws_iam.go
@@ -7,9 +7,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"fmt"
 	"net/url"
-	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


providers/aws/resources/aws_iam.go
```diff
@@ -95,12 +96,8 @@ func (a *mqlAwsIam) credentialReport() ([]interface{}, error) {
 					return nil, err
 				}
 
-				if gresp.State == iamtypes.ReportStateTypeStarted || gresp.State == iamtypes.ReportStateTypeInprogress {
-					// we need to wait
-				} else if gresp.State == iamtypes.ReportStateTypeComplete {
-					// we do not need do do anything
-				} else {
-					// unsupported report state
+				validStates := gresp.State.Values()
```
VasilSirakov: I personally think this is a bit redundant. `gresp.State` comes straight from the AWS API so not sure what this check here does. Thoughts on removing it?

I guess it's only relevant if AWS add a new enum value that's not `started`, `inprogress` or `complete`. In that case, the previous code would have failed but I don't even know if that's something we want here.

PR #5824, opened 2025-08-04
🧹 Capture baseproduct for sles

jaym:
For some versions of SLES SAP, this is where that information is capturued as opposed to VARIANT_ID


imilchev (approved):
LGTM

PR #5825, opened 2025-08-04
⚠️ Warn if an MQL resource does not have id set.

preslavgerchev:
Only warn for now since this will break existing resources that do not have an id set. We should first see how many such resources we have and only make this mandatory after

I need to regen all lr code but id like to first review the actual LR compiler change

### DIFF
--- a/providers-sdk/v1/lr/go.go
+++ b/providers-sdk/v1/lr/go.go
@@ -170,7 +170,7 @@ func NewResource(runtime *plugin.Runtime, name string, args map[string]*llx.RawD
 		if res != nil {
 		  mqlId := res.MqlID()
 			if mqlId == "" {
-			  log.Warn().Msgf("resource %s has no MQL ID defined, this is usually a dev issue with the resource, please open a Github issue", name)
+			  log.Warn().Msgf("resource %s has no MQL ID defined, this is usually a dev issue with the resource, please open a GitHub issue at https://github.com/mondoohq/cnquery/issues", name)
 			}
 			id := name + "\x00" + mqlId
 			if x, ok := runtime.Resources.Get(id); ok {
@@ -190,7 +190,7 @@ func NewResource(runtime *plugin.Runtime, name string, args map[string]*llx.RawD
 
 	mqlId := res.MqlID()
 	if mqlId == "" {
-		log.Warn().Msgf("resource %s has no MQL ID defined, this is usually a dev issue with the resource, please open a Github issue", name)
+		log.Warn().Msgf("resource %s has no MQL ID defined, this is usually a dev issue with the resource, please open a GitHub issue at https://github.com/mondoohq/cnquery/issues", name)
 	}
 	id := name + "\x00" + mqlId
 	if x, ok := runtime.Resources.Get(id); ok {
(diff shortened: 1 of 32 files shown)


providers-sdk/v1/lr/go.go
```diff
@@ -167,7 +168,11 @@ func NewResource(runtime *plugin.Runtime, name string, args map[string]*llx.RawD
 		}
 
 		if res != nil {
-			id := name+"\x00"+res.MqlID()
+		  mqlId := res.MqlID()
+			if mqlId == "" {
+			  log.Warn().Msgf("resource %s has no MQL ID defined, this is usually a dev issue with the resource, please open a Github issue", name)
```
tas50:
```suggestion
			  log.Warn().Msgf("resource %s has no MQL ID defined, this is usually a dev issue with the resource, please open a GitHub issue", name)
```

We probably want to update this to tell them where to open an issue as well

providers-sdk/v1/lr/go.go
```diff
@@ -183,7 +188,11 @@ func NewResource(runtime *plugin.Runtime, name string, args map[string]*llx.RawD
 		return nil, err
 	}
 
-	id := name+"\x00"+res.MqlID()
+	mqlId := res.MqlID()
+	if mqlId == "" {
+		log.Warn().Msgf("resource %s has no MQL ID defined, this is usually a dev issue with the resource, please open a Github issue", name)
```
tas50:
```suggestion
		log.Warn().Msgf("resource %s has no MQL ID defined, this is usually a dev issue with the resource, please open a GitHub issue", name)
```

providers-sdk/v1/lr/go.go
```diff
@@ -206,7 +215,11 @@ func CreateResource(runtime *plugin.Runtime, name string, args map[string]*llx.R
...
```
tas50:
```suggestion
		log.Warn().Msgf("resource %s has no MQL ID defined, this is usually a dev issue with the resource, please open a GitHub issue", name)
```

providers-sdk/v1/lr/go.go
preslavgerchev: nice, added

providers-sdk/v1/lr/go.go
```diff
@@ -167,7 +168,11 @@ func NewResource(runtime *plugin.Runtime, name string, args map[string]*llx.RawD
 		}
 
 		if res != nil {
-			id := name+"\x00"+res.MqlID()
+		  mqlId := res.MqlID()
+			if mqlId == "" {
+			  log.Warn().Msgf("resource %s has no MQL ID defined, this is usually a dev issue with the resource, please open a GitHub issue at https://github.com/mondoohq/cnquery/issues", name)
```
czunker: nit: Perhaps remove this part from the warninig:
```suggestion
			  log.Warn().Msgf("resource %s has no MQL ID defined, this is usually an issue with the resource, please open a GitHub issue at https://github.com/mondoohq/cnquery/issues", name)
```

czunker (approved):
Thanks @preslavgerchev

providers-sdk/v1/lr/go.go
preslavgerchev: done

jaym (approved):

PR #5831, opened 2025-08-07
Fix `&&` short-circuiting

jaym:
Take this example query:
```
asset.platform == 'foo' && azure.subscription.storage.account.containers != empty
```

I have no azure credentials. There is no platform 'foo'. Yet the thing after the `&&` still runs. The reason this happens is because the executor needs a value for each datapoint. We quit executing the chain that was caused by `asset.platform == 'foo' &&`, however needing all datapoints means we still execute
`azure.subscription.storage.account.containers != empty`

The reason we need a value for each datapoint is because the query is only considered finished executing once each datapoint is reported. It was possible at one point the executor spawned a goroutine and generated those in a different goroutine, so we couldn't just say at the end of the executor, report a nil value for each datapoint

### DIFF
--- a/mql/mql_test.go
+++ b/mql/mql_test.go
@@ -464,3 +464,17 @@ func TestArrayConcat(t *testing.T) {
 		},
 	})
 }
+
+func TestAndShortCircuiting(t *testing.T) {
+	x := testutils.InitTester(testutils.LinuxMock())
+	x.TestSimple(t, []testutils.SimpleTest{
+		{
+			Code:        "false && muser.error",
+			Expectation: nil,
+		},
+		{
+			Code:  "true && muser.error",
+			Error: "this is an error from the mockprovider",
+		},
+	})
+}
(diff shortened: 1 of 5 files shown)


llx/builtin_simple.go
```diff
@@ -94,11 +94,28 @@ func boolOrOpV2(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64, fLeft
 // boolAndOp behaves like boolOp, but checks if the left argument is false first
 // (and stop if it is). Only then proceeds to check the right argument.
 func boolAndOpV2(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64, fLeft func(interface{}) bool, fRight func(interface{}) bool) (*RawData, uint64, error) {
+	arg := chunk.Function.Args[0]
+
 	if bind.Value != nil && !fLeft(bind.Value) {
+		// if the left side is false, we can return false immediately
```
preslavgerchev, line 100: can we add a test for this? if we put an invalid query/something that panics in the right side to verify it never gets evaluated

PR #5834, opened 2025-08-08
`auditd` resource: Improve error handling

mm-weber:
## Implemented functionality to the `auditd:` resource:

1. Graceful Error Handling for Missing Files:
  - auditd.config('/nonexistent/path') now returns empty params instead of errors
  - auditd.rules already had graceful error handling for missing directories
  - Queries continue executing instead of failing

2. Failed vs Error State Distinction:
 - Missing configuration files result in graceful handling (empty params)
 - Other errors (like permission denied) still propagate as errors
 - Method chaining works even with missing files
3. Comprehensive Test Coverage:
 - Tests for missing auditd.conf files returning empty params
 - Tests for query continuation with missing files
 - Tests for method chaining capabilities
 - Tests for both auditd.config and auditd.rules resources

## Changes Made:

1. Updated auditd.go:
 - Enhanced parse() method to handle "not found" errors gracefully
 - Modified file() method to handle missing default config files
 - Fixed data types to use map[string]interface{} consistently
2. Comprehensive Test Suite:
 - Added TestResource_AuditdConfig_ErrorHandling with 4 test cases
 - Added TestResource_AuditdRules_ErrorHandling with 4 test cases
 - Added TestResource_AuditdConfig_FailedStateVsError with 3 test cases
 - Updated existing tests to match new graceful behavior

## What it looks like (before)

A missing audit file leads to an error using this `mql`:

```
  - uid: mondoo-linux-security-system-is-disabled-when-audit-logs-are-full
    title: Ensure system is disabled when audit logs are full
    impact: 40
    filters: |
      asset.kind != "container-image"
    mql: |
      auditd.config.params.space_left_action == /email|exec|single|halt/
      auditd.config.params.action_mail_acct == "root"
      auditd.config.params.admin_space_left_action == /halt|single/
```

```
Failing:
! Error:          Ensure system is disabled when audit logs are full
  Message:    1 error occurred:
	* file '/etc/audit/auditd.conf' not found

Scanned 1 asset
```

## What it looks like (after)
```
Failing:
✕ MEDIUM (60):    Ensure system is disabled when audit logs are full
  Query:

  Result:
    [failed] auditd.config.params.space_left_action == /email|exec|single|halt/
    auditd.config.params.action_mail_acct == "root"
    auditd.config.params.admin_space_left_action == /halt|single/

      [failed] auditd.config.params[space_left_action] == /email|exec|single|halt/
        expected: == /email|exec|single|halt/
        actual:   _
      [failed] auditd.config.params[action_mail_acct] == "root"
        expected: == "root"
        actual:   _
      [failed] auditd.config.params[admin_space_left_action] == /halt|single/
        expected: == /halt|single/
        actual:   _



Scanned 1 asset
```

### DIFF
--- a/providers/os/resources/auditd.go
+++ b/providers/os/resources/auditd.go
@@ -103,7 +103,7 @@ func (s *mqlAuditdConfig) parse(file *mqlFile) error {
 
 	ini := parsers.ParseIni(content.Data, "=")
 
-	res := make(map[string]any, len(ini.Fields))
+	res := make(map[string]interface{}, len(ini.Fields))
 	s.Params.Data = res
 	s.Params.State = plugin.StateIsSet
 
--- a/providers/os/resources/auditd.go
+++ b/providers/os/resources/auditd.go
@@ -102,7 +102,7 @@ func (s *mqlAuditdConfig) parse(file *mqlFile) error {
 
 	ini := parsers.ParseIni(content.Data, "=")
 
-	res := make(map[string]interface{}, len(ini.Fields))
+	res := make(map[string]any{}, len(ini.Fields))
 	s.Params.Data = res
 	s.Params.State = plugin.StateIsSet
 
--- a/providers/os/resources/auditd.go
+++ b/providers/os/resources/auditd.go
@@ -7,6 +7,7 @@ package resources
 import (
 	"errors"
 	"fmt"
+	"os"
 	"regexp"
 	"slices"
 	"strings"
@@ -64,6 +65,13 @@ func (s *mqlAuditdConfig) file() (*mqlFile, error) {
 		"path": llx.StringData(defaultAuditdConfig),
 	})
 	if err != nil {
+		// Check if this is a "file not found" type error
+		// errorMsg := err.Error()
+		if errors.Is(err, os.ErrNotExist) {
+			// For missing files, return nil - parse() method will handle this gracefully
+			return nil, nil
+		}
+		// For other errors (like permission issues), propagate the error
 		return nil, err
 	}
 	return f.(*mqlFile), nil
@@ -83,8 +91,8 @@ func (s *mqlAuditdConfig) parse(file *mqlFile) error {
 	content := file.GetContent()
 	if content.Error != nil {
 		// Check if this is a "file not found" type error
-		errorMsg := content.Error.Error()
-		if strings.Contains(errorMsg, "no such file") || strings.Contains(errorMsg, "does not exist") || strings.Contains(errorMsg, "not found") {
+		// errorMsg := content.Error.Error()
+		if errors.Is(content.Error, os.ErrNotExist) {
 			// Handle missing config file gracefully - set empty params
 			s.Params.Data = make(map[string]interface{})
 			s.Params.State = plugin.StateIsSet
--- a/providers/os/resources/auditd.go
+++ b/providers/os/resources/auditd.go
@@ -89,6 +89,7 @@ func (s *mqlAuditdConfig) parse(file *mqlFile) error {
 
 	content := file.GetContent()
 	if content.Error != nil {
+
 		errorMsg := content.Error.Error()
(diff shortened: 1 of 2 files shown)


providers/os/resources/auditd.go
```diff
@@ -62,6 +64,13 @@ func (s *mqlAuditdConfig) file() (*mqlFile, error) {
 		"path": llx.StringData(defaultAuditdConfig),
 	})
 	if err != nil {
+		// Check if this is a "file not found" type error
+		errorMsg := err.Error()
+		if strings.Contains(errorMsg, "no such file") || strings.Contains(errorMsg, "does not exist") || strings.Contains(errorMsg, "not found") {
```
syrull: Is this handling the different OS error responses? I don't think that there is different error responses on different OS, especially for Go..

Asserting the error message on string comparison is not a good idea. I would go for more strict approach here.

providers/os/resources/auditd.go
```diff
@@ -72,17 +81,29 @@ func (s *mqlAuditdConfig) parse(file *mqlFile) error {
 	defer s.lock.Unlock()
 
 	if file == nil {
-		return errors.New("no base auditd config file to read")
+		// Set empty params for missing config file
```
syrull, line 84: No need for excessive comments

providers/os/resources/auditd.go
```diff
@@ -72,17 +81,29 @@ func (s *mqlAuditdConfig) parse(file *mqlFile) error {
...
+		s.Params.Data = make(map[string]interface{})
+		s.Params.State = plugin.StateIsSet
+		return nil
 	}
 
 	content := file.GetContent()
 	if content.Error != nil {
+		// Check if this is a "file not found" type error
+		errorMsg := content.Error.Error()
+		if strings.Contains(errorMsg, "no such file") || strings.Contains(errorMsg, "does not exist") || strings.Contains(errorMsg, "not found") {
```
syrull: Again same as above for errs.

providers/os/resources/auditd.go
```diff
@@ -72,17 +81,29 @@ func (s *mqlAuditdConfig) parse(file *mqlFile) error {
...
+			return nil
+		}
+		// For other errors (like permission denied), propagate the error
 		return content.Error
 	}
 
 	ini := parsers.ParseIni(content.Data, "=")
 
-	res := make(map[string]any, len(ini.Fields))
+	res := make(map[string]interface{}, len(ini.Fields))
```
syrull, line 106: This seems redundant, any is basically an alias for interface

providers/os/resources/auditd.go
mm-weber: see below

providers/os/resources/auditd.go
mm-weber: Hmm, you're right it does not make any sense to do the comparisons against all those strings.
Not sure how to do it more strict.

Tried this with no avail:
```
if errors.Is(content.Error, os.ErrNotExist) {
```

Have you got a suggestions?

providers/os/resources/auditd.go
mm-weber: Hmm, you're right it does not make any sense to do the comparisons against all those strings.
Not sure how to do it more strict.

Tried this with no avail:
```
if errors.Is(content.Error, os.ErrNotExist) {
```

Have you got a suggestions?

providers/os/resources/auditd.go
mm-weber: Ok

mm-weber: The pipeline complains if we do it with any

providers/os/resources/auditd.go
syrull: Well what would be the error, I would just debug it to see, place a breakpoint there.

mm-weber:
Closing for now.

PR #5837, opened 2025-08-11
Refactored provider signing to use trusted signing

username-is-already-taken2:
### Summary
This PR migrates the windows provider file signing process from DigiCert to Azure Trusted Signer.

### Key Changes

- Removed redundant tasks: Some tasks became unnecessary after switching to Azure and were removed to simplify the workflow.

- Azure-specific tasks: Added new tasks to handle authentication required for jsign to work with Azure.  Authentication to micrsoft is done using OIDC

_Note: jsign was upgraded on the runner to version 7.1, which supports trusted signing._

### Workflow Improvements

New inputs were added to the workflow_dispatch to aid with troubleshooting:

- use-test-cert: Signs the binary using our test certificate (not publicly trusted).

### DIFF
--- a/.github/workflows/providers.yaml
+++ b/.github/workflows/providers.yaml
@@ -16,6 +16,11 @@ on:
         type: boolean
         required: false
         default: false
+      use-test-cert:
+        description: "Use test certificate profile (not publicly trusted)"
+        required: false
+        default: false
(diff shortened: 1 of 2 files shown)


sibuthomasmathew (approved):
Lgtm

PR #5838, opened 2025-08-11
✨ Detect .NET Framework as package

czunker:
.NET Framework bundled with Windows does not show up in the default package list. Query the registry keys separatly and add the disciovered version to the packages.

### DIFF
--- a/providers/os/resources/packages/windows_packages.go
+++ b/providers/os/resources/packages/windows_packages.go
@@ -25,6 +25,8 @@ import (
 	"go.mondoo.com/cnquery/v11/providers/os/resources/cpe"
 	"go.mondoo.com/cnquery/v11/providers/os/resources/powershell"
 	"go.mondoo.com/cnquery/v11/providers/os/resources/purl"
+	"go.mondoo.com/ranger-rpc/codes"
+	"go.mondoo.com/ranger-rpc/status"
 )
 
 // ProcessorArchitecture Enum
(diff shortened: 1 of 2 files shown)


providers/os/resources/packages/windows_packages.go
```diff
@@ -278,9 +280,63 @@ func (w *WinPkgManager) getLocalInstalledApps() ([]Package, error) {
...
+	}
+	packages = append(packages, dotNetFramework...)
+
 	return packages, nil
 }
 
+// getDotNetFramework returns the .NET Framework package
+func (w *WinPkgManager) getDotNetFramework() ([]Package, error) {
+	// https://learn.microsoft.com/en-us/dotnet/framework/install/how-to-determine-which-versions-are-installed#net-framework-45-and-later-versions
+	dotNet45plus := "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\NET Framework Setup\\NDP\\v4\\Full"
```
preslavgerchev, line 269: can we not put those in the `pkgs` list in `getLocalInstalledApps` ?

providers/os/resources/packages/windows_packages.go
```diff
@@ -278,9 +280,63 @@ func (w *WinPkgManager) getLocalInstalledApps() ([]Package, error) {
...
+	// https://learn.microsoft.com/en-us/dotnet/framework/install/how-to-determine-which-versions-are-installed#use-registry-editor-older-framework-versions
+	dotNet35 := "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\NET Framework Setup\\NDP\\v3.5"
+
+	return getDotNetFrameworkPackageFromRegistryKeys(dotNet45plus, dotNet35, w.platform)
+}
+
+// getDotNetFrameworkFs returns the .NET Framework package discovered on the filesystem
+func (w *WinPkgManager) getDotNetFrameworkFs() ([]Package, error) {
```
preslavgerchev, line 277: same question, but for fs scanning. do net framework packages present themselves in a different way in the registry? would be better if we can handle this in the general `getFsInstalledApps` and extend the conversion there if required

providers/os/resources/packages/windows_packages.go
```diff
@@ -278,9 +280,63 @@ func (w *WinPkgManager) getLocalInstalledApps() ([]Package, error) {
...
+	// https://learn.microsoft.com/en-us/dotnet/framework/install/how-to-determine-which-versions-are-installed#net-framework-45-and-later-versions
+	dotNet45plus := "Microsoft\\NET Framework Setup\\NDP\\v4\\Full"
+	// https://learn.microsoft.com/en-us/dotnet/framework/install/how-to-determine-which-versions-are-installed#use-registry-editor-older-framework-versions
+	dotNet35 := "Microsoft\\NET Framework Setup\\NDP\\v3.5"
+
+	return getDotNetFrameworkPackageFromRegistryKeys(dotNet45plus, dotNet35, w.platform)
```
preslavgerchev, line 283: is this the connection platform or the local one? e.g. for fs scanning, would this be the detected platform or the one where we're running the scanning from

providers/os/resources/packages/windows_packages.go
czunker: No.
They registry key items are different than the ones used for the packages.
The mapping wouldn't work.

providers/os/resources/packages/windows_packages.go
czunker: No.
They registry key items are different than the ones used for the packages.
The mapping wouldn't work.

providers/os/resources/packages/windows_packages.go
czunker: I don't know.
I used the same platform that we also use for the other packages.

providers/os/resources/packages/windows_packages.go
```diff
@@ -522,6 +586,53 @@ func getPackageFromRegistryKeyItems(children []registry.RegistryKeyItem, platfor
 	return pkg
 }
 
+// getDotNetFrameworkPackageFromRegistryKeyItems returns the .NET Framework package from the registry key items
+func getDotNetFrameworkPackageFromRegistryKeyItems(items []registry.RegistryKeyItem, platform *inventory.Platform) *Package {
```
preslavgerchev, line 534: cant we use `getPackageFromRegistryKeyItems` here? this has the same signature

providers/os/resources/packages/windows_packages.go
```diff
@@ -278,9 +280,63 @@ func (w *WinPkgManager) getLocalInstalledApps() ([]Package, error) {
...
+	// https://learn.microsoft.com/en-us/dotnet/framework/install/how-to-determine-which-versions-are-installed#use-registry-editor-older-framework-versions
+	dotNet35 := "Microsoft\\NET Framework Setup\\NDP\\v3.5"
+
+	return getDotNetFrameworkPackageFromRegistryKeys(dotNet45plus, dotNet35, w.platform)
+}
+
+// getDotNetFrameworkPackageFromRegistryKeys returns the .NET Framework package from the registry keys
+func getDotNetFrameworkPackageFromRegistryKeys(dotNet45plus, dotNet35 string, platform *inventory.Platform) ([]Package, error) {
+	items, err := registry.GetNativeRegistryKeyItems(dotNet45plus)
+	if err != nil && status.Code(err) != codes.NotFound {
```
preslavgerchev, line 289: i dont think that the registry code returns rpc error codes

czunker: It does:
https://github.com/mondoohq/cnquery/blob/9c3ef878966a270b9004692b05f5de9793f1e6c8/providers/os/registry/registrykey_windows.go#L54

PR #5857, opened 2025-08-19
✨  Read tags from vsphere

claytoncasey01:
This PR adds tags to both hosts and vm. 

Screenshot 2025-08-19 at 3 47 50 PM

### DIFF
--- a/providers/vsphere/resources/datacenter.go
+++ b/providers/vsphere/resources/datacenter.go
@@ -10,11 +10,11 @@ import (
 	"github.com/vmware/govmomi/object"
 	"go.mondoo.com/cnquery/v11/llx"
 	"go.mondoo.com/cnquery/v11/providers-sdk/v1/plugin"
+	"go.mondoo.com/cnquery/v11/providers-sdk/v1/util/convert"
 	"go.mondoo.com/cnquery/v11/providers/vsphere/connection"
 	"go.mondoo.com/cnquery/v11/providers/vsphere/resources/resourceclient"
 	"go.mondoo.com/cnquery/v11/types"
 )
-
 
 func newVsphereHostResources(vClient *resourceclient.Client, runtime *plugin.Runtime, vhosts []*object.HostSystem) ([]interface{}, error) {
 	mqlHosts := make([]interface{}, len(vhosts))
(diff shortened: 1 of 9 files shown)


providers/vsphere/resources/resourceclient/vm.go
```diff
@@ -75,6 +80,50 @@ func (c *Client) VirtualMachineByMoid(moid types.ManagedObjectReference) (*objec
 	return nil, errors.New("reference is not a valid virtual machine")
 }
 
+// GetVmTags retrieves tags for a virtual machine using the vSphere vAPI
+// If the vAPI is not available, it will return an empty array instead of an error.
+// This maintains backward compatibility with vSphere environments that don't use tags.
+func (c *Client) GetVmTags(ctx context.Context, vmRef types.ManagedObjectReference, conf *inventory.Config) []string {
```
claytoncasey01: This and the similar one for Host tags had to be added because the existing `vmInfo` and `hostInfo` didn't actually have the tags we wanted, those only seemed to contain system tags, for example: `SYSTEM/COM.VMWARE.VIM.VC`. 

`Mondoo:business criticality` for example was not present on `vmInfo` or `hostInfo`.

providers/vsphere/resources/datacenter.go
```diff
@@ -29,15 +31,31 @@ func newVsphereHostResources(vClient *resourceclient.Client, runtime *plugin.Run
 		}
 
 		var name string
+		var tags []string
 		if hostInfo != nil {
 			name = hostInfo.Name
+			simpleTags := extractTagKeys(hostInfo.Tag)
```
claytoncasey01, line 37: We may want to avoid these tags `SYSTEM/COM.VMWARE.VIM.VC` entirely?

providers/vsphere/resources/discovery.go
```diff
@@ -124,6 +124,22 @@ func discoverDatacenter(conn *connection.VsphereConnection, datacenterResource *
...
+			}
+
+			// Add tags as labels with vsphere.vmware.com/tag/ prefix
+			if mqlHost.Tags.Error == nil {
+				for _, tag := range mqlHost.Tags.Data {
+					if tagStr, ok := tag.(string); ok {
+						labels["vsphere.vmware.com/tag/"+tagStr] = "true"
+					}
+				}
+			}
```
czunker, line 141: This looks to me the same as on L172-L186. Can you extract it into something like `createLabels`?

providers/vsphere/resources/resourceclient/host.go
```diff
@@ -140,6 +145,49 @@ func HostLicenses(client *vim25.Client, hostID string) ([]types.LicenseManagerLi
...
+		if err != nil {
+			// If we can't get category, just use tag name
+			tagStrings[i] = tag.Name
+			continue
+		}
+		tagStrings[i] = fmt.Sprintf("%s:%s", category.Name, tag.Name)
+	}
+
+	return tagStrings
+}
```
czunker: This looks to me the same as `GetVmTags`. Is it possible to have the main logic of these two extracted into one func?

czunker:
Tags on ESXi hosts also work:
```
  1: {
    name: "Mondoo Datacenter-1"
    hosts: [
      0: {
        tags: [
          0: "Mondoo:owner"
        ]
```

PR #5858, opened 2025-08-20
🐛 Reorder the auditd rule resource operator regex parsing

tas50:
Ensure that ">=" and "<=" parse properly

Before this the fields were not getting parses correctly when using >= or <= operators:

```
    fields: [
      0: {
        key: "auid"
        op: ">" <-- This should be ">="
        value: "=1000" <-- This should be "1000"
      }
```


arlimus (approved):
Thank you for the quick fix, we need to separately test the parser in a follow-up

PR #5859, opened 2025-08-20
⭐️ Refactor device connection and partition mounting

preslavgerchev:
* Add new structs
    * `Partition` : simply indicates a partition that should be mounted. It has no mountpoint information as we do not differentiate between a mounted and a non-mounted partition here. The filtering happens when doing `lsblk` and using `BlockDevices`
    * `MountedPartitionInput`: `Partition` + `MountOptions` and `MountDir`
    * `MountedPartition`: `Partition` + `MountOptions` and `MountPoint`. We could unite this and the input struct but I like keeping them split. 
* All connections now do their own mounting in their own tmp folders. This means that connections do not rely on one another anymore. 
* Improve the device manager interface to make the functions more intuitive. 
* Improve the volume mounter interface to make it more intuitive.
* Overall code improvements where possible

### DIFF
--- a/providers/os/connection/device/device_manager.go
+++ b/providers/os/connection/device/device_manager.go
@@ -11,9 +11,9 @@ type DeviceManager interface {
 	// Name returns the name of the device manager
 	Name() string
 	// IdentifyMountTargets returns a list of partitions that match the given options and can be mounted
-	IdentifyMountTargets(opts map[string]string) ([]*snapshot.PartitionInfo, error)
-	// Mounts the partition and returns the directory it was mounted to
-	Mount(pi *snapshot.PartitionInfo) (string, error)
+	IdentifyMountTargets(opts map[string]string) ([]*snapshot.Partition, error)
+	// Mounts the partition and returns the directories it was mounted to
+	Mount(partitions []*snapshot.Partition) ([]*snapshot.MountedPartition, error)
 	// UnmountAndClose unmounts the partitions from the specified dirs and closes the device manager
 	UnmountAndClose()
 }
--- a/providers/os/connection/device/device_manager.go
+++ b/providers/os/connection/device/device_manager.go
@@ -12,7 +12,7 @@ type DeviceManager interface {
 	Name() string
 	// IdentifyMountTargets returns a list of partitions that match the given options and can be mounted
 	IdentifyMountTargets(opts map[string]string) ([]*snapshot.Partition, error)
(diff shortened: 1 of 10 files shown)


providers/os/connection/snapshot/volumemounter.go
```diff
@@ -172,3 +178,10 @@ func (m *volumeMounter) RemoveTempScanDir() error {
 
 	return errors.Join(errs...)
 }
+
+func PathDepth(path string) int {
```
VasilSirakov, line 179: i like that this function is simple but in every comparison of the sorts, this will allocate a string slice due to the `strings.Split`.

if we ever need to optimize this we could also consider `strings.Count(str, "/")` to avoid the extra allocation in case we can assume we can't have paths like `a//b` for example. `path.Clean(str)` can also be useful in this scenario as a more thorough cleanup.

providers/os/connection/device/device_manager.go
```diff
@@ -11,9 +11,9 @@ type DeviceManager interface {
 	// Name returns the name of the device manager
 	Name() string
 	// IdentifyMountTargets returns a list of partitions that match the given options and can be mounted
-	IdentifyMountTargets(opts map[string]string) ([]*snapshot.PartitionInfo, error)
-	// Mounts the partition and returns the directory it was mounted to
-	Mount(pi *snapshot.PartitionInfo) (string, error)
+	IdentifyMountTargets(opts map[string]string) ([]*snapshot.Partition, error)
+	// Mounts the partition and returns the directories it was mounted to
```
slntopp: nit
```suggestion
	// Mounts partitions and returns the directories they were mounted to
```

providers/os/connection/snapshot/blockdevices.go
```diff
@@ -203,15 +207,16 @@ func (device BlockDevice) GetPartitions(includeBoot bool, includeMounted bool) (
 			return false
 		}
 
-		// skip mounted partitions unless includeMounted is true
+		// sltekip mounted partitions unless includeMounted is true
```
slntopp: keyboard slap?

```suggestion
		// skip mounted partitions unless includeMounted is true
```

slntopp (approved):

providers/os/connection/snapshot/volumemounter.go
slntopp: Not urgent, but makes sense to use `strings.Count`, `path.Clean` is unnecessary

PR #5861, opened 2025-08-21
🐛 Fix non-breaking spaces in MS SQL GDR packages

czunker:
MS SQL GDR updates may contain non-breaking spaces:
Screenshot from 2025-08-21 15-30-13
Screenshot from 2025-08-21 15-29-39

### DIFF
--- a/providers/os/resources/packages/windows_packages.go
+++ b/providers/os/resources/packages/windows_packages.go
@@ -72,8 +72,8 @@ var (
 		WinArchX86OnArm64: "x86onarm",
 	}
 
-	sqlGDRUpdateRegExp = regexp.MustCompile(`^GDR \d+ .+ SQL Server \d+ \(KB\d+\)`)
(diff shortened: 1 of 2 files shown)


providers/os/resources/packages/windows_packages.go
```diff
@@ -715,6 +715,8 @@ func createPackage(name, version, format, arch, publisher, installLocation strin
 		purlType = purl.TypeAppx
 	}
 
+	// replace non-breaking spaces with regular spaces
+	name = strings.ReplaceAll(name, " ", " ")
```
czunker: GitHub does not show the difference, but there is one:
image

chris-rock: Can we use the unicode value here, would be much easier to understand the code.

czunker: Done.

PR #5862, opened 2025-08-22
✨ Detect installed .NET Framework runtime version

czunker:
image


providers/os/resources/packages/windows_packages.go
```diff
@@ -107,6 +107,11 @@ Get-ItemProperty (@(
 Select-Object -Property DisplayName,DisplayVersion,Publisher,EstimatedSize,InstallSource,UninstallString,InstallLocation | ConvertTo-Json -Compress
 `
 
+// We need to fill in the path collected from the registry
+const dotNetClrVersionScript = `
```
chris-rock, line 111: I assume this is only for .NET Framework, not .NET Core?

czunker: I'm not sure what you mean by .NET Core, but this is only meant for the .NET Framework

PR #5863, opened 2025-08-25
✨ add missing resources.first and resources.last

arlimus:
So now we can do things like:

```coffee
users.first
services.last
```


scottford-io (approved):
Looks good! Thanks @arlimus ⭐

PR #5872, opened 2025-08-27
🧹 Replace dprecated link checker

czunker:
See https://github.com/gaurav-nelson/github-action-markdown-link-check for more details.


arlimus (approved):
thanks for replacing it, yeah that old one was deprecated for good

PR #5875, opened 2025-08-27
Update goreleaser for v12

username-is-already-taken2:
I’ve made some updates to the goreleaser config to reflect the major version change. I also stumbled upon a few other references to version 11 that were hiding under the carpet.


arlimus:
Thank you @username-is-already-taken2 !!

PR #5893, opened 2025-08-29
🧹 v13: Remove deprecated MQL properties

tas50:
## Summary

Removes deprecated MQL resource properties that were replaced in prior versions. This is the v13 deprecation cleanup — all removed fields already have replacements in place.

## Removed Fields

| Resource | Removed Field | Replacement |
|----------|--------------|-------------|
| `aws.elb.loadbalancer` | `vpcId` | `vpc` |
| `aws.elb.loadbalancer` | `createdTime` | `createdAt` |
| `aws.cloudfront.function` | `createdTime` | `createdAt` |
| `aws.s3.bucket` | `createdTime` | `createdAt` |
| `aws.s3.bucket.policy` | `id` | — |
| `aws.backup.vaultRecoveryPoint` | `creationDate` | `createdAt` |
| `aws.dynamodb.table` | `createdTime` | `createdAt` |
| `aws.rds.dbcluster` | `createdTime` | `createdAt` |
| `aws.rds.dbinstance` | `createdTime` | `createdAt` |
| `azure.subscription.cloudDefenderService.securityContact` | `alertNotifications` | `notificationSources` |
| `gcp.organization` | `lifecycleState` | `state` |
| `gcp.project` | `number` | `id` |
| `gcp.project` | `lifecycleState` | `state` |
| `gcp.project.sqlService.instance` | `currentDiskSize` | — |
| `gcp.project.sqlService.instance` | `project` | `projectId` |
| `gcp.project.gkeService.cluster` | `zone` | `location` |
| `platform` | `vulnerabilityReport` | `vulnmgmt` |
| `parse.openpgp` | `path` | `file.path` |
| `sshd.config` | `content` | `file.content` |
| `docker.image` | `virtualsize` | `size` |
| `yum.repo` | `filename` | `file.path` |
| `windows` | `features` | `windows.serverFeatures` |
| `vcd.externalNetwork` | `id` | `urn` |
| `asset` (vsphere) | `vulnerabilityReport` | `vulnmgmt` |
| `esxi` | `host` | — |
| `esxi` | `vm` | — |

## Removed Resources

| Resource | Replacement |
|----------|-------------|
| `platform.eol` | — |
| `windows.feature` | `windows.serverFeature` |

## Review Fixes

- Fixed S3 bucket `__id` cache collision by passing `__id` explicitly in `CreateResource`
- Removed `"id"` from S3 bucket policy `CreateResource` (field no longer in `.lr`)
- Removed `asset.vulnerabilityReport` from vsphere provider

## Test plan

- [ ] Verify S3 bucket queries return distinct buckets
- [ ] Verify S3 bucket policy queries don't error at runtime
- [ ] `make test/go/plain` passes

### DIFF
--- a/providers/aws/resources/aws.lr.go
+++ b/providers/aws/resources/aws.lr.go
@@ -37715,7 +37715,12 @@ func createAwsS3Bucket(runtime *plugin.Runtime, args map[string]*llx.RawData) (p
 		return res, err
 	}
 
-	// to override __id implement: id() (string, error)
+	if res.__id == "" {
+		res.__id, err = res.id()
+		if err != nil {
+			return nil, err
+		}
+	}
 
 	if runtime.HasRecording {
 		args, err = runtime.ResourceFromRecording("aws.s3.bucket", res.__id)


providers/os/resources/windows.go
```diff
@@ -259,7 +173,7 @@ func initWindowsServerFeature(runtime *plugin.Runtime, args map[string]*llx.RawD
 	}
 	winResource := obj.(*mqlWindows)
 
-	features := winResource.GetFeatures()
+	features := winResource.GetServerFeatures()
```
tas50, line 176: This was calling the other resource the whole time

providers/aws/resources/aws_elb.go
```diff
@@ -74,19 +74,31 @@ func (a *mqlAwsElb) getClassicLoadBalancers(conn *connection.AwsConnection) []*j
...
+					if lb.VPCId != nil {
+						mqlVpc, err := NewResource(a.MqlRuntime, "aws.vpc",
+							map[string]*llx.RawData{
+								"arn": llx.StringData(fmt.Sprintf(vpcArnPattern, region, conn.AccountId(), convert.ToValue(lb.VPCId))),
+							})
+						if err != nil {
+							return nil, err
+						}
+						// update the vpc setting
+						args["vpc"] = llx.ResourceData(mqlVpc, mqlVpc.MqlName())
```
tas50: The `vpc` field was never implemented for the classic LB. This is just that same logic from the app LBs copied over

providers/vsphere/resources/vsphere.lr
```diff
@@ -264,6 +256,9 @@ private vsphere.vmknic @defaults("name") {
   tags []string
 }
 
+// VMware ESXi resource
+esxi {}
```
tas50, line 256: Not sure if this is the right way to do this, but without it things blew up

mqlc/mqlc_test.go
```diff
@@ -1271,10 +1271,6 @@ func TestCompiler_ResourceExpansion(t *testing.T) {
 				Type:    string(types.Range),
 				Binding: (3 << 32) | 1,
 			}, res.CodeV2.Blocks[2].Chunks[3])
-			assertFunction(t, "content", &llx.Function{
```
tas50, line 1318: This field has been removed so we can't use it for mqlc testing anymore

tas50:
Larger question on this one is should we actually remove sshd.config.content. Yes you can access it via the file resource, but the content field is something we have in a number of our resources.

providers/os/resources/windows.go
arlimus: great catch!

providers/vsphere/resources/vsphere.lr
arlimus: yeah this is solid, it exposes the resource on the top layer and fits well into the provider

mqlc/mqlc_test.go
arlimus: nice, no harm removing it, thank you!

arlimus:
> Larger question on this one is should we actually remove sshd.config.content. Yes you can access it via the file resource, but the content field is something we have in a number of our resources.

I think it's ok to remove it, but we may want to wait for a full major version before we do it? (that said, it is rarely used, so I'd be ok to remove it in v12 as well)

tas50:
@czunker Before we kick off cnquery 13 I'd love your insight here on where we can cleanup old vuln stuff within the client. I did my best to touch only the parts that were marked for deprecation, but maybe there's more to do.

providers/vsphere/resources/vsphere.lr
czunker: I think, this can also be removed.

tas50: Done — removed `asset.vulnerabilityReport` from the vsphere `.lr` file, deleted the implementation in `asset_advisories.go`, and regenerated.

providers/aws/resources/aws.lr
```diff
@@ -3,7 +3,7 @@
 
 import "../../network/resources/network.lr"
 
-option provider = "go.mondoo.com/cnquery/v9/providers/aws"
```
tas50, line 6: @arlimus is there a reason these weren't done before?

arlimus: yes, we will change all the ID strings for these in v14. In v13 we are still chained to backwards-compatibility of cross-provider calling, which has forced us to keep these really old IDs around. v13 remedies that and in v14 we can finally remove it. 

We are also planning to remove the version indictator, since we are moving at a fast pace and it hasn't helped in the way we intended it. The providers sdk will remain versioned, i.e. the foundation for all of these, but the rest won't have a version string in their imports anymore, both in go-code as well as in these provider IDs.

tldr: we have to revert this for now

arlimus (changes requested):

PR #5907, opened 2025-09-03
🧹 Update k8s deps to the latest

tas50:
Now that we're not including the PSP resource we can update to the latest dep


czunker (approved):
Thanks @tas50

PR #5919, opened 2025-09-04
:broom: Removes v13 deprecations in `mondoo-vmware-inventory.mql.yaml` 

mm-weber:
Meant to fix: https://github.com/mondoohq/cnquery/issues/5914

Does this
`esxi.host` -> replaced with `vsphere.host`

### DIFF
--- a/content/mondoo-vmware-inventory.mql.yaml
+++ b/content/mondoo-vmware-inventory.mql.yaml
@@ -57,45 +57,45 @@ packs:
         title: VMware ESXi Installed packages
         filters: asset.platform == "vmware-esxi"
         mql: |
-          esxi.host.packages
+          vsphere.host.packages
       - uid: mondoo-vmware-asset-inventory-esxi-services
         title: VMware ESXi Services
         filters: asset.platform == "vmware-esxi"
         mql: |
-          esxi.host.services
+          vsphere.host.services
       - uid: mondoo-vmware-asset-inventory-esxi-acceptance-level
         title: VMware ESXi Acceptance Level
         filters: asset.platform == "vmware-esxi"
         mql: |
-          esxi.host.acceptanceLevel
+          vsphere.host.acceptanceLevel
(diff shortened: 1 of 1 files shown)


atomic111:
this query pack should provide raw data on the vcenter an esxi level.

may be we should apply filters on the query level or work with variant

content/mondoo-vmware-inventory.mql.yaml
```diff
@@ -50,52 +50,52 @@ packs:
...
       - uid: mondoo-vmware-asset-inventory-esxi-kernel-modules
         title: VMware ESXi Kernel modules
-        filters: asset.platform == "vmware-esxi"
+        filters: asset.platform == "vmware-vsphere"
         mql: |
           vsphere.host.kernelModules
       - uid: mondoo-vmware-asset-inventory-esxi-installed-packages
         title: VMware ESXi Installed packages
-        filters: asset.platform == "vmware-esxi"
+        filters: asset.platform == "vmware-vsphere"
```
atomic111: this is not working on the vcenter level

content/mondoo-vmware-inventory.mql.yaml
```diff
@@ -50,52 +50,52 @@ packs:
           vsphere.datacenters { vms { name advancedSettings["guestInfo.detailed.data"] properties["guest"]["guestState"] } }
       - uid: mondoo-vmware-asset-inventory-esxi-kernel-modules
         title: VMware ESXi Kernel modules
-        filters: asset.platform == "vmware-esxi"
```
chris-rock: This does not seem to be correct. Why do we change the asset type? The resources are defined here https://github.com/mondoohq/cnquery/blob/main/providers/vsphere/resources/vsphere.lr#L107-L113

content/mondoo-vmware-inventory.mql.yaml
atomic111: and it gets also not executed on the esxi level

tas50 (approved):

PR #5931, opened 2025-09-08
🧹 Simplify asset detection fns.

VasilSirakov:


providers/os/connection/device/device_connection.go
```diff
@@ -227,54 +227,42 @@ func TryDetectAssetFromPath(connId uint32, path string, conf *inventory.Config,
...
 		return nil, err
 	}
 
-	return tryDetectAsset(fsConn, path, conf, asset)
+	return fsConn, nil
 }
 
-func tryDetectAsset(fsConn *fs.FileSystemConnection, path string, conf *inventory.Config, asset *inventory.Asset) (*fs.FileSystemConnection, error) {
-	p, ok := detector.DetectOS(fsConn)
+func TryDetectAsset(conn shared.Connection, path string, conf *inventory.Config, asset *inventory.Asset) error {
```
VasilSirakov, line 238: this fn is more generic now that it accepts `shared.Connection`. perhaps we should move it to a more appropriate location?

PR #5935, opened 2025-09-12
🐛 Fix recordings, do not store assets if they do not have an MRN or an id

preslavgerchev:
* Do not store assets without MRN or an id. We do not know what to do with those assets, so best to not just store them at all
* Adding an asset by MRN after it's been added by id will now overwrite the asset
* Ensure `runtime.AssetUpdated` gets called when we update the asset's MRN. Recordings that get sent upstream are only retrieved by MRN so we need to ensure that every time the MRN gets updated we need to update the recording
* Some tests for adding an asset to a recording. We should start adding more of those to make the recording bug-free now that we are storing it upstream

To test, first run `make cnquery/install`, then run this qp:
```
packs:
  - uid: rec-test
    name: Recording test
    filters:
      - mql: asset.family.contains("unix")
    queries:
      - uid: packages
        title: packages
        mql: packages
      - uid: users
        title: users
        mql: users
      - uid: services
        title: services
        mql: services
```

```
~/go/bin/cnquery scan -f qp.yaml --record recording.json
```

Then load the recording:
```
~/go/bin/cnquery shell --use-recording recording.json
```

Accessing `packages`, `users` and `services` should be instant

### DIFF
--- a/explorer/scan/local_scanner.go
+++ b/explorer/scan/local_scanner.go
@@ -316,11 +316,15 @@ func (s *LocalScanner) distributeJob(ctx context.Context, job *Job, upstream *up
 			// attach the asset details to the assets list
 			for i := range batch {
 				asset := batch[i].Asset
+				runtime := batch[i].Runtime
 				log.Debug().Str("asset", asset.Name).Strs("platform-ids", asset.PlatformIds).Msg("update asset")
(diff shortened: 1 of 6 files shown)


vjeffrey (approved):
thank u! works beautifully!

PR #5939, opened 2025-09-15
fix link-check job

pratik-mondoo:


.github/workflows/link-check.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v5
       - name: markdown-link-check
-        uses: tcort/markdown-link-check@ef7e09486e579ba7479700b386e7ca90f34cbd0a # v3.13.7
+        uses: gaurav-nelson/github-action-markdown-link-check@v1
```
czunker: Can we also use a hash for this action?
We should migrate to hashes instead of tags. Dependabot is even able to update hashes.

czunker (approved):
Thanks @pratik-mondoo

PR #5947, opened 2025-09-19
✨ Pin GitHub actions to hashes

czunker:
Use hashes instead of mutable tags.

We can still use dependabot for the updates. And according to this [issue](https://github.com/dependabot/dependabot-core/issues/4691), dependabot will also take care of the human-readable comments.


chris-rock:
Yes we should do that! Thank you @czunker

PR #5965, opened 2025-09-23
🧹 fix default tf provider

chris-rock:


providers/defaults.go
```diff
@@ -470,7 +470,7 @@ var DefaultProviders Providers = map[string]*Provider{
 	"terraform": {
 		Provider: &plugin.Provider{
 			Name:            "tailscale",
-			ID:              "go.mondoo.com/cnquery/v12/providers/tailscale",
+			ID:              "go.mondoo.com/cnquery/v9/providers/terraform",
```
jaym, line 473: same doesn't look right here. The name above is terraform

chris-rock: right, there was even more wrong

PR #5978, opened 2025-09-29
🐛 make startup-script attribute multiline string for gcp

imilchev:
`instance/attributes/startup-script` should be treated as a multi-line string and we should not attempt to parse it as json or anything else. This fixes an issue where having a bash script as a `startup-script` becomes an endless loop


username-is-already-taken2 (approved):
LGTM

PR #5987, opened 2025-10-01
:sparkles: extend aws.account. fields

mm-weber:


providers/aws/resources/aws.lr.go
```diff
@@ -27,6 +27,10 @@ func init() {
 			Init: initAwsAccount,
 			Create: createAwsAccount,
 		},
+		"aws.account.alternateContact": {
```
chris-rock, line 231: The code needs to be rebased to latest main and the latest lr compiler should be installed. Then we see the new constants here

providers/aws/resources/aws_account.go
```diff
@@ -102,6 +105,171 @@ func (c *mqlAwsAccount) tags() (map[string]any, error) {
...
+		resp, err := client.GetAlternateContact(context.TODO(), &account.GetAlternateContactInput{
+			AlternateContactType: cType,
+		})
+
+		if err != nil {
+			// Check if contact not configured (ResourceNotFoundException)
+			var notFoundErr *types.ResourceNotFoundException
+			if errors.As(err, &notFoundErr) {
+				// Contact not configured - create resource with exists=false
+				contact, resErr := CreateResource(a.MqlRuntime, "aws.account.alternateContact",
```
chris-rock: We just introduced generated resources so that this resource name is generated by lr compiler, see https://github.com/mondoohq/cnquery/blob/main/providers/aws/resources/aws.lr.go#L19-L26. You get the new resource names when you use the latest lr compiler.

providers/aws/resources/aws_account.go
```diff
@@ -102,6 +105,171 @@ func (c *mqlAwsAccount) tags() (map[string]any, error) {
...
+				}
+				contacts = append(contacts, contact)
+				continue
+			}
+			// Other error - return it
+			return nil, err
+		}
+
+		// Contact is configured - create resource with data
+		contact, err := CreateResource(a.MqlRuntime, "aws.account.alternateContact",
```
chris-rock: Needs to be updated to the go variable for the resource

vjeffrey (approved):

PR #5989, opened 2025-10-01
✨ Extend AWS Provider: aws.iam.samlProviders 

mm-weber:
Added two new private resources:

#### `aws.iam.samlProvider`
```lr
private aws.iam.samlProvider @defaults("arn name") {
  arn string                    // ARN of the SAML provider
  name string                   // Name of the SAML provider
  createdAt time                // Creation timestamp
  validUntil time              // SAML metadata document expiration
  metadataDocument string      // SAML metadata document XML
  tags map[string]string       // Tags on the provider
}
```

#### `aws.iam.oidcProvider`
```lr
private aws.iam.oidcProvider @defaults("arn url") {
  arn string                   // ARN of the OIDC provider
  url string                   // URL of the identity provider
  clientIds []string          // List of client IDs (audiences)
  thumbprints []string        // List of server certificate thumbprints
  createdAt time              // Creation timestamp
  tags map[string]string      // Tags on the provider
}
```

#### Extended `aws.iam` Resource
Added two new fields to the main `aws.iam` resource:
```lr
samlProviders() []aws.iam.samlProvider    // List of SAML 2.0 identity providers
oidcProviders() []aws.iam.oidcProvider    // List of OIDC identity providers
```

### 2. Implementation (`providers/aws/resources/aws_iam.go`)

Implemented four methods:

#### `samlProviders()` - Lines 1457-1509
- Uses AWS API `iam:ListSAMLProviders` to list all SAML providers
- Fetches detailed information using `iam:GetSAMLProvider` for each provider
- Extracts provider name from ARN
- Creates MQL resources with all required fields including tags

#### `oidcProviders()` - Lines 1511-1555
- Uses AWS API `iam:ListOpenIDConnectProviders` to list all OIDC providers
- Fetches detailed information using `iam:GetOpenIDConnectProvider` for each provider
- Creates MQL resources with all required fields including tags

#### Resource ID Methods
- `mqlAwsIamSamlProvider.id()` - Returns the provider ARN
- `mqlAwsIamOidcProvider.id()` - Returns the provider ARN

### 3. Test File (`providers/aws/resources/aws_iam_identity_providers_test.go`)

Created comprehensive unit tests:
- Test listing SAML providers
- Test listing OIDC providers
- Test accessing SAML provider fields
- Test accessing OIDC provider fields

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -718,6 +718,10 @@ aws.iam {
   // List of server certificates stored in IAM
   serverCertificates() []dict
   instanceProfiles() []aws.iam.instanceProfile
+  // List of SAML 2.0 identity providers configured in IAM
+  samlProviders() []aws.iam.samlProvider
+  // List of OpenID Connect (OIDC) identity providers configured in IAM
+  oidcProviders() []aws.iam.oidcProvider
 }
 
 // Entry in AWS IAM credential report
@@ -913,6 +917,38 @@ private aws.iam.virtualmfadevice @defaults("serialNumber") {
   enableDate time
   // User associated with the MFA device
   user() aws.iam.user
+}
+
+// AWS IAM SAML 2.0 identity provider
+private aws.iam.samlProvider @defaults("arn name") {
+  // ARN of the SAML provider
+  arn string
+  // Name of the SAML provider
+  name string
+  // Time when the SAML provider was created
+  createdAt time
+  // SAML metadata document expiration time
+  validUntil time
+  // SAML metadata document in XML format
+  metadataDocument string
+  // Tags associated with the SAML provider
+  tags map[string]string
+}
+
+// AWS IAM OpenID Connect (OIDC) identity provider
+private aws.iam.oidcProvider @defaults("arn url") {
+  // ARN of the OIDC provider
+  arn string
+  // URL of the identity provider
+  url string
+  // List of client IDs (audiences) for the OIDC provider
+  clientIds []string
+  // List of server certificate thumbprints for the OIDC provider
+  thumbprints []string
+  // Time when the OIDC provider was created
+  createdAt time
+  // Tags associated with the OIDC provider
+  tags map[string]string
 }
 
 // AWS IAM Access Analyzer resource (for assessing the configuration of AWS IAM Access Analyzer)
--- a/providers/aws/resources/aws.lr.go
+++ b/providers/aws/resources/aws.lr.go
@@ -247,6 +247,14 @@ func init() {
 			// to override args, implement: initAwsIamVirtualmfadevice(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error)
 			Create: createAwsIamVirtualmfadevice,
 		},
+		"aws.iam.samlProvider": {
+			// to override args, implement: initAwsIamSamlProvider(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error)
+			Create: createAwsIamSamlProvider,
+		},
+		"aws.iam.oidcProvider": {
+			// to override args, implement: initAwsIamOidcProvider(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error)
+			Create: createAwsIamOidcProvider,
+		},
 		"aws.iam.accessAnalyzer": {
 			// to override args, implement: initAwsIamAccessAnalyzer(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error)
 			Create: createAwsIamAccessAnalyzer,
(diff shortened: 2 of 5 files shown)


providers/aws/resources/aws.lr
```diff
@@ -718,6 +718,10 @@ aws.iam {
   // List of server certificates stored in IAM
   serverCertificates() []dict
   instanceProfiles() []aws.iam.instanceProfile
+  // List of SAML 2.0 identity providers configured in IAM
+  samlProviders() []aws.iam.samlProvider
+  // List of OpenID Connect (OIDC) identity providers configured in IAM
+  oidcProviders() []aws.iam.oidcProvider
```
tas50, line 764: toss oidc in expect.txt

mm-weber: Fixed

providers/aws/resources/aws.lr
```diff
@@ -718,6 +718,10 @@ aws.iam {
   // List of server certificates stored in IAM
   serverCertificates() []dict
   instanceProfiles() []aws.iam.instanceProfile
```
chris-rock, line 760: lets use the opportunity to add the missing comment here

providers/aws/resources/aws.lr.go
```diff
@@ -247,6 +247,14 @@ func init() {
 			// to override args, implement: initAwsIamVirtualmfadevice(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error)
 			Create: createAwsIamVirtualmfadevice,
 		},
+		"aws.iam.samlProvider": {
```
chris-rock, line 462: lets use latest lr compiler

providers/aws/resources/aws_iam_identity_providers_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"testing"
+
+	"go.mondoo.com/cnquery/v12/providers-sdk/v1/testutils"
+)
+
+func TestAwsIamSamlProviders(t *testing.T) {
+	t.Run("list SAML providers", func(t *testing.T) {
+		res := testutils.InitTester(t, testutils.LinuxMock())
```
chris-rock: I think we should use an AWS Mock here. This test should not even pass with a LinuxMock

providers/aws/resources/aws_iam.go
```diff
@@ -1453,3 +1453,111 @@ func initAwsIamInstanceProfile(runtime *plugin.Runtime, args map[string]*llx.Raw
...
+		return nil, errors.Wrap(err, "could not gather aws iam oidc providers")
+	}
+
+	// For each OIDC provider, fetch detailed information
+	for _, provider := range listResp.OpenIDConnectProviderList {
+		if provider.Arn == nil {
+			continue
+		}
+
+		getResp, err := svc.GetOpenIDConnectProvider(ctx, &iam.GetOpenIDConnectProviderInput{
```
chris-rock: Since we need to do an additional call, I think it would be best to create the resource with the arn only and then fetch the data dynamically on the resource when the user asks more.

providers/aws/resources/aws_iam.go
```diff
@@ -1453,3 +1453,111 @@ func initAwsIamInstanceProfile(runtime *plugin.Runtime, args map[string]*llx.Raw
...
+		return nil, errors.Wrap(err, "could not gather aws iam saml providers")
+	}
+
+	// For each SAML provider, fetch detailed information
+	for _, provider := range listResp.SAMLProviderList {
+		if provider.Arn == nil {
+			continue
+		}
+
+		getResp, err := svc.GetSAMLProvider(ctx, &iam.GetSAMLProviderInput{
```
chris-rock: I think we should create the resource with the arn only and then only fetch additional data when required. See example here https://github.com/mondoohq/cnquery/blob/main/providers/os/resources/sshd.go#L156-L247

providers/aws/resources/aws_iam.go
```diff
@@ -1453,3 +1453,111 @@ func initAwsIamInstanceProfile(runtime *plugin.Runtime, args map[string]*llx.Raw
...
+
+		getResp, err := svc.GetOpenIDConnectProvider(ctx, &iam.GetOpenIDConnectProviderInput{
+			OpenIDConnectProviderArn: provider.Arn,
+		})
+		if err != nil {
+			log.Warn().Err(err).Str("arn", *provider.Arn).Msg("could not get oidc provider details")
+			continue
+		}
+
+		mqlOidcProvider, err := CreateResource(a.MqlRuntime, "aws.iam.oidcProvider",
```
chris-rock: Lets use resource variable generated by latest lr compiler

providers/aws/resources/aws.lr
syrull: done

providers/aws/resources/aws_iam_identity_providers_test.go
syrull: We are not testing the other resources, this is probably generated by AI so I am removing it

syrull:
I've implemented lazy loading of the `mqlAwsIamSamlProvider` but it has to be tested

providers/aws/resources/aws_iam.go
mm-weber: Fixed, with lazy loading

providers/aws/resources/aws_iam.go
mm-weber: Fixed

PR #5995, opened 2025-10-02
⭐ openeuler platform/family detection + package detection

tas50:
openeuler is RHEL based, but it doesn't have an /etc/os-release file. Detect the package + the family with a special check for openeuler within the redhat family check. Also add the path where it stores the RPM db so we can parse package data.

### DIFF
--- a/providers/os/detector/detector_platform_test.go
+++ b/providers/os/detector/detector_platform_test.go
@@ -889,3 +889,14 @@ func TestElementaryOSDetector(t *testing.T) {
 	assert.Equal(t, "x86_64", di.Arch, "os arch should be identified")
 	assert.Equal(t, []string{"debian", "linux", "unix", "os"}, di.Family)
 }
+
+func TestOpenEulerDetector(t *testing.T) {
+	di, err := detectPlatformFromMock("./testdata/detect-openeuler.toml")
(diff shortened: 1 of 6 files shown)


providers/os/detector/detector_platform_test.go
```diff
@@ -878,3 +878,14 @@ func TestUbiOSDetector(t *testing.T) {
...
+
+func TestOpenEulerDetector(t *testing.T) {
+	di, err := detectPlatformFromMock("./testdata/detect-openeuler.toml")
+	assert.Nil(t, err, "was able to create the provider")
+
+	assert.Equal(t, "openeuler", di.Name, "os name should be identified")
+	assert.Equal(t, "openEuler 24.03 (LTS-SP2)", di.Title, "os title should be identified")
+	assert.Equal(t, "24.03", di.Version, "os version should be identified")
+	assert.Equal(t, "aarch64", di.Arch, "os arch should be identified")
+	assert.Equal(t, []string{"redhat", "linux", "unix", "os"}, di.Family)
```
chris-rock: I am not sure we should it into the redhat family

tas50: Split this out into a euler family that only include openeuler for now, but will also include euleros once that support is validated

PR #6000, opened 2025-10-02
✨ Record assets in shell by platform id.

preslavgerchev:
Ensure assets with no mrn get stored under their platform id, enabling recordings in a shell.

### DIFF
--- a/providers-sdk/v1/recording/recording.go
+++ b/providers-sdk/v1/recording/recording.go
@@ -320,22 +320,15 @@ func (r *recording) finalize() {
 }
 
 func (r *recording) getExistingAsset(asset *inventory.Asset) *Asset {
-	if asset.Mrn != "" || asset.Id != "" {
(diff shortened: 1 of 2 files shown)


providers-sdk/v1/recording/recording.go
```diff
@@ -565,8 +558,9 @@ func RawDataArgsToResultArgs(args map[string]*llx.RawData) (map[string]*llx.Resu
 
 func getAssetIdForRecording(asset *inventory.Asset) string {
 	id := asset.Mrn
-	if id == "" {
-		id = asset.Id
+	if id == "" && len(asset.PlatformIds) > 0 {
```
chris-rock: should we sort the platform ids so that we always pick the first one?

preslavgerchev: makes sense, added

Issue #5969, opened 2025-09-29
Cannot retrieve: `azure.subscription.web.app.functions`


syrull:
**Describe the bug**
An empty list should be present, currently it just goes into 404 not found http request.

**To Reproduce**
Invoked: `azure.subscription.web.apps[0]` 

```
cnspec> azure.subscription.web.apps[0]
azure.subscription.web.apps[0]: azure.subscription.webService.appsite id="/subscriptions/<uid>/resourceGroups/<rg>/providers/Microsoft.Web/sites/<name>" name="<name>" location="<loc>"
cnspec> azure.subscription.web.apps[0].functions
GET https://management.azure.com/subscriptions/<uid>/resourceGroups/<rg>/providers/Microsoft.Web/sites/<name>/functions
--------------------------------------------------------------------------------
RESPONSE 404: 404 Not Found
ERROR CODE: NotFound
--------------------------------------------------------------------------------
{
  "error": {
    "code": "NotFound",
    "message": "Error retrieving functions. <html>\r\n<head><title>404 Not Found</title></head>\r\n<body>\r\n<center><h1>404 Not Found</h1></center>\r\n<hr><center>nginx/1.28.0</center>\r\n</body>\r\n</html>\r\n"
  }
}
--------------------------------------------------------------------------------
azure.subscription.web.apps[0].functions: no data available
```

- Env: Azure

syrull:
This is what happens whenever there is a functionapp:

```
cnspec> azure.subscription.web.apps[1].functions
azure.subscription.web.apps[1].functions: []
```

PR #6006, opened 2025-10-03
:star: Add Azure Batch Service and Account resources

syrull:
```
cnspec> azure.subscription.batch { * }
azure.subscription.batchService: {
  accounts: [
    0: azure.subscription.batchService.account id="/subscriptions/xxx/resourceGroups/xxx-xxx/providers/Microsoft.Batch/batchAccounts/zzz" name="zzzz" location="xxx"
  ]
  subscriptionId: "xxxx"
}
cnspec> azure.subscription.batch { accounts[0] }
azure.subscription.batchService: {
  accounts[0]: azure.subscription.batchService.account id="/subscriptions/xxxx/resourceGroups/xxx/providers/Microsoft.Batch/batchAccounts/xxxx" name="xxx" location="xxx"
}
cnspec> azure.subscription.batchService { accounts[0] { * } }
azure.subscription.batchService: {
  accounts[0]: {
    dedicatedCoreQuotaPerVmFamily: [
      0: {
        coreQuota: 10.000000
        name: "standardAv2Family"
      }
      1: {
        coreQuota: 10.000000
        name: "standardDv3Family"
      }
      2: {
        coreQuota: 10.000000
        name: "standardDSv3Family"
      }
      3: {
        coreQuota: 10.000000
        name: "standardEv3Family"
 ...
 ```

### DIFF
--- a/providers/azure/resources/azure.lr
+++ b/providers/azure/resources/azure.lr
@@ -1130,7 +1130,7 @@ private azure.subscription.webService.appRuntimeStack @defaults("preferredOs run
   // Web App stack minor version name
   minorVersion string
   // Whether the stack version is auto-updated
-  isDefault bool
+  autoUpdate bool
   // Whether the stack is deprecated
   deprecated bool
   // Whether the stack is hidden
--- a/providers/azure/resources/web.go
+++ b/providers/azure/resources/web.go
@@ -61,7 +61,7 @@ type AzureWebAppStackRuntime struct {
 	MinorVersion  string    `json:"minorVersion,omitempty"`
 	IsDeprecated  bool      `json:"isDeprecated,omitempty"`
 	IsHidden      bool      `json:"isHidden,omitempty"`
-	IsDefault     bool      `json:"isDefault,omitempty"`
+	AutoUpdate    bool      `json:"autoUpdate,omitempty"`
 	EndOfLifeDate time.Time `json:"endOfLifeDate,omitempty"`
 }
 
@@ -105,7 +105,7 @@ type runtimeStackDescriptor struct {
 	Name         string
 	MinorVersion string
 	ID           string
-	IsDefault    bool
+	AutoUpdate   bool
 	IsDeprecated bool
 }
 
@@ -134,8 +134,8 @@ func runtimeStackDescriptorFromMap(values map[string]any) *runtimeStackDescripto
 	if id, ok := values["id"].(string); ok {
 		descriptor.ID = id
 	}
-	if isDefault, ok := values["isDefault"].(bool); ok {
-		descriptor.IsDefault = isDefault
+	if autoUpdate, ok := values["autoUpdate"].(bool); ok {
+		descriptor.AutoUpdate = autoUpdate
 	}
 	if isDeprecated, ok := values["isDeprecated"].(bool); ok {
 		descriptor.IsDeprecated = isDeprecated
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: providers/azure/resources/azure.lr [hunk 1]
```diff
@@ -194,6 +196,66 @@ azure.subscription.computeService.disk @defaults("name location properties.osTyp
   properties dict
 }
 
+// Azure batch
```
tas50:
```suggestion
// Azure Batch
```

providers/azure/resources/azure.lr
```diff
@@ -194,6 +196,94 @@ azure.subscription.computeService.disk @defaults("name location properties.osTyp
...
+// Azure Batch
+private azure.subscription.batchService {
+  // Subscription identifier
+  subscriptionId string
+  // Batch accounts in the subscription
+  accounts() []azure.subscription.batchService.account
+}
+
+// Azure Batch account
+azure.subscription.batchService.account @defaults("id name location") {
```
czunker, line 208: I'm not that deep into MQL resources, but when `azure.subscription.batchService` is private, shouldn't the sub-resources not also be private?

providers/azure/resources/azure.lr
```diff
@@ -194,6 +196,94 @@ azure.subscription.computeService.disk @defaults("name location properties.osTyp
...
+  // Batch account private endpoint connections
+  privateEndpointConnections []dict
+  // Batch account pools
+  pools() []azure.subscription.batchService.account.pool
+  // Batch account diagnostic settings
+  diagnosticSettings() []azure.subscription.monitorService.diagnosticsetting
+}
+
+// Azure Batch pool
+azure.subscription.batchService.account.pool @defaults("id name") {
```
czunker, line 264: Same question about whether that should be private or not.

providers/azure/resources/batch.go
```diff
@@ -0,0 +1,379 @@
...
+	if account.Identity != nil {
+		identity, err := convert.JsonToDict(account.Identity)
+		if err != nil {
+			return nil, err
+		}
+		identityData = llx.DictData(identity)
+	}
+
+	propertiesData := llx.NilData
+	var (
```
czunker, line 90: nit: Use the same pattern for both? Otherwise, it reads a bit strange.

providers/azure/resources/web.go
```diff
@@ -99,6 +101,85 @@ func createWebAppResourceFromSite(runtime *plugin.Runtime, resourceType string,
...
+		descriptor.Name = strings.ToLower(name)
+	}
+	if minor, ok := stringFromTValue(&runtime.MinorVersion); ok {
+		descriptor.MinorVersion = strings.ToLower(minor)
+	}
+	if id, ok := stringFromTValue(&runtime.RuntimeVersion); ok {
+		descriptor.ID = id
+	}
+	if isDefault, ok := boolFromTValue(&runtime.AutoUpdate); ok {
+		descriptor.IsDefault = isDefault
```
czunker: `IsDefault` vs. `AutoUpdate`: Is this assignment correct?

providers/azure/resources/web.go
```diff
@@ -99,6 +101,85 @@ func createWebAppResourceFromSite(runtime *plugin.Runtime, resourceType string,
...
+		return descriptor
+	}
+	if name, ok := stringFromTValue(&runtime.Name); ok {
+		descriptor.Name = strings.ToLower(name)
+	}
+	if minor, ok := stringFromTValue(&runtime.MinorVersion); ok {
+		descriptor.MinorVersion = strings.ToLower(minor)
+	}
+	if id, ok := stringFromTValue(&runtime.RuntimeVersion); ok {
+		descriptor.ID = id
```
czunker: Same here:
`ID` vs. `RuntimeVersion`

czunker:
> This fixes this too (it was 2 lines fix): #5969

As this is unrelated to the batch, better split it out into another PR.

providers/azure/resources/azure.lr
syrull: I did this based on pattern I saw within the file, usually I know that when you declare something `private` it shouldn't be accessible in the queries, that does not apply for the sub-resources.

As far as I know this is the way it is because of "aliasing", in this case we used "batch" while the actual object is called `batchService`

providers/azure/resources/azure.lr
syrull: same answer as above

providers/azure/resources/batch.go
syrull: done

providers/azure/resources/web.go
syrull: The ID should be unique, putting it as the runtime version should be fine right?

providers/azure/resources/web.go
syrull: This is actually a change from azure that we didn't apply I will fix the field.

providers/azure/resources/web.go
czunker: How I understand the docs, the name seems to be unique inside a subscription: https://learn.microsoft.com/en-us/azure/app-service/deploy-best-practices#use-github-actions

You could have multiple apps with the same runtime version.

A combination of subscription ID and app name should make it unique.

czunker: @syrull  Did you find a solution for this one?

syrull: yes forgot to  push give me a minute

syrull: donee

PR #6034, opened 2025-10-10
✨ Add generate cmd to mqlr that unifies go code + docs generation.

preslavgerchev:
That unifies 3 commands into one, instead of running:
1. mqlr go
2. mqlr docs yaml
3. mqlr docs json

it now suffices to run `mql generate`. most of the parameters are defaults, they can be exposed if needed

### DIFF
--- a/providers-sdk/v1/mqlr/cmd/generate.go
+++ b/providers-sdk/v1/mqlr/cmd/generate.go
@@ -8,6 +8,14 @@ import (
 	"github.com/spf13/cobra"
 )
 
+func init() {
+	rootCmd.AddCommand(generateCmd)
+	generateCmd.Flags().String("dist", "", "folder for output LR and docs generation")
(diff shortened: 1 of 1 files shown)


providers-sdk/v1/mqlr/cmd/generate.go
```diff
@@ -0,0 +1,49 @@
...
+
+		lrFile := args[0]
+		headerFile := ""
+		runGoCmd(lrFile, dist, headerFile, false)
+		runDocsYamlCmd(lrFile, headerFile, defaultVersionField, docsFile)
+		runDocsJsonCmd(docsFile, dist)
+	},
+}
+
+func init() {
```
chris-rock: nit: i personally prefer if init is before the the code

preslavgerchev: agreed

PR #6040, opened 2025-10-10
⭐ Improve azurelinux platform detection

tas50:
- Don't include the version in the platform name
- Parse the build ID

### DIFF
--- a/providers/os/detector/detector_all.go
+++ b/providers/os/detector/detector_all.go
@@ -125,16 +125,13 @@ var azurelinux = &PlatformResolver{
 	IsFamily: false,
 	Detect: func(r *PlatformResolver, pf *inventory.Platform, conn shared.Connection) (bool, error) {
 		if pf.Name == "azurelinux" {
-			// Microsoft includes the version in the title, which we do not want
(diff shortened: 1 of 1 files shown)


providers/os/detector/detector_all.go
```diff
@@ -120,6 +120,28 @@ var manjaro = &PlatformResolver{
 	},
 }
 
+var azurelinux = &PlatformResolver{
+	Name:     "azurelinux",
+	IsFamily: false,
+	Detect: func(r *PlatformResolver, pf *inventory.Platform, conn shared.Connection) (bool, error) {
+		if pf.Name == "azurelinux" {
+			// Microsoft includes the version in the title, which we do not want
+			pf.Title = "Microsoft Azure Linux"
```
chris-rock: We should use the name from osr[NAME]

tas50: 🤦 yep that makes way more sense

tas50: Technically this makes the data less great if os-release for some reason can't be parsed, but I'm fairly sure we'd never get this far if os-release couldn't be parsed.

PR #6041, opened 2025-10-10
⭐ Improve detection of flatcar container linux 

tas50:
Use the better title so we don't include the version in the platform
title

requires https://github.com/mondoohq/cnquery/pull/6040 to be merged first


providers/os/detector/detector_platform_test.go
```diff
@@ -954,3 +954,27 @@ func TestSteamOSDetector(t *testing.T) {
...
+
+	assert.Equal(t, "azurelinux", di.Name, "os name should be identified")
+	assert.Equal(t, "Microsoft Azure Linux", di.Title, "os title should be identified")
+	assert.Equal(t, "3.0", di.Version, "os version should be identified")
+	assert.Equal(t, "3.0.20241005", di.Build, "os build should be identified")
+	assert.Equal(t, "x86_64", di.Arch, "os arch should be identified")
+	assert.Equal(t, []string{"linux", "unix", "os"}, di.Family)
+}
+
+func TestDetector(t *testing.T) {
```
chris-rock: We should use a more descriptive name for the test function

providers/os/detector/detector_all.go
```diff
@@ -120,6 +120,43 @@ var manjaro = &PlatformResolver{
...
 }
 
+var azurelinux = &PlatformResolver{
+	Name:     "azurelinux",
+	IsFamily: false,
+	Detect: func(r *PlatformResolver, pf *inventory.Platform, conn shared.Connection) (bool, error) {
+		if pf.Name == "azurelinux" {
+			osrd := NewOSReleaseDetector(conn)
+			osr, err := osrd.osrelease()
+			if err != nil {
```
preslavgerchev: nit:
```suggestion
			if err == nil {
			    pf.Title = osr["NAME"]
			    ...
			    }
```

saves you a `return` statement as we do not care about the err

providers/os/detector/detector_platform_test.go
tas50: Clearly I deleted the old platform name and forgot to add back the new one. Oops

providers/os/detector/detector_all.go
tas50: This what you're thinking? https://github.com/mondoohq/cnquery/pull/6041/commits/b8edb3afdabd0120971a6cab709164068419c6c0

PR #6051, opened 2025-10-16
⭐ Add snapshotWindow field to aws.elasticache.cluster

Bajusz15:
Adds snapshotWindow field to aws cache cluster

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -65,6 +65,7 @@ make providers/mqlr
 ./mqlr generate providers/aws/resources/aws.lr --docs-file providers/aws/resources/aws.lr.manifest.yaml --dist providers/aws/resources
 ```
 To quickly install the changed provider plugin run `make providers/build/aws && make providers/install/aws`.
+
 ## Debug providers
 
 `cnquery` uses a plugin mechanism. Each provider has its own go modules. This ensures that dependencies are only used on


Bajusz15:
recheck

providers/aws/resources/aws_elasticache.go
```diff
@@ -23,23 +21,6 @@ func (a *mqlAwsElasticache) id() (string, error) {
 	return "aws.elasticache", nil
 }
 
-// parseSnapshotWindow parses the SnapshotWindow string (format "HH:MM") into a time.Time
-func parseSnapshotWindow(snapshotWindow *string) (*time.Time, error) {
```
imilchev: could you explain why this is removed?

Bajusz15: @imilchev If I remember correctly, I first generated snapshotWindow as time, then when we tested with @syrull, we ran into an error, that this cannot be parsed into a time. The format AWS returns is `"07:30-08:30"`, and we tried to parse it like it's `"07:30"`. Which didn't work because of the format, and also would be incorrect, the `07:30-08:30` value cannot be a time value. Then @syrull made a change to not convert it to time but keep it as string, with which it was working correctly on the tested Elasticache resource. So, in short  we no longer need to parse it as time.

docs/development.md
```diff
@@ -59,6 +59,12 @@ make cnquery/generate
 This generates and updates all required files for the build. At this point you can `make cnquery/install` again as
 outlined above.
 
+If you make update to a provider's lr file, you can generate go files for that provider with this command: 
+```bash
+make providers/mqlr
+./mqlr generate providers/aws/resources/aws.lr --docs-file providers/aws/resources/aws.lr.manifest.yaml --dist providers/aws/resources
+```
+To quickly install the changed provider plugin run `make providers/build/aws && make providers/install/aws`.
```
tas50, line 67:
```suggestion
To quickly install the changed provider plugin run `make providers/build/aws && make providers/install/aws`.

```

PR #6055, opened 2025-10-17
:star: Gather the kernel version on FreeBSD

tas50:
Currently on FreeBSD we return an empty version of the Linux data structure:

```
cnquery> kernel.info
kernel.info: {
  args: null
  device: ""
  path: ""
  version: ""
}
```

Instead let's return the version of the kernel using 99.9% of the same logic we use on AIX:

```
cnquery> kernel.info
kernel.info: {
  version: "FreeBSD 14.3-RELEASE releng/14.3-n271432-8c9ce319fef7 GENERIC"
}


arlimus (approved):
That is nice! kudos

PR #6057, opened 2025-10-19
⭐ Add support for BSD variants in the kernel resource

tas50:
Update the FreebsdKernelManager to be just a generic BsdKernelManager


chris-rock:
Added openbsd kernel module handling and os detection test for freebsd 15

PR #6064, opened 2025-10-20
✨ Add aliases support to aws.kms.key resource

Bajusz15:
Add aliases() method to kms keys.

### DIFF
--- a/providers/os/resources/os.lr.manifest.yaml
+++ b/providers/os/resources/os.lr.manifest.yaml
@@ -50,6 +50,7 @@ resources:
       params: {}
     min_mondoo_version: 9.0.0
   auditd.rule:
+    fields: {}
     is_private: true
     min_mondoo_version: 9.0.0
   auditd.rule.control:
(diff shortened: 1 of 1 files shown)


feature_string.go
```diff
@@ -28,9 +28,9 @@ const _Feature_name = "MassQueriesPiperCodeBoolAssertionsK8sNodeDiscoveryMQLAsse
 var _Feature_index = [...]uint8{0, 11, 20, 34, 50, 65, 81, 99, 116, 132, 152, 167, 186, 201}
 
 func (i Feature) String() string {
-	i -= 1
-	if i >= Feature(len(_Feature_index)-1) {
-		return "Feature(" + strconv.FormatInt(int64(i+1), 10) + ")"
+	idx := int(i) - 1
```
syrull, line 31: Whats going on here?

imilchev: that code is auto-generated. I would assume we generated it with a newer version of the generator

imilchev (approved):
LGTM

feature_string.go
Bajusz15: CI was complaining after I resolved with main, then I regenerated it. The command is `//go:generate go run golang.org/x/tools/cmd/stringer -type=Feature`

PR #6069, opened 2025-10-22
⭐ Adds privateEndpointConnection to azure web apps

Bajusz15:
Extends `azure.subscription.web.apps` with private endpoint connections.

Example query and results for `azure.subscription.web.apps { name privateEndpointConnections {id name type privateLinkServiceConnectionState { description status actionsRequired }  provisioningState ipAddresses } }`:
```
azure.subscription.web.apps: [
  0: {
    privateEndpointConnections: [
      0: {
        privateLinkServiceConnectionState: {
          status: "Approved"
          description: ""
          actionsRequired: "None"
        }
        type: "Microsoft.Web/sites/privateEndpointConnections"
        name: "mate-private-endpoint-REDACTED"
        id: "/subscriptions/REDACTED/resourceGroups/REDACTED/providers/Microsoft.Web/sites/REDACTED/privateEndpointConnections/mate-private-endpoint-REDACTED"
        ipAddresses: [
          0: "10.0.0.4"
        ]
        provisioningState: "Succeeded"
      }
    ]
    name: "REDACTED"
  }
]
```

### DIFF
--- a/providers/azure/resources/azure.lr
+++ b/providers/azure/resources/azure.lr
@@ -1203,7 +1203,7 @@ private azure.subscription.privateEndpointConnection @defaults("id name type") {
   properties dict
 }
 
-// connection state for a private endpoint connection
+// Connection state for a private endpoint connection
 private azure.subscription.privateEndpointConnection.connectionState @defaults("status") {
   // Required actions for the private link service connection
   actionsRequired string


providers/azure/resources/azure.lr
```diff
@@ -1179,6 +1179,8 @@ private azure.subscription.webService.appsite @defaults("id name location") {
   ftp() azure.subscription.webService.appsite.basicPublishingCredentialsPolicies
   // SCM publishing method policies
   scm() azure.subscription.webService.appsite.basicPublishingCredentialsPolicies
+  // Private endpoint connections for the web app site
+  privateEndpointConnections() []dict
```
preslavgerchev: is it possible to model these as as separate resource instead of using a dict? that way we have an idea of what the resource might contain 

e.g. something like 
`azure.subscription.networkService.privateEndpoint`

Bajusz15: I think yes, I will look into it. I got this from an existing example, we use it for accounts too.

imilchev: yeah, when we initially developed mql resources we had to add a lot and fast. We took some shortcuts with dicts. However, they have no docs and no type definition, so they are very hard to use. I would suggest using a proper type as well

syrull: To add on that its hard to keep track on what resources we have and how to use them since they are fetched dynamically and if I am to make a query I have to guess what the dict returns.

Bajusz15: @syrull @preslavgerchev I added privateEndpointConnections, could you review again? Also I suggest creating another issue for accounts, they could also use privateEndpointConnections not as `dict`.

providers/azure/resources/azure.lr
```diff
@@ -1181,6 +1179,36 @@ private azure.subscription.webService.appsite @defaults("id name location") {
   ftp() azure.subscription.webService.appsite.basicPublishingCredentialsPolicies
   // SCM publishing method policies
   scm() azure.subscription.webService.appsite.basicPublishingCredentialsPolicies
+  // Private endpoint connections for the web app site
+  privateEndpointConnections() []azure.subscription.privateEndpointConnection
+}
+
+private azure.subscription.privateEndpointConnection @defaults("id name type") {
```
tas50, line 1187: We need a comment for this resource as well

providers/azure/resources/azure.lr
```diff
@@ -1181,6 +1179,36 @@ private azure.subscription.webService.appsite @defaults("id name location") {
...
+  privateEndpointId string
+  // Private link service connection state details
+  privateLinkServiceConnectionState azure.subscription.privateEndpointConnection.connectionState
+  // Provisioning state of the private endpoint connection
+  provisioningState string
+  // privateEndpointConnection properties
+  properties dict
+}
+
+private azure.subscription.privateEndpointConnection.connectionState @defaults("status") {
```
tas50, line 1207: this one too

providers/azure/resources/azure.lr
```diff
@@ -1181,6 +1179,38 @@ private azure.subscription.webService.appsite @defaults("id name location") {
...
+  privateEndpointId string
+  // Private link service connection state details
+  privateLinkServiceConnectionState azure.subscription.privateEndpointConnection.connectionState
+  // Provisioning state of the private endpoint connection
+  provisioningState string
+  // privateEndpointConnection properties
+  properties dict
+}
+
+// connection state for a private endpoint connection
```
tas50:
```suggestion
// Connection state for a private endpoint connection
```

PR #6073, opened 2025-10-23
⭐ Add azure.subscription.web.hostingEnvironments to list resources

Bajusz15:
Adds following methods and resources to azure provider:

- hostingEnvironments() []azure.subscription.webService.hostingEnvironment
- private azure.subscription.webService.hostingEnvironment.virtualNetwork
- private azure.subscription.webService.hostingEnvironment

### DIFF
--- a/providers/azure/resources/web.go
+++ b/providers/azure/resources/web.go
@@ -1376,7 +1376,7 @@ func (a *mqlAzureSubscriptionWebService) hostingEnvironments() ([]any, error) {
 						"type":   llx.StringDataPtr(props.VirtualNetwork.Type),
 						"subnet": llx.StringDataPtr(props.VirtualNetwork.Subnet),
 					}
-					vnRes, err := CreateResource(a.MqlRuntime, "azure.subscription.webService.hostingEnvironment.virtualNetwork", vnArgs)
+					vnRes, err := CreateResource(a.MqlRuntime, ResourceAzureSubscriptionWebServiceHostingEnvironmentVirtualNetwork, vnArgs)
 					if err != nil {
 						return nil, err
 					}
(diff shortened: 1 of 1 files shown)


providers/azure/resources/web.go
```diff
@@ -1299,3 +1299,112 @@ func (a *mqlAzureSubscriptionWebServiceAppsite) privateEndpointConnections() ([]
...
+					if setting == nil {
+						continue
+					}
+					dict, err := convert.JsonToDict(setting)
+					if err != nil {
+						return nil, err
+					}
+					items = append(items, dict)
+				}
+				args["clusterSettings"] = llx.ArrayData(items, types.Dict)
```
imilchev, line 1397: if this type is not too big, please add an actual resource for it. Don't use a dict. If it's too big we can decide which props are most interesting and implement just those. Either is fine

providers/azure/resources/web.go
```diff
@@ -1299,3 +1299,112 @@ func (a *mqlAzureSubscriptionWebServiceAppsite) privateEndpointConnections() ([]
...
+			}
+
+			mqlRes, err := CreateResource(a.MqlRuntime, "azure.subscription.webService.hostingEnvironment", args)
```
syrull: Use the generated constant in the lr.go file

providers/azure/resources/web.go
Bajusz15: Unfortunately, it's key value pair, I can't add a type for it.

Bajusz15: https://learn.microsoft.com/en-us/rest/api/appservice/app-service-environments/list?view=rest-appservice-2025-03-01&tabs=HTTP#appserviceenvironmentcollection

imilchev (approved):
LGTM

PR #6074, opened 2025-10-24
:star: POC: Extend auditd.rules resource with live-system capabilities

mm-weber:


providers/os/connection/container/registry_connection.go
```diff
@@ -17,13 +17,21 @@ var _ shared.Connection = &RegistryConnection{}
 
 type RegistryConnection struct {
 	plugin.Connection
-	asset *inventory.Asset
+	asset         *inventory.Asset
+	auditProvider *shared.AuditRuleProvider
```
chris-rock, line 21: I do not understand why this would make sense

providers/os/connection/shared/audit_provider.go
```diff
@@ -0,0 +1,211 @@
...
+
+import (
+	"fmt"
+	"io"
+	"strings"
+	"sync"
+)
+
+// AuditRuleData holds parsed audit rules categorized by type
+type AuditRuleData struct {
```
chris-rock, line 14: Why is that required.

providers/os/resources/os.lr.manifest.yaml
```diff
@@ -1002,7 +1000,6 @@ resources:
       warndays: {}
     min_mondoo_version: 5.15.0
   sshd:
-    fields: {}
     min_mondoo_version: 5.15.0
   sshd.config:
```
chris-rock, line 1004: Those providers/requirementsAuditdLiveSystem.md do not feed like that add additional value and should not be added.

providers/os/resources/auditd.go
```diff
@@ -140,132 +141,161 @@ var auditdDowncaseKeywords = []string{
...
-		s.Controls = plugin.TValue[[]any]{State: plugin.StateIsSet, Error: err}
-		s.Files = plugin.TValue[[]any]{State: plugin.StateIsSet, Error: err}
-		s.Syscalls = plugin.TValue[[]any]{State: plugin.StateIsSet, Error: err}
-		return err
+		return nil, err
 	}
 
-	var errors multierr.Errors
+	// Initialize result data
+	data := &shared.AuditRuleData{
```
chris-rock, line 237: Why is this data structure not just in this file? We also need additional test cases for this.

chris-rock (changes requested):

tas50:
We're going to need to do this a bit differently and this PR has diverged main a lot.

PR #6076, opened 2025-10-27
🐛 Order LR resources before generating schema.

preslavgerchev:
Since the AST resources is a map, generating schema from it is not deterministic, we've seen some changes if schema is regenerated. This ensures the map is always sorted and extends some tests

### DIFF
--- a/providers-sdk/v1/mqlr/lrcore/schema.go
+++ b/providers-sdk/v1/mqlr/lrcore/schema.go
@@ -5,6 +5,8 @@ package lrcore
 
 import (
 	"errors"
+	"maps"
+	"slices"
 	"strings"
 
 	"go.mondoo.com/cnquery/v12/providers-sdk/v1/resources"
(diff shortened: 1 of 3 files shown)


providers-sdk/v1/mqlr/lrcore/schema.go
```diff
@@ -60,17 +62,23 @@ func Schema(ast *LR) (*resources.Schema, error) {
 		}
 	}
 
+	sorted := slices.SortedFunc(maps.Keys(res.Resources), func(a, b string) int {
+		aDepth := strings.Count(a, ".")
+		bDepth := strings.Count(b, ".")
+		return aDepth - bDepth
+	})
```
pratik-mondoo, line 69: Out of curiosity: How is order of things of same depth determined? Is it handled later somewhere? 
```
a.x.y
b.x.y
```
This could end up as a* b* or b* a*. Right?

VasilSirakov: that's a good point, though for this particular bug I think the edge case is more along the lines of:
`a.b.c` and `a.b.d` because they have the same depth and share the same (potentially implicit) resources `a` and `a.b` and if `a.b.c` is private and `a.b.d` is public, we might still get nondeterminism.

@preslavgerchev thoughts on adding a tie-break alphabetical sorting?

preslavgerchev: I think same depth is fine, this is more about making sure that any implicit resources that get created are deferred until last and that we first process the outer (shortest depth) resources first

The code below iterates through all parts of the resource name, split by `.`, so for `a.b.c` it would iterate through all 3 and try and create implicit resources. This change ensures that we first get `a` and `b` created instead. if we first access `a.b.c` (since it's a map) we'd create implicit resources for `a` and `b` where some flags might be incorrect

PR #6079, opened 2025-10-27
⭐ Fix ms365 provider microsoft.settings()

Bajusz15:
Replaces `dict` with 2 new structs for querying settings. Removes old mapping for the resource.
This should resolve the issue https://github.com/mondoohq/cnquery/issues/5755

```
➜  cnquery git:(mate/5755/microsoft.settings-bug) ✗ go run ./apps/cnquery/cnquery.go run ms365 --certificate-path ${MS365_CERTIFICATE_PATH} --tenant-id ${MS365_TENANT_ID} --client-id ${MS365_CLIENT_ID} -c "microsoft.groupSettings.where(displayName == \"Group.Unified\")" 
! using builtin provider for ms365
! using builtin provider for ms365
! using builtin provider for ms365
! using builtin provider for ms365
microsoft.settings.where: [
  0: microsoft.setting displayName="Group.Unified" values=[
       0: microsoft.settingValue id = REDACTED
       1: microsoft.settingValue id = REDACTED
       2: microsoft.settingValue id = REDACTED
       3: microsoft.settingValue id = REDACTED
       4: microsoft.settingValue id = REDACTED
       5: microsoft.settingValue id = REDACTED
       6: microsoft.settingValue id = REDACTED
       7: microsoft.settingValue id = REDACTED
       8: microsoft.settingValue id = REDACTED
       9: microsoft.settingValue id = REDACTED
       10: microsoft.settingValue id = REDACTED
       11: microsoft.settingValue id = REDACTED
       12: microsoft.settingValue id = REDACTED
       13: microsoft.settingValue id = REDACTED
       14: microsoft.settingValue id = REDACTED
     ]
]
```

### DIFF
--- a/providers/ms365/resources/settings.go
+++ b/providers/ms365/resources/settings.go
@@ -19,6 +19,7 @@ func (a *mqlMicrosoft) settings() ([]any, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	ctx := context.Background()
 	settings, err := graphClient.GroupSettings().Get(ctx, &groupsettings.GroupSettingsRequestBuilderGetRequestConfiguration{})
 	if err != nil {
--- a/providers/ms365/resources/structs.go
+++ b/providers/ms365/resources/structs.go
@@ -193,6 +193,7 @@ func newSetting(p models.GroupSettingable) GroupSetting {
 }
 
 // structs for AuthorizationPolicy
+
 type Entity struct {
 	Id *string `json:"id"`
 }


providers/ms365/resources/ms365.lr
```diff
@@ -27,7 +27,7 @@ microsoft {
   // List of roles
   roles() microsoft.roles
   // Microsoft 365 settings
-  settings() dict
+  settings() []microsoft.setting
```
imilchev: if this is used in existing policies, what you are doing now is going to be a breaking change. Anyone running the existing policies will encounter errors. You should check with @syrull what the usage for this field is and if it's okay to break it. If we cannot afford changing the type, you would need to create a new function that returns the new type and keep the old one around

tas50: I think we want to have a convo internally if we're ok breaking some of these dict values up outside of a major release. I'm a HUGE fan of getting rid of them, but there is some customer risk and coordination we may need to do with our policies team

imilchev (approved):
LGTM!

PR #6084, opened 2025-10-27
Delete deprecated and dead code ms365 `microsoft.settings`

Bajusz15:
DO NOT MERGE!
Until all policies using the old method are removed. Check with @imilchev before merging.
This should only be merged for a new release.


tas50:
As long as we mark it deprecated we will hit it with 13

Bajusz15:
@imilchev  Closing it 👍

Bajusz15:
Closing it as it is not needed for now.

PR #6090, opened 2025-10-28
⭐ Add groupIdsToMonitorExpirations to ms365 `microsoft.groupLifecyclePolicies[]` 

Bajusz15:
Currently stuck.

I managed to get this working in my PR, but only if i copied the bearer token from my browser. Otherwise I just get 401, as the permissions I assign to graph API doesn't matter for this endpoint. (https://main.iam.ad.ext.azure.com/api/Directories/LcmSettings)

We need to decide if this is a practical approach to use: https://rozemuller.com/use-internal-azure-api-in-automation/ (he user will have to authenticate in the browser, and the user needs permission to see group expiration). 

This requires AADInternals. For now I'm parking it.

### DIFF
--- a/providers/ms365/resources/grouplifecyclepolicies.go
+++ b/providers/ms365/resources/grouplifecyclepolicies.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"sync"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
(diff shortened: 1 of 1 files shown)


providers/ms365/resources/grouplifecyclepolicies.go
```diff
@@ -5,14 +5,50 @@ package resources
...
 	"go.mondoo.com/cnquery/v12/providers/ms365/connection"
+	"go.mondoo.com/cnquery/v12/types"
 )
 
+var lcmSettingsLock sync.Mutex
+
+const lcmSettingsScript = `
+$ErrorActionPreference = "Stop"
+
+$Uri = 'https://main.iam.ad.ext.azure.com/api/Directories/LcmSettings'
```
chris-rock, line 30: We should never call urls via powershell

PR #6091, opened 2025-10-28
🧹 Small shell cleanup.

preslavgerchev:


cli/shell/shell.go
```diff
@@ -119,22 +116,6 @@ func (s *Shell) printWelcome() {
 	fmt.Fprintln(s.out, s.Theme.Welcome)
 }
 
-func (s *Shell) print(msg string) {
-	if msg == "" {
```
preslavgerchev, line 123: this was only ever called in one place `print("")`, so it effectively did nothing

PR #6092, opened 2025-10-28
🐛 Fix connection types for network devices.

preslavgerchev:
This ensures we install the network devices provider if it's missing in case we're using an inventory file with connection type `nd-ssh`

### DIFF
--- a/providers/providers.go
+++ b/providers/providers.go
@@ -83,10 +83,10 @@ func (s ProviderLookup) String() string {
 		res = append(res, "provider="+s.ProviderName)
 	}
 	if s.ConnName != "" {
-		res = append(res, "name="+s.ConnName)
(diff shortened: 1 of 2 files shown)


providers/providers.go
```diff
@@ -83,10 +83,10 @@ func (s ProviderLookup) String() string {
 		res = append(res, "provider="+s.ProviderName)
 	}
 	if s.ConnName != "" {
-		res = append(res, "name="+s.ConnName)
+		res = append(res, "conn name="+s.ConnName)
```
preslavgerchev, line 86: this helps a lot to know how the provider lookup was performed

chris-rock (approved):
Thank you @preslavgerchev

PR #6098, opened 2025-10-29
Fix ms365 authenticationRequirements

Bajusz15:
There was a typo in the resource name. Should be `microsoft.user.authenticationRequirements`.

### DIFF
--- a/providers/ms365/resources/users.go
+++ b/providers/ms365/resources/users.go
@@ -826,7 +826,7 @@ func (a *mqlMicrosoftUser) authenticationRequirements() (*mqlMicrosoftUserAuthen
 		return nil, transformError(err)
 	}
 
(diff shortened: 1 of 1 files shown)


providers/ms365/resources/users.go
```diff
@@ -826,7 +826,7 @@ func (a *mqlMicrosoftUser) authenticationRequirements() (*mqlMicrosoftUserAuthen
 		return nil, transformError(err)
 	}
 
-	mqlAuthRequirements, err := CreateResource(a.MqlRuntime, "microsoft.userAuthenticationRequirements",
+	mqlAuthRequirements, err := CreateResource(a.MqlRuntime, "microsoft.user.AuthenticationRequirements",
```
preslavgerchev: can we use the lr-generated string const instead? that would prevent this from happening again

PR #6100, opened 2025-10-29
⭐ Add new field ms365.exchangeonline.hostedContentFilterPolicy

Bajusz15:
I added it as a dict because it's a huge object with many fields. 

to use in query: `ms365.exchangeonline.hostedContentFilterPolicy`

### DIFF
--- a/providers/ms365/resources/ms365.lr.manifest.yaml
+++ b/providers/ms365/resources/ms365.lr.manifest.yaml
@@ -969,6 +969,18 @@ resources:
       - microsoft365
   ms365.exchangeonline.reportSubmissionPolicy:
     fields:
+      disableQuarantineReportingOption: {}
+      enableCustomNotificationSender: {}
+      enableOrganizationBranding: {}
+      enableReportToMicrosoft: {}
+      enableThirdPartyAddress: {}
+      junkReviewResultMessage: {}
+      notJunkReviewResultMessage: {}
+      notificationFooterMessage: {}
+      notificationSenderAddresses: {}
+      phishingReviewResultMessage: {}
+      postSubmitMessageEnabled: {}
+      preSubmitMessageEnabled: {}
       reportChatMessageEnabled: {}
       reportChatMessageToCustomizedAddressEnabled: {}
       reportJunkAddresses: {}
--- a/providers/ms365/resources/ms365_exchange.go
+++ b/providers/ms365/resources/ms365_exchange.go
@@ -196,6 +196,18 @@ type ReportSubmissionPolicy struct {
 	ReportPhishAddresses                        []string `json:"ReportPhishAddresses"`
 	ReportChatMessageEnabled                    bool     `json:"ReportChatMessageEnabled"`
 	ReportChatMessageToCustomizedAddressEnabled bool     `json:"ReportChatMessageToCustomizedAddressEnabled"`
+	EnableReportToMicrosoft                     bool     `json:"EnableReportToMicrosoft"`
+	PreSubmitMessageEnabled                     bool     `json:"PreSubmitMessageEnabled"`
+	PostSubmitMessageEnabled                    bool     `json:"PostSubmitMessageEnabled"`
+	EnableThirdPartyAddress                     bool     `json:"EnableThirdPartyAddress"`
+	PhishingReviewResultMessage                 string   `json:"PhishingReviewResultMessage"`
+	NotificationFooterMessage                   string   `json:"NotificationFooterMessage"`
+	JunkReviewResultMessage                     string   `json:"JunkReviewResultMessage"`
+	NotJunkReviewResultMessage                  string   `json:"NotJunkReviewResultMessage"`
+	NotificationSenderAddress                   []string `json:"NotificationSenderAddress"`
+	EnableCustomNotificationSender              bool     `json:"EnableCustomNotificationSender"`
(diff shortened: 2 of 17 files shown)


providers/ms365/resources/ms365.lr.manifest.yaml
```diff
@@ -905,6 +905,8 @@ resources:
       authenticationPolicy: {}
       dkimSigningConfig: {}
       externalInOutlook: {}
+      hostedContentFilterPolicy: {}
+      hostedInboundSpamFilterPolicy: {}
```
czunker: Is this a leftover?
The PR is only about `hostedContentFilterPolicy`.

providers/ms365/resources/ms365_exchange.go
```diff
@@ -365,9 +368,12 @@ func (r *mqlMs365Exchangeonline) getExchangeReport() error {
 	malwareFilterPolicy, malwareFilterPolicyErr := convert.JsonToDictSlice(report.MalwareFilterPolicy)
 	r.MalwareFilterPolicy = plugin.TValue[[]any]{Data: malwareFilterPolicy, State: plugin.StateIsSet, Error: malwareFilterPolicyErr}
 
-	hostedOutboundSpamFilterPolicy, hostedOutboundSpamFilterPolicyErr := convert.JsonToDictSlice(report.HostedOutboundSpamFilterPolicy)
+	hostedOutboundSpamFilterPolicy, hostedOutboundSpamFilterPolicyErr := convert.JsonToDictSlice(report.HostedContentFilterPolicy)
 	r.HostedOutboundSpamFilterPolicy = plugin.TValue[[]any]{Data: hostedOutboundSpamFilterPolicy, State: plugin.StateIsSet, Error: hostedOutboundSpamFilterPolicyErr}
 
+	hostedContentFilterPolicy, hostedContentFilterPolicyErr := convert.JsonToDictSlice(report.HostedContentFilterPolicy)
```
czunker: Is it intended that you put the `HostedContentFilterPolicy` into both fields?

providers/ms365/resources/ms365_exchange.go
```diff
@@ -500,6 +506,10 @@ func (r *mqlMs365Exchangeonline) hostedOutboundSpamFilterPolicy() ([]any, error)
 	return nil, r.getExchangeReport()
 }
 
+func (r *mqlMs365Exchangeonline) hostedContentFilterPolicy() ([]any, error) {
+	return nil, r.getExchangeReport()
```
czunker, line 534: This looks like a copy&paste error.

providers/ms365/resources/ms365_exchange.go
Bajusz15: Nope, good catch 🦅  During the rename I probably confused it.

providers/ms365/resources/ms365.lr.manifest.yaml
Bajusz15: Yes, I named it that first, but it was confusing and in the documentation it's called different. I will regenerate this

Bajusz15: @czunker Hmm, I just found out this file is not regenerated but rather "merged" with new fields. So if I don't remove it manually after refactoring it stays there. 

here is the code from the generator:
```
// if we have docs from existing manifest, merge them in
	if existingDocs.Resources != nil {
		maps.Copy(docs.Resources, existingDocs.Resources)
	}
	// ensure default values and fields are set
	for k := range docs.Resources {
		docs.Resources[k] = ensureDefaults(k, docs.Resources[k], currentVersion, defaultVersion)
		mergeFields(docs.Resources[k], fields[k], currentVersion, defaultVersion)
		// Merge in other doc fields from core.lr
		docs.Resources[k].IsPrivate = isPrivate[k]
	}
```

providers/ms365/resources/ms365_exchange.go
Bajusz15: No, this is kindof a "lazy loading" or return catched object method. All of methods of this object return that function. The `r.getExchangeReport()` fetches the whole object. It has a mutex to lock in first fetch, and store the values. My guess is this is expensive to run so we cache the object. 

```
func (r *mqlMs365Exchangeonline) getExchangeReport() error {
	conn := r.MqlRuntime.Connection.(*connection.Ms365Connection)

	r.exchangeReportLock.Lock()
	defer r.exchangeReportLock.Unlock()

	// only fetch once
	if r.fetched {
		return r.fetchErr
	}

```

This lazy loader functions are used like this: we call `plugin.GetOrCompute` which returns cached object or calls the function if value is not set in cache. 

```
func (c *mqlMs365Exchangeonline) GetHostedContentFilterPolicy() *plugin.TValue[[]any] {
	return plugin.GetOrCompute[[]any](&c.HostedContentFilterPolicy, func() ([]any, error) {
		return c.hostedContentFilterPolicy()
	})
}

func GetOrCompute[T any](cached *TValue[T], compute func() (T, error)) *TValue[T] {
	if cached.IsSet() {
		return cached
	}
	x, err := compute()
	if err != nil {
		res := &TValue[T]{Data: x, Error: err}
		if err != NotReady {
			res.State = StateIsSet | StateIsNull
			(*cached) = *res
		}
		return res
	}
	.....
	
```

czunker: Thanks for the clarification

PR #6113, opened 2025-10-31
🐛 Allow ref's to be used as keys in map lookups

jaym:
Fixes the following query:
```
x = {"foo": "bar"}
y = "foo"; z = y;
x[z] == "bar"
```


arlimus (approved):
sweet!! hugekudos for adding this

PR #6119, opened 2025-11-01
🧹 improve mock to allow data loading via constructor

chris-rock:
extracted from https://github.com/mondoohq/cnquery/pull/6096/files to make it easier to use mock with inline test data.


chris-rock:
Turns green once https://github.com/mondoohq/cnquery/pull/6120 is merged

Bajusz15 (approved):

PR #6135, opened 2025-11-04
⭐ Add resource microsoft.tenant.branding

Bajusz15:
We need `keepMeSignedInDisabled`, this information is not available in Graph API. There is a powershell module to fetch the information. 
example query: `microsoft.tenant.branding {*}`
```
microsoft.tenant.branding: [
  0: {
    bannerLogo: ""
    backgroundColor: ""
    domainName: "REDACTED"
    useTransparentLightbox: false
    keepMeSignedInDisabled: false
  }
]
```
and `microsoft.tenant.realm {*}` 
```
microsoft.tenant.realm: [
  0: {
    nameSpaceType: "Managed"
    federationBrandName: "REDACTED"
    login: "REDACTED"
    domainName: "REDACTED"
  }
]
```

### DIFF
--- a/providers/ms365/resources/ms365.lr
+++ b/providers/ms365/resources/ms365.lr
@@ -170,7 +170,7 @@ private microsoft.tenant.brandingInfo @defaults("keepMeSignedInDisabled") {
 }
 
 // Tenant realm information
-private microsoft.tenant.realmInfo {
+private microsoft.tenant.realmInfo @defaults("login nameSpaceType")  {
   // Tenant federation brand name
   federationBrandName string
   // Domain name


providers/ms365/resources/tenant.go
```diff
@@ -63,6 +68,30 @@ var tenantFields = []string{
 	"preferredLanguage",
 }
 
+var tenantBrandingReport = `
+$UserName = '%s'
+
+# Query realm discovery endpoint
+$response = Invoke-RestMethod -UseBasicParsing -Uri ("https://login.microsoftonline.com/common/userrealm/" + $UserName + "?api-version=2.0")
```
imilchev: have we used this approach elsewhere? It looks to me that this is a simple `curl` request. You should be able to do this completely in Go code without using powershell. Is that the case or am I missing something?

If you really can execute this API call, then I suggest using Go code to run this request and parse the response, not powershell

Bajusz15: We used similar approach in ms365_exchange.go, but I think here it was used because the original issue linked to a repo containing a similar powershell script https://github.com/Gerenios/AADInternals/blob/ec04e908e48992bf5b5ff933029614f40e0c3604/AccessToken_utils.ps1#L239
Probbly I can make is as a go http request if that is preferred, after lunch I will push it 👍

chris-rock: Since this is an unauthenticated call, calling this url is ok.

providers/ms365/resources/structs.go
```diff
@@ -156,6 +156,19 @@ func newUnifiedRolePermission(p models.UnifiedRolePermissionable) UnifiedRolePer
 	}
 }
 
+type userRealmResponse struct {
```
chris-rock: Those structs are specific for the resource, therefore I would keep them with tenant.go

providers/ms365/resources/tenant.go
```diff
@@ -256,3 +260,83 @@ func (a *mqlMicrosoftTenant) formsSettings() (*mqlMicrosoftTenantFormsSettings,
 
 	return formSetting.(*mqlMicrosoftTenantFormsSettings), nil
 }
+
+func (a *mqlMicrosoftTenant) getTenantBrandingInfoForDomain(domainName string) (*userRealmResponse, error) {
```
chris-rock: this function name is confusing as it download the full realm information and not just the branding

providers/ms365/resources/ms365.lr
```diff
@@ -149,6 +149,24 @@ microsoft.tenant @defaults("name") {
   technicalNotificationMails []string
   // Preferred language
   preferredLanguage string
+  // Tenant branding information for all verified domains
+  branding() []microsoft.tenant.brandingInfo
```
chris-rock, line 155: The resource should look differently:

```javascript
microsoft.tenant {
   realm() []microsoft.tenant.realmInfo
   branding() []microsoft.tenant.brandingInfo
}

private microsoft.tenant.brandingInfo @defaults("keepMeSignedInDisabled") {
  // Domain name
  domainName string
  // if "Keep me signed in" is disabled for the tenant
  keepMeSignedInDisabled bool
  // Tenant branding background color
  backgroundColor string
  // Tenant branding logo URL
  bannerLogo string
  // Transparent light box
  useTransparentLightbox string
}

private microsoft.tenant.realmInfo { 
  // Tenant federation brand name
  federationBrandName string
  // Domain name
  domainName string
  // Login endpoint
  login string
  // Namespace Type
  nameSpaceType string
}
```

providers/ms365/resources/tenant.go
```diff
@@ -256,3 +281,138 @@ func (a *mqlMicrosoftTenant) formsSettings() (*mqlMicrosoftTenantFormsSettings,
...
+	}
+
+	// Cache the response
+	realmCache[domainName] = &realmResp
+
+	return &realmResp, nil
+}
+
+func (a *mqlMicrosoftTenant) realm() ([]any, error) {
+	verifiedDomains := a.GetVerifiedDomains()
```
chris-rock, line 294: Since the code is duplicated, you should not call this twice. Instead I recommend an approach, where you store both fields with one call, similar to https://github.com/mondoohq/cnquery/blob/main/providers/os/resources/powershell.go#L85-L95

Bajusz15: Using a mutex here to cache the data from the response, but yeah I can rewrite using that approach too 👍 
https://github.com/mondoohq/cnquery/pull/6135/files#diff-5bcdd451832984d37241e3f5a761b8fd4bdbed3af19437c824eec62df4070a4fR25-R28

providers/ms365/resources/tenant.go
```diff
@@ -256,3 +284,129 @@ func (a *mqlMicrosoftTenant) formsSettings() (*mqlMicrosoftTenantFormsSettings,
...
+			if err != nil {
+				logger.DebugDumpJSON("tenant-branding-create-error", []byte(fmt.Sprintf("domain: %s, index: %d, error: %v", domainName, i, err)))
+				a.realmDataErr = err
+				return err
+			}
+			brandingInfos = append(brandingInfos, mqlBranding)
+		}
+	}
+
+	a.realmInfos = realmInfos
```
chris-rock: I think we should set the state with the resources here, see https://github.com/mondoohq/cnquery/blob/3b1546508fd2501046c2ca70d6e31c8ab56bd942/providers/os/resources/sshd.go#L202-L230 as an example

providers/ms365/resources/tenant.go
```diff
@@ -256,3 +284,129 @@ func (a *mqlMicrosoftTenant) formsSettings() (*mqlMicrosoftTenantFormsSettings,
...
+	}
+
+	return &realmResp, nil
+}
+
+func (a *mqlMicrosoftTenant) realm() ([]any, error) {
+	if err := a.fetchRealmData(); err != nil {
+		return nil, err
+	}
+	return a.realmInfos, nil
```
chris-rock: I think this should be `return nil, a.fetchRealData()`

providers/ms365/resources/tenant.go
```diff
@@ -256,3 +284,129 @@ func (a *mqlMicrosoftTenant) formsSettings() (*mqlMicrosoftTenantFormsSettings,
...
+}
+
+func (a *mqlMicrosoftTenant) branding() ([]any, error) {
+	if err := a.fetchRealmData(); err != nil {
+		return nil, err
+	}
+	return a.brandingInfos, nil
```
chris-rock: I think this should be `return nil, a.fetchRealData()`

providers/ms365/resources/ms365.lr
```diff
@@ -149,6 +149,36 @@ microsoft.tenant @defaults("name") {
...
+  // Tenant branding background color
+  backgroundColor string
+  // Tenant branding logo URL
+  bannerLogo string
+  // Transparent light box
+  useTransparentLightbox bool
+}
+
+// Tenant realm information
+private microsoft.tenant.realmInfo {
```
chris-rock: here we are missing the defaults, I think we should use login and nameSpaceType

chris-rock (approved):
Great addition @Bajusz15 Lets update the mqlr definition and then we are ready to get this merged.

chris-rock (approved):

PR #6138, opened 2025-11-06
⭐ Add os resource `network.routes` for Linux, macOS, Windows

Bajusz15:
This PR implements `network.routes` resource (externalIP and primary IP will come in a separate PR). To ensure accuracy and compatibility, `osquery` was used as the reference implementation for comparison across all operating systems.
Part of https://github.com/mondoohq/planning/issues/849 implementation.

Example query: `network.routes{ destination iface gateway }`

## Platform-Specific Implementations

Each platform uses a different approach optimized for best compatibility:

- **Linux**: Uses `ip -json route show table all` to parse JSON-formatted route data, providing comprehensive route information including all routing tables, or as fallback for minimal distritbutions we read from the generated files "/proc/net/route" and "/proc/net/ipv6_route".
- **Darwin (macOS)**: Uses `golang.org/x/net/route` package to parse system routing tables via the `route` syscall, with route flag parsing based on BSD-style `sys/route.h` constants (example logic https://pkg.go.dev/golang.org/x/net/route#example-ParseRIB)
- **Windows**: Uses PowerShell `Get-NetRoute` cmdlet as the primary method, with `netstat -rn` as a fallback for systems where PowerShell is unavailable or returns no results. 

## Verification

Results were verified against osquery using the following query as the standard on all OSes:
`SELECT destination, netmask, gateway, interface, flags, type from routes;`

### Route Flags Handling
Route flags are handled platform-specifically:
- **Linux**: Flags are parsed directly from the `ip route show table all` command output, or as fallback for minimal distritbutions we read from the generated files "/proc/net/route" and "/proc/net/ipv6_route". 
- **Darwin**: Integer route flags are converted to human-readable strings (e.g., `GATEWAY`, `HOST`, `UP`, ....)
- **Windows**: Flags are currently empty (`[]string{}`) as Windows route information doesn't provide equivalent flag data

### Build Tags
Due to platform-specific dependencies, it's not possible to use a shared file and some implementations are separated into different files with a build suffix or build tags:
- `//go:build windows` and `//go:build !windows` for Windows native and remote implementation
- stub files contain empty functions for missing implementations, this way the binary compilation is successful for all platforms. In these cases the real implementation would import packages which are excluded for the given platform, and the binary compilation would fail.

Windows native and remote implementation have different entrypoints, there is _windows_routes_native_entrypoint.go_ and _windows_routes_remote_entrypoint.go_ file, the native version imports package which is excluded from other platforms. This way we can use windows routes on macOS and linux with ssh.

## How Has This Been Tested?
The test configurations mimic real results I got from running osquery locally. 

- [x] TestParseNetstatPowerShellOutput
- [x] TestParsePowerShellGetNetRouteOutput
- [x] Test_parseIpRouteJSON
- [x] Test_parseRouteMessage
- [x] Tested on Linux (Debian, and alpine for containers) - verified routes with cnquery match osquery output
- [x] Tested on macOS/Darwin (macbook) - verified routes with cnquery *mostly* match osquery output 
- [x] Tested on Windows 11 (VM running in UTM on my ARM Mac) - verified native implementation (see latest comment), powershell and netstat outputs seperately, and cnquery output after compilation

### DIFF
--- a/providers/os/resources/networkinterface/linux_routes.go
+++ b/providers/os/resources/networkinterface/linux_routes.go
@@ -4,6 +4,8 @@
 package networkinterface
 
 import (
+	"encoding/binary"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"net"
@@ -276,12 +278,9 @@ func hexToIP(hexStr string) (net.IP, error) {
 		return nil, err
 	}
 	// Convert from little-endian to IP
-	return net.IPv4(
-		byte(val&0xff),
-		byte((val>>8)&0xff),
-		byte((val>>16)&0xff),
-		byte((val>>24)&0xff),
-	), nil
+	bytes := make([]byte, 4)
+	binary.LittleEndian.PutUint32(bytes, uint32(val))
+	return net.IPv4(bytes[0], bytes[1], bytes[2], bytes[3]), nil
 }
 
 // hexToIPv6 converts a hex string (32 hex chars) to net.IP (IPv6)
@@ -291,15 +290,9 @@ func hexToIPv6(hexStr string) (net.IP, error) {
 	}
 
 	ip := make(net.IP, 16)
-	for i := 0; i < 16; i++ {
-		// IPv6 in /proc/net/ipv6_route is stored in network byte order (big-endian)
-		// Each byte is represented by 2 hex chars
-		byteStr := hexStr[i*2 : i*2+2]
-		val, err := strconv.ParseUint(byteStr, 16, 8)
-		if err != nil {
-			return nil, err
-		}
-		ip[i] = byte(val)
+	_, err := hex.Decode(ip, []byte(hexStr))
+	if err != nil {
+		return nil, err
 	}
 
 	return ip, nil


Bajusz15:
> @Bajusz15 Can you post the output of the routes on the different platform here?

I will upload in files, it can be a big output with flags.

providers/os/resources/networkinterface/windows_routes_pwsh.go
```diff
@@ -0,0 +1,419 @@
...
+	P2 string `json:"P2"` // Destination (IPv4) or If (IPv6)
+	P3 string `json:"P3"` // Netmask (IPv4) or Metric (IPv6)
+	P4 string `json:"P4"` // Gateway (IPv4) or Network Destination (IPv6)
+	P5 string `json:"P5"` // Interface (IPv4) or Gateway (IPv6)
+	P6 string `json:"P6"` // Metric (IPv4) or Gateway (IPv6, when "On-link" appears)
+}
+
+// UnmarshalJSON handles case-insensitive field names (P1/P2/P3 vs p1/p2/p3)
+// and converts numeric values to strings
+func (n *NetstatRoute) UnmarshalJSON(data []byte) error {
```
Bajusz15, line 198: For anyone wondering why this custom unmarshal is needed, it's because the netstat output is "unstable" meaning sometimes the same fields are string, number, or null in different elements, so there is no one struct we can unmarshal to. The columns are different for ipv4 and ipv6, and the output contains headers of the tables which we don't want to parse.

Bajusz15:
Here uploading the proc/net sources for linux test outputs from alpine container and debian: 
[alpine_ipv4.txt](https://github.com/user-attachments/files/23524176/proc_net_route.txt)

[debian_ipv4.txt](https://github.com/user-attachments/files/23524173/ipv4.txt)
[alpine_ipv6.txt](https://github.com/user-attachments/files/23524175/proc_net_ipv6_route.txt)

[debian_ipv6.txt](https://github.com/user-attachments/files/23524174/ipv6.txt)

providers/os/id/networki/windows_routes.go
```diff
@@ -0,0 +1,414 @@
+//go:build windows
```
chris-rock: This will not be correct, as this powershell implementation is also required for remote ssh connection to Windows.

providers/os/resources/os.lr
```diff
@@ -1864,6 +1867,21 @@ private networkInterface @defaults("name mac active") {
...
+private networkRoutes {
+  []networkRoute
+  
+  // Default Route
+  defaults() []networkRoute
+}
+
+// information on the network routes
+private networkRoute @defaults("destination flags"){
+  destination string
```
chris-rock, line 1883: We need comments on top of each field

providers/os/resources/networkinterface/darwin_routes.go
```diff
@@ -0,0 +1,195 @@
+// Copyright (c) Mondoo, Inc.
```
chris-rock, line 1: thise package should be a subdirectory in resources, not id

providers/os/id/networki/darwin_routes.go
```diff
@@ -0,0 +1,195 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+//go:build darwin
```
chris-rock, line 4: Since we only need this rarely, i recommend to use file suffix for darwin, see https://dave.cheney.net/tag/build-tags

providers/os/id/networki/linux_routes.go
```diff
@@ -0,0 +1,313 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+//go:build linux
```
chris-rock: this should not be a build tag since it will fail when you ssh from macos to a linux box

providers/os/id/networki/windows_routes.go
```diff
@@ -0,0 +1,414 @@
...
+		return routes, nil
+	}
+	log.Debug().Err(err).Int("routeCount", len(routes)).Msg("PowerShell Get-NetRoute failed or returned no routes, trying netstat")
+
+	// fallback to netstat
+	return n.detectWindowsRoutesViaNetstat()
+}
+
+// detectWindowsRoutesViaPowerShell uses PowerShell Get-NetRoute command
+func (n *neti) detectWindowsRoutesViaPowerShell() ([]Route, error) {
```
chris-rock: I think we should also add a native Windows implementation, as this would make it better for local execution

chris-rock (commented):
Great first iteration @Bajusz15

providers/os/id/networki/windows_routes.go
Bajusz15: I will look into a solution, I remember had to split into 3 build files with tags because the dependencies I import 
`golang.org/x/net/route` and `golang.org/x/sys/unix` has build tags that exclude windows, and with the conditional function like this it was failing to build:

```

	if pf.IsFamily(inventory.FAMILY_LINUX) {
		return n.detectLinuxInterfaces()
	}
	if pf.IsFamily(inventory.FAMILY_DARWIN) {
		return n.detectDarwinInterfaces()
	}
	if pf.IsFamily(inventory.FAMILY_WINDOWS) && conn.Capabilities().Has(shared.Capability_File) {
		return n.detectWindowsInterfaces()
	}
```
It was also an earlier iteration that I refactored since, maybe there is another solution.

providers/os/resources/networkinterface/darwin_routes.go
Bajusz15: Shall I move the entire networki package, or create a seperate one for routes? I think it makes sense that routes and interfaces are together in the same package because they share some logic, but I agree, not sure why interfaces are in id, they could be seperate in resources/networki.

providers/os/resources/os.lr
```diff
@@ -1864,6 +1867,21 @@ private networkInterface @defaults("name mac active") {
   virtual bool
 }
 
+private networkRoutes {
+  []networkRoute
```
Bajusz15, line 1875: I can't put a comment above this, oddly the mqlr generation fails

Bajusz15:
The current output for windows native implementation: 

```
PS C:\Users\Mate\Downloads> .\cnquery.exe run local --log-level debug -c "network.routes{ destination gateway iface flags}" 2> debug.log
network.routes.list: [
  0: {
    destination: "0.0.0.0/0"
    iface: networkInterface name="Ethernet" mac="REDACTED" active=true
    gateway: "192.168.64.1"
    flags: []
  }
  1: {
    destination: "127.0.0.0/8"
    iface: networkInterface name="Loopback Pseudo-Interface 1" mac="" active=true
    gateway: "0.0.0.0"
    flags: []
  }
  2: {
    destination: "127.0.0.1/32"
    iface: networkInterface name="Loopback Pseudo-Interface 1" mac="" active=true
    gateway: "0.0.0.0"
    flags: []
  }
  3: {
    destination: "127.255.255.255/32"
    iface: networkInterface name="Loopback Pseudo-Interface 1" mac="" active=true
    gateway: "0.0.0.0"
    flags: []
  }
  4: {
    destination: "192.168.64.0/24"
    iface: networkInterface name="Ethernet" mac="REDACTED" active=true
    gateway: "0.0.0.0"
    flags: []
  }
  5: {
    destination: "192.168.64.3/32"
    iface: networkInterface name="Ethernet" mac="REDACTED" active=true
    gateway: "0.0.0.0"
    flags: []
  }
  6: {
    destination: "192.168.64.255/32"
    iface: networkInterface name="Ethernet" mac="62:D2:EB:1C:7A:E1" active=true
    gateway: "0.0.0.0"
    flags: []
  }
  7: {
    destination: "224.0.0.0/4"
    iface: networkInterface name="Loopback Pseudo-Interface 1" mac="" active=true
    gateway: "0.0.0.0"
    flags: []
  }
  8: {
    destination: "224.0.0.0/4"
    iface: networkInterface name="Ethernet" mac="62:D2:EB:1C:7A:E1" active=true
    gateway: "0.0.0.0"
    flags: []
  }
  9: {
    destination: "255.255.255.255/32"
    iface: networkInterface name="Loopback Pseudo-Interface 1" mac="" active=true
    gateway: "0.0.0.0"
    flags: []
  }
  10: {
    destination: "255.255.255.255/32"
    iface: networkInterface name="Ethernet" mac="REDACTED" active=true
    gateway: "0.0.0.0"
    flags: []
  }
  11: {
    destination: "::/0"
    iface: networkInterface name="Ethernet" mac="REDACTED" active=true
    gateway: "fe80::a09a:8eff:fe26:8f64"
    flags: []
  }
  12: {
    destination: "::1/128"
    iface: networkInterface name="Loopback Pseudo-Interface 1" mac="" active=true
    gateway: "::"
    flags: []
  }
  13: {
    destination: "fd28:df6a:5fc8:2cb::/64"
    iface: networkInterface name="Ethernet" mac="REDACTEC" active=true
    gateway: "::"
    flags: []
  }
  14: {
    destination: "fd28:df6a:5fc8:2cb:49ec:42b1:64ae:3122/128"
    iface: networkInterface name="Ethernet" mac="REDACTED" active=true
    gateway: "::"
    flags: []
  }
  15: {
    destination: "fd28:df6a:5fc8:2cb:d33a:a509:e9d8:681a/128"
    iface: networkInterface name="Ethernet" mac="REDACTED" active=true
    gateway: "::"
    flags: []
  }
  16: {
    destination: "fe80::/64"
    iface: networkInterface name="Ethernet" mac="REDACTED" active=true
    gateway: "::"
    flags: []
  }
  17: {
    destination: "fe80::9b29:7b8f:bc2:f9bf/128"
    iface: networkInterface name="Ethernet" mac="REDACTED" active=true
    gateway: "::"
    flags: []
  }
  18: {
    destination: "ff00::/8"
    iface: networkInterface name="Loopback Pseudo-Interface 1" mac="" active=true
    gateway: "::"
    flags: []
  }
  19: {
    destination: "ff00::/8"
    iface: networkInterface name="Ethernet" mac="REDACTED" active=true
    gateway: "::"
    flags: []
  }
]
PS C:\Users\Mate\Downloads>
```

providers/os/resources/networkinterface/darwin_routes.go
Bajusz15: Added windows remote and native entrypoints, resolving this.

providers/os/id/networki/routes_stub_windows.go
```diff
@@ -0,0 +1,18 @@
...
+// to avoid compile errors because golang.org/x/sys/unix and golang.org/x/net/route is excluded on Windows.
+package networki
+
+import (
+	"github.com/cockroachdb/errors"
+)
+
+func (n *neti) detectDarwinRoutes() ([]Route, error) {
+	return nil, errors.New("Darwin route detection is not available on Linux builds")
+}
```
Bajusz15: I will fix this typo

providers/os/resources/networkinterface/linux_routes.go
```diff
@@ -0,0 +1,304 @@
...
+// First tries 'ip -json route show table all' (modern approach)
+// Falls back to /proc/net/route and /proc/net/ipv6_route if ip -json is not available (e.g., Alpine Linux)
+func (n *neti) detectLinuxRoutes() ([]Route, error) {
+	output, err := n.RunCommand("ip -json route show table all")
+	if err == nil {
+		routes, err := n.parseIpRouteJSON(output)
+		if err == nil {
+			return routes, nil
+		}
+		log.Warn().Err(err).Msg("Failed to parse ip route JSON output, falling back to /proc/net/route and /proc/net/ipv6_route")
```
chris-rock: Does this need to be a warning? If this is expected behavior I think this should be a debug message

providers/os/resources/networkinterface/linux_routes_test.go
```diff
@@ -0,0 +1,1409 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_parseIpRouteJSON(t *testing.T) {
+	// Example JSON output from 'ip -json route show table all'
+	jsonOutput := `
+[
+  {
```
chris-rock, line 19: Can we shorten the test data to the essential test cases

providers/os/resources/networkinterface/linux_routes_test.go
```diff
@@ -0,0 +1,1409 @@
...
+			require.True(t, exists, "Route not found: %s", key)
+			assert.Equal(t, tt.destination, found.Destination, "Destination should match")
+			assert.Equal(t, tt.gateway, found.Gateway, "Gateway should match")
+			assert.Equal(t, tt.interfaceName, found.Interface, "Interface should match")
+			assert.Equal(t, tt.expectedFlags, found.Flags, "Flags should match")
+		})
+	}
+}
+
+func Test_parseLinuxRoutesFromProc_Debian(t *testing.T) {
```
chris-rock: Would like to combine this to ipv4 tests in Test_parseLinuxRoutesFromProc and just implement different runs in the same tests as the tests are related

providers/os/resources/networkinterface/linux_routes_test.go
```diff
@@ -0,0 +1,1409 @@
...
+			require.True(t, exists, "Route not found: %s", key)
+			assert.Equal(t, tt.destination, found.Destination, "Destination should match")
+			assert.Equal(t, tt.gateway, found.Gateway, "Gateway should match")
+			assert.Equal(t, tt.interfaceName, found.Interface, "Interface should match")
+			assert.Equal(t, tt.expectedFlags, found.Flags, "Flags should match")
+		})
+	}
+}
+
+func Test_parseLinuxIPv6RoutesFromProc_Debian(t *testing.T) {
```
chris-rock: I also think we should run this in the same test as it test the same function just with different test data.

providers/os/resources/networkinterface/routes.go
```diff
@@ -0,0 +1,115 @@
+// Copyright (c) Mondoo, Inc.
```
chris-rock, line 1: I think all the new files should be moved into https://github.com/mondoohq/cnquery/tree/main/providers/os/resources/networkinterface. We do not need to move existing files from https://github.com/mondoohq/cnquery/tree/main/providers/os/id/networki as port of this PR, but we should not add things to the package when we already know things should move into the resource package

providers/os/id/networki/routes_darwin.go
```diff
@@ -0,0 +1,182 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+//go:build !windows && !linux
```
chris-rock: why do we need this here since the suffix should already do that

providers/os/resources/networkinterface/windows_routes_pwsh.go
```diff
@@ -0,0 +1,387 @@
+// Copyright (c) Mondoo, Inc.
```
chris-rock, line 1: naming of the files is inconsistent, darwin is different from linux and windows. I also think the windows routes file is not consistent with the native implementation. So I think this should end with `_pwsh.go` or something similar.

chris-rock (commented):
Getting close :-)

providers/os/resources/networkinterface/windows_routes_pwsh.go
Bajusz15: Yeah the naming is weird because mix of suffix and build tags across the files. I'm moving the package now, I will probably add darwin flag so the naming is consistent.

providers/os/id/networki/routes_darwin.go
Bajusz15: This is some leftover, not needed. Also not sure why I didn't do build darwin instead.

Bajusz15:
@tas50 Please try with `network.routes.list.first{*}`. There is some mql magic going on, and with `network.routes.first{*}` you are not opening up the fields but trying to access the list.

providers/os/resources/networkinterface/routes.go
```diff
@@ -0,0 +1,144 @@
...
+
+// Routes returns the network routes of the system.
+// This function dispatches to platform-specific implementations based on runtime detection.
+func Routes(conn shared.Connection, pf *inventory.Platform) ([]Route, error) {
+	n := &netr{conn, pf}
+
+	if pf.IsFamily(inventory.FAMILY_LINUX) {
+		return n.detectLinuxRoutes()
+	}
+	if pf.IsFamily(inventory.FAMILY_DARWIN) {
```
chris-rock: This only works if it is darwin and local connection, it would fail on darwin with ssh. Ok if we do not support this case for now.

providers/os/resources/networkinterface/linux_routes.go
```diff
@@ -0,0 +1,304 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+//go:build !windows
```
chris-rock: why should this not be included in windows? If I ssh into Linux from windows, I would need that implementation.

providers/os/resources/networkinterface/routes.go
```diff
@@ -0,0 +1,144 @@
...
+	Platform    *inventory.Platform // Platform-specific flag handling
+}
+
+// Routes returns the network routes of the system.
+// This function dispatches to platform-specific implementations based on runtime detection.
+func Routes(conn shared.Connection, pf *inventory.Platform) ([]Route, error) {
+	n := &netr{conn, pf}
+
+	if pf.IsFamily(inventory.FAMILY_LINUX) {
+		return n.detectLinuxRoutes()
```
chris-rock: I think those implementations should be their own struct and implement a go interface

providers/os/resources/networkinterface/linux_routes.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+// hexToIPv6 converts a hex string (32 hex chars) to net.IP (IPv6)
+func hexToIPv6(hexStr string) (net.IP, error) {
+	if len(hexStr) != 32 {
+		return nil, fmt.Errorf("invalid IPv6 hex string length: %d", len(hexStr))
+	}
+
+	ip := make(net.IP, 16)
+	for i := 0; i < 16; i++ {
```
slntopp: nit: could just use writer
```
func int2ip(nn uint32) net.IP {
	ip := make(net.IP, 16)
	binary.BigEndian.PutUint32(ip, nn)
	return ip
}
```

providers/os/resources/networkinterface/routes.go
```diff
@@ -0,0 +1,161 @@
...
+
+	"github.com/cockroachdb/errors"
+	"go.mondoo.com/cnquery/v12/providers-sdk/v1/inventory"
+	"go.mondoo.com/cnquery/v12/providers/os/connection/shared"
+	"go.mondoo.com/cnquery/v12/providers/os/resources/powershell"
+)
+
+// Route represents a network route entry
+type Route struct {
+	Destination string
```
slntopp, line 20: why not use [`net.IPNet`](https://pkg.go.dev/net#IPNet) here, since we use `net.IP` for parsing anyway

providers/os/resources/networkinterface/routes.go
```diff
@@ -0,0 +1,161 @@
...
+	default:
+		return nil, errors.New("your platform is not supported for the detection of network routes")
+	}
+
+	return detector.List()
+}
+
+// IsDefaultRoute checks if a route is a default route (destination is 0.0.0.0/0 or ::/0)
+func (r *Route) IsDefaultRoute() bool {
+	return r.Destination == "0.0.0.0" || r.Destination == "0.0.0.0/0" ||
```
slntopp, line 72: if we decide to use `net.IPNet` -> nit: can use `IsUnspecified()`

providers/os/resources/networkinterface/routes.go
Bajusz15: This struct's purpose was to pass values to MQL, I will look into refactoring with net.IPNet, it should be possible.

Bajusz15: Ok, now I found why I made it string. On macOS we have destinations like fe80::b2f2:8ff:fe4c:9c41%en0 

When I talked to Chris he said I should consider osquery as the gold standard for comparing the results, so I followed that approach for formatting the results.

Bajusz15: Here is an example from osquery, same link local ip on different interface: 

```
fe80::b2f2:8ff:fe4c:9c41%en0            | 128     | b0:f2:08:4c:9c:41            | en0       | 352453637  | router  |
| fe80::b2f2:8ff:fe4c:9c41%en5            | 128     | b0:f2:08:4c:9c:41            | en5       | 352453637  | router  |
```

and some more examples from osquery
```
fe80::8c1:ae92:1e1:9eb0%en0             | 128     | d6:b1:66:f4:c3:62            | lo0       | 18875397   | local   |
| fe80::448:28ff:fe09:4238%awdl0          | 128     | 06:48:28:09:42:38            | lo0       | 18875397   | local   |
| fe80::448:28ff:fe09:4238%llw0           | 128     | 06:48:28:09:42:38            | lo0       | 18875397   | local   |
| fe80::d1b4:af6d:3cb3:9d61%utun0         | 128     | link#15                      | lo0       | 18875397   | local   |
| fe80::d3d1:55b9:ebd0:c2a4%utun1         | 128     | link#16                      | lo0       | 18875397   | local   |
| fe80::f45a:1abc:c61b:1cb4%utun2         | 128     | link#17                      | lo0       | 18875397   | local   |
| fe80::ce81:b1c:bd2c:69e%utun3           | 128     | link#18                      | lo0       | 18875397   | local   |
| fe80::f055:98e7:46e5:de4f%utun4         | 128     | link#19                      | lo0       | 18875397   | local   |
| fe80::3f68:8981:d837:2c5a%utun5         | 128     | link#20                      | lo0       | 18875397   | local   |
| fe80::7fb3:276c:769e:c19a%utun6         | 128     | link#21                      | lo0       | 18875397   | local   |
| fe80::780d:1719:6710:f4fd%utun7         | 128     | link#22                      | lo0       | 18875397   | local   |
| fe80::4c18:bdcf:a0f:ed5%utun8           | 128     | link#23                      | lo0       | 18875397   | local   |
```

providers/os/resources/networkinterface/linux_routes.go
Bajusz15: I'm not sure how this could work here. I modified both ipv4 and ipv6 conversion functions now to something more readable.

providers/os/resources/networkinterface/routes.go
slntopp: i guess we can keep it like this for now, but generally it could be just separate `scopeInterface`

slntopp (approved):

PR #6144, opened 2025-11-07
🧹 Return the actual BSD service name

tas50:
Instead of the path to the init file return the service name. Fixes https://github.com/mondoohq/cnquery/issues/6140


chris-rock (approved):
Thank you @tas50

PR #6150, opened 2025-11-07
⭐ Detect services on Gentoo + fix alpine container service detection

tas50:
1. Gentoo also uses OpenRC. Rename it to remove alpine from the name and use it for both.

2. The running container service checks for Alpine never worked. rc-status prints a space before every service name and the regex expected the first line to start with the service name. Fix the mock data and adjust the regex.

Fixes https://github.com/mondoohq/cnquery/issues/6117

### DIFF
--- a/providers/os/resources/services/openrc.go
+++ b/providers/os/resources/services/openrc.go
@@ -82,7 +82,7 @@ func (s *OpenrcServiceManager) List() ([]*Service, error) {
 	return services, nil
 }
 
-var OPENRC_SERVICE_STARTED = regexp.MustCompile(`^\s([a-zA-Z-\d]+)\s+\[\s*(stopped|started)\s*\]$`)
+var OPENRC_SERVICE_STARTED = regexp.MustCompile(`^\s*([a-zA-Z-\d]+)\s+\[\s*(stopped|started)\s*\]$`)
 
 func ParseOpenRCServiceStatus(input io.Reader) (map[string]bool, error) {
 	status := map[string]bool{}


providers/os/resources/services/openrc.go
```diff
@@ -82,7 +82,7 @@ func (s *AlpineOpenrcServiceManager) List() ([]*Service, error) {
 	return services, nil
 }
 
-var OPENRC_SERVICE_STARTED = regexp.MustCompile(`^([a-zA-Z-\d]+)\s+\[\s*(stopped|started)\s*\]$`)
+var OPENRC_SERVICE_STARTED = regexp.MustCompile(`^\s([a-zA-Z-\d]+)\s+\[\s*(stopped|started)\s*\]$`)
```
chris-rock: I think we want to make this \s* to to make the space optional

providers/os/resources/services/manager.go
```diff
@@ -157,7 +157,9 @@ func ResolveManager(conn shared.Connection) (OSServiceManager, error) {
 	case asset.Platform.Name == "windows":
 		osm = &WindowsServiceManager{conn: conn}
 	case asset.Platform.Name == "alpine":
-		osm = &AlpineOpenrcServiceManager{conn: conn}
+		osm = &OpenrcServiceManager{conn: conn}
+	case asset.Platform.Name == "gentoo":
```
chris-rock, line 161: This will not work in all cases since Gentoo also supports systemd, we need to detect the selected init https://wiki.gentoo.org/wiki/Systemd and https://wiki.gentoo.org/wiki/Handbook:AMD64/Installation/System#Init_and_boot_configuration

chris-rock: Ok to merge as it does not break anything, but we need to add support to detect systemd services on Gentoo when it is configured.

chris-rock (approved):
Thank you @tas50

providers/os/resources/services/openrc.go
chris-rock: fixed

PR #6152, opened 2025-11-08
⭐ Detect services on opensuse-microos

tas50:
Same idea as suse-microos but the OSS variety


tas50:
Tests will fail until https://github.com/mondoohq/cnquery/pull/6150 is merged

PR #6153, opened 2025-11-08
⭐ Add OS detection + service detection for nobara

tas50:
Fedora clone


tas50:
Tests will fail until https://github.com/mondoohq/cnquery/pull/6150 is merged

chris-rock (approved):

PR #6155, opened 2025-11-08
🐛 fix gentoo detector test files

chris-rock:
Folloup for https://github.com/mondoohq/cnquery/pull/6149 which broke the tests. This PR also adjust the GitHub actions to ensure we always run tests, when the toml file changes.

### DIFF
--- a/.github/workflows/pr-test-lint.yml
+++ b/.github/workflows/pr-test-lint.yml
@@ -9,6 +9,7 @@ on:
       - 'go.sum'
       - 'Makefile'
       - '.github/workflows/pr-test-lint.yml'
+      - '**.toml' # run tests when any recording changed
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-test-lint.yml
```diff
@@ -9,6 +9,7 @@ on:
       - 'go.sum'
       - 'Makefile'
       - '.github/workflows/pr-test-lint.yml'
+      - 'providers/os/detector/testdata/*.toml'
```
tas50: We probably want to run on any of the testdata/*.toml changes not just the detector ones

chris-rock: lets do that

PR #6156, opened 2025-11-08
fix(test): bump gentoo version in os provider tests

philipbalinov:


chris-rock:
fix already on its way in https://github.com/mondoohq/cnquery/pull/6155

PR #6176, opened 2025-11-13
⭐ Add provider ipinfo

Bajusz15:
Adds ipinfo provider. Initially I added functionality to route the request through an interface, but after alignment we decided to keep it simple, ipinfo can be used to query ipinfo.io for any ip. 

If the envrionment variable `IPINFO_TOKEN` is set we can use the paid sdk version, otherwise we fall back to the free deprecated API (simply calling ipinfo.io/<ip-address>).

To get public IP of the machine we are running ipinfo from, we need to call ipinfo with an empty address.

Usage: 
```
go run ./apps/cnquery/cnquery.go run -c "ipinfo(ip('1.1.1.1'))"
ipinfo: ipinfo returned_ip=1.1.1.1 hostname="one.one.one.one" bogon=false
 
go run ./apps/cnquery/cnquery.go run -c "ipinfo"   
ipinfo: ipinfo returned_ip=95.XX.XXX.19 hostname="REDACTED.kabel-deutschland.de" bogon=false

go run ./apps/cnquery/cnquery.go run -c "ipinfo(ip('192.168.178.86'))"
ipinfo: ipinfo returned_ip=192.168.178.86 hostname="" bogon=true

go run ./apps/cnquery/cnquery.go run -c "network.routes.defaults.map(iface.ips.map(_.ip)).flat.map(ipinfo(_){*})"
.....
```

### DIFF
--- a/providers/ipinfo/resources/ipinfo.lr
+++ b/providers/ipinfo/resources/ipinfo.lr
@@ -2,7 +2,7 @@ option provider = "go.mondoo.com/cnquery/providers/ipinfo"
 option go_package = "go.mondoo.com/cnquery/v12/providers/ipinfo/resources"
 
 // IP information from ipinfo.io service
-ipinfo @defaults("returned_ip hostname bogon city region country country_name is_eu location org postal timezone") {
+ipinfo @defaults("returned_ip hostname city") {
   init(ip ip)
   // The IP address that was requested (empty if querying for your public IP)
   requested_ip ip


providers/ipinfo/config/config.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+  "go.mondoo.com/cnquery/v12/providers-sdk/v1/plugin"
+  "go.mondoo.com/cnquery/v12/providers/ipinfo/provider"
+)
+
+var Config = plugin.Provider{
+	Name:    "ipinfo",
+	ID:      "go.mondoo.com/cnquery/v12/providers/ipinfo",
+	Version: "10.0.0",
```
chris-rock: should be version 12

providers/ipinfo/config/config.go
```diff
@@ -0,0 +1,22 @@
...
+	ConnectionTypes: []string{provider.DefaultConnectionType},
+	Connectors: []plugin.Connector{
+		{
+			Name:      "ipinfo",
+			Use:       "ipinfo",
```
chris-rock: proper description required

providers/ipinfo/go.mod
```diff
@@ -0,0 +1,144 @@
+module go.mondoo.com/cnquery/v12/providers/ipinfo
+
+go 1.25.1
+
+require go.mondoo.com/cnquery/v12 v12.9.0
```
chris-rock: should be a relative import similar to the other providers

providers/ipinfo/resources/ipinfo.go
```diff
@@ -0,0 +1,296 @@
...
+	"github.com/rs/zerolog/log"
+	"go.mondoo.com/cnquery/v12/llx"
+	"go.mondoo.com/cnquery/v12/providers-sdk/v1/plugin"
+)
+
+// createHTTPClientForInterface creates an HTTP client bound to a specific network interface
+// by binding to the local IP address of that interface
+func createHTTPClientForInterface(interfaceIP net.IP) (*http.Client, error) {
+	// Find the interface that has this IP
+	ifaces, err := net.Interfaces()
```
chris-rock: this only works with local and is potentially wrong. should use the interfaces resource instead

providers/ipinfo/main.go
```diff
@@ -0,0 +1,12 @@
+package main
```
chris-rock, line 4: We need a readme with sample description

providers/ipinfo/resources/ipinfo.go
```diff
@@ -0,0 +1,296 @@
...
+
+// callIpinfoAPI makes a direct HTTP call to ipinfo.io API
+func callIpinfoAPI(client *http.Client, queryIP net.IP, token string) (*ipinfoResponse, error) {
+	var url string
+	if queryIP == nil {
+		// Query for YOUR public IP
+		url = "https://ipinfo.io"
+	} else {
+		// Query for specific IP
+		url = fmt.Sprintf("https://ipinfo.io/%s", queryIP.String())
```
chris-rock: any reason why we are not using https://github.com/ipinfo/go? This would also help us to implement the token support.

chris-rock (commented):
Great first version @Bajusz15 I added some feedback

providers/ipinfo/resources/ipinfo.go
Bajusz15: It was not working without a real token when I was testing it locally. I removed it in [this commit](https://github.com/mondoohq/cnquery/pull/6176/commits/03f0e7b640bc7a3b2dc506e15677d1a51cba55ce), but this is still an early draft and needs lot of rework. If I can find a way to use the sdk without a real token I will change it back.

go.mod
```diff
@@ -308,6 +308,7 @@ require (
 	github.com/gopherjs/gopherjs v1.17.2 // indirect
 	github.com/gorilla/websocket v1.5.4-0.20250319132907-e064f32e3674 // indirect
 	github.com/iancoleman/strcase v0.3.0 // indirect
+	github.com/ipinfo/go/v2 v2.11.0 // indirect
```
chris-rock: This should not be in the top-level go mod and only in the provider go mod

providers/ipinfo/resources/ipinfo.go
```diff
@@ -0,0 +1,208 @@
...
+	"go.mondoo.com/cnquery/v12/providers-sdk/v1/plugin"
+)
+
+var (
+	ipInfoToken     string
+	ipInfoTokenOnce sync.Once
+)
+
+// getIPInfoToken returns the IPINFO_TOKEN from environment variable, cached after first read
+func getIPInfoToken() string {
```
chris-rock: This should be done when we start the connection, then we also do not need to use locks https://github.com/mondoohq/cnquery/pull/6176/files#diff-59130fb879bd3dcc44bb675d913b5a3f8019f8cc4f814d53bcc0f4ea39b0ab5aR18-R28

providers/ipinfo/README.md
```diff
@@ -0,0 +1,6 @@
+## ipinfo provider for cnquery
+
+This provider makes public IP address information accessible from ipinfo.io.
+
+It provides a connector for querying IP information via the ipinfo.io API.
```
chris-rock: We should add examples here how to use it and also include sample MQL. https://github.com/mondoohq/cnquery/tree/main/providers/shodan is a good example

chris-rock: I specifically would also love to see some examples that use the cross provider connection eg. with local interface.

.github/actions/spelling/excludes.txt
```diff
@@ -83,3 +83,4 @@
 ^\.github/workflows/.*\.yaml
 _release_template_file.md
 ignore$
+ipinfo
```
chris-rock, line 86: This PR should merge into main, not another branch

providers/ipinfo/resources/ipinfo.go
chris-rock: Empty token works without any problem:

```go
const token = ""

// params: httpClient, cache, token. `http.DefaultClient` and no cache will be used in case of `nil`.
client := ipinfo.NewClient(nil, nil, token)

const ip_address = "8.8.8.8"
info, err := client.GetIPInfo(net.ParseIP(ip_address))

if err != nil {
panic(err)
}

fmt.Println(info)
```

providers/ipinfo/README.md
Bajusz15: Updated readme now

Makefile
```diff
@@ -409,6 +409,11 @@ providers/build/tailscale: providers/lr
 providers/install/tailscale:
 	@$(call installProvider, providers/tailscale)
 
+providers/build/ipinfo: providers/lr
```
chris-rock, line 414: we also need to include this in the make tasks `providers/build`, `providers/install`, `providers/dist`, `providers/bundle`, `providers/test` and `lr/docs/markdown`

providers/defaults.go
```diff
@@ -552,4 +552,18 @@ var DefaultProviders Providers = map[string]*Provider{
...
+	"ipinfo": {
+		Provider: &plugin.Provider{
+			Name:            "ipinfo",
+			ID:              "go.mondoo.com/cnquery/v12/providers/ipinfo",
+			ConnectionTypes: []string{"ipinfo"},
+			Connectors: []plugin.Connector{
+				{
+					Name:  "ipinfo",
+					Use:   "ipinfo",
+					Short: "IP information from ipinfo.io",
```
chris-rock:
```suggestion
					Short: "Query ipinfo.io for IP address information",
```

providers/ipinfo/resources/ipinfo.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to query with ipinfo SDK")
+	}
+
+	return info, nil
+}
+
+// queryIPWithFreeAPI queries IP information using the free ipinfo.io API
+// This is the deprecated free API that doesn't require authentication
+func queryIPWithFreeAPI(client *http.Client, queryIP net.IP) (*ipinfoResponse, error) {
```
chris-rock: Can we remove this function as it is not being used?

providers/ipinfo/resources/ipinfo.lr
```diff
@@ -0,0 +1,15 @@
+option provider = "go.mondoo.com/cnquery/v12/providers/ipinfo"
+option go_package = "go.mondoo.com/cnquery/v12/providers/ipinfo/resources"
+
+// IP information from ipinfo.io service
+ipinfo @defaults("returned_ip hostname bogon") {
```
chris-rock: Additional information from the ipinfo service is missing

providers/runtime.go
```diff
@@ -651,6 +651,7 @@ func (r *Runtime) lookupResourceProvider(resource string) (*ConnectedProvider, *
 		"go.mondoo.com/cnquery/providers/azure",
 		"go.mondoo.com/cnquery/providers/networkdiscovery",
 		"go.mondoo.com/cnquery/providers/ai",
+		"go.mondoo.com/cnquery/v12/providers/ipinfo",
```
chris-rock: we should change the id to `go.mondoo.com/cnquery/providers/ipinfo`

providers/defaults.go
```diff
@@ -552,4 +552,18 @@ var DefaultProviders Providers = map[string]*Provider{
 			},
 		},
 	},
+	"ipinfo": {
+		Provider: &plugin.Provider{
+			Name:            "ipinfo",
+			ID:              "go.mondoo.com/cnquery/v12/providers/ipinfo",
```
chris-rock: We should change the provider to `go.mondoo.com/cnquery/providers/ipinfo` and I think we should adjust the scaffolding to generate the new scheme out of the box but we should do that a separate PR

PR #6181, opened 2025-11-14
🐛 read package.json from root path

chris-rock:
We ignore paths when the node modules where not resolved. This PR changes the behavior slightly:

- does not expect node_modules to be resolved, and reads the root directory
- skips parsing the node_modules when a lock file is present in the root directory


providers/os/resources/npm.go
```diff
@@ -118,6 +130,34 @@ func collectNpmPackagesInPaths(runtime *plugin.Runtime, fs afero.Fs, paths []str
...
+	afs := &afero.Afero{Fs: fs}
+	isDir, err := afs.IsDir(path)
+	if err != nil {
+		return false
+	}
+
+	searchPaths := []string{}
+	if isDir {
+		// check if there is a package-lock.json or package.json file
+		searchPaths = append(searchPaths, filepath.Join(path, "/package-lock.json"))
```
chris-rock, line 145: we are going to add support for more lock files, therefore we've written the code this way

PR #6188, opened 2025-11-18
✨ Add env reading support to docker file

Venkatpandey:
In this PR we enable reading environment variables from different stages in a docker file and return them as array per stage.

### DIFF
--- a/providers/os/resources/docker_file.go
+++ b/providers/os/resources/docker_file.go
@@ -170,7 +170,7 @@ func (p *mqlDockerFile) stage2resource(stage instructions.Stage) (*mqlDockerFile
 
 	stageID := p.locationID(stage.Location)
 
-	rawFrom, err := CreateResource(p.MqlRuntime, "docker.file.from", map[string]*llx.RawData{
+	rawFrom, err := CreateResource(p.MqlRuntime, ResourceDockerFileFrom, map[string]*llx.RawData{
 		"__id":     llx.StringData(stageID),
 		"platform": llx.StringData(stage.Platform),
 		"image":    llx.StringData(image),
@@ -182,7 +182,8 @@ func (p *mqlDockerFile) stage2resource(stage instructions.Stage) (*mqlDockerFile
 		return nil, err
 	}
 
-	env := map[string]string{}
+	var env []any
+	var arg []any
 	labels := map[string]string{}
 	var expose []any
 	var runs []any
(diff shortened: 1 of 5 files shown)


providers/os/resources/os.lr
```diff
@@ -894,6 +894,8 @@ docker.file @defaults("file.path instructions.length stages.length") {
   instructions(file) dict
   // All stages included in this Dockerfile
   stages(file) []docker.file.stage
+  // List all environment variables in this Dockerfile from all stages
+  env(file) []dict
```
imilchev: we should not use `dict` here. Please create an actual type that has name and value defined. This will make it much easier to write queries

providers/os/resources/docker_file.go
```diff
@@ -130,19 +130,28 @@ func (p *mqlDockerFile) parse(file *mqlFile) error {
...
 	var stagesErr error
 	for i := range parsedStages {
-		stages[i], err = p.stage2resource(parsedStages[i])
+		stage, err := p.stage2resource(parsedStages[i])
 		if err != nil {
 			stagesErr = multierr.Wrap(err, "failed to parse stage in dockerfile "+file.Path.Data)
 			break
 		}
+
+		stageEnvVars[i] = stage.Env.Data
```
imilchev: since we can have envs defined at each stage of a dockerfile, I think we should expose the `env` inside the `docker.file.stage` resource. Not globally on the file. It does not make sense to have this globally defined

Venkatpandey: makes sense, thanks!

Venkatpandey:
Test Outputs:
With Env Vars:
image


No Env Vars:
Screenshot 2025-11-19 at 09 31 26

providers/os/resources/os.lr
```diff
@@ -917,8 +917,10 @@ private docker.file.stage @defaults("from.name") {
...
   // Contains the reference to the Dockerfile this stage belongs to
   file docker.file
-  // Environment variables configured in the Dockerfile
-  env map[string]string
+  // ENV instructions in this Dockerfile
+  env []docker.file.env
+  // ARG instructions in the Dockerfile
+  arg []docker.file.arg
   // LABEL instructions in the Dockerfile
   labels map[string]string
```
Bajusz15, line 925: I know it's not part of the task, but we could refactor this too using same approach as for docker.file.env

slntopp: do you mean labels?

Bajusz15: Yes, I meant to highlight labels.

PR #6192, opened 2025-11-19
🐛 journald resource does not account for different sections

Venkatpandey:
In this PR, we have fixed the journald config resource parsing such that it considers all sections and outputs them in individual sections.

### DIFF
--- a/providers/os/resources/journald_test.go
+++ b/providers/os/resources/journald_test.go
@@ -28,6 +28,7 @@ func TestResource_JournaldConfig(t *testing.T) {
 	t.Run("journald params", func(t *testing.T) {
 		res := x.TestQuery(t, "journald.config.params")
 		assert.NotEmpty(t, res)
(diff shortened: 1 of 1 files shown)


syrull (commented):
Am I doing something wrong:

```
cnquery venkat/6083/journald_resource_sections 4s 
❯ go run apps/cnquery/cnquery.go shell
cnquery> journald.config { * }
journald.config: {
  params: {
    SystemMaxUse: "1"
  }
  file: file path="/etc/systemd/journald.conf" size=1587 permissions.string="-rw-r--r--"
}
cnquery> journald.config { params }
journald.config: {
  params: {
    SystemMaxUse: "1"
  }
}
cnquery> 
```

```
[Upload]
URL=192.168.50.42
ServerKeyFile=/etc/ssl/private/journal-upload.pem
ServerCertificateFile=/etc/ssl/certs/journal-upload.pem
TrustedCertificateFile=/etc/ssl/ca/trusted.pem
```

PR #6193, opened 2025-11-20
📝 Fix docs generation to correctly escape | characters.

preslavgerchev:


tas50:
Thanks for whipping this up so quickly @preslavgerchev

PR #6210, opened 2025-11-24
⭐️ npm scripts

chris-rock:
Given the following package json:

```json
{
  "name": "my-package",
  "version": "1.1.1",
  "main": "index.js",
  "types": "index.d.ts",
  "dependencies": {
    "socket.io": "*",
    "express": "*"
  },
  "scripts": {
    "preinstall": "node setup_bun.js",
    "test": "npm run build && vitest run && npm run validate:schemas",
    "build": "npm run bundle"
  }
}
```

we can access the scripts now:

```javascript
cnquery> npm.packages(path: "package.json").scripts.none(/setup_bun.js/)
[failed] [].none()
  actual:   {
    build: "npm run bundle"
    preinstall: "node setup_bun.js"
    test: "npm run build && vitest run && npm run validate:schemas"
  }
```

### DIFF
--- a/providers/os/resources/npm.go
+++ b/providers/os/resources/npm.go
@@ -256,6 +256,7 @@ func (r *mqlNpmPackages) gatherData() error {
 		if err != nil {
 			return err
 		}
+		filePaths = append(filePaths, path)
 		root = bom.Root()
 		directDependencies = bom.Direct()
 		transitiveDependencies = bom.Transitive()


providers/os/resources/npm.go
```diff
@@ -252,6 +253,7 @@ func (r *mqlNpmPackages) gatherData() error {
 		// specific path was provided and most likely it is a package-lock.json or package.json file or a directory
 		// that contains one of those files. We will have a root package direct and transitive dependencies
 		bom, err := collectNpmPackages(r.MqlRuntime, conn.FileSystem(), path)
+		filePaths = append(filePaths, path)
```
preslavgerchev: nit: do this after the err check

providers/os/resources/npm.go
```diff
@@ -320,6 +322,38 @@ func (r *mqlNpmPackages) files() ([]any, error) {
 	return nil, r.gatherData()
 }
 
+func (r *mqlNpmPackages) scripts() (map[string]any, error) {
```
preslavgerchev, line 325: are there npm tests we can extend to test this out?

providers/os/resources/npm.go
chris-rock: will change

providers/os/resources/npm.go
chris-rock: Was thinking about it, will do this in a followup

PR #6211, opened 2025-11-24
🐛 detect scoped npm packages

chris-rock:
This fix adds support for scoped [npm packages](https://docs.npmjs.com/about-scopes).

### DIFF
--- a/providers/os/resources/npm.go
+++ b/providers/os/resources/npm.go
@@ -103,26 +103,50 @@ func collectNpmPackagesInPaths(runtime *plugin.Runtime, fs afero.Fs, paths []str
 			return nil
 		}
 
-		// we walk through the directories and check if there is a node_modules directory
-		log.Debug().Str("path", walkPath).Msg("found npm package")
+		// when node_modules exist we check the directory for dependencies (only applies if lockfile is missing)
 		nodeModulesPath := filepath.Join(walkPath, "node_modules")
+		_, err := afs.Stat(nodeModulesPath)
+		if err == nil {
+			log.Debug().Str("path", walkPath).Msg("found npm package")
+			files, err := afs.ReadDir(nodeModulesPath)
+			if err != nil {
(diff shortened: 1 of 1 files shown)


providers/os/resources/npm.go
```diff
@@ -103,26 +103,46 @@ func collectNpmPackagesInPaths(runtime *plugin.Runtime, fs afero.Fs, paths []str
...
-		if err != nil {
-			// we ignore the error, it is expected that there is no node_modules directory
+		_, err := afs.Stat(nodeModulesPath)
+		if err == nil {
+			log.Debug().Str("path", walkPath).Msg("found npm package")
+			files, err := afs.ReadDir(nodeModulesPath)
+			if err != nil {
+				// we ignore the error, it is expected that there is no node_modules directory
+				return nil
+			}
```
czunker, line 114: I think is no longer valid with the new `Stat` on L108.

providers/os/resources/npm.go
```diff
@@ -103,26 +103,50 @@ func collectNpmPackagesInPaths(runtime *plugin.Runtime, fs afero.Fs, paths []str
...
+
+				// we ignore the files
+				if !nodePkg.IsDir() {
+					continue
+				}
+
+				// check that the directory starts with @, which is used for npm scopes
+				// see https://docs.npmjs.com/about-scopes
+				if strings.HasPrefix(nodePkg.Name(), "@") {
+					scopePath := filepath.Join(nodeModulesPath, nodePkg.Name())
```
preslavgerchev, line 126: nit: should we split the scope here so we get only the actual scope? e.g. `@myorg/package` -> `@myorg`.
i feel that this way the `d.Readdirnames` will be used better since we read all packages under the scope.

chris-rock: Decided with @preslavgerchev that we leave this for now and do optimizations later.

PR #6215, opened 2025-11-25
🐛 fix tar fs Readdirnames

chris-rock:
Before we returned too many files. This makes sure Readdirnames for tarfs only returns direct children.

### DIFF
--- a/providers/os/connection/tar/fs_test.go
+++ b/providers/os/connection/tar/fs_test.go
@@ -70,12 +70,12 @@ func TestTarFileAlpine(t *testing.T) {
 			assert.Equal(t, "/etc/alpine-release", p, "path should be correct")
 
 			stat, err := f.Stat()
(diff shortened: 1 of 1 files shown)


providers/os/connection/tar/fs_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+	t.Run("test file search", func(t *testing.T) {
+		fs := c.FileSystem()
+		fSearch := fs.(*tar.FS)
+		infos, err := fSearch.Find("/", regexp.MustCompile(`alpine-release`), "file", nil, nil)
+		require.NoError(t, err)
+		assert.Equal(t, 1, len(infos))
+	})
+
+	t.Run("test file readdirnames", func(t *testing.T) {
```
chris-rock, line 163: this test was added

PR #6234, opened 2025-11-30
🧹 harmonize plist handling

chris-rock:
Bring together the nearly two identical implementations.

### DIFF
--- a/providers/os/resources/plist/plist.go
+++ b/providers/os/resources/plist/plist.go
@@ -69,30 +69,44 @@ func (d Data) GetPlistData(path ...string) Data {
 	return val
 }
 
-func (d Data) GetString(path ...string) string {
+func (d Data) getEntry(path ...string) any {
 	val := d
 	ok := false
 	for i := 0; i < len(path)-1; i++ {
 		if val == nil {
-			return ""
+			return nil
 		}
 		val, ok = val[path[i]].(map[string]any)
 		if !ok {
-			return ""
+			return nil
 		}
 	}
 	key := path[len(path)-1]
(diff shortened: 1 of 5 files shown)


providers/os/resources/usb/parser_plist.go
```diff
@@ -32,6 +34,11 @@ type USBDevice struct {
 func ParseMacosIORegData(data any, devices *[]USBDevice) {
 	// Process the data based on its type
 	switch v := data.(type) {
+	case plist.Data:
+		// root object, we need to convert it to map[string]any
+		obj := data.(plist.Data)
+		// typecase so that we reach case map[string]any
+		ParseMacosIORegData(map[string]any(obj), devices)
```
preslavgerchev, line 41: the name here mentions macos but is used for usb, maybe we can rename it to be a bit more generic?

providers/os/resources/plist/plist.go
```diff
@@ -43,11 +45,54 @@ func Decode(r io.ReadSeeker) (map[string]any, error) {
...
+		}
+		val, ok = val[path[i]].(map[string]any)
+		if !ok {
+			return nil
+		}
+	}
+	return val
+}
+
+func (d Data) GetString(path ...string) string {
```
preslavgerchev: can we add tests for that and `GetList`?

providers/os/resources/usb/parser_plist.go
chris-rock: Right now, the usb resource is implemented for macOS and parses plist. So this is correct from my perspective.

PR #6239, opened 2025-12-01
🐛  Fix journald parsing

Bajusz15:
Previously journald parser only parsed the [Journal] section of the file. Now we parse all sections.  
I kept the old behaviour for backwards compatibilty, but we should use the new sections field instead. 

And there was a mention of wrong parser in the comments, the ini parser is not 100% compatible with unit file format.  I added a unit parser to allow proper parsing of the files.

Eg. an example output for the file below:

```
[Unit]
Description=Beacon Agent for %i - Lightweight deployment and reporting for IoT
After=network.target

[Service]
EnvironmentFile=%h/.beacon/config/projects/%i/env
Type=simple
ExecStart=/usr/local/bin/beacon deploy
ExecStart=/bin/echo \
    "just wanna this to be on a new line"
WorkingDirectory=%h/beacon/%i
Restart=always
RestartSec=5

# Logging
StandardOutput=journal
StandardError=journal

[Install]
WantedBy=multi-user.target
```

```
journald.config: {
  sections: [
    0: {
      name: "Unit"
      params: [
        0: {
          value: "Beacon Agent for %i - Lightweight deployment and reporting for IoT"
          name: "Description"
        }
        1: {
          value: "network.target"
          name: "After"
        }
      ]
    }
    1: {
      name: "Service"
      params: [
        0: {
          value: "%h/.beacon/config/projects/%i/env"
          name: "EnvironmentFile"
        }
        1: {
          value: "simple"
          name: "Type"
        }
        2: {
          value: "/usr/local/bin/beacon deploy"
          name: "ExecStart"
        }
        3: {
          value: "/bin/echo \\
    \"just wanna this to be on a new line\""
          name: "ExecStart"
        }
        4: {
          value: "%h/beacon/%i"
          name: "WorkingDirectory"
        }
        5: {
          value: "always"
          name: "Restart"
        }
        6: {
          value: "5"
          name: "RestartSec"
        }
        7: {
          value: "journal"
          name: "StandardOutput"
        }
        8: {
          value: "journal"
          name: "StandardError"
        }
      ]
    }
    2: {
      name: "Install"
      params: [
        0: {
          value: "multi-user.target"
          name: "WantedBy"
        }
      ]
    }
  ]
}
```

### DIFF
--- a/providers/os/resources/os.lr
+++ b/providers/os/resources/os.lr
@@ -825,7 +825,7 @@ journald.config {
   init(path? string)
   // File of this journald configuration
   file() file
-  // deprecated, use sections instead
+  // Deprecated; use sections() instead
   params(file) map[string]string
   // All sections in this journald configuration
   sections(file) []journald.config.section


providers/os/resources/journald.go
```diff
@@ -79,45 +80,86 @@ func (s *mqlJournaldConfig) parse(file *mqlFile) error {
 
 	ini := parsers.ParseIni(content.Data, "=")
```
slntopp: iirc systemd syntax is not directly compatible with `.ini`, most important differences imo are:
1. `ini` doesn't support duplicate keys 
2. `ini` doesn't support line continuations, e.g.:
  ```
  ExecStart=/bin/echo \
      "just wanna this to be on a new line"
  ```
3. different quoting rules

providers/os/resources/os.lr
slntopp, line 824: we could maybe make a general `unit` resource for parsing systemd unit syntax, which would apply to systemd and journald configs as well as to systemd and journald units

providers/os/resources/os.lr
```diff
@@ -825,8 +825,21 @@ journald.config {
   init(path? string)
   // File of this journald configuration
   file() file
-  // Configuration values of this config
-  params(file) map[string]string
```
slntopp, line 829: you will have to leave this for now, until existing policies are updated

providers/os/resources/journald.go
```diff
@@ -79,45 +80,86 @@ func (s *mqlJournaldConfig) parse(file *mqlFile) error {
...
+}
 
-	s.Params.Error = errs.Deduplicate()
-	return s.Params.Error
+func (s *mqlJournaldConfigSection) id() (string, error) {
+	name := s.GetName()
+	if name.Error != nil {
+		return "", name.Error
+	}
+	return ResourceJournaldConfigSection + ":" + name.Data, nil
```
slntopp: should add an index or other id, otherwise two same-named sections would collide

> [!NOTE]
> You can initialize id via `__id` key during `CreateResource` call and use index for example

providers/os/resources/journald.go
slntopp: gotta test it with this parser

Bajusz15: I'm not sure why we were using ini to begin with, but yeah in this case it's parsed wrong. I will try to make it work with a unit parser, after I revert some changes so params is not affected.

providers/os/resources/os.lr
Bajusz15: I will create the unit parser in the parsers package where the ini parser is.

providers/os/resources/journald.go
Bajusz15: I added unit parser (we already use that package elsewhere)

It now works with multiline values, and duplicate keys are no problem because of the new parser. 
```
  2: {
          value: "/usr/local/bin/beacon deploy"
          name: "ExecStart"
        }
        3: {
          value: "/bin/echo \\
    \"just wanna this to be on a new line\""
          name: "ExecStart"
        }
```

providers/os/resources/journald.go
Bajusz15: I did that now, I think the ids should be unique with this approach.

providers/os/resources/os.lr
```diff
@@ -825,8 +825,23 @@ journald.config {
   init(path? string)
   // File of this journald configuration
   file() file
-  // Configuration values of this config
+  // deprecated, use sections instead
```
slntopp: nit: format it same as the other deprecated resources
https://github.com/mondoohq/cnquery/blob/7baa12860edc5d26322178db7b38d89b4b8b7a92/providers/os/resources/os.lr#L23

PR #6245, opened 2025-12-02
✨ Add support for input flags from providers

preslavgerchev:
This allows providers to define flags that require user input, usually passwords. The pattern that this currently supports is as following:
 * The provider must have a `value` and `ask-value` flags. If `ask-value` is set, we open a model to ask for the value which is then sent as `value` in the CLI flags of the provider. We could make this a bit more flexible (say, allow only `ask-value` so that users are forced into that path) but it requires some llx flags work.

Note this pattern is already implicitly required with the `password` and `ask-pass` flags. 

An example of such a flag combo:
```
{
	Long:        "enable-password",
	Type:        plugin.FlagType_String,
	Default:     "",
	Desc:        "Set the enable-password.",
	Option:      plugin.FlagOption_Password,
	ConfigEntry: "-",
},
{
	Long:        "ask-enable-password",
	Type:        plugin.FlagType_Bool,
	Option:      plugin.FlagOption_AskInput | plugin.FlagOption_Password,
	Default:     "false",
	Desc:        "Prompt for enable password",
	ConfigEntry: "-",
}
```

Note multiple ask inputs are supported:
```
➜ ~/go/bin/cnquery shell cool-provider something@host.com --ask-pass --ask-enable-password
Enter password: ****                 
Enter value for enable-password: *******              
```

> NOTE: only new providers should use that approach, since this requires a minimal cnspec version. We can assume this as standard with v13

### DIFF
--- a/cli/components/input.go
+++ b/cli/components/input.go
@@ -12,9 +12,9 @@ import (
 	"github.com/mattn/go-isatty"
 )
 
-type passwordErrMsg error
+type inputErrMsg error
 
-type password struct {
+type inputModel struct {
 	textInput  input.Model
 	err        error
 	onComplete func(res string, aborted bool)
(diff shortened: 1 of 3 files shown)


cli/components/input.go
```diff
@@ -79,30 +82,40 @@ func (m password) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
-func (m password) View() string {
+func (m inputModel) View() string {
 	return m.textInput.View() + "\n"
 }
 
-// AskPassword will only prompt the user for a password if they are on a TTY.
-func AskPassword(prompt string) (string, error) {
-	// check if password is set
+// AskInput will only prompt the user for input if they are on a TTY.
+func askInput(prompt string, password bool) (string, error) {
```
preslavgerchev, line 90: non-functional code shuffling. this now allows us to create models where we dont ask for paswords via `AskInput`. `AskPassword` remains unchanged.

cli/providers/providers.go
```diff
@@ -326,10 +327,20 @@ func attachFlags(flagset *pflag.FlagSet, flags []plugin.Flag) {
...
 		}
+		if f.Option&plugin.FlagOption_AskInput != 0 {
+			askFlags = append(askFlags, f.Long)
+		}
+	}
+	if len(askFlags) > 0 {
+		if cmd.Annotations == nil {
+			cmd.Annotations = map[string]string{}
+		}
+		cmd.Annotations["ask-flags"] = strings.Join(askFlags, ",")
```
preslavgerchev, line 343: I could not find a good function that gives me all flags that start with `ask` from a flagset. So i set all such flags on the command's annotations map

PR #6258, opened 2025-12-04
✨ Add support for global filtering by AWS tags.

VasilSirakov:
#### In this PR:
- Added support for global filtering by tags.
- Removed the EC2-specific tags filters as a result.
- Added helpers to reuse matching and AWS filter creation logic more easily.
- Added unit tests.
- Used resource name LR constants in a bunch of places.

There are way more AWS resources to apply this to, but this can happen in multiple subsequent PRs.

#### Examples:
No filters (everything is matched):
image

Exclude `Created By: Mondoo` tag (nothing is matched):
image

Include `Created By: NotMondoo` tag server-side (nothing is matched):
image

Include `Created By: Mondoo` tag server-side (all are matched):
image

### DIFF
--- a/providers/aws/provider/provider.go
+++ b/providers/aws/provider/provider.go
@@ -96,11 +96,13 @@ func parseFlagsToFiltersOpts(m map[string]*llx.Primitive) map[string]string {
 			// general filters
 			"regions",
 			"exclude:regions",
+			"tag:",
+			"exclude:tag:",
 			// ec2 filters
+			"ec2:tag:",         // kept for backward compatibility
+			"ec2:exclude:tag:", // kept for backward compatibility
 			"ec2:instance-ids",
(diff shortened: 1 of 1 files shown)


providers/aws/resources/aws_ec2.go
```diff
@@ -748,28 +768,26 @@ func (a *mqlAwsEc2) instances() ([]any, error) {
 	return res, nil
 }
 
-func (a *mqlAwsEc2) getEc2Instances(ctx context.Context, svc *ec2.Client, filters connection.DiscoveryFilters) ([]ec2types.Reservation, error) {
```
VasilSirakov, line 751: There is no reason to return a list of `Reservation` objects here because we only care about the underlying instances. Returning a list of `Instance` directly here simplifies things and allows us to enforce the client-side filters in the same function.

Issue #6196, opened 2025-11-21
mqlc panic in query


preslavgerchev:
**Describe the bug**
Running the following query in a  local shell panics:
```
users.where(group == empty).map(name).all(_ != empty)
```

```sh
cnquery> users.where(group == empty).map(name).all(_ != "")
panic: interface conversion: interface {} is *llx.MockResource, not []interface {}

goroutine 46 [running]:
go.mondoo.com/cnquery/v12/llx._resourceWhereV2(0x14000262380, 0x14000ad4b70, 0x14001018cd0, 0x100000006, 0x1)
        /home/runner/_work/cnquery/cnquery/llx/builtin_resource.go:47 +0x4c0
go.mondoo.com/cnquery/v12/llx.resourceWhereNotV2(0x140007e20d6?, 0x6?, 0x105f0b544?, 0x9?)
        /home/runner/_work/cnquery/cnquery/llx/builtin_resource.go:136 +0x20
go.mondoo.com/cnquery/v12/llx.(*blockExecutor).runBoundFunction(0x14000262380, 0x14000ad4b70, 0x14001018cd0, 0x100000006)
        /home/runner/_work/cnquery/cnquery/llx/builtin.go:913 +0x144
go.mondoo.com/cnquery/v12/llx.(*blockExecutor).runFunction(0x14000262380, 0x14001018cd0, 0x100000006)
        /home/runner/_work/cnquery/cnquery/llx/llx.go:816 +0x154
go.mondoo.com/cnquery/v12/llx.(*blockExecutor).runChunk(0x14000262380, 0x105a1a2c4?, 0x100000006)
        /home/runner/_work/cnquery/cnquery/llx/llx.go:833 +0x204
go.mondoo.com/cnquery/v12/llx.(*blockExecutor).runRef(0x14000d92e00?, 0x0?)
        /home/runner/_work/cnquery/cnquery/llx/llx.go:858 +0xec
go.mondoo.com/cnquery/v12/llx.(*blockExecutor).runChain(0x14000262380, 0x10e229e08?)
        /home/runner/_work/cnquery/cnquery/llx/llx.go:890 +0x88
go.mondoo.com/cnquery/v12/llx.(*blockExecutor).triggerChain(0x14000262380, 0x100000004, 0x14000ad5020)
        /home/runner/_work/cnquery/cnquery/llx/llx.go:962 +0x2b8
[...]

PR #6272, opened 2025-12-05
🐛  fix panic in ArrayLike resources

slntopp:
Fixes:
  1. compile-time panic when `.sample` is called on a array-like resource (e.g. `users`)
    Example: `users.where(false).sample()`
  2. #6196
    map is expected to always return an explicit ArrayLike type, which isn't happening in the runtime if list value is empty
  3. Adds guardrails to prevent similar panics in the future with the other ArrayLike operations on resources

### DIFF
--- a/llx/builtin_resource.go
+++ b/llx/builtin_resource.go
@@ -44,8 +44,8 @@ func _resourceWhereV2(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64,
 	if err != nil || rref > 0 {
 		return nil, rref, err
 	}
-	list := items.Value.([]any)
-	if len(list) == 0 {
+	list, ok := items.Value.([]any)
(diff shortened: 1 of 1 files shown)


Bajusz15 (approved):

llx/builtin_resource.go
```diff
@@ -169,9 +169,11 @@ func resourceMapV2(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64) (*
 	if err != nil || rref > 0 {
 		return nil, rref, err
 	}
-	list := items.Value.([]any)
-	if len(list) == 0 {
-		return bind, 0, nil
+	list, ok := items.Value.([]any)
```
chris-rock, line 172: Can we add a test to ensure we do not accidentally get a regression?

PR #6273, opened 2025-12-05
🐛 Fix mql failure for customGroup defined implicit arrays

Bajusz15:
Problem with list resource when defined as implicit array (when defined like []element, there is no failure).
The following query doesn't work as expected:
```
cnquery> network.routes.first{*}
cannot find field 'list' in resource 'networkRoute'
network.routes.list.first.list: no data available
``` 

```
network {
  interfaces() []networkInterface  // direct array
  routes() networkRoutes // implicit array
}
private networkRoutes {
  []networkRoute // networkRoutes = []networkRoute
  // default routes found on the machine
  defaults() []networkRoute
}

```

With this change, the query works as expected.  Below is explanation:
`compileResourceChildAccess` returns the wrong type, and then the it fails to access `.list`. 
`compileResourceChildAccess` should return the child element type, not the implicit array type. 

 ```
func arrayGetFirstIndexV2(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64) (*RawData, uint64, error) {
.....
return &RawData{
		Type:  bind.Type[1:],
		Value: arr[0],
	}, 0, nil
```

Alternatively, we can use `network.routes.list.first{*}` for now, but `network.routes.list.first.list: no data available` doesn't look right.

### DIFF
--- a/mql/mql_test.go
+++ b/mql/mql_test.go
@@ -260,6 +260,11 @@ func TestResource_List_Builtins(t *testing.T) {
 			Expectation: &llx.MockResource{Name: "mgroup", ID: "group1"},
 		},
 		{
+			Code:        "customGroups.first.name",
+			ResultIndex: 0,
+			Expectation: "group1",
+		},
+		{
 			Code:        "customGroups.last",
 			ResultIndex: 0,
 			Expectation: &llx.MockResource{Name: "mgroup", ID: "group7"},

### CODE REVIEW

File: mql/mql_test.go [hunk 1]
```diff
@@ -259,6 +259,11 @@ func TestResource_List_Builtins(t *testing.T) {
 			ResultIndex: 0,
 			Expectation: &llx.MockResource{Name: "mgroup", ID: "group1"},
 		},
+		{
+			Code:        "customGroups.first.name",
+			ResultIndex: 0,
+			Expectation: "group1",
+		},
```
Bajusz15, line 266: Without the fix, this would fail with the following error: 

```
--- FAIL: TestResource_List_Builtins (0.01s)
    --- FAIL: TestResource_List_Builtins/customGroups.first.name (0.00s)
        /Users/matebajusz/go/src/go.mondoo.com/cnquery/mql/testutils.go:107: failed to compile code: cannot find field 'name' in customGroups
```
Same problem that @tas50 noticed with `network.routes`

PR #6275, opened 2025-12-08
✨ Add tag filtering to AWS resources (es,elb,efs,redshift,rds,sagemaker,apigateway).

VasilSirakov:


providers/aws/resources/aws_apigateway.go
```diff
@@ -69,7 +69,11 @@ func (a *mqlAwsApigateway) getRestApis(conn *connection.AwsConnection) []*jobpoo
 				}
 
 				for _, restApi := range restApisResp.Items {
-					mqlRestApi, err := CreateResource(a.MqlRuntime, "aws.apigateway.restapi",
+					if conn.Filters.General.IsFilteredOutByTags(restApi.Tags) {
```
preslavgerchev, line 72: can we also add debug logs here?

VasilSirakov: yeah, forgot about that 😄

PR #6286, opened 2025-12-10
✨ Add CodeQL actions

czunker:


.github/workflows/codeql.yml
czunker, line 1: This is a simplified version of what is generated here:

https://github.com/mondoohq/cnquery/new/main?filename=.github%2Fworkflows%2Fcodeql.yml&workflow_template=code-scanning%2Fcodeql

I also replaced the tags with hashes.

.github/workflows/dependency-review.yml
czunker, line 1: We already have this action here: https://github.com/mondoohq/ranger-rpc/blob/main/.github/workflows/dependency-review.yml

It might also be useful in this repo.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      # Add any setup steps before running the `github/codeql-action/init` action.
+      # This includes steps like installing compilers or runtimes (`actions/setup-node`
+      # or others). This is typically only required for manual builds.
+      # - name: Setup runtime (example)
+      #   uses: actions/setup-example@v1
+
+      # Initializes the CodeQL tools for scanning.
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@cf1bb45a277cb3c205638b2cd5c984db1c46a412 # v4.31.7
```
philipbalinov, line 71: don't we need to explicitly call `github/codeql-action/analyze` as well?

czunker: Nice catch. I over-simplified the workflow.

I pushed an update.

philipbalinov (approved):

PR #6307, opened 2025-12-15
:star: Detect `euleros` as family `euler`

pratik-mondoo:


providers/os/resources/packages/packages.go
```diff
@@ -110,6 +110,7 @@ func ResolveSystemPkgManagers(conn shared.Connection) ([]OperatingSystemPkgManag
 	case asset.Platform.IsFamily("redhat") ||
 		asset.Platform.IsFamily("euler") ||
 		asset.Platform.Name == "hce" ||
+		asset.Platform.Name == "euleros" ||
```
tas50: We should make sure that `euleros` ends up int he `euler` family and then we can skip this

tas50: If we add euleros in https://github.com/mondoohq/cnquery/blob/959da289d977db7350378fc9df043e1c696521a0/providers/os/detector/detector_all.go and then it can get added to the family here https://github.com/mondoohq/cnquery/blob/959da289d977db7350378fc9df043e1c696521a0/providers/os/detector/detector_all.go#L1089

providers/os/detector/detector_all.go
```diff
@@ -817,6 +817,17 @@ var hce = &PlatformResolver{
 	},
 }
 
+var euleros = &PlatformResolver{
```
tas50, line 820: Can we get a test for this using the contents of the /etc/os-release file

pratik-mondoo: Added.

tas50 (approved):

PR #6310, opened 2025-12-16
🧹 Improve discovery filters parsing in the AWS provider.

VasilSirakov:


providers/aws/connection/filters.go
```diff
@@ -107,3 +137,58 @@ type EcsDiscoveryFilters struct {
...
+func parseBoolOpt(opts map[string]string, key string, defaultVal bool) bool {
+	for k, v := range opts {
+		if k == key {
+			parsed, err := strconv.ParseBool(v)
+			if err == nil {
+				return parsed
+			}
+		}
+	}
+	return defaultVal
```
czunker, line 154: Do we need a defaultValue as a separate parameter?

I only see `false` which would be the default anyway.

VasilSirakov: You are right, we only need `false` as a default but I exposed it just to make the fn more generic. I can remove it if you think it's not worth it.

czunker: I don't know much about the AWS part.
Let's keep it.

PR #6311, opened 2025-12-16
✨🐛 Fix panic in cyclonedx conversion. Improve decoder interface.

preslavgerchev:
* Fix a panic when there are no components under tools
* Improve the decoder interface to use `ReadSeeker`, implement multi-decoder


chris-rock (approved):
Thank you @preslavgerchev

PR #6317, opened 2025-12-17
✨ Add support for operating_system/container asset parsing in cyclonedx sboms

preslavgerchev:
This PR ensures that operating_system and container sboms get an improved platform/platform ids so that we can improve the asset detection


sbom/cyclonedx.go
```diff
@@ -168,30 +169,59 @@ func (ccx *CycloneDX) convertCycloneDxToSbom(bom *cyclonedx.BOM) (*Sbom, error)
...
+					Name:    component.Name,
+					Version: component.Version,
+					Vendor:  component.Author,
+				}
+			}
+		}
+
+		// if we have no generator info, fallback to trying tools. these are deprecated
+		// but might still be present
+		if sbom.Generator == nil && bom.Metadata.Tools.Tools != nil {
```
preslavgerchev, line 208: this is purely fallback in case the tools are present and we couldnt get a generator

PR #6318, opened 2025-12-18
🐛 Fix sshd sshd config check (globPathPattern used as file)

Bajusz15:
Issue: sshd config parser stores file patterns in Context.Path (as path), causing "file not found" errors, because the pattern is treated as a file.

When parsing sshd configuration files that use Include directives with glob patterns (e.g., Include /etc/ssh/sshd_config.d/*.conf), the parser was storing the literal glob pattern string in the `MatchBlock.Context.Path` field instead of the actual expanded file paths.
**Because of this:**
When sshd.config.blocks is accessed, matchBlocks2Resources() attempts to create file resources using `cur.Context.Path`, since `Context.Path` contained the literal glob pattern (e.g., /etc/ssh/sshd_config.d/*.conf) instead of the actual file path (e.g., /etc/ssh/sshd_config.d/01-mondoo.conf), it resulted in errors like:
  file '/etc/ssh/sshd_config.d/*.conf' not found

**Root Cause:**
ParseBlocks() was setting Context.Path to the rootPath parameter (which could be a glob pattern) without tracking which actual files were expanded
expandGlob() returned an error when directories didn't exist, instead of treating it as "no matches" (standard glob behavior)

**Example:**
SSHD config contains: Include /etc/ssh/sshd_config.d/*.conf
Directory exists with file: /etc/ssh/sshd_config.d/01-mondoo.conf
Parser creates MatchBlock with Context.Path = "/etc/ssh/sshd_config.d/*.conf" (literal glob)

**How to reproduce:**
Run the following script:
```
#!/bin/bash
# Test script to reproduce the exact customer scenario
# Customer has /etc/ssh/sshd_config.d/01-mondoo.conf with their config
# The error occurs when running the specific MQL queries

set -e

# Get the script directory
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"

# Set up logging
LOG_FILE="$SCRIPT_DIR/test-customer-scenario-$(date +%Y%m%d-%H%M%S).log"
exec > >(tee -a "$LOG_FILE") 2>&1

echo "=== Customer Scenario Test ==="
echo "Log file: $LOG_FILE"
echo ""

# Build cnquery for Linux
echo "=== Building cnquery for Linux ==="
CNQUERY_BIN="$SCRIPT_DIR/cnquery-linux"
if [ ! -f "$CNQUERY_BIN" ]; then
    echo "Building cnquery for Linux (amd64)..."
    cd "$SCRIPT_DIR"
    GOOS=linux GOARCH=amd64 go build -o "$CNQUERY_BIN" ./apps/cnquery
    if [ $? -ne 0 ]; then
        echo "ERROR: Failed to build cnquery"
        exit 1
    fi
    echo "✓ Built cnquery for Linux"
else
    echo "✓ Using existing Linux binary"
fi
echo ""

# Create test directory
TEST_DIR=$(mktemp -d)
echo "Test directory: $TEST_DIR"
cd "$TEST_DIR"

# Copy binary
cp "$CNQUERY_BIN" ./cnquery-linux

# Create Dockerfile with customer's exact scenario
cat > Dockerfile << 'EOF'
FROM ubuntu:22.04

# Install OpenSSH server
RUN apt-get update && \
    apt-get install -y openssh-server && \
    mkdir -p /var/run/sshd

# Create the directory and customer's exact config file
RUN mkdir -p /etc/ssh/sshd_config.d && \
    echo "KexAlgorithms sntrup761x25519-sha512@openssh.com,curve25519-sha256@libssh.org,diffie-hellman-group18-sha512" > /etc/ssh/sshd_config.d/01-mondoo.conf && \
    echo "MACs umac-128-etm@openssh.com,hmac-sha2-256-etm@openssh.com,hmac-sha2-512-etm@openssh.com,umac-128@openssh.com,hmac-sha2-256,hmac-sha2-512" >> /etc/ssh/sshd_config.d/01-mondoo.conf && \
    echo "HostbasedAuthentication no" >> /etc/ssh/sshd_config.d/01-mondoo.conf && \
    echo "PermitEmptyPasswords no" >> /etc/ssh/sshd_config.d/01-mondoo.conf && \
    echo "PermitUserEnvironment no" >> /etc/ssh/sshd_config.d/01-mondoo.conf && \
    echo "PermitRootLogin no" >> /etc/ssh/sshd_config.d/01-mondoo.conf && \
    echo "ClientAliveInterval 300" >> /etc/ssh/sshd_config.d/01-mondoo.conf && \
    echo "ClientAliveCountMax 0" >> /etc/ssh/sshd_config.d/01-mondoo.conf && \
    echo "LoginGraceTime 60" >> /etc/ssh/sshd_config.d/01-mondoo.conf && \
    echo "MaxAuthTries 4" >> /etc/ssh/sshd_config.d/01-mondoo.conf && \
    echo "IgnoreRhosts yes" >> /etc/ssh/sshd_config.d/01-mondoo.conf && \
    echo "LogLevel VERBOSE" >> /etc/ssh/sshd_config.d/01-mondoo.conf && \
    echo "X11Forwarding no" >> /etc/ssh/sshd_config.d/01-mondoo.conf && \
    echo "Banner /etc/issue.net" >> /etc/ssh/sshd_config.d/01-mondoo.conf && \
    echo "AllowGroups ssh_users" >> /etc/ssh/sshd_config.d/01-mondoo.conf && \
    echo "Match all" >> /etc/ssh/sshd_config.d/01-mondoo.conf && \
    echo "  ClientAliveCountMax 3" >> /etc/ssh/sshd_config.d/01-mondoo.conf && \
    chmod 644 /etc/ssh/sshd_config.d/01-mondoo.conf

# Ensure Include directive is in main config
RUN grep -q "Include /etc/ssh/sshd_config.d/\*\.conf" /etc/ssh/sshd_config || \
    echo "Include /etc/ssh/sshd_config.d/*.conf" >> /etc/ssh/sshd_config

# Copy cnquery binary
COPY cnquery-linux /usr/local/bin/cnquery
RUN chmod +x /usr/local/bin/cnquery

# Keep container running
CMD ["/bin/bash", "-c", "while true; do sleep 3600; done"]
EOF

echo "=== Building Docker image ==="
docker build -t test-customer-scenario . 2>&1 | tail -10

CONTAINER=$(docker run -d test-customer-scenario)
echo "Container ID: $CONTAINER"
echo ""

echo "=== Verifying setup ==="
docker exec "$CONTAINER" test -f /etc/ssh/sshd_config.d/01-mondoo.conf && echo "✓ Config file exists" || echo "✗ Config file missing"
docker exec "$CONTAINER" grep -q "Include /etc/ssh/sshd_config.d/\*\.conf" /etc/ssh/sshd_config && echo "✓ Include directive found" || echo "✗ Include directive missing"
echo ""

echo "=== Running customer's exact failing query ==="
echo "Query:"
echo "  defaultBlock = sshd.config.blocks.where(criteria == \"\");"
echo "  userBlocks = sshd.config.blocks.where(criteria != \"\");"
echo "  userBlocks.all(params.ClientAliveInterval >= 1);"
echo "  defaultBlock.all(params.ClientAliveInterval >= 1);"
echo ""

# Simplified query without props (since we don't have props defined)
QUERY="defaultBlock = sshd.config.blocks.where(criteria == \"\"); userBlocks = sshd.config.blocks.where(criteria != \"\"); userBlocks.all(params.ClientAliveInterval >= 1); defaultBlock.all(params.ClientAliveInterval >= 1); defaultBlock; userBlocks"

docker exec "$CONTAINER" cnquery run local -c "$QUERY" 2>&1 | tee /tmp/customer-query-output.log
	


```

PS. Since then, it was discovered for docker connection some file system methods were not supported, so I added implementations.

### DIFF
--- a/providers/os/resources/sshd/params.go
+++ b/providers/os/resources/sshd/params.go
@@ -85,23 +85,32 @@ func (m MatchBlocks) Flatten() map[string]any {
 
 func mergeIncludedBlocks(matchConditions map[string]*MatchBlock, blocks MatchBlocks, curBlock string) {
 	for _, block := range blocks {
-		// meaning:
-		// 1. curBlock == "", we can always add all subblocks
-		// 2. if block == "", we can add it to whatever current block is
-		// 3. in all other cases the block criteria must match, or we move on
-		if block.Criteria != curBlock && curBlock != "" && block.Criteria != "" {
+		if block.Criteria == "" {
+			// Default block: merge into the current block
+			existing := matchConditions[curBlock]
+			if existing == nil {
+				existing = &MatchBlock{
+					Criteria: curBlock,
+					Params:   map[string]any{},
+					Context:  block.Context,
+				}
+				matchConditions[curBlock] = existing
+			}
+			if existing.Params == nil {
+				existing.Params = map[string]any{}
+			}
+			for k, v := range block.Params {
+				if _, ok := existing.Params[k]; !ok {
+					existing.Params[k] = v
+				}
+			}
 			continue
 		}
 
+		// Match block: always add to global map
 		existing, ok := matchConditions[block.Criteria]
-		if block.Criteria == "" {
-			existing = matchConditions[curBlock]
-		} else if !ok {
-			existing = block
-			matchConditions[block.Criteria] = existing
-		}
-
-		if existing == nil {
+		if !ok {
+			// Create a new Match block
 			existing = &MatchBlock{
 				Criteria: block.Criteria,
 				Params:   map[string]any{},
--- a/providers/os/resources/sshd/params_test.go
+++ b/providers/os/resources/sshd/params_test.go
@@ -145,3 +145,67 @@ func TestSSHParseWithGlob(t *testing.T) {
 	assert.Equal(t, "no", adminBlock.Params["PasswordAuthentication"])
 	assert.Contains(t, adminBlock.Context.Path, "02_security.conf")
 }
+
+func TestSSHParseIncludeInsideMatchBlock(t *testing.T) {
+	// Test edge case: Include directive inside a Match block
+	// The included file has a different Match block - both should be present
+	// This verifies that Match blocks from included files are always added
+	// to the global map, regardless of where the Include directive appears.
+
+	fileContent := func(path string) (string, error) {
+		content, err := os.ReadFile(path)
+		if err != nil {
+			return "", err
+		}
+		return string(content), nil
+	}
+
+	globExpand := func(glob string) ([]string, error) {
+		// Handle the nested include file
+		if glob == "match_include_nested.conf" {
+			return []string{"./testdata/match_include_nested.conf"}, nil
+		}
+		// For other paths, check if file exists
+		if _, err := os.Stat(glob); err == nil {
+			return []string{glob}, nil
+		}
+		return []string{glob}, nil
+	}
+
+	mainConfigPath := "./testdata/match_include_main.conf"
+	blocks, err := ParseBlocksWithGlob(mainConfigPath, fileContent, globExpand)
+	require.NoError(t, err)
+	assert.NotNil(t, blocks)
+
+	// Find both Match blocks
+	var specialBlock *MatchBlock
+	var adminBlock *MatchBlock
+	var defaultBlock *MatchBlock
+
+	for _, block := range blocks {
+		switch block.Criteria {
+		case "Group special":
+			specialBlock = block
+		case "User admin":
+			adminBlock = block
+		case "":
+			defaultBlock = block
+		}
+	}
+
+	// Verify default block exists
+	require.NotNil(t, defaultBlock, "should have default block")
+	assert.Equal(t, "22", defaultBlock.Params["Port"])
+
+	// Verify the Match block containing the Include exists
+	require.NotNil(t, specialBlock, "should have 'Group special' match block")
+	assert.Equal(t, "30", specialBlock.Params["ClientAliveInterval"])
+
+	// Verify the Match block from the included file exists (this is the edge case)
+	// Before the fix, this block would be filtered out because Include was inside
+	// a Match block with different criteria.
+	require.NotNil(t, adminBlock, "should have 'User admin' match block from included file")
+	assert.Equal(t, "yes", adminBlock.Params["PermitRootLogin"])
+	assert.Equal(t, "no", adminBlock.Params["PasswordAuthentication"])
+	assert.Contains(t, adminBlock.Context.Path, "match_include_nested.conf")
+}
--- a/providers/os/resources/sshd/testdata/match_include_main.conf
+++ b/providers/os/resources/sshd/testdata/match_include_main.conf
@@ -1,0 +1,6 @@
+# Main config with Match block containing Include
+Port 22
+Match Group special
+    ClientAliveInterval 30
+    Include match_include_nested.conf
+
--- a/providers/os/resources/sshd/testdata/match_include_nested.conf
+++ b/providers/os/resources/sshd/testdata/match_include_nested.conf
@@ -1,0 +1,6 @@
+# Included file with different Match block
+# This Match block should still be added even though Include is inside another Match block
+Match User admin
+    PermitRootLogin yes
+    PasswordAuthentication no
+


providers/os/resources/sshd/params.go
```diff
@@ -112,17 +112,25 @@ func mergeIncludedBlocks(matchConditions map[string]*MatchBlock, blocks MatchBlo
...
+func ParseBlocks(rootPath string, globPathContent globPathContentFunc) (MatchBlocks, error) {
+	content, actualPaths, err := globPathContent(rootPath)
 	if err != nil {
 		return nil, err
 	}
 
+	// This ensures that glob patterns use the actual file path instead of the glob pattern
+	actualPath := rootPath
+	if len(actualPaths) > 0 {
+		actualPath = actualPaths[0]
```
czunker: What about the other matches?

Taking the example from your PR description, let's assume the glob `*.conf` would match `01_mondoo.conf` and `02_mondoo.conf`.

From reading the code, I think we would not detect the 2nd file.

What do you think about expanding the glob outside this function, and calling the function for every file found?

providers/os/resources/sshd/params_test.go
```diff
@@ -15,8 +15,8 @@ func TestSSHParser(t *testing.T) {
 	raw, err := os.ReadFile("./testdata/sshd_config")
```
czunker, line 15: Would it be possible to expand the testdata and split the single file up into at least two includes in subdirectories?

This way, we could add the glob case to the test cases.

Bajusz15: @czunker I changed the approach now, and added test for multiple includes. I also verified using the script to reproduce. When I rebuild cnquery with my changes, the problem is gone.

czunker:
I tried it with a container, but couldn't parse the config:
```
cnquery run docker 14df8ca5f2d0 -c 'sshd.config.blocks.where(criteria != "")'
! unable to expand Include directive error="not implemented" path=/etc/ssh/sshd_config.d/*.conf
sshd.config.blocks.where: []
```

The file layout looks like this:
```
./sshd_config.d
./sshd_config.d/10-mondoo.conf
./sshd_config.d/50-redhat.conf
./sshd_config.d/20-mondoo.conf
./sshd_config.d/40-redhat-crypto-policies.conf
./sshd_config
```
The sshd config has an `Include`:
```
Include /etc/ssh/sshd_config.d/*.conf
```

czunker:
I checked the config. One of the included files has another `Include` directive:
```
[root@14df8ca5f2d0 ssh]# grep Include * -R
sshd_config:Include /etc/ssh/sshd_config.d/*.conf
sshd_config.d/40-redhat-crypto-policies.conf:Include /etc/crypto-policies/back-ends/opensshserver.config
```

I commented that one out, but no change.

Bajusz15:
> cnquery run docker 14df8ca5f2d0 -c 'sshd.config.blocks.where(criteria != "")'

I think the unimplemented error here is because you are running it via docker connection, not locally, and the method is not implemented. But I will make a fix for this too as it's a valid use case.

providers/os/connection/ssh/cat/cat_file.go
```diff
@@ -95,7 +96,13 @@ func (f *File) Readdir(count int) (res []os.FileInfo, err error) {
...
 	res = make([]os.FileInfo, len(names))
 	for i, name := range names {
-		res[i], err = f.catfs.Stat(name)
+		var statPath string
+		if filepath.IsAbs(name) {
+			statPath = name
+		} else {
+			statPath = filepath.Join(f.path, name)
+		}
+		res[i], err = f.catfs.Stat(statPath)
```
Bajusz15, line 105: This function was not used until now, but I made the change so it's backward compatible.

czunker:
> @czunker I got same not implemented error when checking with docker connection. I added implementation to the missing method. Can you try again with `cnquery run docker 14df8ca5f2d0 -c 'sshd.config.blocks.where(criteria != "")'`?

@Bajusz15 Thanks for adding this to the container connection.

The error is gone, but the blocks are empty:
```
cnquery run docker 67f1fc72c9fd -c 'sshd.config.blocks.where(criteria != "")'
sshd.config.blocks.where: []
```
The default block works:
```
cnquery run docker 67f1fc72c9fd -c 'sshd.config.blocks.where(criteria == "")'
sshd.config.blocks.where: [
  0: sshd.config.matchBlock criteria=""
     in /etc/ssh/sshd_config:1-132
...
```

Anything else I can try?

PR #6320, opened 2025-12-18
🐛 Fix sbom panic

Bajusz15:
Issue: we try to access nil values in `convertCycloneDxToSbom` function


Bajusz15:
Fixed in another PR

PR #6330, opened 2025-12-22
✨ Add platform data to SBOMs

czunker:
In case an SBOM is not related to an OS, use the SBOM as platform.

### DIFF
--- a/sbom/cyclonedx.go
+++ b/sbom/cyclonedx.go
@@ -170,16 +170,34 @@ func (ccx *CycloneDX) convertCycloneDxToSbom(bom *cyclonedx.BOM) (*Sbom, error) 
 	}
 
 	rootComponent := bom.Metadata.Component
+	title := rootComponent.Description
+	version := rootComponent.Version
+	if title == "" {
+		title = "CycloneDX"
+	}
+	if version == "" {
+		version = bom.SpecVersion.String()
+	}
 	sbom := &Sbom{
 		Asset: &Asset{
 			Name: rootComponent.Name,
 			Platform: &Platform{
 				Name:    "cyclonedx",
-				Version: bom.SpecVersion.String(),
-				Title:   "CycloneDX",
+				Version: version,
+				Title:   title,
 			},
 		},
 		Packages: make([]*Package, 0),
+	}
+
+	switch rootComponent.Type {
+	case cyclonedx.ComponentTypeOS:
+		hostnameId := "//platformid.api.mondoo.app/hostname/" + rootComponent.Name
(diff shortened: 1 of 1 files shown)


sbom/spdx_test.go
```diff
@@ -71,3 +71,17 @@ func TestSpdxJsonDecoder(t *testing.T) {
 	require.NoError(t, err)
 	assert.NotNil(t, sbomReport)
 }
+
+func TestSpdxJsonDecoder_GitHub_DependecyGraph(t *testing.T) {
+	f, err := os.Open("testdata/vercel_next.js_937412.json")
```
chris-rock, line 79: lets reduce the test sample to the minimal set that needs to be tested

sbom/spdx.go
```diff
@@ -191,12 +191,17 @@ func (s *Spdx) convertToSbom(doc *spdx.Document) *Sbom {
 		},
 		Asset: &Asset{
 			Name: doc.DocumentName,
+			Platform: &Platform{
+				Name:    "spdx",
+				Version: doc.SPDXVersion,
+				Title:   "SPDX",
```
chris-rock, line 197: should we have the same behavior as for cyclonedx?

czunker: When I understand correctly what you are asking for, we already have this:
https://github.com/mondoohq/cnquery/pull/6330/changes#diff-6672e9266c0b7c5144539ef3a2c78e1c22d668bb2a997a5a887d3062d5d10445R234

In case we find an OS asset type, the platform data is replaced with the OS information.

chris-rock: looks like the cyclonedx implementation is not in line with spdx https://github.com/mondoohq/cnquery/pull/6330/files#diff-119d34da701c1094b5a2488c64ec950a418b885ac815938bf0e3fbd49c8b5a8bR173-R183

sbom/cyclonedx.go
```diff
@@ -256,6 +247,15 @@ func (ccx *CycloneDX) convertCycloneDxToSbom(bom *cyclonedx.BOM) (*Sbom, error)
 				sbom.Asset.Platform.Cpes = []string{component.CPE}
 			}
 			sbom.Packages = append(sbom.Packages, pkg)
+			switch rootComponent.Type {
```
chris-rock: i am not sure why we should use the root component here. We cannot assume that cyclone dx include a os type component

chris-rock: Then it makes sense to keep the previous implementation as is.

czunker: Reverted

chris-rock (approved):
Thank you @czunker

PR #6350, opened 2026-01-05
⭐️ shell rewrite

chris-rock:
This is a rewrite of the shell for the modern world:

- search of history with ctrl+r
- multi-line paste
- simple syntax highlighting for MQL queries
- shows spinner for slower queries
- filter resources by connected provider
- shows asset information
- Removes github.com/c-bata/go-prompt

**macOS**

Screenshot 2026-01-05 at 20 34 03

**Windows Server 2025**

shell_windows_2
shell_windows

**Windows 2019**

Screenshot 2026-01-05 at 23 16 00

TODO:
- [x] Test in CMD
- [x] Test in Powershell

### DIFF
--- a/cli/shell/model.go
+++ b/cli/shell/model.go
@@ -372,6 +372,13 @@ func (m *shellModel) handleKeyMsg(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		// Show asset information
 		return m, m.showAssetInfo()
 
+	case "?":
+		// Show keybindings help (only when input is empty to avoid interfering with queries)
+		if m.input.Value() == "" {
+			helpText := m.theme.SecondaryText("Keyboard Shortcuts:") + m.keyMap.FormatFullHelp()
+			return m, tea.Println(helpText)
+		}
+
 	case "ctrl+r":
 		// Enter history search mode
 		if len(m.history) > 0 {
@@ -836,8 +843,8 @@ func (m *shellModel) View() string {
 		}
 	}
 
-	// Help bar at the bottom
-	b.WriteString("\n")
+	// Help bar at the bottom (with empty line separator)
+	b.WriteString("\n\n")
 	b.WriteString(m.renderHelpBar())
 
 	return b.String()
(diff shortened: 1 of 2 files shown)


cli/shell/model.go
```diff
@@ -0,0 +1,1043 @@
...
+		}
+
+		return historyLoadedMsg{history: filtered}
+	}
+}
+
+// saveHistory saves command history to disk
+func (m *shellModel) saveHistory() {
+	if m.historyPath == "" {
+		homeDir, _ := homedir.Dir()
```
syrull, line 209: If this fails, I think we would have `.mondoo_history` in some relative location of the execution path, also the err is handled in other parts of the code.

cli/shell/model.go
```diff
@@ -0,0 +1,1043 @@
...
+		tea.Println(m.theme.Welcome),
+	)
+}
+
+// loadHistory loads command history from disk
+func (m *shellModel) loadHistory() tea.Cmd {
+	return func() tea.Msg {
+		homeDir, err := homedir.Dir()
+		if err != nil {
+			return historyLoadedMsg{history: []string{}}
```
syrull, line 184: Maybe log a warning that it cannot load the history? I am not sure how this is usually handled.

cli/shell/completer.go
```diff
@@ -4,65 +4,77 @@
...
 
+	// Get MQL suggestions
 	var query string
 	if c.queryPrefix != nil {
 		query = c.queryPrefix()
 	}
-	query += doc.TextBeforeCursor()
+	query += text
 
 	bundle, _ := mqlc.Compile(query, nil, mqlc.NewConfig(c.schema, c.features))
```
syrull, line 68: nit: Let's not disregard the err and log it in a Debug statement instead, I think it would be useful for whenever we lose some autocomplete due whatever reason with the Compile

cli/shell/model.go
chris-rock: history is only supported in the home directory and I do not think we should change that. We handle the error for critical things like configuration where things fail, but history is nice to have. So i think that is ok.

cli/shell/completer.go
chris-rock: not needed for the suggestions as the compile error is printed seperately

cli/shell/model.go
chris-rock: We can add a log warning.

cli/shell/model.go
```diff
@@ -0,0 +1,1076 @@
...
+			m.searchMode = true
+			m.searchQuery = ""
+			m.searchMatches = nil
+			m.searchIdx = 0
+			// Save current input as draft
+			m.historyDraft = m.input.Value()
+		}
+		return m, nil
+
+	case "ctrl+j":
```
jaym, line 394: thoughts on making this shift+return?

chris-rock: I tried shift+return but some terminals just return "return" so i was not able to catch the difference. Probably I missed something.

PR #6353, opened 2026-01-06
⭐ AWS: EC2 AMI images collection

florinutz:
## Summary

We need to add a collection for AWS EC2 AMI images in the AWS provider. Currently `aws.ec2.image` exists as a resource type but there's no `aws.ec2.images` collection to iterate over all AMIs in an account. This will allow querying AMI encryption status, age, and other image attributes.

## AWS API Reference

The data is available via the AWS EC2 API:
- [DescribeImages](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeImages.html)

```bash
# List all AMIs owned by the account
aws ec2 describe-images --owners self

# List AMIs with specific filters
aws ec2 describe-images --owners self --filters "Name=state,Values=available"
```

Response includes:
```json
{
  "Images": [
    {
      "ImageId": "ami-12345678",
      "Name": "my-custom-ami",
      "State": "available",
      "CreationDate": "2024-01-15T10:30:00.000Z",
      "BlockDeviceMappings": [
        {
          "DeviceName": "/dev/xvda",
          "Ebs": {
            "Encrypted": true,
            "VolumeSize": 8,
            "VolumeType": "gp3",
            "SnapshotId": "snap-12345678"
          }
        }
      ],
      "RootDeviceType": "ebs",
      "VirtualizationType": "hvm",
      "Architecture": "x86_64",
      "Public": false,
      "OwnerId": "123456789012",
      "Tags": [
        {"Key": "Environment", "Value": "Production"}
      ]
    }
  ]
}
```

## Suggested MQL

Extend the existing `aws.ec2` resource:

```
aws.ec2 {
  images []aws.ec2.image
}
```

Extend `aws.ec2.image` with additional fields:

```
aws.ec2.image {
  // Existing fields...
  id string
  name string
  arn string
  architecture string
  createdAt time
  
  // New fields needed
  state string
  encrypted bool  // derived: true if all block devices are encrypted
  public bool
  rootDeviceType string
  virtualizationType string
  blockDeviceMappings []dict
  tags map[string]string
}
```

### Expected Usage

```mql
// Check all AMIs have encrypted block devices
aws.ec2.images.all(encrypted == true)

// Alternative: Check all block device mappings are encrypted
aws.ec2.images.all(
  blockDeviceMappings.where(ebs != null).all(ebs.Encrypted == true)
)

// Check AMIs are not older than 90 days
aws.ec2.images.all(time.now - createdAt < 90 * time.day)

// Check no AMIs are public
aws.ec2.images.all(public == false)
```

## Note

The existing `aws.ec2.image` resource is accessed via `aws.ec2.instance.image` for individual instances. This request adds a collection to list all AMIs in the account for security auditing purposes.

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -3454,8 +3454,6 @@ private aws.ec2.image @defaults("ownerAlias id name") {
 
 // Amazon EC2 image block device mapping
 private aws.ec2.image.blockDeviceMapping @defaults("deviceName") {
-  // Internal ID for this block device mapping
-  id string
   // Device name (e.g., /dev/sda1, /dev/xvda)
   deviceName string
   // Virtual device name for instance store (ephemeral0, ephemeral1, etc.)
--- a/providers/aws/resources/aws_ec2.go
+++ b/providers/aws/resources/aws_ec2.go
@@ -687,7 +687,7 @@ func createBlockDeviceMappings(runtime *plugin.Runtime, imageArn string, mapping
 		mappingID := fmt.Sprintf("%s/device/%s", imageArn, deviceName)
 
 		args := map[string]*llx.RawData{
-			"id":          llx.StringData(mappingID),
+			"__id":        llx.StringData(mappingID),
 			"deviceName":  llx.StringDataPtr(mapping.DeviceName),
 			"virtualName": llx.StringDataPtr(mapping.VirtualName),
 			"noDevice":    llx.BoolData(mapping.NoDevice != nil && *mapping.NoDevice != ""),
@@ -1222,14 +1222,6 @@ func (i *mqlAwsEc2Instance) keypair() (*mqlAwsEc2Keypair, error) {
 
 func (i *mqlAwsEc2Image) id() (string, error) {
 	return i.Arn.Data, nil
-}
-
-func (b *mqlAwsEc2ImageBlockDeviceMapping) id() (string, error) {
-	return b.Id.Data, nil
-}
-
-func (e *mqlAwsEc2ImageEbsBlockDevice) id() (string, error) {
-	return e.__id, nil
 }
 
 func initAwsEc2Image(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error) {
--- a/providers/aws/resources/aws.lr.go
+++ b/providers/aws/resources/aws.lr.go
@@ -5171,9 +5171,6 @@ var getDataFields = map[string]func(r plugin.Resource) *plugin.DataRes{
 	"aws.ec2.image.region": func(r plugin.Resource) *plugin.DataRes {
 		return (r.(*mqlAwsEc2Image).GetRegion()).ToDataRes(types.String)
 	},
-	"aws.ec2.image.blockDeviceMapping.id": func(r plugin.Resource) *plugin.DataRes {
-		return (r.(*mqlAwsEc2ImageBlockDeviceMapping).GetId()).ToDataRes(types.String)
-	},
 	"aws.ec2.image.blockDeviceMapping.deviceName": func(r plugin.Resource) *plugin.DataRes {
 		return (r.(*mqlAwsEc2ImageBlockDeviceMapping).GetDeviceName()).ToDataRes(types.String)
 	},
@@ -11988,10 +11985,6 @@ var setDataFields = map[string]func(r plugin.Resource, v *llx.RawData) bool{
 	},
 	"aws.ec2.image.blockDeviceMapping.__id": func(r plugin.Resource, v *llx.RawData) (ok bool) {
 		r.(*mqlAwsEc2ImageBlockDeviceMapping).__id, ok = v.Value.(string)
-		return
-	},
-	"aws.ec2.image.blockDeviceMapping.id": func(r plugin.Resource, v *llx.RawData) (ok bool) {
-		r.(*mqlAwsEc2ImageBlockDeviceMapping).Id, ok = plugin.RawToTValue[string](v.Value, v.Error)
 		return
 	},
 	"aws.ec2.image.blockDeviceMapping.deviceName": func(r plugin.Resource, v *llx.RawData) (ok bool) {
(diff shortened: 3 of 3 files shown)


providers/aws/resources/aws_ec2.go
```diff
@@ -661,6 +661,119 @@ func initAwsEc2Keypair(runtime *plugin.Runtime, args map[string]*llx.RawData) (m
...
+	for _, region := range regions {
+		f := func() (jobpool.JobResult, error) {
+			log.Debug().Str("region", region).Msgf("ec2>getImagesJob>calling aws with region")
+
+			svc := conn.Ec2(region)
+			ctx := context.Background()
+			var res []any
+
+			// Only fetch images owned by this account
+			images, err := svc.DescribeImages(ctx, &ec2.DescribeImagesInput{Owners: []string{"self"}})
```
syrull: This is unpaginated call, https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeImages.html 

according to the docs we need to use pagination calls.

Something like this:

```go
  params := &ec2.DescribeImagesInput{
      Owners:  []string{"self"}
  }
  paginator := ec2.NewDescribeImagesPaginator(svc, params)
  for paginator.HasMorePages() {
      images, err := paginator.NextPage(ctx)
      if err != nil {
          if Is400AccessDeniedError(err) {
              log.Warn().Str("region", region).Msg("error accessing region for AWS API")
              return res, nil
          }
          return nil, err
      }

      for _, image := range images.Images {
          // ... process images
      }
  }
```

florinutz: done, good catch

providers/aws/resources/aws_ec2.go
```diff
@@ -1062,6 +1181,25 @@ func (i *mqlAwsEc2Image) id() (string, error) {
...
+	for _, mapping := range blockDeviceMappings {
+		if mapping.Ebs != nil {
+			hasEbsDevices = true
+			if mapping.Ebs.Encrypted == nil || !*mapping.Ebs.Encrypted {
+				return false
+			}
+		}
+	}
+
+	return hasEbsDevices
```
preslavgerchev: nit: if all the block devices must be encrypted, you can just return true here and remove `hasEbsDevices` since if there's a single non-encrypted block you return false (line 1195)

providers/aws/resources/aws.lr
```diff
@@ -3434,6 +3436,22 @@ private aws.ec2.image @defaults("ownerAlias id name") {
...
+  // Current state of the image (available, pending, failed, etc.)
+  state string
+  // Whether the image is public
+  public bool
+  // Root device type (ebs or instance-store)
+  rootDeviceType string
+  // Virtualization type (hvm or paravirtual)
+  virtualizationType string
+  // Block device mappings for the image
+  blockDeviceMappings []dict
```
preslavgerchev: can we make this a proper mql resource instead of dict? then we can also drop `encrypted` since we'd be able to write a check, asking if all block devices in the image are encrypted

syrull: Agree

Context: We kinda have this problem with the microsoft resources where it is extremely hard to find anything because we used dicts for a lot of the stuff and now whenever we have a nested properties that we need in a query, we don't know if we have them or not.

I think the general way to handle this would be:

- If the properties are k-v where the key is some user defined variable -> use dict
- If there are some specific AWS properties -> better create explicit resources for them

This eases the policy work a lot.

tas50: Dict values make the docs pretty terrible too

providers/aws/resources/aws_ec2.go
```diff
@@ -748,7 +792,7 @@ func (a *mqlAwsEc2) getImagesJob(conn *connection.AwsConnection) []*jobpool.Job
 					}
 					mqlImage, err := CreateResource(a.MqlRuntime, ResourceAwsEc2Image,
 						map[string]*llx.RawData{
-							"arn":                 llx.StringData(fmt.Sprintf(imageArnPattern, region, conn.AccountId(), convert.ToValue(image.ImageId))),
+							"arn":                 llx.StringData(imageArn),
```
imilchev, line 795: does the arn already encode the region?

florinutz: yup, line 765

providers/aws/resources/aws.lr
```diff
@@ -3445,15 +3445,47 @@ private aws.ec2.image @defaults("ownerAlias id name") {
...
   // Tags associated with the image
   tags map[string]string
   // Region where the image is located
   region string
 }
 
+// Amazon EC2 image block device mapping
+private aws.ec2.image.blockDeviceMapping @defaults("deviceName") {
+  // Internal ID for this block device mapping
+  id string
```
preslavgerchev: it is not required to expose this, since we only want this for uniqueness during the runtime, this will not server other purpose

in such cases, it is enough to add this to the list of args:
```
args["__id"] = your-id
```
the `__id` is a special keyword that's used to to directly create ids for mql resources

providers/aws/resources/aws_ec2.go
```diff
@@ -1181,23 +1224,12 @@ func (i *mqlAwsEc2Image) id() (string, error) {
...
-				return false
-			}
-		}
-	}
+func (b *mqlAwsEc2ImageBlockDeviceMapping) id() (string, error) {
+	return b.Id.Data, nil
+}
 
-	return hasEbsDevices
+func (e *mqlAwsEc2ImageEbsBlockDevice) id() (string, error) {
```
preslavgerchev: nit: if you want to use the underlying `__id` you do not need to define `id()` here, the behaviour will be the same

providers/aws/resources/aws_ec2.go
preslavgerchev: nit: i dont think ami images actually have an ARN. this might be confusing, we should look into deprecating this if possible. (does not have to be in this pr)

florinutz: you mean it might be counterintuitive for users to see arns for amis because they're not used often, right?
should I remove the arn now or later in another pr, wdyt? or maybe a better question: how do we deprecate? :D

preslavgerchev: i dont think AMIs have ARNs at all, only AMI ids. Some AWS resources dont have ARNs

I would not change anything in this PR, I was only making a note that we need to take a proper look at it. Deprecation wise: we can just leave a comment above the field and remove it in the second next major version (so that would be v14)

preslavgerchev: @florinutz I will open an issue to track this as a follow-up

PR #6367, opened 2026-01-08
⭐️ support files.find on windows with powershell

chris-rock:


providers/os/resources/filesfind/powershell_find.go
```diff
@@ -0,0 +1,103 @@
...
+	"strconv"
+	"strings"
+)
+
+func BuildPowershellCmd(from string, xdev bool, fileType string, regex string, permission int64, search string, depth *int64) string {
+	var script strings.Builder
+
+	// Use here-strings to safely embed all user input
+	script.WriteString("$Path = @'\n")
+	script.WriteString(from)
```
jaym, line 20: Do we still need to do some escaping here? For example, if someone figures out how to get from = `\n'@\nrm -rf /`

chris-rock: Good idea, lets just strip any \n and \r as those should not be in the file paths anyway.

PR #6368, opened 2026-01-08
⭐ extend secretmanager with KmsKey field

LittleSalkin1806:
What was done in this MR:

- [ - ] I added "KmsKey" field to secretmanager secrets .

- [ - ] Additionally i changed the KMS Key default values to retrieve arn and aliases.

I choosed arn + aliases against id, region , description because of the following reasons.

Aliases:
AWS recommends using aliases for customer-managed keys ([AWS KMS Best Practices](https://docs.aws.amazon.com/kms/latest/developerguide/kms-alias.html))
Aliases are the primary way to reference keys in a human-readable manner
Can be used directly in API calls, eliminating the need to remember key IDs
Support for cross-account references when properly configured
Immutable per region after creation but can be reassigned to different keys

Arn:
Already includes ID & region
often has to be build manually
often needed by api as input -> therefore we dont need region again

Removed Description:
Free-form text with no validation or uniqueness requirements
Optional metadata field
Often left empty or filled with generic boilerplate text
Can have useful information but can be queried at any point, alias is shorter and often used in an expressive manner.

Question:
Does it break anything to change default of KmsKey?

Any tests or things i overlooked after first look into this repo?

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -716,7 +716,7 @@ aws.kms @defaults("keys") {
 }
 
 // AWS Key Management Service (KMS) key
-private aws.kms.key @defaults("arn aliases") {
+private aws.kms.key @defaults("id region aliases metadata.Description") {
   // Unique identifier for the key
   id string
   // ARN of the key


providers/aws/resources/aws.lr
```diff
@@ -716,7 +716,7 @@ aws.kms @defaults("keys") {
 }
 
 // AWS Key Management Service (KMS) key
-private aws.kms.key @defaults("id region metadata.Description") {
+private aws.kms.key @defaults("arn aliases") {
```
tas50: In general we try to keep the ARN out of the defaults since it's a bit cryptic

LittleSalkin1806: Telekom is currently checking CLA before i will sign it (by our Open Source Program Office) 
Sure i will stick to mondoos decisions
Is then id, region, aliases fine ?

Any tests or things i overlooked after first look into this repo?

tas50: Any reason not to do `id region aliases metadata.Description` since the defaults are really meant to present the useful complete output you'd want in cnquery shell or a query pack?

As for the tests I'll leave that part up to the engineers on the team. I'll make sure this gets a proper review.

LittleSalkin1806: if the defaults is for the user and its not to much information thats given back then sure.

The reason why i think description is not the best i have explained above but if its properly used its very explanatory for the user.

for me its just important to add aliases as its very descriptive and often the only human readable part to know what the key is used for other than the description.

LittleSalkin1806:

PR #6370, opened 2026-01-08
🐛 AWS: ECS cluster ARN construction bug

### LINKED ISSUE #6266: AWS: Fix ECS cluster ARN construction bug
## Summary

There is a bug in the AWS provider when accessing ECS cluster data. The provider incorrectly constructs the cluster ARN, causing API calls to fail with an ARN syntax exception.

## Error Observed

```
rpc error: code = Unknown desc = operation error ECS: DescribeClusters, https response error StatusCode: 400, 
RequestID: a51e6ddf-80de-47ab-9321-ce4a83cbf356, ClientException: com.amazon.arn.ARNSyntaxException: 
Fifth colon (namespace/relative-id delimiter) not found: arn:aws:sts::162854405951
aws.ecs.cluster: no data available
```

## Expected Behavior

The provider should construct a proper ECS cluster ARN in the format:
```
arn:aws:ecs:us-east-1:123456789012:cluster/my-cluster
```

## Actual Behavior

The provider appears to be using an STS ARN format instead:
```
arn:aws:sts::162854405951
```

This malformed ARN causes the `DescribeClusters` API call to fail.

## Steps to Reproduce

1. Configure AWS credentials for an account with ECS clusters
2. Run cnquery/cnspec against AWS
3. Attempt to query ECS cluster data:
```bash
cnquery run aws -c "aws.ecs.clusters"
```

## Environment

- cnquery/cnspec version: (current)
- AWS Region: us-east-1
- ECS clusters exist in the account

## Additional Context

This bug blocks the ability to:
- Query ECS cluster settings (including Container Insights configuration)
- Access cluster tags
- List container instances in clusters

## AWS API Reference

The correct API call should be:
- [DescribeClusters](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeClusters.html)

```bash
# Working CLI example
aws ecs describe-clusters --clusters my-cluster --include SETTINGS
```

florinutz:
## Summary

There is a bug in the AWS provider when accessing ECS cluster data. The provider incorrectly constructs the cluster ARN, causing API calls to fail with an ARN syntax exception.

## Error Observed

```
rpc error: code = Unknown desc = operation error ECS: DescribeClusters, https response error StatusCode: 400, 
RequestID: a51e6ddf-80de-47ab-9321-ce4a83cbf356, ClientException: com.amazon.arn.ARNSyntaxException: 
Fifth colon (namespace/relative-id delimiter) not found: arn:aws:sts::162854405951
aws.ecs.cluster: no data available
```

## Expected Behavior

The provider should construct a proper ECS cluster ARN in the format:
```
arn:aws:ecs:us-east-1:123456789012:cluster/my-cluster
```

## Actual Behavior

The provider appears to be using an STS ARN format instead:
```
arn:aws:sts::162854405951
```

This malformed ARN causes the `DescribeClusters` API call to fail.

## Steps to Reproduce

1. Configure AWS credentials for an account with ECS clusters
2. Run cnquery/cnspec against AWS
3. Attempt to query ECS cluster data:
```bash
cnquery run aws -c "aws.ecs.clusters"
```

## Environment

- cnquery/cnspec version: (current)
- AWS Region: us-east-1
- ECS clusters exist in the account

## Additional Context

This bug blocks the ability to:
- Query ECS cluster settings (including Container Insights configuration)
- Access cluster tags
- List container instances in clusters

## AWS API Reference

The correct API call should be:
- [DescribeClusters](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeClusters.html)

```bash
# Working CLI example
aws ecs describe-clusters --clusters my-cluster --include SETTINGS
```

---

fixes #6266 

---

This was also affecting `task`, not just `cluster`.

### DIFF
--- a/providers/aws/resources/aws.go
+++ b/providers/aws/resources/aws.go
@@ -216,10 +216,6 @@ func getAssetIdentifier(runtime *plugin.Runtime) *assetIdentifier {
 			}
 		}
 	}
-	// Only return an assetIdentifier if we have a valid ARN
-	if arnStr == "" {
-		return nil
-	}
 
 	return &assetIdentifier{name: a.Name, arn: arnStr}
 }


providers/aws/resources/aws_ecs.go
```diff
@@ -166,12 +166,17 @@ func initAwsEcsCluster(runtime *plugin.Runtime, args map[string]*llx.RawData) (m
 	a := args["arn"].Value.(string)
 	conn := runtime.Connection.(*connection.AwsConnection)
 
+	// Validate and parse ARN if provided
+	parsedARN, err := validateAndParseARN(a, "ecs")
```
VasilSirakov, line 170: the old code seems a bit strange. if it does not parse the ARN successfully, it just proceeds into the `DescribeClusters` and `DescribeTasks` calls.

I think it would be better if we simply short-circuit and return a descriptive error when the ARN is for whatever reason invalid. that would also simplify the code below as `parsedARN` can never be nil.

providers/aws/resources/aws_ecs.go
```diff
@@ -487,3 +494,25 @@ func ecsTagsToMap(tags []ecstypes.Tag) map[string]any {
 	}
 	return res
 }
+
+// validateAndParseARN validates that the given string is a valid ECS ARN
+// and returns the parsed ARN structure. Returns an error if the ARN is malformed
+// or does not belong to the expectedService.
+func validateAndParseARN(arnStr, expectedService string) (*arn.ARN, error) {
```
VasilSirakov, line 495: we could write a couple of simple test cases for this function

VasilSirakov (approved):

providers/aws/resources/aws.go
```diff
@@ -204,13 +205,23 @@ func getAssetIdentifier(runtime *plugin.Runtime) *assetIdentifier {
...
+			if _, err := arn.Parse(id); err == nil {
+				arnStr = id
+			} else {
+				log.Debug().Str("invalid_arn", id).Err(err).Msg("skipping invalid ARN in asset PlatformIds")
+			}
 		}
 	}
-	return &assetIdentifier{name: a.Name, arn: arn}
+	// Only return an assetIdentifier if we have a valid ARN
+	if arnStr == "" {
```
VasilSirakov: could we remove this in case it is just a sanity check?

there is a bunch of code throughout the provider that uses this function, so it would be safer for now to keep the old return value and evaluate this separately.

florinutz: You're right. s3, for example, uses `name` even when `arn` is missing: https://github.com/mondoohq/cnquery/blob/d0e0eabf0b0f8bcfd03bfd32d325e7a6e38bd12f/providers/aws/resources/aws_s3.go#L178

Removing this if seems to keep us on the safe side.

Thanks for holding your ground!

VasilSirakov (approved):

PR #6378, opened 2026-01-09
🧹shell suggestions

chris-rock:
This change removes the idea of a filtered schema. The reason is that you may want to use a MQL query that comes from a resource provider like yara. Filtering the schema for the compiler is not really useful. Instead we change the approach by returning the provider of the resource suggestion from MQL compiler. Then the shell can re-order the suggestions by the connected provider.

### DIFF
--- a/cli/shell/completer.go
+++ b/cli/shell/completer.go
@@ -22,17 +22,19 @@ type Suggestion struct {
 
 // Completer is an auto-complete helper for the shell
 type Completer struct {
-	schema             resources.ResourcesSchema
-	features           cnquery.Features
-	connectedProviders []string
+	schema   resources.ResourcesSchema
+	features cnquery.Features
+	sortFn   func(a, b *llx.Documentation) int
 }
 
 // NewCompleter creates a new Mondoo completer object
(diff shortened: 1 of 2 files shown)


cli/shell/completer.go
```diff
@@ -46,27 +49,30 @@ func (c *Completer) Complete(text string) []Suggestion {
...
-	if c.queryPrefix != nil {
-		query = c.queryPrefix()
-	}
-	query += text
-
-	bundle, _ := mqlc.Compile(query, nil, mqlc.NewConfig(c.schema, c.features))
+	bundle, _ := mqlc.Compile(text, nil, mqlc.NewConfig(c.schema, c.features))
 	if bundle != nil && len(bundle.Suggestions) > 0 {
+		// reorder suggestions to put the ones from connected providers first
+		slices.SortFunc(bundle.Suggestions, func(a, b *llx.Documentation) int {
```
preslavgerchev: nit: can we extend the mql suggestions tests to also test this?

a test case where shell is connected to one provider and we give it resources from that one and another (not connected) one

chris-rock: added a new test

PR #6404, opened 2026-01-16
🧹 Update go-atlassian to v2.11.0 and add nil safety guards

tas50:
## Summary
- Update `go-atlassian` from v1 to v2.11.0 (latest)
- Add nil guards for all pointer fields in the v2 library (`Creator`, `AvatarURLs`, `Project`, `Status`, `IssueType`, `Created`) to prevent nil pointer panics
- Rebase onto main (cnquery→mql v13 rename)

## Test plan
- [ ] Build the atlassian provider: `make providers/build/atlassian`
- [ ] Verify Jira issues query works with valid credentials
- [ ] Verify no panic when issue fields are nil (e.g., issues without a creator)


providers/atlassian/resources/atlassian_jira.go
```diff
@@ -221,9 +220,10 @@ func (a *mqlAtlassianJira) issues() ([]any, error) {
 			return nil, err
 		}
 		for _, issue := range issues.Issues {
-			created, err := time.Parse(JIRA_TIME_FORMAT, issue.Fields.Created)
```
czunker, line 224: I assume we still need to include the format for the parsing.

tas50: Previously they returned a string and we had to do the conversion. Now they return *models.DateTimeScheme which Claude claims is similar to time.Time

providers/atlassian/resources/atlassian_jira.go
czunker, line 268: This might now be `nil` and cause a panic.

providers/atlassian/resources/atlassian_jira.go
```diff
@@ -221,9 +220,9 @@ func (a *mqlAtlassianJira) issues() ([]any, error) {
...
-			if err != nil {
-				return nil, err
+			var created time.Time
+			if issue.Fields.Created != nil {
+				created = time.Time(*issue.Fields.Created)
```
czunker, line 225: We should check that type conversion and also do something when `Created` is nil to later prevent the possible panic.

tas50: Addressed - added nil guards for all pointer fields from the v2 library. `Creator`, `Creator.AvatarURLs`, `Project`, `Status`, `IssueType`, and `Fields` are now all checked for nil before dereferencing. When `Created` is nil, it defaults to the zero-value `time.Time`. When other pointer fields are nil, their string values default to empty strings.

providers/atlassian/resources/atlassian_jira.go
tas50: Fixed - added nil check for `issue.Fields.Creator` and `creator.AvatarURLs` before accessing their fields. Also added nil guards for `issue.Fields.Project`, `issue.Fields.Status`, and `issue.Fields.IssueType` to prevent any nil pointer panics with the v2 library's pointer types.

providers/atlassian/resources/atlassian_jira.go
```diff
@@ -232,7 +231,7 @@ func (a *mqlAtlassianJira) issues() ([]any, error) {
 					"id":      llx.StringData(creator.AccountID),
 					"name":    llx.StringData(creator.DisplayName),
 					"type":    llx.StringData(creator.AccountType),
-					"picture": llx.StringData(creator.AvatarUrls.One6X16),
+					"picture": llx.StringData(creator.AvatarURLs.One6X16),
```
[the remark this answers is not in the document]
tas50: Addressed — added nil guards for `Creator`, `AvatarURLs`, `Project`, `Status`, and `IssueType` before dereferencing. When nil, string fields default to empty strings and `Created` defaults to zero-value `time.Time`.

PR #6410, opened 2026-01-17
⭐ add launchPermissions field to aws.ec2.image

### LINKED ISSUE #6362: Add `launchPermissions` field to `aws.ec2.image` resource
## Summary

The `aws.ec2.image` resource is missing a field to determine whether an AMI is shared with other accounts.

**Note:** The `public` field already exists and works correctly.

## AWS API Reference

- `DescribeImageAttribute` with `Attribute=launchPermission` returns sharing details

## Proposed MQL Field

```go
"launchPermissions": {
    Type:      types.List(types.Dict),
    Doc:       "Launch permissions for the image (user IDs and groups)",
}
```

## Use Case

```mql
// Find AMIs shared with all users
aws.ec2.images.where(launchPermissions.any(Group == "all"))
```

## Additional Context

This field is needed to query AMI sharing configuration beyond just the public/private status.

anurag-2911:
Add launchPermissions() field to the aws.ec2.image resource that returns launch permission information for AMIs. This allows users to determine who can launch instances from an AMI beyond the simple public/private boolean.

The field returns a list of permission dictionaries containing:
- userId: AWS account ID that has launch permission
- group: Special group ("all" for public AMIs)
- organizationArn: AWS Organization ARN with permission
- organizationalUnitArn: AWS OU ARN with permission

Closes #6362

### DIFF
--- a/providers/aws/resources/aws.lr.go
+++ b/providers/aws/resources/aws.lr.go
@@ -30320,12 +30320,7 @@ func createAwsEc2ImageLaunchPermission(runtime *plugin.Runtime, args map[string]
 		return res, err
 	}
 
-	if res.__id == "" {
-		res.__id, err = res.id()
-		if err != nil {
-			return nil, err
-		}
-	}
+	// to override __id implement: id() (string, error)
 
 	if runtime.HasRecording {
 		args, err = runtime.ResourceFromRecording("aws.ec2.image.launchPermission", res.__id)


providers/aws/resources/aws.lr
```diff
@@ -3470,6 +3470,8 @@ private aws.ec2.image @defaults("ownerAlias id name") {
   tags map[string]string
   // Region where the image is located
   region string
+  // Launch permissions for the image (user IDs and groups that can launch instances from this AMI)
+  launchPermissions() []dict
```
imilchev: I think you should add a new resource, instead of using `dict`. The downside of `dict` is that it has 0 docs and it's pretty much guessing what the thing contains. If you define a new resource for this, you can define the properties and add docs

anurag-2911: Updated the implementation as per the feedback,Please check now

providers/aws/resources/aws_ec2.go
```diff
@@ -1224,6 +1224,62 @@ func (i *mqlAwsEc2Image) id() (string, error) {
...
+		ImageId:   aws.String(imageId),
+		Attribute: ec2types.ImageAttributeNameLaunchPermission,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	permissions := make([]interface{}, 0, len(result.LaunchPermissions))
+	for _, perm := range result.LaunchPermissions {
+		mqlPermission, err := CreateResource(i.MqlRuntime, "aws.ec2.image.launchPermission",
```
imilchev: as of recently, we have `const` defined for the resource name. You can use the const instead of typing in the string manually. `ResourceAwsEc2ImageLaunchPermission` (or something similar)

providers/aws/resources/aws_ec2.go
```diff
@@ -1224,6 +1224,62 @@ func (i *mqlAwsEc2Image) id() (string, error) {
...
+			map[string]*llx.RawData{
+				"userId":                llx.StringDataPtr(perm.UserId),
```
imilchev, line 1264: every resource needs to have `__id` defined. That `__id` needs to be globally unique because it is used for storing mql resources in memory (for caching purposes). If you don't define `__id`, it will be an empty string and you will be overriding cache entries because all your instances of this MQL resource will have the same id. 

Search for `__id` in the codebase to see a few examples of how we create those

providers/aws/resources/aws_ec2.go
```diff
@@ -1224,6 +1224,62 @@ func (i *mqlAwsEc2Image) id() (string, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		permissions = append(permissions, mqlPermission)
+	}
+
+	return permissions, nil
+}
+
+func (p *mqlAwsEc2ImageLaunchPermission) id() (string, error) {
```
imilchev: ah, I see you added this to a func. I think it would save you some lines of code if you just set the `__id` property when you create the resource

PR #6415, opened 2026-01-19
 ⭐ MS365: Add `externalAccessWithTrialTenants` field

### LINKED ISSUE #6268: MS365: Add `externalAccessWithTrialTenants` field to `ms365.teams.csTenantFederationConfiguration`
The `ms365.teams.csTenantFederationConfiguration` resource is missing the `ExternalAccessWithTrialTenants` property which controls federation with trial Teams tenants.

This property is available via Teams PowerShell:
```powershell
Get-CsTenantFederationConfiguration | Select-Object ExternalAccessWithTrialTenants
```

**Expected values:** `Allowed` or `Blocked`

**Desired MQL:**
```
ms365.teams.csTenantFederationConfiguration.externalAccessWithTrialTenants
```

Since cnquery already uses Teams PowerShell for the `csTenantFederationConfiguration` resource, this field should be added to the existing `CsTenantFederationConfiguration` struct in `ms365_teams.go`.

**Reference:** https://learn.microsoft.com/en-us/powershell/module/microsoftteams/get-cstenantfederationconfiguration

florinutz:
The `ms365.teams.csTenantFederationConfiguration` resource is missing the `ExternalAccessWithTrialTenants` property which controls federation with trial Teams tenants.

This property is available via Teams PowerShell:
```powershell
Get-CsTenantFederationConfiguration | Select-Object ExternalAccessWithTrialTenants
```

**Expected values:** `Allowed` or `Blocked`

**Desired MQL:**
```
ms365.teams.csTenantFederationConfiguration.externalAccessWithTrialTenants
```

Since cnquery already uses Teams PowerShell for the `csTenantFederationConfiguration` resource, this field should be added to the existing `CsTenantFederationConfiguration` struct in `ms365_teams.go`.

**Reference:** https://learn.microsoft.com/en-us/powershell/module/microsoftteams/get-cstenantfederationconfiguration

---

closes #6268

### DIFF
--- a/providers/ms365/resources/ms365_teams.go
+++ b/providers/ms365/resources/ms365_teams.go
@@ -66,7 +66,7 @@ ConvertTo-Json -Depth 4 $msteams
 `
 
 type MsTeamsReport struct {
-	CsTeamsClientConfiguration      any                      `json:"CsTeamsClientConfiguration"`
+	CsTeamsClientConfiguration      any                              `json:"CsTeamsClientConfiguration"`
 	CsTenantFederationConfiguration *CsTenantFederationConfiguration `json:"CsTenantFederationConfiguration"`
 	CsTeamsMeetingPolicy            *CsTeamsMeetingPolicy            `json:"CsTeamsMeetingPolicy"`
 	CsTeamsMessagingPolicy          *CsTeamsMessagingPolicy          `json:"CsTeamsMessagingPolicy"`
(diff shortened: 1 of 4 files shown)


providers/ms365/resources/ms365_teams.go
```diff
@@ -200,6 +201,7 @@ func (r *mqlMs365Teams) gatherTeamsReport() error {
 				"treatDiscoveredPartnersAsUnverified":         llx.BoolData(tenantConfig.TreatDiscoveredPartnersAsUnverified),
 				"sharedSipAddressSpace":                       llx.BoolData(tenantConfig.SharedSipAddressSpace),
 				"restrictTeamsConsumerToExternalUserProfiles": llx.BoolData(tenantConfig.RestrictTeamsConsumerToExternalUserProfiles),
+				"externalAccessWithTrialTenants":              llx.StringData(tenantConfig.ExternalAccessWithTrialTenants),
```
czunker, line 204: nit:
I don't know how we do this for other properties, but would it make sense to convert the string to lowercase?

florinutz: We don't really lowercase stuff in cnquery or in this provider.

I would keep the value as it is when it arrives from the endpoint. If users want case-insensitive    
  comparisons in their queries, they can use `.downcase`

PR #6419, opened 2026-01-20
🐛  fix nil interface and return types for time

LittleSalkin1806:
Two things are fixed here.

1. Added "if bind.Value == nil check" before the type assertion to prevent the panic
 I found this bug while accessing a optional field -> rotationperiod of kms keys which was not set.
 
 The error
 ```golang
 panic: interface conversion: interface {} is nil, not *time.Time [recovered, repanicked]
 ```

2.Fixed the return type from types.Array(types.Time) to the correct type (types.Int for most, types.Bool for timeInRange)

Here you can see the types of time
```golang
types.Time: {
	"seconds": {typ: intType, signature: FunctionSignature{}},
	"minutes": {typ: intType, signature: FunctionSignature{}},
	"hours":   {typ: intType, signature: FunctionSignature{}},
	"days":    {typ: intType, signature: FunctionSignature{}},
	"unix":    {typ: intType, signature: FunctionSignature{}},
	"inRange": {typ: boolType, compile: compileTimeInRange},
},
```

### DIFF
--- a/llx/builtin_simple.go
+++ b/llx/builtin_simple.go
@@ -2552,9 +2552,12 @@ func DurationToTime(i int64) time.Time {
 }
 
 func timeSecondsV2(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64) (*RawData, uint64, error) {
+	if bind.Value == nil {
+		return &RawData{Type: types.Int}, 0, nil
+	}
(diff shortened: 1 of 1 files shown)


arlimus (approved):
@LittleSalkin1806 thank you!! :pray:

PR #6422, opened 2026-01-20
⭐ MS365: Add `microsoft.security.informationProtection.sensitivityLabels` resource

### LINKED ISSUE #6269: MS365: Add `microsoft.security.informationProtection.sensitivityLabels` resource
We need access to sensitivity labels from Microsoft Purview Information Protection via the Graph API.

**API Endpoint:**
```
GET https://graph.microsoft.com/beta/security/informationProtection/sensitivityLabels
```

**Response:**
```json
{
  "value": [
    {
      "id": "guid",
      "name": "Confidential",
      "displayName": "Confidential",
      "isEnabled": true,
      "isActive": true,
      "assignedPolicies": [...]
    }
  ]
}
```

**Desired MQL:**
```
microsoft.security.informationProtection.sensitivityLabels {
  id
  name
  displayName
  isEnabled
  isActive
  assignedPolicies
}
```

**Permissions required:** `InformationProtectionPolicy.Read.All`

**Reference:** https://learn.microsoft.com/en-us/graph/api/resources/security-sensitivitylabel

florinutz:
We need access to sensitivity labels from Microsoft Purview Information Protection via the Graph API.

**API Endpoint:**
```
GET https://graph.microsoft.com/beta/security/informationProtection/sensitivityLabels
```

**Response:**
```json
{
  "value": [
    {
      "id": "guid",
      "name": "Confidential",
      "displayName": "Confidential",
      "isEnabled": true,
      "isActive": true,
      "assignedPolicies": [...]
    }
  ]
}
```

**Desired MQL:**
```
microsoft.security.informationProtection.sensitivityLabels {
  id
  name
  displayName
  isEnabled
  isActive
  assignedPolicies
}
```

**Permissions required:** `InformationProtectionPolicy.Read.All`

**Reference:** https://learn.microsoft.com/en-us/graph/api/resources/security-sensitivitylabel

---

PS: watch out for [this comment](https://github.com/mondoohq/cnquery/pull/6422#issuecomment-3773731742). Basically, the beta endpoint response has changed since December and I agreed with @syrull to just fetch whatever's there atm.

---

closes #6269

### DIFF
--- a/providers/ms365/resources/security_informationprotection.go
+++ b/providers/ms365/resources/security_informationprotection.go
@@ -37,10 +37,13 @@ func (r *mqlMicrosoftSecurityInformationProtection) sensitivityLabels() ([]any, 
 		return nil, transformError(err)
 	}
 
+	labels, err := iterate[security.SensitivityLabelable](ctx, resp, betaClient.GetAdapter(), security.CreateSensitivityLabelCollectionResponseFromDiscriminatorValue)
+	if err != nil {
+		return nil, err
+	}
+
 	var res []any
-	labels := resp.GetValue()
-	for i := range labels {
-		label := labels[i]
+	for _, label := range labels {
 		if label == nil {
 			continue
 		}


florinutz:
The beta endpoint might have changed since the writing of this ticket, because these are the fields that I could use at this point:

```go
		map[string]*llx.RawData{
			"__id":           llx.StringDataPtr(label.GetId()),
			"id":             llx.StringDataPtr(label.GetId()),
			"name":           llx.StringDataPtr(label.GetName()),
			"description":    llx.StringDataPtr(label.GetDescription()),
			"toolTip":        llx.StringDataPtr(label.GetTooltip()),
			"color":          llx.StringDataPtr(label.GetColor()),
			"contentFormats": llx.ArrayData(contentFormats, types.String),
			"isAppliable":    llx.BoolDataPtr(label.GetIsAppliable()),
			"hasProtection":  llx.BoolDataPtr(label.GetHasProtection()),
			"isActive":       llx.BoolDataPtr(label.GetIsActive()),
			"sensitivity":    llx.IntDataPtr(label.GetSensitivity()),
			"parent":         llx.ResourceData(parentResource, ResourceMicrosoftSecurityInformationProtectionSensitivityLabel),
		})
```

syrull:
Yes, this seems okay

providers/ms365/resources/security_informationprotection.go
```diff
@@ -0,0 +1,95 @@
...
+	for i := range labels {
+		label := labels[i]
+		mqlResource, err := createSensitivityLabelResource(r.MqlRuntime, label)
+		if err != nil {
+			return nil, err
+		}
+		res = append(res, mqlResource)
+	}
+	return res, nil
+}
```
Bajusz15, line 57: Not a big deal if you leave it like this, but I think you don't need to call the API every time we use this field. There is a caching pattern used across ms365 resources, you can cache the values once in informationProtection() instead of creating empty resource, then sensitivityLabels() would just return the cached value.

providers/ms365/resources/security_informationprotection.go
```diff
@@ -0,0 +1,95 @@
...
+
+func createSensitivityLabelResource(runtime *plugin.Runtime, label security.SensitivityLabelable) (plugin.Resource, error) {
+	if label == nil {
+		return nil, nil
+	}
```
Bajusz15: I think you should not call this to create a resource, if label is nil.

providers/ms365/resources/security_informationprotection.go
```diff
@@ -0,0 +1,95 @@
...
+	"go.mondoo.com/cnquery/v12/types"
+)
+
+func (r *mqlMicrosoftSecurity) informationProtection() (*mqlMicrosoftSecurityInformationProtection, error) {
+	resource, err := CreateResource(r.MqlRuntime, ResourceMicrosoftSecurityInformationProtection, map[string]*llx.RawData{})
+	if err != nil {
+		return nil, err
+	}
+	return resource.(*mqlMicrosoftSecurityInformationProtection), nil
+}
```
Bajusz15, line 25: Also, don't create an empty resource without an id. Even if it's not used because the child is used, it's better to give it a unique id so it stays cached by the runtime. 

Like `<subscription_id>/information_protection` or similar.

providers/ms365/resources/security_informationprotection.go
Bajusz15: I think it's enough if you fix this, the rest of my comments are nit.

providers/ms365/resources/security_informationprotection.go
tas50: This follows the same pattern as `exchange()` which also creates a resource and lets sub-fields lazy-load via the runtime's `GetOrCompute` caching. Keeping it consistent for now.

providers/ms365/resources/security_informationprotection.go
tas50: Fixed — nil labels are now skipped in the loop before calling `createSensitivityLabelResource`.

providers/ms365/resources/security_informationprotection.go
tas50: Fixed — added `__id` using the tenant ID: `microsoft.security.informationProtection/<tenantId>`.

providers/ms365/resources/security_informationprotection.go
```diff
@@ -0,0 +1,97 @@
...
+		if label == nil {
+			continue
+		}
+		mqlResource, err := createSensitivityLabelResource(r.MqlRuntime, label)
+		if err != nil {
+			return nil, err
+		}
+		res = append(res, mqlResource)
+	}
+	return res, nil
```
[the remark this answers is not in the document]
tas50: Fixed — now using the existing `iterate[T]` helper with `security.CreateSensitivityLabelCollectionResponseFromDiscriminatorValue` to handle pagination.

PR #6428, opened 2026-01-21
🐛 bug + enhancement - prevent false positives findings for user errors

LittleSalkin1806:
## Problem

While fixing [cnspec#2037](https://github.com/mondoohq/cnspec/issues/2037), I discovered that MQL silently returns `nil` when accessing dict/map keys that don't exist due to case mismatch.

**Example:** `gcloud.storage.bucket.iamConfiguration.UniformBucketLevelAccess.enabled`

The key `UniformBucketLevelAccess` doesn't exist — it should be `uniformBucketLevelAccess` (lowercase `u`). Instead of an error, the query silently returns `nil`, causing false positives.

For me personally, false positives in security queries should be avoided. They aren't as bad as false negatives, but they still sit there silently, occupy resources, and need to be handled. At worst, they trigger incidents for issues that aren't actually happening.

## Solution

**Commit 1: Case-mismatch detection**
- Added "did you mean?" suggestions when a key exists with different casing
- Truly non-existent keys still return `nil` (backward compatible)

**Commit 2: Error propagation in block operations**
- `where`, `all`, `any`, `one`, `none` now propagate errors instead of silently absorbing them
- Aligns with how `runBlock` already works (line 639-640 of llx.go)
- `filterList` (used by `recurse`) remains error-absorbing since `recurse` intentionally traverses heterogeneous structures

**Commit 3: Tests**
- Added comprehensive tests for key-not-found behavior

## Before / After

**Before:**

    [failed] gcp.storage.bucket.iamConfiguration.UniformBucketLevelAccess.enabled == true
    expected: == true
    actual: _

**After:**

    [failed] gcp.storage.bucket.iamConfiguration.UniformBucketLevelAccess.enabled == true
    expected: == true
    actual: key 'UniformBucketLevelAccess' not found, did you mean 'uniformBucketLevelAccess'?

## Why propagate ALL errors (not just case-mismatch)?

1. **Consistency**: `runBlock` already propagates errors via `err.Deduplicate()`
2. **Silent failures are dangerous**: The previous behavior led to the false positive issue
3. **Debugging aid**: Type conflicts, null operations, and cast errors inside predicates should be visible to users

## Learnings & Open Questions

After implementing these changes, I learned that we only deal with the actual API response of each resource and do not use any provider schema. Therefore, `nil` returns are mandatory to keep the current logic working.

I'm aware that safety can be achieved with better-written MQL queries (e.g., using `has`), but I wanted to address this at a lower level (in the code). So I assume the current behavior is intentional / a feature.

**Potential improvements:**
- **Fuzzy matching**: Could catch more typos beyond just case-sensitivity
- **Explicit error handling**: Absorb some errors intentionally while surfacing others

With case-insensitive checks and fuzzy matching, we could likely reduce many false positives.

## Conclusion

There are still some open questions, but at minimum this is a finding: the current code can produce many false positives without additional checks. Safety can be achieved today using explicit MQL query functions (like `has`), but there may be room for improvement at the engine level.

Especially fuzzy matching and case insensitive is the case for nested dictionaries where no schema from .lr can be used inside the interactive shell and scans.

### DIFF
--- a/llx/builtin_map.go
+++ b/llx/builtin_map.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/lithammer/fuzzysearch/fuzzy"
 	"go.mondoo.com/cnquery/v12/types"
 	"go.mondoo.com/cnquery/v12/utils/multierr"
 )
@@ -80,6 +81,15 @@ func mapGetIndexV2(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64) (*
 				}, 0, nil
 			}
 		}
+
+		// No case-insensitive match - try fuzzy matching for typos
+		if suggestion := findFuzzyMapKey(key, m); suggestion != "" {
+			return &RawData{
+				Type:  childType,
+				Error: errors.New("key '" + key + "' not found, did you mean '" + suggestion + "'?"),
+			}, 0, nil
+		}
+
 		// Key doesn't exist and no similar keys found - return nil for backward compatibility
 		return &RawData{
 			Type:  childType,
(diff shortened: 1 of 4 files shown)


LittleSalkin1806:
@arlimus  i think the fuzzy search is theoretically the only point not introduced but anway do you see any value to let this PR open or should i close it ?
Many thanks for these nice features :)

PR #6454, opened 2026-01-24
Aws bugs fixes and enhancements (fix 4+m initialization of aws shell/scans)

LittleSalkin1806:
## Summary

In this PR I addressed several findings in the AWS provider that caused **5+ minutes to initialize the shell** when running without any options (e.g., `cnspec/cnquery shell aws`, `cnspec/cnquery scan aws`) + when handeling filters `--filters regions=eu-central-1`.

## Root Cause Analysis

Without specifying any discovery options or filters, the AWS provider performs a **full account discovery** across **all regions** and **all resource types**. Combined with the three problems below, this resulted in extremely slow initialization times:

| Problem | Description | Impact |
|---------|-------------|--------|
| **Problem 1** | Filters are not propagated to child connections | Child connections ignored parent-level filters, scanning unintended resources |
| **Problem 2** | Child connections run default discover behavior (`auto`) even when parent used a dedicated discover option | Each child connection re-discovered resources instead of respecting parent's targeted scope |
| **Problem 3** | Default HTTP client configuration (long timeouts and per-request retry/backoff) is applied for each region and asset |  Cumulative retry/backoff delays across regions and assets cause repeated timeouts and massively slow initialization especially because auto discovery option was used for most of the child connections. |
| Problem 4| Some resources are designed in a way to ignore filters regions. Which is not visible for the user directly. | Problems 1-3 occur even with filter regions for S3 Buckets especially.

## Consequences

These three problems combined caused:

- **5+ minute shell initialization** for a simple `cnspec shell aws` command
- **Excessive AWS API calls** - child connections discovering resources in "auto" mode instead of respecting parent's targeted discovery option
- **Multiplied timeout delays** - each region and asset retrying independently
- **Scanning of unintended resources** - filters specified at parent level not passed to child connections
- **Poor user experience** - scans appearing to hang with no feedback
- **Potential AWS rate limiting** - broader discovery scope combined with retry attempts
- **Resource waste** - filtering logic re-evaluated (or ignored) at each child connection level

## Additional Changes

1. **Documentation fix** - Corrected `--regions` CLI option documentation
2. **AWS S3 API optimization** - Reduced API calls when `--filters regions=xx` is provided (+ added TODO for best practice)
3. **Race condition fix** - Cloned connection pointer was always pointing to parent variable

## Open Topics

- [x] **Tests** - Need to add test coverage for the fixes
- [ ]  **Assets Ignoring Filters** S3 Behaviour fixed. Find out and fix other services which behave similar -> WAF ?.
- run before changes [full_run.log](https://github.com/user-attachments/files/24832514/full_run.log)
- Tested most of the stuff locally after fixed  can provide logs if needed

### DIFF
--- a/providers/aws/resources/aws_s3.go
+++ b/providers/aws/resources/aws_s3.go
@@ -57,8 +57,6 @@ func (a *mqlAwsS3) id() (string, error) {
 
 func (a *mqlAwsS3) buckets() ([]any, error) {
 	conn := a.MqlRuntime.Connection.(*connection.AwsConnection)
-
-	svc := conn.S3("")
 	ctx := context.Background()
 
 	configuredRegions, err := conn.Regions()
@@ -73,6 +71,7 @@ func (a *mqlAwsS3) buckets() ([]any, error) {
 	var bucketsWithRegions []bucketWithRegion
 
 	for _, region := range configuredRegions {
+		svc := conn.S3(region)
 		log.Debug().Str("region", region).Msg("listing S3 buckets in region")
 		params := &s3.ListBucketsInput{BucketRegion: aws.String(region)}
 		paginator := s3.NewListBucketsPaginator(svc, params, func(o *s3.ListBucketsPaginatorOptions) {


providers-sdk/v1/inventory/inventory.go
```diff
@@ -455,8 +455,10 @@ func (cfg *Config) Clone(opts ...CloneOption) *Config {
 			clonedObject.Discover = &Discovery{}
 		}
 		clonedObject.Discover.Filter = make(map[string]string)
-		for k, v := range cfg.Discover.Filter {
-			clonedObject.Discover.Filter[k] = v
+		if cfg.Discover != nil {
```
VasilSirakov: nit: can be written as:
```go
maps.Copy(clonedObject.Discover.Filter, cfg.Discover.GetFilter())
```

`GetFilter()` here is nil-safe.

providers/aws/resources/aws_s3.go
```diff
@@ -61,47 +61,107 @@ func (a *mqlAwsS3) buckets() ([]any, error) {
...
-			continue
+	hasIncludeRegionFilter := len(conn.Filters.General.Regions) > 0
+
+	type bucketWithRegion struct {
+		bucket s3types.Bucket
+		region string
+	}
+	var bucketsWithRegions []bucketWithRegion
+
+	if hasIncludeRegionFilter {
```
VasilSirakov: Nice find!

I think we can improve this even further because `conn.Regions()` already applies the include/exclude region filters. In that case, `configuredRegions` should already have only the applicable regions so the entire `else` case along with `hasIncludeRegionFilter` can be dropped.

providers/aws/resources/aws_s3.go
```diff
@@ -61,47 +61,107 @@ func (a *mqlAwsS3) buckets() ([]any, error) {
...
-		o.Limit = 100
-	})
-	for paginator.HasMorePages() {
-		output, err := paginator.NextPage(ctx)
-		if err != nil {
-			return nil, err
-		}
-		totalBuckets = append(totalBuckets, output.Buckets...)
+	// Get configured regions
+	configuredRegions, err := conn.Regions()
```
preslavgerchev, line 62: nice one @LittleSalkin1806

providers/aws/connection/connection.go
```diff
@@ -75,9 +76,11 @@ func NewAwsConnection(id uint32, asset *inventory.Asset, conf *inventory.Config)
 	for _, opt := range opts {
 		opt(c)
 	}
-	// custom retry client
+	// custom retry client with reduced retries and shorter backoff
+	// to avoid excessive delays when regions are unreachable
 	retryClient := retryablehttp.NewClient()
-	retryClient.RetryMax = 5
+	retryClient.RetryMax = 2              // reduced from 5 to avoid long delays on unreachable regions
```
preslavgerchev, line 82: +1, makes sense to reduce the time we wait on retries

providers/aws/provider/provider.go
```diff
@@ -289,7 +292,9 @@ func (s *Service) detect(asset *inventory.Asset, conn plugin.Connection) error {
 }
 
 func (s *Service) discover(conn *connection.AwsConnection) (*inventory.Inventory, error) {
-	if conn.Conf.Discover == nil {
+	// Skip discovery for child connections (they have a parent connection ID set).
+	// and should not run discovery again.
+	if conn.Conf.ParentConnectionId != 0 {
```
preslavgerchev: I think we should turn this into a check whether the discover config has any targets (len(conn.Conf.GetDiscover().GetTargets()) == 0) and fix this differently: Ensure that `ParseCLI` always assigns `auto` if no targets are set at all, just like azure does it:
https://github.com/mondoohq/cnquery/blob/main/providers/azure/provider/provider.go#L102


While this would work for now, we might start doing additional discovery from child connections in the future, e.g. fanning out discovery across a large AWS organization to each child AWS account, in that case we would need to run discovery from the child connection

providers-sdk/v1/inventory/inventory.go
LittleSalkin1806: Thanks for the hint :)

providers/aws/resources/aws_s3.go
LittleSalkin1806: Thats something i should have seen....
Sadly at 2 am my brain seem to have lost the ability to think.

Good catch.
Feels like playing baseball

providers/aws/connection/connection.go
```diff
@@ -328,6 +331,10 @@ func (h *AwsConnection) Regions() ([]string, error) {
 	svc := h.Ec2(h.cfg.Region)
 	ctx := context.Background()
 
+	// DescribeRegions works to get the list of enabled regions for the account ( each account of organization)
+	// but this does not mean the respective service endpoint is available in that region. They will timeout instead of failing fast
+	// (e.g. EKS,KMS,Sagemaker is for example not available in ap-southeast-1 etc)
+	// This also does not cover SCPs that might block access to certain regions.
 	res, err := svc.DescribeRegions(ctx, &ec2.DescribeRegionsInput{})
```
LittleSalkin1806, line 338: Added a Remark where i just want to state for situation where its just not possible to define filters is really not ideal.

LittleSalkin1806: AWS Organizational Rollout: Scaling & Performance Issues
=========================================================

Current Behavior
----------------

1. Initial Connection (Management Account)
   - Connects using credentials (role assumption, profile, etc.)
   - Calls DescribeRegions to get list of "enabled" regions for that account
   - Caches this region list

2. Organization Discovery
   - Calls organizations.ListAccounts() to discover all member accounts
   - Creates an "asset" for each account with cloned configuration from the parent

3. Per-Account Scanning
   - Each discovered account gets its own connection
   - Filters (including regions=) are inherited from the parent connection
   - Each account then iterates through the region list for every AWS service (S3, RDS, EC2, EKS, etc.)


Key Problems
------------

Problem 1: Region Availability ≠ Service Endpoint Availability

DescribeRegions returns regions that are enabled for the account, but this does not mean service endpoints are available in those regions. When a service endpoint is unavailable, the API call will timeout instead of failing fast.


Problem 2: SCPs Not Reflected in DescribeRegions

Organizations using Service Control Policies (SCPs) with deny rules will still see all regions reported as "active" by DescribeRegions. The SCP restrictions are only enforced at the API call level, causing:
- Timeouts on every blocked region
- Timeouts multiplied by every service
- Timeouts multiplied by every account


Problem 3: No Platform-Level Filter Options

For Mondoo-hosted/platform-integrated scans, there is currently no way to provide scan inputs such as:
- Region include/exclude filters
- Account include/exclude filters
- Service-specific configurations

This makes it impossible for operators to optimize scans for their specific organizational setup.


Impact at Scale
---------------

500 accounts × 5 SCP-blocked regions × 30+ services × ~1s timeout = Potential long wait times

The current implementation determines regions via a single ec2.DescribeRegions call rather than per-service availability checks. This creates significant overhead:

1. Unnecessary API calls to regions where services are blocked
2. Timeouts instead of fast failures for unavailable endpoints
3. No optimization path without platform-level filter support


Summary
-------

For organizational rollout or account-based Mondoo-hosted integration, the current design creates excessive overhead in API calls. This scales especially poorly for organizations because:

1. SCPs are invisible to the DescribeRegions check that determines scan scope
2. Unavailable service endpoints timeout rather than fail fast
3. No filter options can be provided at the platform level to limit scan scope

Without the ability to configure region filters at the platform level, organizations with SCPs will experience significant scan delays that scale with the number of accounts, blocked regions, and services.

What i dont know is how bad this scales

providers/aws/resources/aws.lr
```diff
@@ -2739,7 +2739,7 @@ aws.rds.dbcluster @defaults("id region engine engineVersion") {
   engine string
   // Version of the database engine for this DB cluster
   engineVersion string
-  // Whether the cluster is publicly accessible
+  // Whether the cluster is publicly accessible. Note: This will only return a value for non-Aurora Multi-AZ DB clusters
```
LittleSalkin1806, line 2742: This was just a finding by concidence.
I did wanted to start a whole PR for this.
its cluster related not db instance related as i could see from the types.go just switched the text.

providers/aws/resources/discovery.go
```diff
@@ -148,11 +148,11 @@ func Discover(runtime *plugin.Runtime) (*inventory.Inventory, error) {
 }
 
 func getDiscoveryTargets(config *inventory.Config) []string {
-	targets := config.Discover.Targets
-
-	if len(targets) == 0 {
+	// as we dont always initalize Discover in ParseCLI we provide both checks
+	if config.Discover == nil || len(config.Discover.Targets) == 0 {
```
preslavgerchev: following https://github.com/mondoohq/cnquery/pull/6454#discussion_r2727723763 we could just remove this check since we ensure that `ParseCli` would always provide back a target (either auto if nothing is specified or something more specific if the user used the --targets flag)

providers/aws/connection/connection.go
LittleSalkin1806: @preslavgerchev @VasilSirakov  could you give me feedback to this remark. How bad you think this scales ?
We have in our AWS org round about 1800 accounts if we include everything

providers/aws/resources/discovery.go
LittleSalkin1806: I created this after i got a panic with no inputs provided after some of the changes in this PR.
So dont know the root cause anymore. Then i will change it.

providers/aws/provider/provider.go
LittleSalkin1806: Tbh maybe i was just not updating my provider again with the build one
I am confused ...
With this fix its working exactly as intended

So the problem on org is that it only has accounts as connections and this will default to auto discover right ?

providers/aws/connection/connection.go
LittleSalkin1806: For now i could only see a single account took 4+m  instead of some seconds when we could apply filters and not ask for service endpoints that time out.
That this regions discovery scaled bad especially if it is being redone on every account within an org.

providers/aws/provider/provider.go
preslavgerchev: @LittleSalkin1806 It's more of a design choice than anything, your fix would still work but we are generally trying to control discovery only through the discover targets in the inventory and not via properties of the connection itself

I did a small patch to show the suggested changes where
1. We ensure that auto is set if there are 0 targets in `ParseCLI`
2. Discover now looks at the discovery targets instead. There's a subtle difference that we now check if the are targets instead of whether the struct is nil
3. The AWS provider discovery now only works with the provided list and does not fallback to auto as that's handled by the `ParseCLI` function

```sh
diff --git a/providers/aws/provider/provider.go b/providers/aws/provider/provider.go
index 94ceb13a2..be271a339 100644
--- a/providers/aws/provider/provider.go
+++ b/providers/aws/provider/provider.go
@@ -54,9 +54,6 @@ func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error)
                return &plugin.ParseCLIRes{Asset: asset}, nil
        }
 
-       inventoryConfig := &inventory.Config{
-               Type: req.Connector,
-       }
        // discovery flags
        discoverTargets := []string{}
        if x, ok := flags["discover"]; ok && len(x.Array) != 0 {
@@ -67,9 +64,13 @@ func (s *Service) ParseCLI(req *plugin.ParseCLIReq) (*plugin.ParseCLIRes, error)
        }
        filterOpts := parseFlagsToFiltersOpts(flags)
 
-       // skip unecessary discovery objects for child connections where we dont set both flags
-       if len(discoverTargets) > 0 || len(filterOpts) > 0 {
-               inventoryConfig.Discover = &inventory.Discovery{Targets: discoverTargets, Filter: filterOpts}
+       if len(discoverTargets) == 0 {
+               discoverTargets = []string{resources.DiscoveryAuto}
+       }
+
+       inventoryConfig := &inventory.Config{
+               Type:     req.Connector,
+               Discover: &inventory.Discovery{Targets: discoverTargets, Filter: filterOpts},
        }
        asset := inventory.Asset{
                Connections: []*inventory.Config{inventoryConfig},
@@ -292,9 +293,7 @@ func (s *Service) detect(asset *inventory.Asset, conn plugin.Connection) error {
 }
 
 func (s *Service) discover(conn *connection.AwsConnection) (*inventory.Inventory, error) {
-       // Skip discovery for child connections (they have a parent connection ID set).
-       // and should not run discovery again.
-       if conn.Conf.ParentConnectionId != 0 {
+       if len(conn.Conf.GetDiscover().GetTargets()) == 0 {
                return nil, nil
        }
 
diff --git a/providers/aws/resources/discovery.go b/providers/aws/resources/discovery.go
index 0c5ad556e..105cb583c 100644
--- a/providers/aws/resources/discovery.go
+++ b/providers/aws/resources/discovery.go
@@ -148,11 +148,7 @@ func Discover(runtime *plugin.Runtime) (*inventory.Inventory, error) {
 }
 
 func getDiscoveryTargets(config *inventory.Config) []string {
-       // as we dont always initalize Discover in ParseCLI we provide both checks
-       if config.Discover == nil || len(config.Discover.Targets) == 0 {
-               return Auto
-       }
-       targets := config.Discover.Targets
+       targets := config.GetDiscover().GetTargets()
 
        if stringx.Contains(targets, DiscoveryAll) {
                // return the All list + All Api Resources list
```

providers/aws/connection/connection.go
LittleSalkin1806: This config is used for every call

This feature makes for me a lot of sense in aws.

Adding an endpoint cache to AwsConnection
Wrapping service discovery calls with the health check
Using a shorter timeout for initial endpoint probes

providers/aws/provider/provider.go
LittleSalkin1806: Thanks that helped me to get it.
Let me change it.

For me the important part was to prevent automatic re-discovery on child connections in general.

providers/aws/connection/connection.go
LittleSalkin1806: The http config for api calls is used for every api call.

This feature makes for me a lot of sense in aws.

Adding an endpoint cache to AwsConnection
Wrapping service discovery calls with the health check
Using a shorter timeout for initial endpoint probes

Give back Region list based on inital endpoint health check
Either no permission as SCP in place (already handeled)
But not for timeouts for dedicated service endpoints only on EC2 list

providers/aws/resources/aws_s3.go
```diff
@@ -61,47 +61,45 @@ func (a *mqlAwsS3) buckets() ([]any, error) {
...
-	for _, bucket := range totalBuckets {
-		location, err := svc.GetBucketLocation(ctx, &s3.GetBucketLocationInput{
-			Bucket: bucket.Name,
+	type bucketWithRegion struct {
+		bucket s3types.Bucket
+		region string
+	}
+	var bucketsWithRegions []bucketWithRegion
+
+	for _, region := range configuredRegions {
```
preslavgerchev, line 73: I think we need to bring the client init to inside the region loop here (line 61: `svc := conn.S3("")`)  and make it `svc := conn.S3(region)` so that we we can re-initialize it for each region correctly. Else I see this when running discovery:
```
! could not list S3 buckets in region error="operation error S3: ListBuckets, https response error StatusCode: 400, RequestID: SJYNJZ76B2R1GN2M, HostID: M1L6lyP4BqlqCmG8EHKccY41gBh9UmwOcUWrkVzQKj42uIMSNjiM9gjAOG52grMf3Zv9hpl75CA=, api error InvalidArgument: Requests with bucket-region specified must be made to the corresponding regional endpoint" region=ap-south-1
```

VasilSirakov: +1 reproduced it as well. I think PR can be approved once that's fixed

PR #6457, opened 2026-01-24
⭐ Add containerd.containers resource

tas50:
This mirrors docker.containers as much as possible

```
> containerd.containers
containerd.containers: [
  0: containerd.container id="98deb1bb7adca2cce91ae97270238814436c4de3b8b15da64922bea1948f8671" status="RUNNING"
  1: containerd.container id="a5e26880c8937d5d0ee37ffc9c3e3605448ff11a6eab018162beac0be6e66919" status="RUNNING"
  2: containerd.container id="f0cd63c6710ff702d9209ebfabc49084fb8fc82e77e411b38816a1b674ec7539" status="RUNNING"
]
```

```
> containerd.containers.first{*}
containerd.containers.first: {
  pid: 3440318
  id: "98deb1bb7adca2cce91ae97270238814436c4de3b8b15da64922bea1948f8671"
  image: ""
  namespace: "moby"
  status: "RUNNING"
  snapshotter: ""
  labels: {
    com.docker/engine.bundle.path: "/var/run/docker/containerd/98deb1bb7adca2cce91ae97270238814436c4de3b8b15da64922bea1948f8671"
  }
  runtime: "io.containerd.runc.v2"
}
```

### DIFF
--- a/providers/os/resources/containerd.go
+++ b/providers/os/resources/containerd.go
@@ -10,6 +10,8 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/kballard/go-shellquote"
+	"github.com/rs/zerolog/log"
 	"go.mondoo.com/cnquery/v12/llx"
 	"go.mondoo.com/cnquery/v12/types"
 )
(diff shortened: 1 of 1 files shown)


providers/os/resources/containerd.go
```diff
@@ -0,0 +1,201 @@
...
+	if err := json.Unmarshal(jsonData, &info); err != nil {
+		return nil, err
+	}
+	return &info, nil
+}
+
+func (p *mqlContainerd) containers() ([]any, error) {
+	// Get all namespaces using ctr CLI
+	o, err := CreateResource(p.MqlRuntime, "command", map[string]*llx.RawData{
+		"command": llx.StringData("ctr namespaces list -q"),
```
chris-rock: Generally I think we should not require the containerd command cli, but instead use the go sdk. We can add this as a followup improvement

tas50: I tried it with the go SDK but wasn't able to do that remotely which seemed like an important one. I'm down for using the SDK if we can perform that locally and remote.

PR #6462, opened 2026-01-25
⭐ Add Safari extensions detection for macOS

chris-rock:
Add support for querying Safari extensions on macOS systems.

- Add safari resource with extensions() method
- Add safari.extension private resource with metadata fields
- Use pluginkit to enumerate Safari web-extension, extension, and content-blocker types
- Parse Info.plist from .appex bundles for extension metadata
- Include container app information (path and name)

Usage:

```javascript
safari.extensions { name version identifier extensionType }
```

### DIFF
--- a/providers/os/resources/os.lr.manifest.yaml
+++ b/providers/os/resources/os.lr.manifest.yaml
@@ -1176,7 +1176,7 @@ resources:
   safari:
     fields:
       extensions: {}
-    min_mondoo_version: latest
+    min_mondoo_version: 9.0.0
     platform:
       name:
       - macos
@@ -1194,7 +1194,7 @@ resources:
       path: {}
       version: {}
     is_private: true
-    min_mondoo_version: latest
+    min_mondoo_version: 9.0.0
     platform:
       name:
       - macos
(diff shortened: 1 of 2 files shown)


providers/os/resources/os.lr.manifest.yaml
```diff
@@ -1024,6 +1024,31 @@ resources:
         min_mondoo_version: latest
       settings: {}
     min_mondoo_version: 5.15.0
+  safari:
+    fields:
+      extensions: {}
+    min_mondoo_version: latest
```
czunker: The Chrome extension had 9.0.0 here.

providers/os/resources/safari.go
```diff
@@ -0,0 +1,138 @@
...
+	seen := make(map[string]bool)
+	var extensions []any
+
+	for _, extType := range safariExtensionTypes {
+		// Run pluginkit to list extensions of this type
+		cmd, err := conn.RunCommand("pluginkit -mAvvv -p " + extType)
+		if err != nil {
+			// pluginkit might not be available or no extensions of this type
+			continue
+		}
```
czunker, line 62: Should we check the command exists? Or show a warning here?

How do we know whether no extension is installed or the command is missing?

chris-rock: Great point, I extracted the command existence check.

PR #6463, opened 2026-01-25
⭐ Add a sudoers resource

tas50:
This is a new resource that allows us to avoid complex patterns where we have to iterate over /etc/sudoers and all the files in /etc/sudoers.d and then apply an often crytic regex. Instead we can write simpler queries against just user entries or defaults.

Before

```
    props:
      - uid: mondooLinuxSecuritySudoersFiles
        title: Return the files from /etc/sudoers.d
        mql: |
          sudoersFiles = files.find(from: "/etc/sudoers.d/", type: 'file').list.map(path) + ["/etc/sudoers"]
          return sudoersFiles.map(file(_).content.lines.where(_ == /^[^#]/)).flat
    mql: |
      props.mondooLinuxSecuritySudoersFiles
        .where(_ == /Defaults/)
        .any(_ == /logfile=\"\/var\/log\/sudo\.log\"/)
```

After

```
mql: sudoers.defaults.contains(parameter == "logfile" && scope == "global" && value == "/var/log/sudo.log")
```

---

```coffee
> sudoers.defaults
sudoers.defaults: [
  0: sudoers.default parameter="env_reset"
  1: sudoers.default parameter="mail_badpass"
  2: sudoers.default parameter="secure_path"
  3: sudoers.default parameter="use_pty"
  4: sudoers.default parameter="log_host, log_year, logfile"
  5: sudoers.default parameter="env_keep"
  6: sudoers.default parameter="logfile"
]
> sudoers.userSpecs
sudoers.userSpecs: [
  0: sudoers.userSpec users=[
       0: "bob"
     ] hosts=[
       0: "ALL"
     ] commands=[
       0: "ALL"
     ]
]
```


providers/os/resources/sudoers.go
```diff
@@ -0,0 +1,942 @@
...
+		}
+
+		// If a custom path is provided, just use that file
+		files, err := getSortedPathFiles(runtime, path)
+		if err != nil {
+			return nil, nil, err
+		}
+		args["files"] = llx.ArrayData(files, types.Resource("file"))
+		delete(args, "path")
+	}
```
czunker, line 38: Same as for limits. I think we need to fall back to the default here.

providers/os/resources/sudoers.go
```diff
@@ -0,0 +1,942 @@
...
+	// Remove quotes from value if present
+	value = strings.Trim(value, "\"")
+
+	return scope, target, parameter, value, operation, negated
+}
+
+// parseSudoersLine parses a single line from a sudoers file
+func parseSudoersLine(line string) *sudoersLine {
+	// Filter out comments
+	if strings.HasPrefix(line, "#") {
```
czunker: That would skip
```
#include
```
From the docs:
> For compatibility with sudo versions prior to 1.9.1

We might still come across these versions.

tas50: I added a special case for #include

providers/os/resources/sudoers.go
tas50: Fixed

PR #6464, opened 2026-01-25
⭐ Add a limits resource

tas50:
Another resource to replace some very complex and hard to understand parsing of multiple files in a .d directory.

Old:

```
-      securityLimitFiles = files.find(from: '/etc/security/limits.d/', type: 'file')
-        .list.where(basename == /\.(conf)$/)
-        .map(path)
-
-      securityLimitAllFiles = securityLimitFiles + ["/etc/security/limits.conf"]
-
-      securityLimitAllFiles.map(file(_).content.lines.where( _ == /^[^#]/ )).flat.one(_ == /\*\s+hard\s+core\s+0/)
````

New:

```
mql: limits.entries.contains(type == "hard" && item == "core" && domain == "*" && value == "infinity")
```

---

```coffee
> limits.files
limits.files: [
  0: file path="/etc/security/limits.conf" size=2766 permissions.string="-rw-r--r--"
  1: file path="/etc/security/limits.d/10-coredump-debian.conf" size=182 permissions.string="-rw-r--r--"
]
> limits.entries
limits.entries: [
  0: limits.entry domain="*" item="core"
  1: limits.entry domain="*" item="core"
  2: limits.entry domain="root" item="core"
  3: limits.entry domain="*" item="core"
  4: limits.entry domain="root" item="core"
]
>
```


providers/os/resources/limits.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+
+		// If a custom path is provided, just use that file
+		files, err := getSortedPathFiles(runtime, path)
+		if err != nil {
+			return nil, nil, err
+		}
+		args["files"] = llx.ArrayData(files, types.Resource("file"))
+		delete(args, "path")
+	}
```
czunker: Should we fall back to the default here?

When I run it without `path` I don't get the content:
```
cnquery run container image ubuntu:24.04 -c 'limits.entries{ * }'
limits.entries: []

```

But:
```
cnquery run container image ubuntu:24.04 -c 'limits.files{ path }'
limits.files: [
  0: {
    path: "/etc/security/limits.conf"
  }
]
```

tas50: Fixed

PR #6465, opened 2026-01-25
⭐ Add a modprobe resource

tas50:
Right now we have to write very specific checks that look for modprobe entries in specific file paths. These are really incorrect checks which only work if you follow our remediation steps and use our expected paths. If your system was already secure and you already had blacklist entries for these modules, the checks wouldn't catch those.

Example of how we do it today:

```coffeescript
file("/etc/modprobe.d/hfs.conf").exists &&
file("/etc/modprobe.d/hfs.conf").content.contains("install hfs /bin/false") &&
file("/etc/modprobe.d/hfs.conf").content.contains("blacklist hfs")
```

How this allows us to do it instead which works with any .d file

```coffeescript
modprobe.blacklists.contains(module == "hfs")
modprobe.installs.contains(module == "hfs" && command == "/bin/false")
```

### DIFF
--- a/providers/os/resources/modprobe.go
+++ b/providers/os/resources/modprobe.go
@@ -603,9 +603,10 @@ func parseSoftdeps(runtime *plugin.Runtime, filePath string, content string) ([]
 				continue
 			}
 
-			if currentSection == "pre" {
+			switch currentSection {
+			case "pre":
 				pre = append(pre, part)
-			} else if currentSection == "post" {
+			case "post":
 				post = append(post, part)
 			}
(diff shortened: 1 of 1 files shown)


providers/os/resources/os.lr
```diff
@@ -1260,6 +1260,99 @@ logindefs {
   params(content) map[string]string
 }
 
+// Kernel module configuration from modprobe.d
+modprobe {
```
chris-rock: we should have defaults for this resource

PR #6466, opened 2026-01-26
⭐ Add a new openBSMAudit resource

tas50:
This exposes macOS/FreeBSD audit information:

```
openBSMAudit: {
  file: file path="/etc/security/audit_control" size=378 permissions.string="-rw-------"
  flags: [
    0: "lo"
    1: "ad"
    2: "fm"
    3: "-all"
    4: "fd"
  ]
  naflags: [
    0: "lo"
    1: "aa"
  ]
  dir: "/var/audit"
  superuserSetSflagsMask: [
    0: "has_authenticated"
    1: "has_console_access"
  ]
  path: "/etc/security/audit_control"
  minfree: 5
  filesz: "2M"
  superuserClearSflagsMask: [
    0: "has_authenticated"
    1: "has_console_access"
  ]
  policy: [
    0: "cnt"
    1: "argv"
  ]
  params: {
    dir: "/var/audit"
    expire-after: "60d OR 1024M"
    filesz: "2M"
    flags: "lo,ad,fm,-all,fd"
    member-clear-sflags-mask: "has_authenticated"
    member-set-sflags-mask: ""
    minfree: "5"
    naflags: "lo,aa"
    policy: "cnt,argv"
    superuser-clear-sflags-mask: "has_authenticated,has_console_access"
    superuser-set-sflags-mask: "has_authenticated,has_console_access"
  }
  content: "#
dir:/var/audit
flags:lo,ad,fm,-all,fd
minfree:5
naflags:lo,aa
policy:cnt,argv
filesz:2M
expire-after:60d OR 1024M
superuser-set-sflags-mask:has_authenticated,has_console_access
superuser-clear-sflags-mask:has_authenticated,has_console_access
member-set-sflags-mask:
member-clear-sflags-mask:has_authenticated
"
  memberSetSflagsMask: []
  memberClearSflagsMask: [
    0: "has_authenticated"
  ]
  expireAfter: "60d OR 1024M"
}
```

This allows us to replace an existing query like this:

```
file("/etc/security/audit_control").exists;
["/etc/security/audit_control"].where(file(_).exists) {
  file(_).content.lines.where( _ == /^expire-after/) {
    _.split(":")[1] == /[6-9]\dd|\d{3,}d/ || _.split(":")[1] == /\d+0G|[1-9G]/
  }
}
```

With this much easier to understand query:

```
openBSMAudit.expireAfter == "60d OR 1024M" <--- We could also regex this
if we cared
```

### DIFF
--- a/providers/os/resources/openbsmaudit.go
+++ b/providers/os/resources/openbsmaudit.go
@@ -21,11 +21,7 @@ func initOpenBSMAudit(runtime *plugin.Runtime, args map[string]*llx.RawData) (ma
 	conn := runtime.Connection.(shared.Connection)
 	platform := conn.Asset().Platform
 
-	supported := false
-	if platform.IsFamily("darwin") || platform.Name == "freebsd" {
-		supported = true
-	}
-
+	supported := platform.IsFamily("darwin") || platform.Name == "freebsd"
 	if !supported {
 		return nil, nil, errors.New("openBSMAudit resource is only supported on macOS and FreeBSD")
 	}


providers/os/resources/openbsmaudit.go
```diff
@@ -0,0 +1,258 @@
...
+	conn := runtime.Connection.(shared.Connection)
+	platform := conn.Asset().Platform
+
+	supported := false
+	if platform.IsFamily("darwin") || platform.Name == "freebsd" {
+		supported = true
+	}
+
+	if !supported {
+		return nil, nil, errors.New("openBSMAudit resource is only supported on macOS and FreeBSD")
```
tas50, line 26: Curious if folks think this is a pattern worth using or if it should get removed. We have it in a few places.

chris-rock: I think it makes sense to have it here. Therefore I would consider it a good pattern.

providers/os/resources/openbsmaudit.go
```diff
@@ -0,0 +1,258 @@
...
+
+		f, err := CreateResource(runtime, "file", map[string]*llx.RawData{
+			"path": llx.StringData(path),
+		})
+		if err != nil {
+			return nil, nil, err
+		}
+		args["file"] = llx.ResourceData(f, "file")
+		args["path"] = llx.StringData(path)
+		delete(args, "path")
```
chris-rock: This should be removed.

PR #6475, opened 2026-01-26
⭐ Update recording to store resource ids from GetData calls

preslavgerchev:
When we use recordings to read from them, sometimes we run into the following issue: A resource id stored under an identifier that was computed in the resource's provider and that the recording does not know about. The flow for this is roughly as following:
 1. A request towards `GetData` is made with the resource name + id that the runtime is asking for. 
 2. `GetData` responds with the resource. It might also include a different resource id since this is now the proper (resolved) id of the resource.

This PR stores the originally requested resource id alongside the response from the provider to allow us to do this lookup when we read from recordings later on.

### DIFF
--- a/providers-sdk/v1/recording/recording.go
+++ b/providers-sdk/v1/recording/recording.go
@@ -58,6 +58,7 @@ func NewAssetRecording(asset *inventory.Asset) *Asset {
 		Asset:       ai,
 		connections: map[string]*connection{},
 		resources:   map[string]*Resource{},
+		IdsLookup:   map[string]string{},
 	}
 }
 
@@ -99,7 +100,7 @@ func (n *readOnly) EnsureAsset(asset *inventory.Asset, provider string, connecti
 	}
 }
 
-func (n *readOnly) AddData(connectionID uint32, resource string, id string, field string, data *llx.RawData) {
+func (n *readOnly) AddData(req llx.AddDataReq) {
 }
 
 type RecordingOptions struct {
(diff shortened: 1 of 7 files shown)


providers-sdk/v1/recording/recording.go
```diff
@@ -383,24 +383,32 @@ func (r *recording) EnsureAsset(asset *inventory.Asset, providerID string, conne
...
 	if !exist {
 		obj = &Resource{
-			Resource: resource,
-			ID:       id,
+			Resource: req.Resource,
+			ID:       req.ResourceID,
 			Fields:   map[string]*llx.RawData{},
 		}
-		asset.resources[resource+"\x00"+id] = obj
+		asset.resources[req.Resource+"\x00"+req.ResourceID] = obj
```
preslavgerchev, line 407: this holds the resource type + id so that in theory we can support multiple resources of the same type

providers-sdk/v1/recording/recording.go
```diff
@@ -99,7 +100,7 @@ func (n *readOnly) EnsureAsset(asset *inventory.Asset, provider string, connecti
 	}
 }
 
-func (n *readOnly) AddData(connectionID uint32, resource string, id string, field string, data *llx.RawData) {
+func (n *readOnly) AddData(req llx.AddDataReq) {
```
chris-rock, line 103: Is this a breaking change for existing recordings?

PR #6477, opened 2026-01-26
⭐ Recording provider

preslavgerchev:
* Extract the inventory asset to MQL asset in a function under `providers/core` so we can reuse in the recording.
 * Add tests for the function above
 * Simplify recording's `ensureAssetMetadata` to use the same function that the core provider uses, ensuring we always get the same data. 
 * Add a new recording provider that simply loads a recording and forwards all `GetData` calls to it.
 * Add a new `StaticProvider` interface that, if implemented, prevents the runtime from doing resource/field provider lookups. This is the way we avoid creating new providers when running queries, that are usually meant for other providers

### DIFF
--- a/providers-sdk/v1/recording/recording.go
+++ b/providers-sdk/v1/recording/recording.go
@@ -539,6 +539,5 @@ func ensureAssetMetadata(resources map[string]*Resource, asset *inventory.Asset)
 		resources[id] = existing
 	}
 
-	assetArgs := core.CreateAssetResourceArgs(asset)
-	existing.Fields = assetArgs
+	existing.Fields = core.CreateAssetResourceArgs(asset)
 }


providers/runtime.go
```diff
@@ -672,6 +672,12 @@ func (r *Runtime) lookupResourceProvider(resource string) (*ConnectedProvider, *
 		return provider, info, provider.ConnectionError
 	}
 
+	staticProvider, ok := r.Provider.Instance.Plugin.(plugin.StaticProvider)
```
preslavgerchev, line 675: Im also open to hardcoding the name here but feels wrong to add provider-specific logic here

providers-sdk/v1/recording/recording.go
```diff
@@ -513,30 +539,6 @@ func ensureAssetMetadata(resources map[string]*Resource, asset *inventory.Asset)
...
-	}
-
-	if _, ok := existing.Fields["ids"]; !ok {
-		arr := make([]any, len(asset.PlatformIds))
-		for i := range asset.PlatformIds {
-			arr[i] = asset.PlatformIds[i]
-		}
-		existing.Fields["ids"] = llx.ArrayData(arr, types.String)
-	}
+	assetArgs := core.CreateAssetResourceArgs(asset)
```
arlimus: nit, vars that aren't used we can probably just assign, i'll fix

providers/runtime.go
arlimus: let's keep it this way for now, it's a good starting point

PR #6479, opened 2026-01-26
⭐ Azure auto discovery should include all resources

LittleSalkin1806:
As done for AWS and GCP now also for Azure as i am now using Azure Integration as well.

I have one question:
For AWS and GCP as i sticked to the AWS changes you did. And therfore not added certain resources to auto.

From my understanding it is crucial to make auto == all. As on the integrations used on the mondoo plattform we cant define any options as input for the scans right ?

When i look at `gcp` and `aws` there are some things excluded.
I had this problem just today wanted to check ec2 instances but `ec2-instance-api` was not included into auto behaviour and also not included in `all` but worked perfectly fine on the shell on my laptop.

I also dont understand the difference between `instances` and `ec2-instance-api`

Could we align why this is the case and if it does not has any reasons also include it into the scans?

[AWS PR](https://github.com/mondoohq/cnquery/pull/6397/changes)
[GCP PR](https://github.com/mondoohq/cnquery/pull/6451/changes)

### DIFF
--- a/providers/azure/resources/discovery.go
+++ b/providers/azure/resources/discovery.go
@@ -48,7 +48,21 @@ var All = []string{
 	DiscoveryInstances,
 }
 
-var Auto = []string{DiscoverySubscriptions}
+var Auto = []string{
+	DiscoverySubscriptions,
+	DiscoveryInstances,
+	//DiscoveryInstancesApi,
+	DiscoverySqlServers,
+	DiscoveryPostgresServers,
+	DiscoveryPostgresFlexibleServers,
+	DiscoveryMySqlServers,
+	DiscoveryMySqlFlexibleServers,
+	DiscoveryMariaDbServers,
+	DiscoveryStorageAccounts,
+	DiscoveryStorageContainers,
+	DiscoveryKeyVaults,
+	DiscoverySecurityGroups,
+}
 
 func allDiscovery() []string {
 	return append(All, AllAPIResources...)
--- a/providers/azure/resources/discovery_test.go
+++ b/providers/azure/resources/discovery_test.go
@@ -40,4 +40,8 @@ func TestGetDiscoveryTargets(t *testing.T) {
 	// test random
 	config.Discover.Targets = []string{"postgres-servers", "keyvaults-vaults", "instances"}
 	require.Equal(t, []string{DiscoveryPostgresServers, DiscoveryKeyVaults, DiscoveryInstances}, getDiscoveryTargets(config))
+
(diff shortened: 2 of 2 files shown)


providers/azure/resources/discovery.go
```diff
@@ -48,7 +48,21 @@ var All = []string{
 	DiscoveryInstances,
 }
 
-var Auto = []string{DiscoverySubscriptions}
+var Auto = []string{
+	DiscoverySubscriptions,
+	DiscoveryInstances,
+	//DiscoveryInstancesApi,
```
tas50, line 54: @preslavgerchev can you chime in here on the difference between these two and which things we should default to?

vjeffrey: instances will discover the os - like something you can connect to - the instancesAPI one is about instance configuration. we don't want both on right now for any cloud discovery, it will create double assets

LittleSalkin1806:
Current State:

We are currently writing custom policies and leveraging existing CIS policies. These policies will be enforced centrally through integrations across all major cloud providers:

Azure – Organization-wide
AWS – Organization-wide (likely via Lambda)
GCP – Organization-wide
These integrations scan both our centrally enforced policies and any user-defined policies within individual workspaces.

Previous Behavior (before my changes):

For Azure (which I've already updated for GCP, and you guys the handling for AWS), Mondoo only scanned at the Account / Subscription / Project level by default via scheduled scans.

I kept the structure you first introduced in AWS PR

[AWS PR](https://github.com/mondoohq/cnquery/pull/6397/changes)
[GCP PR](https://github.com/mondoohq/cnquery/pull/6451/changes)

This happens because:

--discover defaults to "auto" when no parameters are specified
--filters regions= defaults to "every active region in the account"
The Requirement:

We need the ability to pass custom inputs (--filters, --discover) to the Mondoo platform scans for these integrations.

Current Blocker:

I received feedback via Arthur Ferdinand (who contacted you directly on Slack) that this capability is currently not available.

This means the only way to see scan results for all assets on the Mondoo platform is through local scanning—there's no way to customize the integration scans yet.

I don't yet have visibility into what's being tracked on the T-Sec / Mondoo feature board, but I should have access soon.

tas50:
@LittleSalkin1806 We've added a good amount of discovery but also walked back doing everything. Some of the discovery causes duplicate assets right now and that's quite confusing to folks. I'm going to close this out at this point. Let me know if there's specific assets other than instances you want to enable discovery for.

PR #6492, opened 2026-01-27
⭐ Expand the data collected on AWS secrets manager secrets

tas50:
- Add last access date
- Add owning service
- Add the lambda function for rotation
- Add a sub-resource for rotation rules

I thought these might be helpful if you're trying to get information on the secrets manager secrets @LittleSalkin1806

### DIFF
--- a/providers/aws/resources/aws_secretsmanager.go
+++ b/providers/aws/resources/aws_secretsmanager.go
@@ -61,7 +61,7 @@ func initAwsSecretsmanagerSecret(runtime *plugin.Runtime, args map[string]*llx.R
 		return nil, nil, errors.New("arn required to fetch secretsmanager secret")
 	}
 
-	obj, err := CreateResource(runtime, "aws.secretsmanager", map[string]*llx.RawData{})
+	obj, err := CreateResource(runtime, ResourceAwsSecretsmanager, map[string]*llx.RawData{})
 	if err != nil {
 		return nil, nil, err
 	}
@@ -129,12 +129,12 @@ func (a *mqlAwsSecretsmanager) getSecrets(conn *connection.AwsConnection) []*job
 
 					// add kms key if there is one
 					if secret.KmsKeyId != nil {
-						mqlKeyResource, err := NewResource(a.MqlRuntime, "aws.kms.key",
+						mqlKeyResource, err := NewResource(a.MqlRuntime, ResourceAwsKmsKey,
 							map[string]*llx.RawData{
 								"arn": llx.StringDataPtr(secret.KmsKeyId),
 							})
 						if err != nil {
-							args["kmsKey"] = llx.NilData
+							args["kmsKey"] = &llx.RawData{Type: types.Resource(ResourceAwsKmsKey), Error: err}
 						} else {
 							mqlKey := mqlKeyResource.(*mqlAwsKmsKey)
 							args["kmsKey"] = llx.ResourceData(mqlKey, mqlKey.MqlName())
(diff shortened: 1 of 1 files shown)


providers/aws/resources/aws.lr
```diff
@@ -1699,14 +1701,30 @@ aws.secretsmanager.secret @defaults("arn name") {
...
   // Date of the next secret rotation
   nextRotationDate time
+  // AWS service that created this secret (e.g., RDS, Redshift)
+  owningService string
   // Primary region of the secret
   primaryRegion string
   // Whether rotation is enabled for the secret
   rotationEnabled bool
+  // Lambda function used for automatic rotation
+  rotationLambda() aws.lambda.function
```
czunker:
```suggestion
  rotationLambda aws.lambda.function
```

I don't think we need these, as the data is already added when the `secret` resource is created.

providers/aws/resources/aws.lr
```diff
@@ -1699,14 +1701,30 @@ aws.secretsmanager.secret @defaults("arn name") {
...
+  // Rotation schedule configuration
+  rotationRules() aws.secretsmanager.secret.rotationRules
```
czunker: Same here

providers/aws/resources/aws_secretsmanager.go
```diff
@@ -86,6 +86,18 @@ func (a *mqlAwsSecretsmanagerSecret) kmsKey() (*mqlAwsKmsKey, error) {
...
+	return a.RotationLambda.Data, nil
+}
+
+func (a *mqlAwsSecretsmanagerSecret) rotationRules() (*mqlAwsSecretsmanagerSecretRotationRules, error) {
+	return a.RotationRules.Data, nil
+}
+
+func (a *mqlAwsSecretsmanagerSecretRotationRules) id() (string, error) {
+	return a.__id, nil
+}
```
czunker: I think the generated code should handle this.

tas50: Removed

providers/aws/resources/aws_secretsmanager.go
```diff
@@ -141,6 +151,45 @@ func (a *mqlAwsSecretsmanager) getSecrets(conn *connection.AwsConnection) []*job
...
 					}
 
+					// add rotation lambda if there is one
+					if secret.RotationLambdaARN != nil {
+						mqlLambdaResource, err := NewResource(a.MqlRuntime, "aws.lambda.function",
+							map[string]*llx.RawData{
+								"arn": llx.StringDataPtr(secret.RotationLambdaARN),
+							})
+						if err != nil {
+							args["rotationLambda"] = llx.NilData
```
preslavgerchev: can we set this to an err field to not ignore the actual err?

providers/aws/resources/aws_secretsmanager.go
```diff
@@ -86,6 +86,14 @@ func (a *mqlAwsSecretsmanagerSecret) kmsKey() (*mqlAwsKmsKey, error) {
 	return a.KmsKey.Data, nil
 }
 
+func (a *mqlAwsSecretsmanagerSecret) rotationLambda() (*mqlAwsLambdaFunction, error) {
+	return a.RotationLambda.Data, nil
```
preslavgerchev: i think these fns are not required given that they are not computed fields in the LR schema

tas50: Removed

providers/aws/resources/aws_secretsmanager.go
tas50: any particular suggestion for how you'd want to handle this? Just logging or more?

PR #6503, opened 2026-01-28
🐛  Fix ms365 nested resource with init func

Bajusz15:
Closes https://github.com/mondoohq/cnquery/issues/6185

Users couldn't use dot notation for nested resources like:
`microsoft.policies.externalIdentitiesPolicy.allowExternalIdentitiesToLeave == true`
`microsoft.policies.authenticationMethodsPolicy.authenticationMethodConfigurations == empty`
But block notation worked:
`microsoft.policies { externalIdentitiesPolicy.allowExternalIdentitiesToLeave == true }`

### Root cause
The MQL compiler's compileResource function resolves resources by extending identifiers:
Starts with microsoft -> extends to microsoft.policies -> Tries to extend to microsoft.policies.externalIdentitiesPolicy

For implicit resources the compiler couldn't instantiate them directly because they had no init() method. Without init(), the compiler treats them as non-instantiable, so dot notation fails. 

There are 3 solutions to make dot notation (and instantiation) possible:

1. Renaming so there is no naming config between method name and resource name
2. Adding init() methods to these resources can make them be instantiated directly.
3. Make the resource non private, this way it doesn't need init method.

### DIFF
--- a/providers/ms365/resources/ms365.lr
+++ b/providers/ms365/resources/ms365.lr
@@ -1287,7 +1287,6 @@ microsoft.security.exchange.antispam {
 
 // Microsoft Security Exchange Antispam Hosted Connection Filter Policy
 microsoft.security.exchange.antispam.hostedConnectionFilterPolicy @defaults("identity enableSafeList") {
-  init(identity? string)
   // Policy identity
   identity string
   // Admin display name for the policy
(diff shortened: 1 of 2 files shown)


Bajusz15:
@syrull Could you check, which of the 3 solutions are preferred? 

After talking to Jay I went with renaming for most of the resources, but, I don't want to break any existing policies we have. Alternatively I can add init method to each of them.

PR #6519, opened 2026-01-30
🐛 make aws.es.domain work

LittleSalkin1806:
refers to this [issue](https://github.com/mondoohq/cnquery/issues/6515)

Todos:
- [ ] Add potential helper function as the code is quite repetetive for many aws services

### DIFF
--- a/providers/aws/resources/aws_es.go
+++ b/providers/aws/resources/aws_es.go
@@ -86,7 +86,7 @@ func (a *mqlAwsEs) getDomains(conn *connection.AwsConnection) []*jobpool.Job {
 }
 
 func initAwsEsDomain(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error) {
-	if len(args) >= 2 {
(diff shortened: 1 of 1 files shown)


LittleSalkin1806:
If you guys could be so kind and do the renaming from es to os @tas50 after reviewing the rest  ?

providers/aws/resources/aws_es.go
```diff
@@ -84,12 +86,43 @@ func (a *mqlAwsEs) getDomains(conn *connection.AwsConnection) []*jobpool.Job {
 }
 
 func initAwsEsDomain(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error) {
-	if len(args) > 2 {
+	if len(args) >= 2 {
```
LittleSalkin1806:
```suggestion
	if len(args) > 2 {
```
this needs of course to be reverted
i tested this and forgot to revert it

tas50:
Thaks @LittleSalkin1806. I'll get this provider updated so you can test this out in your infra

PR #6531, opened 2026-02-02
⭐ V13: Improve recordings, allow looking up resources by asset mrn, platform ids or connection id.

preslavgerchev:
While technically not breaking since we do not change the recording format, we only change the lookup code, I think this is a good thing to consider v13.

This improves recordings asset lookup and mimics providers lookup: We now support either an asset mrn, any of the platform ids or the asset's connection id too. There are use cases where we cannot easily do the connection lookup which was in place until now, but we already have asset mrn or platform ids.

Other improvements:
 * Move read-only impl to its own file.
 * Cleanup `recording.go`. `GetData` and `GetResource` use the same code path underneath. `GetData` only checks the fields too since it supports field fetching.
 * Consistent tests for looking up things by mrn/platform ids/connection id.

### DIFF
--- a/llx/runtime.go
+++ b/llx/runtime.go
@@ -23,6 +23,14 @@ type Runtime interface {
 	AssetUpdated(asset *inventory.Asset)
 }
 
+// Allows looking up data for assets, based on different asset identifiers.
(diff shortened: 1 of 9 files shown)


arlimus (approved):
love this improvement, cleaner lookups in the recordings, thank you!

PR #6533, opened 2026-02-02
S3 staticWebsiteHosting now includes RedirectAllRequestsTo and RoutingRules

### LINKED ISSUE #6505: AWS: S3 staticWebsiteHosting does not catch setups where only a redirect is used
**Describe the bug**

We try to implement a check like this:

This control checks whether an S3 bucket is hosting a static website, which must not be used.
It is not allowed to enable the static web page feature in the S3 bucket settings, because it will publish your S3 bucket objects to the internet.
Motivation: S3 buckets that are only internally used are not allowed to use this feature. The bucket must be treated as an external bucket if the objects are to be publicly accessible (external S3 Bucket requirements for open data).

This should be done based on this MQL query, which returns an empty set:

```
> aws.s3.buckets.where(name == "rr-test-mondoo").first.staticWebsiteHosting
aws.s3.buckets.where.first.staticWebsiteHosting: {}
```

Config looks like this:

```yml
{
    "ErrorDocument": null,
    "IndexDocument": null,
    "RedirectAllRequestsTo": {
      "HostName": "www.telekom.com",
      "Protocol": "https"
    },
    "RoutingRules": null,
    "ResultMetadata": {}
  }
```

The issue seems to be here:
https://github.com/mondoohq/cnquery/blob/338b2dc54a2fd204aa534cb20c01145148398803/providers/aws/resources/aws_s3.go#L706

Comment by @LittleSalkin1806 who found this:
They only return something in the code when "hosting a website" is ticked but they do not account for "redirect requests for an object". If hosting is set "errordocument / indexdocument" will be set but as only redirect is set nothing nil will be returned.

**To Reproduce**

- Setup a S3 bucket with "redirect requests for an object" but without "hosting a website"
- Execute the MQL query given above

**Expected behavior**

- We expect the MQL output to include the bucket where redirection is setup and staticWebsiteHosting should be included in the output.

anurag-2911:
Fixes #6505

The staticWebsiteHosting field was returning an empty map when an S3 bucket was configured with only RedirectAllRequestsTo (redirect requests for an object) without IndexDocument/ErrorDocument. This made it impossible to detect that static website hosting was enabled on such buckets.

Now includes:
- RedirectAllRequestsTo.HostName and RedirectAllRequestsTo.Protocol
- RoutingRules indicator when routing rules are configured

### DIFF
--- a/providers/aws/resources/aws_s3.go
+++ b/providers/aws/resources/aws_s3.go
@@ -711,6 +711,17 @@ func (a *mqlAwsS3Bucket) staticWebsiteHosting() (map[string]any, error) {
 		if website.IndexDocument != nil {
 			res["IndexDocument"] = convert.ToValue(website.IndexDocument.Suffix)
 		}
+
+		if website.RedirectAllRequestsTo != nil {
+			res["RedirectAllRequestsTo.HostName"] = convert.ToValue(website.RedirectAllRequestsTo.HostName)
+			if website.RedirectAllRequestsTo.Protocol != "" {
+				res["RedirectAllRequestsTo.Protocol"] = string(website.RedirectAllRequestsTo.Protocol)
+			}
+		}
+
+		if len(website.RoutingRules) > 0 {
+			res["RoutingRules"] = "configured"
+		}
 	}
 
 	return res, nil


providers/aws/resources/aws_s3.go
```diff
@@ -711,6 +711,17 @@ func (a *mqlAwsS3Bucket) staticWebsiteHosting() (map[string]any, error) {
...
+		if website.RedirectAllRequestsTo != nil {
+			res["RedirectAllRequestsTo.HostName"] = convert.ToValue(website.RedirectAllRequestsTo.HostName)
+			if website.RedirectAllRequestsTo.Protocol != "" {
+				res["RedirectAllRequestsTo.Protocol"] = string(website.RedirectAllRequestsTo.Protocol)
+			}
+		}
+
+		if len(website.RoutingRules) > 0 {
+			res["RoutingRules"] = "configured"
+		}
```
LittleSalkin1806, line 724: This would not show the rules details.
we would loose quite some visibility.

https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3/types#Redirect

https://docs.aws.amazon.com/AmazonS3/latest/userguide/how-to-page-redirect.html#advanced-conditional-redirects

We return  [map[string]string] but routing Rules are important to get checked.

LittleSalkin1806: If this is the way  in theory
we have v13 for breaking changes :)

PR #6534, opened 2026-02-02
🐛 aws-secret-manager-edgecases-enhancement

LittleSalkin1806:
KMS Keys API calls use KeyID = support multiple formats aliases, UUID(ID) or arn format.

I tried to stick as much as possible to the current code structure with out refactoring  but let KMS Key handle normalization.

When we look at types
KmsKeyId is already an ARN from the AWS API for secetmanager and probably most of the others.

But as you can see from the screenshot eventbridge connection and potentially others have edge cases.
types.secretsmanager states that only arn should be rerturned but  UUId was returned.

image

I could replicate this by creating event bridge connection using the basic auth and then selecting a key
IMPORTANT!  kms key arn is returned if i select an alias or kmskeyarn but if i just search and select kmskey id then only the UUID is returned by go sdsk

image

I did some normalization on kms side to handle such cases as we dont always know if this case will happen or not for the responding service where the key is linked.

Todos:
- []: I also found some policy limiations. Either for the same account or cross acccount
I would still need this information and distungish on mql level.
Therefore i added some todo  with the idear `accessible: bool`
Every Key which exist and we dont have access return accessible: false

This should probably be an another issue.

### DIFF
--- a/.github/workflows/spell-check.yaml
+++ b/.github/workflows/spell-check.yaml
@@ -28,7 +28,7 @@ jobs:
     steps:
       - name: check-spelling
         id: spelling
-        uses: check-spelling/check-spelling@c635c2f3f714eec2fcf27b643a1919b9a811ef2e # v0.0.25
+        uses: check-spelling/check-spelling@cfb6f7e75bbfc89c71eaa30366d0c166f1bd9c8c # v0.0.26
         with:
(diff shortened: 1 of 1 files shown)


tas50:
I'll see if we can get some subject matter experts on this one for you @LittleSalkin1806

PR #6567, opened 2026-02-04
🧹 Improve the resource docs

tas50:
Comment updates to improve the generated docs


VasilSirakov (approved):
We could start to enforce backticks in some places instead of single quotes depending on how we want it to look in the README.

PR #6573, opened 2026-02-05
🧹 add generic report error function

vjeffrey:


providers-sdk/v1/upstream/health/errors.go
```diff
@@ -75,6 +75,47 @@ func sendPanic(product, version, build string, r any, stacktrace []byte) {
...
 }
 
+func ReportError(product, version, build, err string) {
+	reportError(product, version, build, err)
+}
+
+// reportError sends an error to the mondoo platform for further analysis if the
+// service account is configured.
+// This function does not return an error as it is not critical to send the error to the platform.
+func reportError(product, version, build string, errMsg string) {
```
preslavgerchev, line 85: can't we just make this pubic and drop the other one?

vjeffrey: ya i was just following the pattern 🤷 we totally can

PR #6586, opened 2026-02-06
:star: Add new fields / resources to api gateway / autoscaling

tas50:
Add new fields to `aws.autoscaling.group` and
`aws.applicationAutoscaling.target` including autoscaling policies and scheduled actions.

Add new fields to `aws.apigateway.stage` and `aws.apigateway.restapi`


arlimus (approved):
more data!

PR #6587, opened 2026-02-06
⭐ Add fargateEphemeralStorageKmsKey to aws.ecs.cluster

tas50:
Expose the fargate storage KMS key. Also fix up the defaults for the service so the shell output is more useful


arlimus (approved):
thank you!

PR #6591, opened 2026-02-06
:star: Add additional Linux distros + fix detection logic

tas50:
Make sure we properly return true on gentoo and bottlerocket. Add addition common distros on the desktop

Simplify Gentoo OS detection. It's a rolling release and everyone has a simple /etc/os-release file at this point so we can rely on that.

### DIFF
--- a/providers/os/detector/detector_all.go
+++ b/providers/os/detector/detector_all.go
@@ -588,28 +588,26 @@ var bottlerocket = &PlatformResolver{
 			return false, nil
 		}
 
-		content := strings.TrimSpace(string(c))
-		osr, err := ParseOsRelease(content)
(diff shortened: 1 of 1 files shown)


providers/os/detector/detector_all.go
```diff
@@ -556,7 +589,7 @@ var bottlerocket = &PlatformResolver{
 			pf.Build = osr["BUILD_ID"]
 		}
 
-		return false, nil
+		return true, nil
```
tas50: Curious here how this worked without a clear return true before

tas50: Refactored to make this logic better match the other platforms

PR #6606, opened 2026-02-09
⭐ Add new GCP resource for secrets

tas50:
Expose secrets resource

```
gcp.project.secretmanager.secrets
gcp.project.secretmanager.secrets { name labels replication created expireTime }
gcp.project.secretmanager.secrets { versions { name state created } }
gcp.project.secretmanager.secrets { iamPolicy { role members } }
```

### DIFF
--- a/providers/gcp/resources/gcp.lr
+++ b/providers/gcp/resources/gcp.lr
@@ -3265,7 +3265,7 @@ private gcp.project.secretmanagerService {
 }
 
 // Google Cloud (GCP) Secret Manager secret
-private gcp.project.secretmanagerService.secret @defaults("name created") {
+private gcp.project.secretmanagerService.secret @defaults("name createTime") {
   // Project ID
   projectId string
   // Full resource path (projects/*/secrets/*)
@@ -3273,7 +3273,7 @@ private gcp.project.secretmanagerService.secret @defaults("name created") {
   // Secret name
   name string
   // Time the secret was created
-  created time
+  createTime time
   // User-assigned labels
   labels map[string]string
   // Replication policy configuration
@@ -3287,7 +3287,7 @@ private gcp.project.secretmanagerService.secret @defaults("name created") {
   // Rotation policy configuration
   rotation dict
   // Version aliases mapping alias names to version numbers
-  versionAliases dict
+  versionAliases map[string]int
   // Custom metadata annotations
   annotations map[string]string
   // Version destroy TTL duration
--- a/providers/gcp/resources/gcp.lr.go
+++ b/providers/gcp/resources/gcp.lr.go
@@ -4802,8 +4802,8 @@ var getDataFields = map[string]func(r plugin.Resource) *plugin.DataRes{
 	"gcp.project.secretmanagerService.secret.name": func(r plugin.Resource) *plugin.DataRes {
 		return (r.(*mqlGcpProjectSecretmanagerServiceSecret).GetName()).ToDataRes(types.String)
 	},
-	"gcp.project.secretmanagerService.secret.created": func(r plugin.Resource) *plugin.DataRes {
-		return (r.(*mqlGcpProjectSecretmanagerServiceSecret).GetCreated()).ToDataRes(types.Time)
+	"gcp.project.secretmanagerService.secret.createTime": func(r plugin.Resource) *plugin.DataRes {
+		return (r.(*mqlGcpProjectSecretmanagerServiceSecret).GetCreateTime()).ToDataRes(types.Time)
 	},
 	"gcp.project.secretmanagerService.secret.labels": func(r plugin.Resource) *plugin.DataRes {
 		return (r.(*mqlGcpProjectSecretmanagerServiceSecret).GetLabels()).ToDataRes(types.Map(types.String, types.String))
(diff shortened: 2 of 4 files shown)


tas50:
@LittleSalkin1806 does the change I made make sense?

LittleSalkin1806:
Hey @tas50 i would say fields returned should differ based on the google data structure.

1. There is still a open scenario.
We for example don´t allow automatic replication by an organizational policy. So we need to set the location for our CMK´s
Each replication can have a dedicated key.
So the replication information needs to have the `customerManagedEncryption` information

```text
Secret: "my-secret"
Replication: USER_MANAGED
  Replica 1: location=us-east1,  kmsKey=projects/p/locations/us-east1/keyRings/kr/cryptoKeys/key-a
  Replica 2: location=eu-west1,  kmsKey=projects/p/locations/eu-west1/keyRings/kr/cryptoKeys/key-b
  Replica 3: location=asia-east1, kmsKey=projects/p/locations/asia-east1/keyRings/kr/cryptoKeys/key-c
```

`customerManagedEncryption` is a string and returned but inside the replication dict depending on the replication type
[the Rest Endpoint](https://docs.cloud.google.com/secret-manager/docs/reference/rest/v1/projects.secrets) shows the structure quite good.


here you see the example where we CANT use automatic replication type but one with google managed key one with custom keys (each can be different)



 
image

LittleSalkin1806:
The current idear is a nice quick hack for easy mql queries is this secret encrypted !

Maybe this can be kept if it retrieves a list of kms keys :)

But this is a design choice. If you want to keep data struct of provider or add custom fields to make queries simpler :D

providers/gcp/resources/gcp.lr
```diff
@@ -3037,3 +3039,69 @@ private gcp.project.binaryAuthorizationControl.admissionRule {
...
+  // Pub/Sub topics for event notifications
+  topics []string
+  // Expiration time (if set)
+  expireTime time
+  // Etag of the secret
+  etag string
+  // Rotation policy configuration
+  rotation dict
+  // Version aliases mapping alias names to version numbers
+  versionAliases dict
```
LittleSalkin1806:
```suggestion
  versionAliases map[string]int
```
if you want to be precise :D for documentation which can be importent for mql checks
needs to be changed potentially in types of secretsmanager.go as well

providers/gcp/resources/gcp.lr
```diff
@@ -3037,3 +3039,69 @@ private gcp.project.binaryAuthorizationControl.admissionRule {
   // The resource names of the attestors that must attest to a container image
   requireAttestationsBy []string
 }
+
+// Google Cloud (GCP) Secret Manager
+private gcp.project.secretmanagerService {
+  // Project ID
+  projectId string
```
arlimus, line 3262: do we need this? we can solve linking back to the parent as a general mql thing, I don't think this is necessary unless I'm missing something? ditto below

LittleSalkin1806: The projectId field follows the existing convention used by all other GCP resources in this providers
Removing it would make it an outliner.
But you are right, functionality would be the same just compile errors as the projectId could be used from the connection directly.

In addition the code uses secretsmanagerService projectId in two or three ocassions.
lines 29,30,32,70,73  needs to be changed where projectID of SecretsmanagerService is referenced
initGcpProjectSecretmanagerService then also dont need to set projectID into args anymore.
lines 48,61,143  set an arg the schema doesn't recognize

```text
resources/secretmanager.go:29:7: g.ProjectId undefined (type *mqlGcpProjectSecretmanagerService has no field or method ProjectId)
resources/secretmanager.go:30:16: g.ProjectId undefined (type *mqlGcpProjectSecretmanagerService has no field or method ProjectId)
resources/secretmanager.go:32:17: g.ProjectId undefined (type *mqlGcpProjectSecretmanagerService has no field or method ProjectId)
resources/secretmanager.go:68:7: g.ProjectId undefined (type *mqlGcpProjectSecretmanagerService has no field or method ProjectId)
resources/secretmanager.go:69:17: g.ProjectId undefined (type *mqlGcpProjectSecretmanagerService has no field or method ProjectId)
resources/secretmanager.go:71:17: g.ProjectId undefined (type *mqlGcpProjectSecretmanagerService has no field or method ProjectId)
```

LittleSalkin1806: I am not so deep into your code but i would suggest to rethink this properly and change everything at once if needed :)

Your feature idear is valid and a more mature approach as the runtime already knows the hierarchy already

LittleSalkin1806:
I will properly review this tomorrow or the day after. But right now its time to sleep ;)

providers/gcp/resources/gcp.lr
LittleSalkin1806: I suppose you mean somethign like this ?

```yaml
gcp.project.secretmanager.secrets[0].parent         -> secretmanagerService resource
gcp.project.secretmanager.secrets[0].parent.parent  -> gcp.project resource
gcp.project.secretmanager.secrets[0].project       _> gcp.project (shortcut)
```

arlimus: yeah, I just noticed that we do this everywhere, sorry, that's on me!!
I'll get it merged as is

PR #6617, opened 2026-02-12
🧹 make sure windows is detected as os family

imilchev:
The current OS provider detects windows as its own thing (not an os) which causes issues with asset filters.

image

With this change windows is now properly detected as os:
image

### DIFF
--- a/providers/os/detector/detector_all.go
+++ b/providers/os/detector/detector_all.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"io"
 	"regexp"
+	"runtime"
 	"strings"
 
 	"github.com/rs/zerolog/log"
@@ -1315,7 +1316,8 @@ var unixFamily = &PlatformResolver{
 	Detect: func(r *PlatformResolver, pf *inventory.Platform, conn shared.Connection) (bool, error) {
 		// in order to support linux container image detection, we cannot run
 		// processes here, lets just read files to detect a system
-		return true, nil
+		// We don't want to run unix detection on local windows connections
+		return conn.Type() != shared.Type_Local || runtime.GOOS != "windows", nil
 	},
 }
 


providers/os/detector/detector.go
```diff
@@ -4,20 +4,12 @@
...
-	"runtime"
-
 	"go.mondoo.com/cnquery/v12/providers-sdk/v1/inventory"
 	"go.mondoo.com/cnquery/v12/providers/os/connection/shared"
 )
 
 func DetectOS(conn shared.Connection) (*inventory.Platform, bool) {
-	var res *inventory.Platform
-	var ok bool
-	if conn.Type() == shared.Type_Local && runtime.GOOS == "windows" {
```
chris-rock, line 16: We should not remove that since it will help us to not run unix detection on a local windows machine

providers/os/detector/detector.go
```diff
@@ -4,20 +4,12 @@
...
-		res, ok = WindowsFamily.Resolve(conn)
-	} else {
-		res, ok = OperatingSystems.Resolve(conn)
-	}
+	res, ok := OperatingSystems.Resolve(conn)
```
imilchev, line 12: the `OperatingSystems` family includes `WindowsFamily` so I wouldn't expect any other changes than adding the os family to windows https://github.com/mondoohq/cnquery/blob/main/providers/os/detector/detector_all.go#L1486

providers/os/detector/detector.go
imilchev: I added it to the unix detection explicitly. That should make things a little more transparent

chris-rock: Great improvement.

providers/os/detector/detector.go
chris-rock: Makes sense.

chris-rock (approved):
Thank you @imilchev

PR #6675, opened 2026-02-24
🐛 Fix GCP network/subnetwork discovery not working + several crashes

tas50:
A few different issues here. 
- Fix some crashes I experienced trying to list out networks + a few other places where we weren't doing nil checks
- Get discovery of networks/subnetworks working
- Update the deps to the latest
- Don't make the individual resources private since we have asset platforms for these

### DIFF
--- a/providers/gcp/go.mod
+++ b/providers/gcp/go.mod
@@ -92,7 +92,7 @@ require (
 	github.com/olekukonko/errors v1.2.0 // indirect
 	github.com/olekukonko/ll v0.1.7 // indirect
 	github.com/olekukonko/tablewriter v1.1.3 // indirect
-	github.com/package-url/packageurl-go v0.1.4 // indirect
+	github.com/package-url/packageurl-go v0.1.3 // indirect
 	github.com/pandatix/go-cvss v0.6.2 // indirect
 	github.com/patrickmn/go-cache v2.1.0+incompatible // indirect
 	github.com/planetscale/vtprotobuf v0.6.1-0.20240319094008-0393e58bdf10 // indirect
--- a/providers/gcp/go.sum
+++ b/providers/gcp/go.sum
@@ -773,8 +773,8 @@ github.com/opencontainers/go-digest v1.0.0 h1:apOUWs51W5PlhuyGyz9FCeeBIOUDA/6nW8
 github.com/opencontainers/go-digest v1.0.0/go.mod h1:0JzlMkj0TRzQZfJkVvzbP0HBR3IKzErnv2BNG4W4MAM=
 github.com/opencontainers/image-spec v1.1.1 h1:y0fUlFfIZhPF1W537XOLg0/fcx6zcHCJwooC2xJA040=
 github.com/opencontainers/image-spec v1.1.1/go.mod h1:qpqAh3Dmcf36wStyyWU+kCeDgrGnAve2nCC8+7h8Q0M=
-github.com/package-url/packageurl-go v0.1.4 h1:RHfiiN1SSY+Kic537DXch6fy593rxGJW6WDzAiOwNdk=
-github.com/package-url/packageurl-go v0.1.4/go.mod h1:nKAWB8E6uk1MHqiS/lQb9pYBGH2+mdJ2PJc2s50dQY0=
+github.com/package-url/packageurl-go v0.1.3 h1:4juMED3hHiz0set3Vq3KeQ75KD1avthoXLtmE3I0PLs=
+github.com/package-url/packageurl-go v0.1.3/go.mod h1:nKAWB8E6uk1MHqiS/lQb9pYBGH2+mdJ2PJc2s50dQY0=
 github.com/pandatix/go-cvss v0.6.2 h1:TFiHlzUkT67s6UkelHmK6s1INKVUG7nlKYiWWDTITGI=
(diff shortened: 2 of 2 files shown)


providers/gcp/resources/compute.go
```diff
@@ -1363,39 +1403,35 @@ func initGcpProjectComputeServiceSubnetwork(runtime *plugin.Runtime, args map[st
...
 	if err != nil {
 		return nil, nil, err
 	}
-	computeSvc := obj.(*mqlGcpProjectComputeService)
-	subnetworks := computeSvc.GetSubnetworks()
-	if subnetworks.Error != nil {
-		return nil, nil, subnetworks.Error
+
+	ctx := context.Background()
+	subnetSvc, err := computev1.NewSubnetworksRESTClient(ctx, option.WithHTTPClient(client))
```
imilchev, line 1644: this doesn't look right. Before we were loading the subnetworks from the MQL cache. Now we are going to be using the GCP API every time. I think we should use this as a fallback, not always

providers/gcp/resources/iam.go
```diff
@@ -52,6 +52,18 @@ func initGcpProjectIamServiceServiceAccount(runtime *plugin.Runtime, args map[st
 		return args, nil, nil
 	}
 
+	if args["projectId"] == nil {
```
imilchev: i think this may be hiding some diff issue and it's not fixing the real problem

providers/gcp/resources/compute.go
tas50: Addressed in d0464ff6b — the init functions now try the MQL cache first (original pattern) and only fall back to a direct GCP API call if the resource isn't found in cache. Applied the same cache-first + API fallback pattern to all three init functions (instance, network, subnetwork).

providers/gcp/resources/iam.go
tas50: Removed this change — the commit (bd845c1bd) that added the nil check has been dropped from the branch entirely.

providers/gcp/resources/compute.go
```diff
@@ -1395,7 +1498,37 @@ func initGcpProjectComputeServiceSubnetwork(runtime *plugin.Runtime, args map[st
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	gcpSubnetwork, err := subnetSvc.Get(ctx, &computepb.GetSubnetworkRequest{
+		Project:    projectId,
+		Region:     region,
+		Subnetwork: subnetworkName,
+	})
```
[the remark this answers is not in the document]
tas50: Not an issue — `region` is defined earlier in the function (line 1453) before both the cache lookup and this fallback path. It is populated either from `getAssetIdentifier` or from `args["regionUrl"]`.

providers/gcp/resources/compute.go
```diff
@@ -1262,7 +1313,58 @@ func initGcpProjectComputeServiceNetwork(runtime *plugin.Runtime, args map[strin
 			return args, network, nil
 		}
 	}
-	return nil, nil, errors.New("network not found")
+
+	// Fallback: fetch directly from the GCP API
+	networkName := args["name"].Value.(string)
+	projectId := args["projectId"].Value.(string)
+
+	conn := runtime.Connection.(*connection.GcpConnection)
```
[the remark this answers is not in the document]
tas50: This follows the established pattern across all GCP init functions. The `len(args) == 0` branch populates these from `getAssetIdentifier` (or returns an error). When `len(args) > 0`, the MQL framework always provides the args defined in the `.lr` schema, so `name` and `projectId` will be present.

providers/gcp/resources/compute.go
```diff
@@ -1395,7 +1498,37 @@ func initGcpProjectComputeServiceSubnetwork(runtime *plugin.Runtime, args map[st
 			return args, subnetwork, nil
 		}
 	}
-	return nil, nil, errors.New("subnetwork not found")
+
+	// Fallback: fetch directly from the GCP API
+	subnetworkName := args["name"].Value.(string)
+	projectId := args["projectId"].Value.(string)
+
```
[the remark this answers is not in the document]
tas50: Same as above — this follows the established pattern across all GCP init functions. Args are always populated by either `getAssetIdentifier` or the MQL framework.

providers/gcp/resources/compute.go
```diff
@@ -363,7 +364,49 @@ func initGcpProjectComputeServiceInstance(runtime *plugin.Runtime, args map[stri
...
+		"name":        llx.StringData(zoneData.Name),
+		"description": llx.StringData(zoneData.Description),
+		"status":      llx.StringData(zoneData.Status),
+		"created":     llx.TimeDataPtr(parseTime(zoneData.CreationTimestamp)),
+	})
+	if err != nil {
+		return nil, nil, err
+	}
+
+	gcpInstance, err := gcpComputeSvc.Instances.Get(projectId, zoneName, instanceName).Do()
```
imilchev: this also feels wrong. On L343 we call `GetInstances()`. That should list all instances from GCP, cache them as mql resources and then find the one you are looking for in the list and return it. Now we are unnecessarily doing a singe Get as well. I think if we cannot find the instance above, then this call here would always fail too

PR #6679, opened 2026-02-25
:bug: Fix ELB classic load balancer field alignment and filtering

LittleSalkin1806:
## Summary

This PR fixes several issues with classic (v1) ELB load balancers so they work the same way as v2 (ALB/NLB) load balancers.

### What was wrong

- **Missing fields on classic LBs**: Querying `availabilityZones`, `securityGroups`, `hostedZoneId`, or `vpc` on a classic load balancer would fail with a type conversion error because those fields were never set during creation.
- **Duplicate results**: Every load balancer (classic and v2) appeared twice in query results due to being appended to the list before and after the tag filter check.
- **Crashes with tag filters**: When using `--filters "tag:..."`, looking up security groups or VPCs could fail if those resources didn't match the tag filter, breaking the entire query.
- **Unnecessary API calls**: The `targetGroups` field on classic LBs tried to call the v2 API, which doesn't recognize classic ARNs. Now it returns an empty list immediately.
- **Init function called for wrong resources**: The ELB init function was being triggered for non-ELB resources (like S3 buckets or IAM users), wasting API calls. Now it exits early if the ARN isn't an ELB.
- **Discovery region**: The discovery code was parsing the region from the ARN instead of using the field already on the resource.

### Changes

- **`aws_elb.go`**: Added missing fields to classic LBs, removed duplicate appends, made SG/VPC resolution graceful, added early exit for non-ELB ARNs in init, added classic LB guard on targetGroups.
- **`discovery.go`**: Simplified region handling to use `f.Region.Data` directly.

### DIFF
--- a/providers/aws/config/config.go
+++ b/providers/aws/config/config.go
@@ -132,7 +132,7 @@ Notes:
 					Long:    "filters",
 					Type:    plugin.FlagType_KeyValue,
 					Default: "",
-					Desc:    "Filter options, e.g., --filters regions=us-east-2,us-east-1 , --filters ec2:instance-ids=i-093439483935",
+					Desc:    "Filter options, e.g., --filters regions=us-east-2,us-east-1 --filters ec2:instance-ids=i-093439483935 --filters tag:Environment=production --filters tag:Team=platform --filters exclude:tag:Environment=dev",
 				},
 			},
 		},


chris-rock (approved):
Thank you @LittleSalkin1806

PR #6686, opened 2026-02-26
✨ Add Intune device ID detection to Windows clients

czunker:
- [x] rebase and rerun tests after https://github.com/mondoohq/cnquery/pull/6687 merged

This finds the Intune device ID for Windows client systems and adds it to the platform metadata.


AdamVB:
@vjeffrey @czunker can we get this into the product metrics somehow? See how many of our customer compute assets have an intune ID?

PR #6781, opened 2026-03-03
:star: Add nftables resource

tas50:
## Summary
- Add `nftables` resource for querying nftables firewall tables, chains, and rules via `nft -j list ruleset`
- Accessible through `os.linux.nftables` for Linux systems
- Includes `nftables.table`, `nftables.chain`, and `nftables.rule` resources with full structured data

## Test plan
- [x] Unit tests for JSON parsing (tables, chains, rules, flags, empty/invalid input)
- [x] Interactive verification with `mql shell local -c "os.linux.nftables.tables"` on a Linux host with nftables

### DIFF
--- a/providers/os/resources/nftables.go
+++ b/providers/os/resources/nftables.go
@@ -11,6 +11,7 @@ import (
 
 	"go.mondoo.com/mql/v13/llx"
 	"go.mondoo.com/mql/v13/providers-sdk/v1/plugin"
+	"go.mondoo.com/mql/v13/providers/os/connection/shared"
 	"go.mondoo.com/mql/v13/types"
 )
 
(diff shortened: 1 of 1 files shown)


providers/os/resources/nftables.go
```diff
@@ -0,0 +1,231 @@
...
+			continue
+		}
+		if chain != "" && r.Chain != chain {
+			continue
+		}
+
+		exprDicts := make([]any, len(r.Expr))
+		for i, e := range r.Expr {
+			exprDicts[i] = e
+		}
```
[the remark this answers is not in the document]
tas50: Fixed in baf6f77 — simplified to `copy(exprDicts, r.Expr)`.

providers/os/resources/nftables.go
```diff
@@ -0,0 +1,261 @@
...
+
+func (n *mqlNftables) tables() ([]any, error) {
+	o, err := CreateResource(n.MqlRuntime, "command", map[string]*llx.RawData{
+		"command": llx.StringData("nft -j list ruleset"),
+	})
+	if err != nil {
+		return nil, err
+	}
+	cmd := o.(*mqlCommand)
+	if exit := cmd.GetExitcode(); exit.Data != 0 {
```
czunker, line 157: A connection might not be able to execute commands.

In other resources, we have a safeguard for that case. I think we should add one here, too.

tas50: @czunker I put in a fix for this. Let me know if that's what you were thinking

czunker: 👍

PR #6782, opened 2026-03-03
:star: Add apache2 resource

tas50:
## Summary
- Add `apache2` and `apache2.conf` resources for querying Apache HTTP Server configuration
- Parses Apache config files including directives, `LoadModule`, `VirtualHost`, and `Directory` blocks
- Supports `Include`/`IncludeOptional` with glob expansion for fragmented configs

## Test plan
- [x] Unit tests for config parsing (directives, modules, virtual hosts, directories, includes, comments, continuation lines, globs)
- [x] Interactive verification with `mql shell local -c "apache2.conf.params"` on a host with Apache installed

### DIFF
--- a/providers/os/resources/apache2.go
+++ b/providers/os/resources/apache2.go
@@ -107,24 +107,20 @@ type mqlApache2ConfInternal struct {
 	serverRoot string
 }
 
-// apacheConfPaths maps platform names to their default Apache config location.
-// Debian/Ubuntu use apache2, RHEL/CentOS use httpd.
-var apacheConfPaths = map[string]string{
-	"ubuntu":   "/etc/apache2/apache2.conf",
-	"debian":   "/etc/apache2/apache2.conf",
-	"redhat":   "/etc/httpd/conf/httpd.conf",
-	"centos":   "/etc/httpd/conf/httpd.conf",
-	"fedora":   "/etc/httpd/conf/httpd.conf",
-	"rocky":    "/etc/httpd/conf/httpd.conf",
-	"alma":     "/etc/httpd/conf/httpd.conf",
-	"oracle":   "/etc/httpd/conf/httpd.conf",
-	"amazon":   "/etc/httpd/conf/httpd.conf",
-	"suse":     "/etc/apache2/httpd.conf",
-	"opensuse": "/etc/apache2/httpd.conf",
-	"freebsd":  "/usr/local/etc/apache24/httpd.conf",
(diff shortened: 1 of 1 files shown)


providers/os/resources/apache2.go
```diff
@@ -0,0 +1,423 @@
...
+
+// apacheVersionBinaries lists commands to try for getting the Apache version.
+// Debian/Ubuntu use apache2ctl, RHEL/CentOS use httpd, others may use apachectl.
+var apacheVersionBinaries = []string{"apache2ctl", "httpd", "apachectl"}
+
+func (s *mqlApache2) version() (string, error) {
+	conn := s.MqlRuntime.Connection.(shared.Connection)
+
+	for _, bin := range apacheVersionBinaries {
+		cmd, err := conn.RunCommand(bin + " -v")
```
czunker, line 64: Is there a way to read a file or something?

We might not always be able to run commands, and file lookup is preferred.

providers/os/resources/apache2.go
```diff
@@ -0,0 +1,423 @@
...
+	if file.Error != nil {
+		return "", file.Error
+	}
+
+	return file.Data.Path.Data, nil
+}
+
+func (s *mqlApache2Conf) file() (*mqlFile, error) {
+	conn := s.MqlRuntime.Connection.(shared.Connection)
+	path := apacheConfPath(conn)
```
czunker: When I read this correctly, I think it skips the path set via the init function.

providers/os/resources/apache2.go
tas50: Addressed — version detection now prefers file-based approach: reads the httpd binary directly from well-known paths (`/usr/sbin/apache2`, `/usr/sbin/httpd`, etc.) and extracts the embedded `Apache/x.y.z` version string. RunCommand is kept only as a fallback for non-standard install paths.

providers/os/resources/apache2.go
tas50: The `file()` getter only runs when `apache2.conf` is created without a path argument — it's the default fallback. When a path IS provided, `initApache2Conf` converts it to a file resource and sets `args["file"]` directly, so the framework stores that value and never calls this getter. Same pattern as `sshd.config`. Added a clarifying comment.

providers/os/resources/apache2.go
tas50: I added some very complex logic to avoid shelling out. If folks are ok with that we can use this pattern in other places.

PR #6785, opened 2026-03-04
Fix integration tests failing with go test -cover

czunker:
## Summary
- Strip `GOCOVERDIR` from the environment passed to child build commands in integration test setup
- When `go test -cover` runs, Go sets `GOCOVERDIR` which is inherited by child processes (`go build`, `go run`, `make`). These aren't built with `-cover` and fail at exit when trying to write coverage data
- Fixes the `program not built with -cover` error in `test/commands` and `test/providers`

## Test plan
- [ ] CI `go-test-integration` job passes with `-cover` flag
- [x] Normal `make providers/build/os` still works without changes

### DIFF
--- a/test/cli_test_runner.go
+++ b/test/cli_test_runner.go
@@ -7,7 +7,9 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"os"
 	"os/exec"
+	"strings"
 )
 
 type Runner interface {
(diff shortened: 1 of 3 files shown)


czunker:
One example: https://github.com/mondoohq/cnquery/actions/runs/22656767020/job/65668124281?pr=6690

PR #6824, opened 2026-03-06
✨ Add Windows 10 ESU detection

czunker:
## Summary

- Detect Windows 10 Extended Security Updates (ESU) enrollment and set platform metadata `windows.mondoo.com/support-type: esu`
- Covers both **subscription-based ESU** (registry key `Win10CommercialW365ESUEligible`) and **MAK-activated ESU** (WMI `SoftwareLicensingProduct`)
- Only runs for Windows 10 22H2 (build 19045), the only version eligible for ESU per [Microsoft docs](https://learn.microsoft.com/en-us/windows/whats-new/enable-extended-security-updates-virtual)
- LTSC editions are correctly excluded (different build numbers)
- Follows the same pattern as hotpatch and Intune detection (native Windows registry + WMI fast path, PowerShell fallback for remote connections)

Fixes https://github.com/mondoo-community/customer-issues/issues/89

## Test plan

- [ ] Build and install the OS provider, run against a Windows 10 22H2 machine with ESU subscription enabled — verify `windows.mondoo.com/support-type: esu` appears in platform metadata
- [ ] Run against a Windows 10 22H2 machine with MAK-activated ESU — verify same metadata
- [ ] Run against a Windows 10 22H2 machine without ESU — verify no `support-type` metadata is set
- [ ] Run against a Windows 11 machine — verify ESU check is skipped entirely
- [ ] Run against an older Windows 10 version (e.g. 21H2) — verify ESU check is skipped

### DIFF
--- a/providers/os/detector/windows/esu_unix.go
+++ b/providers/os/detector/windows/esu_unix.go
@@ -1,8 +1,7 @@
 // Copyright (c) Mondoo, Inc.
 // SPDX-License-Identifier: BUSL-1.1
 
-//go:build linux || darwin || netbsd || openbsd || freebsd
-// +build linux darwin netbsd openbsd freebsd
+//go:build !windows
 
 package windows
 
--- a/providers/os/detector/windows/esu_windows.go
+++ b/providers/os/detector/windows/esu_windows.go
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: BUSL-1.1
 
 //go:build windows
-// +build windows
 
 package windows
 


providers/os/detector/detector_win.go
```diff
@@ -226,6 +201,32 @@ func detectIntuneDeviceID(pf *inventory.Platform, conn shared.Connection) {
...
+// ESU is only relevant for Windows 10 (build < 22000).
+func detectESU(pf *inventory.Platform, conn shared.Connection) {
+	buildNumber, err := strconv.Atoi(pf.Version)
+	if err != nil {
+		return
+	}
+	// ESU is only for Windows 10 (build < 22000)
+	if buildNumber >= 22000 {
+		return
+	}
```
czunker: We can add another safe guard for versions before Windows 10.
Some old Windows server might still be used.

PR #6839, opened 2026-03-07
✨ os: add ipv4, ipv6, primaryIPv4, primaryIPv6 to network resource

tas50:
Add convenience fields to the network resource for quickly accessing IPv4/IPv6 addresses and determining primary IPs from default routes.

This completes the work we had outlined in: https://github.com/mondoohq/planning/issues/849

### DIFF
--- a/providers/os/resources/os.lr
+++ b/providers/os/resources/os.lr
@@ -2381,16 +2381,16 @@ network {
   routes() networkRoutes
 
   // All IPv4 addresses detected on host interfaces
-  ipv4() []ipAddress
+  ipv4() []ip
 
   // All IPv6 addresses detected on host interfaces
-  ipv6() []ipAddress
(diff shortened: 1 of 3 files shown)


providers/os/resources/os.lr
```diff
@@ -2376,9 +2376,21 @@ ipAddress @defaults("ip") {
...
+  ipv4() []ipAddress
+
+  // All IPv6 addresses detected on host interfaces
+  ipv6() []ipAddress
+
+  // Primary IPv4 address determined by the default route
+  primaryIPv4() ipAddress
+
+  // Primary IPv6 address determined by the default route
+  primaryIPv6() ipAddress
```
arlimus: let's use the native `ip` type in these places

tas50: Updated

PR #6865, opened 2026-03-09
🐛 Fix GitHub Actions workflow bugs and typos

tas50:
## Summary
- **reusable-lint-providers.yml**: `BASE_SHA` was never set, so the `only-new-issues` input (used by `pr-extended-linting.yml`) silently did nothing. Now passes `github.event.pull_request.base.sha` so `--new-from-rev` actually filters to new issues.
- **providers.yaml**: Quoted `${{ steps.providers.outputs.* }}` in the debug echo to prevent shell word-splitting of JSON arrays. Also fixed "select" → "selected" typo.
- **Typos**: "Provider PAth" → "Provider Path" (pr-test-lint.yml), "becasue" → "because" (release-providers.yml)

## Test plan
- [ ] Verify `pr-extended-linting` now shows only new lint issues on a PR with known lint changes
- [ ] Confirm provider build scoping debug output is properly formatted

### DIFF
--- a/.github/workflows/pr-test-lint.yml
+++ b/.github/workflows/pr-test-lint.yml
@@ -11,15 +11,6 @@ on:
       - ".github/workflows/pr-test-lint.yml"
       - ".github/workflows/reusable-lint-providers.yml"
       - "**.toml" # run tests when any recording changed
-  pull_request:
-    paths:
-      - "**.go"
-      - "**.mod"
-      - "go.sum"
-      - "Makefile"
-      - ".github/workflows/pr-test-lint.yml"
-      - ".github/workflows/reusable-lint-providers.yml"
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-test-lint.yml
```diff
@@ -11,6 +11,15 @@ on:
       - ".github/workflows/pr-test-lint.yml"
       - ".github/workflows/reusable-lint-providers.yml"
       - "**.toml" # run tests when any recording changed
+  pull_request:
```
philipbalinov: this would allow running the workflow against forked PR - i'd be hesitant to do it as-is; at the very least, we have steps that run on our self-hosted runners, and allowing 3rd party code to execute against those is probably not a good idea

PR #7214, opened 2026-04-15
✨ Propagate AWS account tags to discovered assets

vjeffrey:
## Summary

- During AWS discovery, every asset discovered from the primary AWS account now inherits the account's tags as labels.
- Per-asset labels win on key collisions — account tags only fill gaps.
- Sibling account assets emitted by the organization discovery target are skipped (they belong to other accounts).

Account tags are fetched once via \`awsAccount.GetTags()\` at the start of \`Discover()\` and applied in a post-processing merge pass at the end. The primary account asset itself also receives its tags as labels. If fetching account tags fails (e.g. missing IAM permission), a warning is logged and discovery proceeds with no account-level propagation.

Spec: \`docs/superpowers/specs/2026-04-15-aws-account-tags-propagation-design.md\`
Plan: \`docs/superpowers/plans/2026-04-15-aws-account-tags-propagation.md\`

## Test plan

- [x] \`go test ./providers/aws/resources/...\` — all pass
- [x] \`go vet ./providers/aws/resources/...\` — clean
- [x] \`gofmt\` — clean
- [x] Manual verification against a live AWS account with account-level tags: \`mql run aws --discover accounts,ec2-instances -c "asset { name labels }"\` — confirm account tags appear on both the account asset and EC2 instance assets, with per-asset labels preserved on collision

### DIFF
--- a/docs/superpowers/plans/2026-04-15-aws-account-tags-propagation.md
+++ b/docs/superpowers/plans/2026-04-15-aws-account-tags-propagation.md
@@ -1,0 +1,543 @@
+# AWS Account Tag Propagation Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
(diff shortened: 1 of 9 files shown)


vjeffrey:
note: permission required for the tags to be applied - organization:ListTagsForResource - if permission does not exist, we log the permission error and skip the the tag propagation

PR #7217, opened 2026-04-15
🐛 Fix CloudTrail cross-account scan errors for S3 and CloudWatch

tas50:
## Summary
- When scanning organizational CloudTrail trails from a member account (`--discover cloudtrail-trails`), cross-account references to S3 buckets and CloudWatch log groups caused errors because those resources live in the management account and can't be discovered/accessed from the member account.
- **CloudWatch log group**: `initAwsCloudwatchLoggroup` now detects cross-account ARNs and creates a placeholder resource with fields extracted from the ARN (name, region, etc.) instead of returning `"cloudwatch log group does not exist"`. This follows the same pattern used for cross-account KMS keys.
- **S3 bucket**: Added `exists` guards to `location()`, `logging()`, `fetchPublicAccessBlock()`, and `public()` so they return nil/empty gracefully for placeholder buckets (`exists=false`) instead of attempting S3 API calls that fail with `PermanentRedirect` due to unknown region.

## Test plan
- [x] Build the AWS provider: `make providers/build/aws && make providers/install/aws`
- [x] Scan an organizational CloudTrail trail from a member account: `cnspec scan aws --discover cloudtrail-trails`
- [x] Verify no `PermanentRedirect` errors for S3 bucket operations
- [x] Verify no `"cloudwatch log group does not exist"` errors
- [x] Verify the cross-account CloudWatch log group warning appears in logs
- [x] Verify `aws.cloudtrail.trail.logGroup != empty` check passes for trails with log groups configured
- [x] Run existing tests: `go test ./providers/aws/resources/ -run "TestParseLogGroupArn|TestS3"`

### DIFF
--- a/providers/aws/resources/aws.permissions.json
+++ b/providers/aws/resources/aws.permissions.json
@@ -1,7 +1,7 @@
 {
   "provider": "aws",
-  "version": "13.12.0",
-  "generated_at": "2026-04-13T09:58:50-07:00",
+  "version": "13.13.1",
+  "generated_at": "2026-04-15T11:18:18-07:00",
   "permissions": [
     "access-analyzer:ListAnalyzers",
     "access-analyzer:ListFindingsV2",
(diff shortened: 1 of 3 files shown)


tas50:
This resolves failures we were seeing when scanning our own account

vjeffrey (approved):

PR #7241, opened 2026-04-16
🐛 Gracefully handle SageMaker regions where service is unavailable

tas50:
## Summary
- Add `IsServiceNotAvailableInRegionError(err)` alongside the existing `Is400AccessDeniedError(err)` checks in every SageMaker list/describe path across `aws_sagemaker_{hub,infra,labeling,lineage,mlops}.go`.
- Regions that don't offer a given SageMaker surface now log a warning and skip, matching the existing behavior for access-denied errors, instead of failing the whole query.

## Test plan
- [ ] `mql shell aws` against an account with regions where SageMaker is partially available (e.g. opt-in regions) and confirm listing resources (hubs, apps, images, labeling jobs, lineage artifacts, experiments, etc.) no longer errors out.
- [ ] `make providers/build/aws && make providers/install/aws`
- [ ] `make test/lint`


tas50:
Discovered via a full integration test run of all sagemaker resources

PR #7260, opened 2026-04-17
✨ Pin goreleaser edge to v2.15.3 and add auto-update workflow

philipbalinov:
## Summary
- Pin edge workflow goreleaser from `latest` to `v2.15.3` to prevent supply-chain risk from unpinned tool versions
- Add `# auto-update: source=github-releases repo=goreleaser/goreleaser updates=minor` annotation so version bumps are detected automatically (skips major bumps)
- Add `update-pinned-versions.yml` caller workflow (runs weekly) that references the reusable workflow in mondoohq/actions

**Note:** The stable workflow (`goreleaser.yml`) stays pinned at `v2.5.1` without an auto-update annotation due to the RPM signing bug with `rpm 4.17.0` on Ubuntu 22.04 runners.

## How it works
1. Annotate any version string in a workflow file:
   ```yaml
   # auto-update: source=github-releases repo=goreleaser/goreleaser updates=minor
   version: v2.15.3
   ```
2. The weekly workflow scans for these annotations, checks for newer releases, and opens a PR
3. `updates=minor` means only minor and patch bumps are proposed (major bumps are skipped)

## Dependencies
- mondoohq/actions#137 (merged)

## Test plan
- [ ] Verify edge workflow still builds containers successfully with v2.15.3
- [ ] Manually trigger `Update pinned versions` workflow to validate

### DIFF
--- a/.github/workflows/goreleaser-edge.yml
+++ b/.github/workflows/goreleaser-edge.yml
@@ -67,7 +67,8 @@ jobs:
         uses: goreleaser/goreleaser-action@ec59f474b9834571250b370d4735c50f8e2d1e29 # v7.0.0
         with:
           distribution: goreleaser
-          version: latest
+          # auto-update: source=github-releases repo=goreleaser/goreleaser updates=minor
+          version: v2.15.3
           args: release -f .github/.goreleaser-edge.yml --clean --timeout 120m
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/update-pinned-versions.yml
+++ b/.github/workflows/update-pinned-versions.yml
@@ -1,0 +1,18 @@
+name: Update pinned versions
+
+on:
+  schedule:
+    - cron: "22 8 * * 1" # every monday
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  update-versions:
+    uses: mondoohq/actions/.github/workflows/update-pinned-versions.yml@main
+    permissions:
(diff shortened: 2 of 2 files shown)


philipbalinov (commented):
Security review notes:

1. **`@main` ref for reusable workflow** — `mondoohq/actions/.github/workflows/update-pinned-versions.yml@main` is referenced by branch, not SHA. This is a trade-off: pinning by SHA would require updating every time any workflow in `mondoohq/actions` changes, but `@main` means a compromise of that repo's main branch would affect this workflow. Acceptable since:
   - `mondoohq/actions` is an internal org repo with restricted write access
   - This workflow only runs on a weekly schedule or manual dispatch, not on PRs
   - Blast radius is limited to `contents: write` and the SSH deploy key

   If the org wants stricter pinning later, this should be updated to a SHA ref.

2. **SSH deploy key passed to reusable workflow** — `CNQUERY_DEPLOY_KEY_PRIV` is forwarded as a secret. This is consistent with `update-deps.yaml` which does the same. The key is needed for `peter-evans/create-pull-request` to push branches that trigger other workflows (GITHUB_TOKEN-created PRs don't trigger CI).

3. **goreleaser pin looks good** — `latest` → `v2.15.3` eliminates the supply-chain risk of auto-adopting untested releases. The `auto-update` annotation ensures this version won't go stale silently.

philipbalinov:
Dependency updated: mondoohq/actions#136 was closed and replaced with mondoohq/actions#137 (same code, fresh PR to clear stale bot review state).

PR #7329, opened 2026-04-21
🐛 Replace null byte separator with \x1e in recording package

preslavgerchev:
## Summary
- Postgres does not allow null bytes (`\x00`) in text/JSON columns, which breaks when recording data is stored upstream
- Replaced the composite map key separator with ASCII Record Separator (`\x1e`) throughout the recording package
- Added a `keySep` constant so the separator is defined in one place

## Test plan
- [x] All existing recording package tests pass (`go test ./providers-sdk/v1/recording/ -v`)
- [ ] Verify upstream recording storage works with Postgres


czunker (approved):
LGTM

Thanks @preslavgerchev

PR #7364, opened 2026-04-24
🐛 guard against nil asset.Platform in CreateAssetResourceArgs

tas50:
## Summary
Any query that references the \`asset\` resource against a GCP connection crashes mql with a nil pointer dereference:

\`\`\`
...
go.mondoo.com/mql/v13/providers-sdk/v1/recording.CreateAssetResourceArgs
    providers-sdk/v1/recording/recording.go:463 +0x30
go.mondoo.com/mql/v13/providers/core/provider.(*Service).Connect
    providers/core/provider/provider.go:67 +0x8c
\`\`\`

The panic is at \`mapx.Merge(asset.Platform.Labels, asset.Labels)\` — \`asset.Platform\` is nil when the core provider's \`Connect\` runs before the connecting provider has populated platform metadata. \`CreateAssetResourceArgs\` then dereferences the nil pointer in 9 places (\`platform\`, \`kind\`, \`runtime\`, \`version\`, \`arch\`, \`title\`, \`family\`, \`build\`, \`platformMetadata\`, plus the \`Labels\` merge).

Reproduces today against any GCP project with:
\`\`\`
mql run gcp project <id> -c 'asset.name'
mql run gcp project <id> --discover alloydb-clusters -c 'asset.name'
\`\`\`
Found while verifying PRs #7356 / #7355 / #7351 — none of those PRs caused this, it's pre-existing.

## Fix
Fall back to a zero-value \`inventory.Platform\` when \`asset.Platform == nil\` so the asset-level fields we *do* know (\`name\`, \`ids\`, \`labels\`, \`annotations\`, \`fqdn\`) still populate and platform-derived fields return their empty zero values instead of panicking.

## Test plan
- [x] New unit test \`TestCreateAssetResourceArgs/nil_platform_does_not_panic\` covers the nil path; existing cases still pass.
- [x] Rebuilt mql locally (\`make mql/install\`) and confirmed \`mql run gcp project <id> -c 'asset.name'\` and \`mql run gcp project <id> -c 'asset { name platform kind runtime title }'\` both return data instead of panicking.
- [x] Confirmed discovery path (\`--discover alloydb-clusters\`) no longer panics.
- [x] \`go test ./providers-sdk/v1/recording/...\` passes.

### DIFF
--- a/providers-sdk/v1/recording/asset_resource_args_test.go
+++ b/providers-sdk/v1/recording/asset_resource_args_test.go
@@ -103,6 +103,37 @@ func TestCreateAssetResourceArgs(t *testing.T) {
 		assert.Equal(t, types.Map(types.String, types.String), result["labels"].Type)
 	})
 
+	t.Run("nil platform does not panic", func(t *testing.T) {
+		asset := &inventory.Asset{
+			Name:        "no-platform-asset",
+			PlatformIds: []string{"platform-id-1"},
+			Labels:      map[string]string{"env": "prod"},
+			Platform:    nil,
+		}
+
+		result := CreateAssetResourceArgs(asset)
+		require.NotNil(t, result)
+
+		// Platform-derived fields should be their zero values.
+		assert.Equal(t, "", result["platform"].Value)
+		assert.Equal(t, "", result["kind"].Value)
+		assert.Equal(t, "", result["runtime"].Value)
+		assert.Equal(t, "", result["version"].Value)
+		assert.Equal(t, "", result["arch"].Value)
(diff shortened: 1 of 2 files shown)


tas50:
Closing as obsolete: PR #7652 (`🐛 Prevent panics on conversions for bare types & accessing kind on nil platform`) already fixed the same nil-Platform panic in `CreateAssetResourceArgs`, using nil-safe `asset.GetPlatform().Get*()` getters throughout. `PrettyTitle()` is also nil-safe. Closing this PR; the bug is fixed on main.

PR #7371, opened 2026-04-25
🧹 Bump Azure SDK majors and add update-azure-deps skill

tas50:
## Summary

- **`armcontainerregistry`** v1.2.0 → v3.0.0 (skipping v2 to land on latest stable major)
- **`armoperationalinsights`** v1.2.0 → v2.0.2
- New `update-azure-deps` skill at `.claude/skills/update-azure-deps/` so future SDK bumps are token-efficient

## Breaking changes addressed

`operationalinsights` v2 changed `WorkspaceProperties.CreatedDate` and `ModifiedDate` from `*string` to `*time.Time`. Swapped `parseAzureDateString(...)` for `llx.TimeDataPtr(...)` at the two affected call sites in `loganalytics.go`. Compile-clean otherwise.

`containerregistry` v1→v3 (combined) breaking changes — Registry Tasks moved to `armcontainerregistrytasks`, `SystemData.LastModifiedByType` renamed, `RegistryNameCheckRequest`/`RegistryNameStatus`/`RegistryProperties` field removals — none touch our usage. `WorkspaceSKU.CapacityReservationLevel` changed from `*CapacityReservationLevel` (int32-backed) to `*int32` — the existing `int64(*v)` cast still works.

Audited via `Deprecated:` grep on the new SDKs; no deprecated identifiers we use.

## New skill

`probe.py` classifies every Azure dep in a `go.mod` as **UP** / **BETA** / **OK** against `proxy.golang.org`, defaulting to stable-only. `changelog.py` prints CHANGELOG sections between two versions so breaking changes surface before the build runs. `SKILL.md` captures the silent breaking-change patterns (date pointer-type swaps, LRO operation flips, subpackage extractions) that compile cleanly but break behavior.

## Test plan

- [x] `go build ./...` in `providers/azure/`
- [x] `go vet ./...`
- [x] `go test ./resources/...`
- [x] `gofmt -l` on changed files (clean)
- [x] `go mod tidy` clean
- [x] Skill smoke-tested end-to-end against `providers/azure/go.mod` (45 deps classified correctly)


tas50:
Both review points addressed:

- **probe.py dead code** — fixed in 2814f06b7. The `consecutive_missing` counter was incremented to 1 then immediately compared `>= 1`, so the `break` always fired on the first miss and the `continue` was unreachable. Replaced with a plain `break`, which is what the comment said the code was supposed to do.

- **DataExport date fields in loganalytics.go:317** — verified against the v2 SDK source: `DataExportProperties.CreatedDate` and `LastModifiedDate` are still `*string` (only `WorkspaceProperties` got the `*time.Time` change). No code change required; the existing `parseAzureDateString` call is correct, and the build/vet/tests already passed on this PR.

Issue #7385, opened 2026-04-26
🧹 Cloudflare: migrate provider to cloudflare-go v4 (or v6)


tas50:
## Background

The `cloudflare` provider currently uses [`github.com/cloudflare/cloudflare-go`](https://github.com/cloudflare/cloudflare-go) **v0.116.0** — the original handcrafted SDK. Cloudflare has since published auto-generated (Stainless-based) successors:

| Major | Latest | Released |
|---|---|---|
| v0 | v0.116.0 | 2025-09-05 ← current |
| v2 | v2.4.0 | 2024-07-16 |
| v3 | v3.1.0 | 2024-09-24 |
| v4 | v4.6.0 | 2025-07-10 |
| v5 | v5.1.0 | 2025-08-14 |
| v6 | v6.10.0 | 2026-04-23 |

v0 is in maintenance mode; v3+ are auto-generated from Cloudflare's OpenAPI spec and reflect the current API surface (including newer Zero Trust, R2, Workers, and Email Routing endpoints we may want to expose later).

## Goal

Migrate the cloudflare provider to a modern major (target **v4.6.0** as a stable middle ground, or v6.x if it has stabilized by the time this work is picked up) without changing the MQL `.lr` schema or observable query behavior. This is a pure internal refactor.

## Why now

- v0 is no longer the recommended client; new endpoints are added to v3+ first
- Auto-generated SDKs reduce ongoing maintenance (request/response types track the API automatically)
- Avoids stacking divergence as we add more Cloudflare resources (e.g. Zero Trust expansion deferred from #7384)

## Scope

- **No** changes to `providers/cloudflare/resources/cloudflare.lr` or `.lr.versions`
- **No** changes to the MQL query surface — same fields, same return values
- **Yes** rewrites of every cloudflare-go call site across `providers/cloudflare/resources/*.go` and `providers/cloudflare/connection/*.go`
- **Yes** updates to `providers/cloudflare/go.mod`

### Inventory of cloudflare-go methods called today (29)

Account/Identity: `Accounts`, `ListAccountRoles`, `GetAccessOrganization`, `ListAccessApplications`, `ListAccessGroups`, `ListAccessIdentityProviders`, `ListAccessPolicies`, `ListAccessServiceTokens`
Zone/DNS: `ListZones`, `ListDNSRecords`, `ZoneSettings`, `ListPageRules`
[...]

PR #7424, opened 2026-04-28
🧹 Cloudflare: introduce v6 client + migrate cloudflare.go (#7385 phase 1)

tas50:
## Summary

Phase 1 of the cloudflare-go v0 → v6 migration tracked in [#7385](https://github.com/mondoohq/mql/issues/7385). Doing the whole migration in one PR is large (44 call sites, ~22 files) and risky for a "no observable behavior change" refactor. This PR establishes the foundation; subsequent phase PRs migrate one batch of resource files at a time, each shippable with the existing 48-test regression net (#7419, #7423) green.

### `connection.go` — both clients side-by-side
```go
type CloudflareConnection struct {
    Cf       *cloudflare.Client    // v6 — new code uses this
    LegacyCf *v0cloudflare.API     // v0 — kept until last file migrates
    ...
}
```
Both are constructed from the same API token. `LegacyCf` is dropped in the final phase PR.

### `cloudflare.go` — migrated to v6
- `cloudflare.zones()` → `client.Zones.ListAutoPaging(ctx, zones.ZoneListParams{})`
- `cloudflare.accounts()` → `client.Accounts.ListAutoPaging(ctx, accounts.AccountListParams{})`
- All other resource methods (`zone.dns()`, `zone.wafRules()`, `zone.r2()`, `one.accessGroups()`, etc.) still use `LegacyCf` — bulk-renamed mechanically with `sed`. No behaviour change for those paths.

### Test harness
- `helper_test.go` constructs both v0 and v6 clients pointed at the same `httptest.Server`. Existing tests (using `LegacyCf`) work unchanged; migrated tests use the v6 client.
- `TestZones` + `TestAccounts` updated for v6 reality (see below).

## Behaviour notes — v6 differences worth flagging

1. **Pagination contract** — v6's `V4PagePaginationArrayAutoPager` keeps fetching pages until the result array is empty; it ignores `result_info.total_pages`. The test mocks now return `[]` for any `?page=N` where `N > 1` so the loop terminates. Real Cloudflare returns the same shape, so this is a test-fixture concern only.
2. **`zone.owner.email` is always empty under v6** — Cloudflare's OpenAPI spec doesn't expose `email` on the zone-level Owner struct, so v6's `zones.ZoneOwner` has no Email field. The MQL schema field is preserved (no `.lr` change), but it always populates `""`. In practice the v0 SDK rarely returned a real value here either, so this is a near-no-op for real responses; the test fixture is updated accordingly.
3. **`zone.account.type` is always empty under v6** — same reason: `zones.ZoneAccount` only has `ID` and `Name`. Field preserved, populates `""`.
4. **`acc.Settings.EnforceTwofactor`** (lowercase 'f' in v6) vs `EnforceTwoFactor` (v0). Just a Go field-name difference; same JSON.

## What's NOT in this PR

These files still reference `conn.LegacyCf.<method>(...)` and will migrate in follow-up phase PRs:

`accounts.go`, `api_tokens.go`, `certificates.go`, `custom_hostnames.go`, `devices.go`, `dns.go`, `dnssec.go`, `email_routing.go`, `firewall.go`, `gateway.go`, `logpush.go`, `mtls.go`, `one.go`, `r2.go`, `rate_limits.go`, `settings.go`, `stream.go`, `tunnels.go`, `waf.go`, `workers.go`.

## Test plan
- [x] `go test ./providers/cloudflare/...` — all 48 tests green (existing tests use LegacyCf, TestZones+TestAccounts use the new v6 path)
- [x] `go vet ./providers/cloudflare/...` clean
- [x] `gofmt -w` on all changed files
- [x] No changes to `cloudflare.lr` or `.lr.versions`
- [ ] Manual verification with a real Cloudflare account (deferred to reviewer with a test token; `cloudflare.zones { ... }` and `cloudflare.accounts { ... }` are the two paths to spot-check)

## Phasing note for #7385

This is the recommended approach over a single mega-PR because:
1. Each phase ships green-tests
2. Smaller diffs are reviewable
3. If a v6 field-name regression slips through, blast radius is one batch of files instead of all 22
4. Phases can land on different days as reviewers have time

### DIFF
--- a/providers/cloudflare/connection/connection.go
+++ b/providers/cloudflare/connection/connection.go
@@ -6,7 +6,9 @@ import (
 	"errors"
 	"os"
 
-	"github.com/cloudflare/cloudflare-go"
+	v0cloudflare "github.com/cloudflare/cloudflare-go"
+	cloudflare "github.com/cloudflare/cloudflare-go/v6"
+	"github.com/cloudflare/cloudflare-go/v6/option"
 	"go.mondoo.com/mql/v13/providers-sdk/v1/inventory"
 	"go.mondoo.com/mql/v13/providers-sdk/v1/plugin"
 )
@@ -20,7 +22,13 @@ type CloudflareConnection struct {
 	Conf  *inventory.Config
 	asset *inventory.Asset
 
-	Cf *cloudflare.API
+	// Cf is the cloudflare-go v6 client. New resource code should use this.
+	Cf *cloudflare.Client
+	// LegacyCf is the cloudflare-go v0 client kept alongside Cf during the
+	// staged v0→v6 migration tracked in #7385. Resource files that haven't
+	// been migrated yet use this; once a file is migrated, its usages move
+	// to Cf. The field will be dropped once the last migration PR lands.
+	LegacyCf *v0cloudflare.API
 }
 
 func NewCloudflareConnection(id uint32, asset *inventory.Asset, conf *inventory.Config) (*CloudflareConnection, error) {
@@ -30,7 +38,6 @@ func NewCloudflareConnection(id uint32, asset *inventory.Asset, conf *inventory.
 		asset:      asset,
 	}
 
-	// initialize your connection here
 	token := conf.Options[OPTION_API_TOKEN]
 	if token == "" {
 		token = os.Getenv("CLOUDFLARE_TOKEN")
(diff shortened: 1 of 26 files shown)


tas50:
Strategy change — instead of staged migration now, we'll ship a deprecation PR for the v6-incompatible fields (owner.email, account.type, the entire zone.plan resource), let users see the deprecation warnings ahead of time, and run the actual migration after the next major release.

The branch will stay around for reference; closing this PR. New deprecation PR coming next.

PR #7498, opened 2026-05-05
✨ Add new openstack provider

tas50:
## Summary

Adds a new mql provider for OpenStack, built on `gophercloud/v2`. v0 covers seven OpenStack services and 34 resources — broad enough to write meaningful audits across IAM, compute, networking, block storage, images, secrets, and load balancers in a single project.

## Auth

Single-project scope. Auth is resolved from (highest precedence first):
1. CLI flags (`--auth-url`, `--username`, `--password`, `--project-name`, `--user-domain-name`, `--project-domain-name`, `--region`, ...)
2. `--cloud <name>` selecting a `clouds.yaml` entry
3. `OS_*` environment variables

Keystone v3 application credentials are also supported via `--application-credential-id` / `--application-credential-secret` (mutually exclusive with username/password).

The asset platform is `openstack-project` (the connection is mandatory-scoped to a single project), with `Family: ["openstack"]` so cross-provider rules still match.

### Identity — Keystone v3 (4)
- `openstack.project`
- `openstack.user`
- `openstack.role`
- `openstack.domain`

### Compute — Nova v2 (4)
- `openstack.compute.server`
- `openstack.compute.flavor`
- `openstack.compute.keypair`
- `openstack.compute.serverGroup`

### Image — Glance v2 (1)
- `openstack.image`

### Networking — Neutron v2 (13)

Core Neutron:
- `openstack.network`
- `openstack.subnet`
- `openstack.router`
- `openstack.port`
- `openstack.floatingIp`
- `openstack.securityGroup` (embedded `openstack.securityGroup.rule`)

Neutron extensions:
- `openstack.subnetPool`
- `openstack.qosPolicy` (rules embedded as `[]dict`)
- `openstack.trunk`

FWaaS v2:
- `openstack.firewall.group`
- `openstack.firewall.policy`
- `openstack.firewall.rule`

### Block Storage — Cinder v3 (2)
- `openstack.blockstorage.volume`
- `openstack.blockstorage.snapshot`

### Key Manager — Barbican v1 (3)
- `openstack.keymanager.secret` (raw `payload` is intentionally not exposed)
- `openstack.keymanager.container`
- `openstack.keymanager.order`

### Load Balancer — Octavia v2 (7)
- `openstack.octavia.loadBalancer`
- `openstack.octavia.listener`
- `openstack.octavia.pool` (embedded `openstack.octavia.member`)
- `openstack.octavia.member`
- `openstack.octavia.healthMonitor`
- `openstack.octavia.l7Policy` (embedded `openstack.octavia.l7Rule`)
- `openstack.octavia.l7Rule`

## Cross-references

Resources expose typed accessors instead of raw ID fields wherever a target is modeled. Selected highlights:

- `server.{flavor,image,keypair,securityGroups,volumes}()`
- `subnet.{network,subnetPool}()`, `port.{network,securityGroups}()`
- `floatingIp.{floatingNetwork,port,router}()`, `router.externalNetwork()`
- `securityGroup.rule.{securityGroup,remoteGroup}()`
- `project.{domain,parent}()`, `user.{domain,defaultProject,roles}()`, `role.domain()`
- `serverGroup.memberServers()`
- `volume.{project,user,sourceVolume,sourceSnapshot,servers}()`
- `snapshot.{volume,project,user}()`
- `image.owner()`
- `keymanager.container.{secrets,certificate,privateKey,intermediates}()`, `keymanager.order.{secret,container}()`
- `octavia.loadBalancer.{vipPort,vipNetwork,vipSubnet,listeners,pools}()`
- `octavia.listener.{loadBalancer,defaultPool,defaultTlsContainer,sniContainers,clientCATlsContainer,clientCRLContainer,l7Policies}()`
- `octavia.pool.{loadBalancer,listeners,healthMonitor,subnet,caTlsContainer,crlContainer,clientTlsContainer}()`
- `octavia.l7Policy.{listener,redirectPool}()`, `octavia.l7Rule.l7Policy()`
- `firewall.group.{ports,ingressPolicy,egressPolicy}()`, `firewall.policy.rules()`, `firewall.rule.policies()`
- `trunk.{parentPort,subportPorts}()`

`user.roles()` calls Keystone `/role_assignments?user.id=…&effective=true` and deduplicates role IDs across project / domain / system scopes (group-derived assignments included).

`octavia.listener.defaultTlsContainer()` (and the SNI / pool TLS accessors) parse the Barbican ref URL and resolve to `openstack.keymanager.container` when the ref points at a container; null when the ref points at a bare secret.

## Notable resource fields

- **`user.passwordExpiresAt`** — null for federated/passwordless users.
- **`user.ignoreLockoutFailureAttempts`** — Keystone option for users exempt from the lockout policy.
- **`project.isDomain`** — projects acting as a Keystone domain.
- **`subnet.{dnsPublishFixedIp,ipv6AddressMode,ipv6RaMode}`** — DNS leak surface and IPv6 posture (slaac, dhcpv6-stateful, dhcpv6-stateless).
- **`securityGroup.stateful`** — stateless groups don't track connection state.
- **`port.propagateUplinkStatus`** — uplink-status propagation flag.
- **`network|subnet|router|port|floatingIp|securityGroup|securityGroup.rule.{createdAt,updatedAt}`** — drift / forensics timestamps.
- **`volume.{encrypted,multiAttach,bootable,volumeType,replicationStatus,attachments}`** — encryption-at-rest audits, multi-attach surface, image-source forensics.
- **`snapshot.{progress,consumesQuota,groupSnapshotId}`** — snapshot lifecycle and quota tracking.
- **`image.{visibility,protected,hidden,checksum,properties}`** — `properties` carries Glance signing fields (`img_signature`, `img_signature_hash_method`, `img_signature_key_type`, `img_signature_certificate_uuid`).
- **`keymanager.secret.{algorithm,bitLength,mode,expiresAt}`** — cert/key expiry and algorithm-strength audits.
- **`octavia.listener.{tlsCiphers,tlsVersions,alpnProtocols,clientAuthentication,allowedCidrs,hstsIncludeSubdomains}`** — TLS posture; the headline LBaaS audit surface.
- **`octavia.pool.{tlsEnabled,tlsCiphers,tlsVersions,alpnProtocols,persistence}`** — backend TLS and session-stickiness.
- **`octavia.healthMonitor.{type,delay,timeout,maxRetries,httpMethod,urlPath,expectedCodes}`** — uptime probe configuration.
- **`firewall.policy.audited`** — whether an admin has reviewed the policy.
- **`firewall.rule.{action,enabled,sourceIpAddress,destinationIpAddress,sourcePort,destinationPort}`** — full rule introspection.
- **`qosPolicy.rules`** — heterogeneous rule list (`bandwidth_limit`, `dscp_marking`, `minimum_bandwidth`, `packet_rate_limit`).
- **`subnetPool.{prefixes,defaultPrefixLen,minPrefixLen,maxPrefixLen,shared,isDefault}`** — IP-allocation policy.
- **`trunk.{subports,parentPort}`** — trunk-port topology (multi-VLAN attachment surface).

## Behavioral notes

- 401 / 403 / 404 from gophercloud are translated to empty results (mirrors AWS `Is400AccessDeniedError`), so a non-admin token doing `openstack.users` or hitting an unavailable service like Octavia/Barbican doesn't fail the whole query — it just yields nothing.
- Nova reports server security groups by name; Neutron is the source of truth for IDs. Cross-resolution uses a process-wide Neutron group cache so we list once.
- Network and Port lists pull provider, external, MTU, and port-security extensions in a single call via `Extract*Into`.
- Provider segmentation IDs are normalized to int64 (Neutron returns string in some clouds).
- Server volumes are populated from Nova's `os-extended-volumes:volumes_attached` and resolved via the Cinder cache, so listing volumes once via `openstack.volumes` lets `server.volumes` return fully-populated refs.
- `user.roles()` requires permission to read role assignments; non-admin tokens may only see their own assignments.
- Barbican container refs from Octavia listeners can also point at bare secrets in some deployments. We resolve to `keymanager.container` when the ref URL contains `/containers/` and return null for `/secrets/`-style refs.
- FWaaS v2, Octavia, and Barbican are not deployed in every cloud; the defensive 401/403/404 → empty path handles that gracefully.

## Test plan

- [x] `go build ./...` clean
- [x] `make providers/build/openstack && make providers/install/openstack`
- [x] `mql shell openstack --help` shows full flag set
- [x] No-creds path produces clean `Missing input for argument [auth_url]` error
- [x] Bogus `--auth-url` reaches Keystone POST and fails at TCP/DNS (full auth pipeline verified)
- [ ] Smoke against a real OpenStack cloud (e.g. devstack or a tenant's deployment) — needs reviewer creds


tas50:
### Code review

Found 5 issues. The user explicitly requested a deep-dive review on this draft, with extra attention to logic errors, nil handling, and pagination. Pagination looks correct everywhere — every gophercloud `List(...)` call is followed by `.AllPages(ctx())` and the matching `Extract*`. Nil handling on `StateIsNull` returns is consistently correct on singular-resource accessors. The findings below are all logic / wiring issues.

1. **Process-global `sgNameCache` leaks across connections and silently picks the wrong group on duplicate names.** `sgNameCache`/`sgNameCacheDone` are package-level globals populated from the first `NetworkClient` to call `lookupSecurityGroupIDByName`; the `client` argument is ignored after that. A second connection (different cloud / region / project) in the same provider process gets the first connection's name→ID map, and admin-scoped tenants where two projects each have a `default` security group land on whichever was inserted last. Move the cache to `OpenstackConnection` (or `mqlOpenstackInternal`) keyed by client.

   https://github.com/mondoohq/mql/blob/a143149410355926ea29071808af91b04acba8fe/providers/openstack/resources/network.go#L25-L31

   https://github.com/mondoohq/mql/blob/a143149410355926ea29071808af91b04acba8fe/providers/openstack/resources/network.go#L709-L740

2. **`serverFlavorID` falls back to `original_name`, which is a flavor name, not an ID.** From Nova microversion 2.47 onward the embedded `flavor` object on a server response no longer has `id`; only `original_name`, `vcpus`, `ram`, etc. are returned. The fallback stashes the *name* (e.g. `m1.small`) into `cacheFlavorID`, which is then passed as the `id` to `flavors.Get` in `initOpenstackComputeFlavor` — that endpoint expects a UUID, so the typed `flavor()` reference resolves to a stub on every modern OpenStack cloud. Resolve the name to an ID first, or store the name separately and look it up by name.

   https://github.com/mondoohq/mql/blob/a143149410355926ea29071808af91b04acba8fe/providers/openstack/resources/compute.go#L243-L256

3. **`OPTION_INSECURE` is declared but never wired up.** The constant is defined but never read by `resolveAuth`, never appears in the provider's CLI flag list, and the provider client is never built with a custom `*http.Client` setting `InsecureSkipVerify`. Users passing `--insecure` see it silently ignored — TLS verification still applies. Either wire it through (custom `*http.Client` on the gophercloud `ProviderClient`) or remove the constant.

   https://github.com/mondoohq/mql/blob/a143149410355926ea29071808af91b04acba8fe/providers/openstack/connection/options.go#L19-L23

4. **Single-LB queries return empty `listeners`/`pools`.** `initOpenstackOctaviaLoadBalancer` is a stub (`return args, nil, nil`), and `listeners()` / `pools()` only iterate `r.cacheListenerIDs` / `r.cachePoolIDs` — fields that are populated solely from the list path. A targeted query like `openstack.octavia.loadBalancer(id: "...")` therefore reports zero listeners/pools even when Octavia has them. Either populate the caches lazily inside the accessors (e.g. `loadbalancers.GetStatuses` or `listeners.List(WithLoadBalancerID)`), or have the init fetch the LB and prime them.

   https://github.com/mondoohq/mql/blob/a143149410355926ea29071808af91b04acba8fe/providers/openstack/resources/octavia.go#L57-L62

   https://github.com/mondoohq/mql/blob/a143149410355926ea29071808af91b04acba8fe/providers/openstack/resources/octavia.go#L154-L185

5. **`providers/openstack/README.md` uses the pre-v13 binary name `cnquery`.** The repo CLAUDE.md explicitly notes the v13 rename: \"the CLI binary is `mql`\" (other recently-merged provider READMEs use `mql shell <provider>`). The OpenStack README has five `cnquery shell openstack` examples plus one prose mention.

   https://github.com/mondoohq/mql/blob/a143149410355926ea29071808af91b04acba8fe/providers/openstack/README.md#L1-L42

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #7603, opened 2026-05-08
🐛 mqlc: record [].join(...) chunk type as string, not the input array

vjeffrey:
## Summary

`compileArrayJoin` was setting the join chunk's `Function.Type` to the input array type (e.g. `[]string`) instead of the actual output type (`string`). Direct mql evaluation was unaffected because the runtime `arrayJoin` returns its own `RawData{Type: String, ...}`, but cnspec's policy resolver reads the chunk type via `DereferencedTypeV2` to register each datapoint's expected type. When `.join(...)` was the entrypoint of a data query, the executor saw `expectedType=[]string ≠ actualType=string` and routed the result through `CastResult([]string)`, which fed a Go string into `array2result` and failed with `could not convert string to []string`.

The bug only surfaced when `join` was itself the data-query entrypoint — in most policies it sits inside a comparison whose `Bool` is the actual datapoint, which is why it went unnoticed since the function was introduced in #6058.

Also switches the runtime's `typHint` lookup from `chunk.Type()` to `bind.Type` so regex/version/etc. element formatting still works once the chunk type is no longer the array.

## Test plan

- [x] New mqlc test asserts the join chunk's recorded type is `string`
- [ ] `make test/go/plain` passes
- [ ] Verify cnspec policy with `.join(...)` as a data-query entrypoint resolves cleanly end-to-end

### DIFF
--- a/llx/builtin_array.go
+++ b/llx/builtin_array.go
@@ -558,8 +558,8 @@ func arrayJoin(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64) (*RawD
 	}
 
 	typHint := types.Any
-	if chunk.Type().IsArray() {
-		typHint = chunk.Type().Child()
+	if bind.Type.IsArray() {
+		typHint = bind.Type.Child()
 	}
 
 	var res strings.Builder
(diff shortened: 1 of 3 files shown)


arlimus (approved):
Thank you for fixing this!

PR #7654, opened 2026-05-11
🐛 kernel.installed: harden running-kernel detection across all RPM-based families

czunker:
## Summary

`kernel.installed` decides whether each kernel package is the currently running one by string-comparing a derived key against the value parsed from `/proc/version`. The exact key shape differs per platform family — RHEL / Amazon Linux / Oracle / Photon / SUSE all have their own join — but every RPM-based branch funnels the package's `pkg.Version` through the comparison. When the underlying kernel rpm declares a non-zero `Epoch`, `providers/os/resources/packages/rpm_packages.go` prefixes that epoch into `pkg.Version` (e.g. `"1:6.1.170-210.320.amzn2023"`), while `/proc/version` never carries the epoch — so every comparison silently fails and the asset reports every installed kernel with `running: false`.

This is reproducible today on Amazon Linux 2023, whose `kernel` rpm ships with epoch 1. The same shape is latent on Oracle Linux, Photon, and SUSE if any of them ever pick up an epoch on their kernel rpm.

## Change

- Introduced `stripRPMEpoch(version)` to drop a leading `<epoch>:` from any RPM package version. Used by every per-platform matcher.
- Extracted three per-platform comparison helpers, each with the epoch primitive on its hot path:
  - `rpmKernelMatchesRunning(pkgVersion, pkgArch, runningKernel)` — RHEL / Amazon Linux / Oracle (incl. UEK)
  - `photonKernelMatchesRunning(pkgVersion, pkgName, runningKernel)` — Photon's `linux[-flavor]` join
  - `suseKernelMatchesRunning(pkgVersion, pkgName, runningKernel)` — SUSE's `HasSuffix(-flavor)` + `HasPrefix(trimmed-running)` shape, handling the extra dpkg-release segment SUSE carries on `pkg.Version`
- Wired each helper into its branch in `kernel.installed`. Closure bodies become one-liners. The output `KernelVersion.Version` field stays unchanged on every platform, so downstream consumers that pattern-match against it keep working.
- Debian / Ubuntu's branch is unaffected by the bug class — it matches on `pkg.Name` (which never carries an epoch), not on `pkg.Version` — and is left as-is.

## Test plan

- [x] `go test ./providers/os/resources/ -run 'TestRpmKernelMatchesRunning|TestStripRPMEpoch|TestPhotonKernelMatchesRunning|TestSuseKernelMatchesRunning' -v` — 21 internal-package cases pass:
  - `rpmKernelMatchesRunning` × 6: AL2023 epoch-1 match (regression case), AL2023 older ABI mismatch, legacy RHEL with no epoch (backward compat), Oracle UEK with epoch, different architectures, empty `/proc/version`
  - `stripRPMEpoch` × 5: no epoch, epoch 1, multi-digit epoch, empty input, bare colon
  - `photonKernelMatchesRunning` × 5: bare `linux`, `linux-esx` flavor, older inactive kernel, wrong flavor, hypothetical epoch-1 Photon
  - `suseKernelMatchesRunning` × 5: `kernel-default` running, older inactive `kernel-default`, `kernel-rt` vs `-default` running, hypothetical epoch-1 SUSE, empty running
- [x] Verified against a live Amazon Linux 2023 EC2 instance — `kernel.installed` now returns `running: true` on the kernel package whose version matches `/proc/version`, where prior to the fix every entry came back with `running: false`.

### DIFF
--- a/providers/os/resources/kernel.go
+++ b/providers/os/resources/kernel.go
@@ -41,23 +41,59 @@ type KernelVersion struct {
 	Running bool   `json:"running"`
 }
 
+// stripRPMEpoch removes a leading "<epoch>:" prefix from an RPM package
+// version string. /proc/version (and the rest of kernel.info) never carries
+// the epoch, so it must be stripped before comparing a package version to
+// the running kernel version. providers/os/resources/packages/rpm_packages.go
+// concatenates a non-zero / non-"(none)" epoch into pkg.Version, so this
+// matters whenever the underlying kernel rpm declares an Epoch.
+func stripRPMEpoch(version string) string {
+	if idx := strings.IndexByte(version, ':'); idx >= 0 {
+		return version[idx+1:]
+	}
+	return version
+}
+
 // rpmKernelMatchesRunning reports whether the given RPM kernel package
 // describes the currently running kernel, identified by the value
 // /proc/version reports (e.g. "6.1.170-210.320.amzn2023.x86_64").
 //
-// The package version string may carry an RPM epoch prefix like
-// "1:6.1.170-210.320.amzn2023" — see rpm_packages.go where epoch is
-// concatenated into the version when it is non-zero / not "(none)". The
-// running-kernel string from /proc/version never carries the epoch, so a
-// plain string-equality check between "<version>.<arch>" and that string
-// drops the running flag for every package whose epoch is set. AL2023's
-// `kernel` package has epoch 1, so the bug is reproducible there for every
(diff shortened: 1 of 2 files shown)


czunker:
```
cnspec run ssh -i ~/.ssh/id.pem ec2-user@1.2.3.4 -c "kernel.installed"
kernel.installed: [
  0: {
    name: "kernel"
    running: true
    version: "1:6.1.170-210.320.amzn2023"
  }
]
```

PR #7677, opened 2026-05-13
✨ aws: surface SG rule descriptions and typed prefix lists

tas50:
## Summary

Three additive fields on `aws.ec2.securitygroup.ippermission` so audits can reach data we currently drop or only expose as raw dicts:

- **`ipRangeDetails []dict`** and **`ipv6RangeDetails []dict`** — parallel to the existing `ipRanges` / `ipv6Ranges`, but each entry is `{cidr, description}`. AWS returns a per-CIDR description (e.g. "SSH from office") that today we discard at [aws_ec2.go:562](providers/aws/resources/aws_ec2.go#L562); these fields surface it.
- **`prefixLists() []aws.ec2.managedPrefixList`** — typed reference resolving `PrefixListIds` against the existing managed-prefix-list resource. Unlocks traversal like `sg.ipPermissions.prefixLists.entries.cidr`.

A new `initAwsEc2ManagedPrefixList` resolves a managed prefix list by ARN via `DescribeManagedPrefixLists` (region parsed from the ARN) so the typed ref above works even when the user hasn't pre-enumerated the full collection. `entries()` switched to read the schema-level `Region` field, so init-created instances work too.

## Backwards compatibility

- No existing fields change shape or semantics. `ipRanges`, `ipv6Ranges`, and `prefixListIds` keep their current types and values.
- The synthetic `ippermission.id` (`groupId-index[-egress]`) is intentionally **unchanged** in this PR — switching to real `sgr-…` IDs is a separate body of work because it requires a different data source (`DescribeSecurityGroupRules`) and a new IAM permission.
- No new IAM permissions required for this PR — `ec2:DescribeManagedPrefixLists` was already in `aws.permissions.json`.
- New entries live at `13.23.2` in `aws.lr.versions`.

## Test plan

Verified live against AWS account `177043759486`:

- [x] `ipRangeDetails` returns `[{cidr, description}]` — confirmed with a rule whose description is `"SSH"`
- [x] `prefixLists()` resolves AWS-managed `pl-45a6432c` (`com.amazonaws.firewall.regional-prod-only`) end-to-end: id, name, addressFamily, and `entries.length=9` populated lazily via the new `initAwsEc2ManagedPrefixList`
- [x] Existing fields (`ipRanges`, `ipv6Ranges`, `prefixListIds`) unchanged in shape and values
- [x] `make providers/build/aws` clean; `gofmt -l` silent

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -12559,10 +12559,16 @@ private aws.ec2.securitygroup.ippermission @defaults("id toPort fromPort ipProto
   ipProtocol string
   // IPv4 address ranges
   ipRanges []string
+  // IPv4 address ranges with per-range descriptions (each entry: {cidr, description})
+  ipRangeDetails []dict
   // IPv6 address ranges
   ipv6Ranges []string
+  // IPv6 address ranges with per-range descriptions (each entry: {cidr, description})
+  ipv6RangeDetails []dict
   // List of prefix IDs
   prefixListIds []dict
+  // Managed prefix lists this rule targets
+  prefixLists() []aws.ec2.managedPrefixList
   // List of user ID group pairs
   userIdGroupPairs []dict
 }
--- a/providers/aws/resources/aws.lr.go
+++ b/providers/aws/resources/aws.lr.go
@@ -112635,7 +112635,7 @@ func (c *mqlAwsEc2TransitgatewayPeeringAttachment) GetStatusMessage() *plugin.TV
 type mqlAwsEc2ManagedPrefixList struct {
 	MqlRuntime *plugin.Runtime
 	__id       string
-	mqlAwsEc2ManagedPrefixListInternal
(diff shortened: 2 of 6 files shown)


tas50:
Addressed bot review in ceec7f7:

- **Init defensive args** — populate `id` and `region` from the parsed ARN before `DescribeManagedPrefixLists` runs, so the access-denied and empty-response paths leave a queryable resource (rather than empty `Region.Data` that would break `entries()`).
- **Helper extraction** — `buildIpRangeDetails` / `buildIpv6RangeDetails` shared between `ipPermissions()` and `ipPermissionsEgress()`.
- **Dead-code removal** — dropped the no-longer-read `mqlAwsEc2ManagedPrefixListInternal` struct entirely (not just the assignment), now that `entries()` reads `a.Region.Data` directly.

Re-verified against AWS: `prefixLists` still resolves `pl-45a6432c` end-to-end (id, name, addressFamily, `entries.length=9`).

PR #7681, opened 2026-05-13
✨ gcp: typed match fields on firewallPolicy.rule

tas50:
## Summary

The modern hierarchical / network firewall policy rule kept all match data buried in a single \`match dict\`, so any audit that needed source CIDRs, layer-4 ports, or secure-tag references had to spelunk the dict. This PR flattens the high-value match data into typed fields on \`gcp.project.computeService.firewallPolicy.rule\`.

## New fields (all additive)

\`\`\`
srcIpRanges       []string
destIpRanges      []string
layer4Configs     []dict             // {ipProtocol, ports[]}
srcAddressGroups  []string
destAddressGroups []string
srcSecureTags     map[string]string  // tag name -> state (EFFECTIVE / INEFFECTIVE)
targetSecureTags  map[string]string  // modern replacement for instance targetTags
\`\`\`

Symmetric source/destination coverage for the high-value fields (\`IpRanges\`, \`AddressGroups\`). Structurally single fields (\`layer4Configs\`, \`srcSecureTags\` — the GCP SDK only has \`SrcSecureTags\`, no \`DestSecureTags\` — and \`targetSecureTags\` which lives on the rule itself, not on \`match\`) come along as-is. \`srcSecureTags\` and \`targetSecureTags\` use \`map[string]string\` (tag name → state) per the CLAUDE.md guidance for two-string-field structs.

## Not in this PR

- **Predicate methods.** An earlier revision added \`openToInternet()\` / \`allowsSshFromInternet()\` / \`allowsRdpFromInternet()\` to mirror what already exists on the legacy \`gcp.project.computeService.firewall\` ([compute.go:3290](providers/gcp/resources/compute.go#L3290), 3600, 3604). Pulled back out — users can derive these inline in MQL against the new typed fields, and we should decide on the cross-cloud predicate shape (AWS / Azure / OCI counterparts) before locking GCP into a Go-side helper. Legacy firewall predicates unchanged.
- **Niche source/destination match fields.** \`srcFqdns\`, \`destFqdns\`, \`srcRegionCodes\`, \`destRegionCodes\`, \`srcThreatIntelligences\`, \`destThreatIntelligences\` remain in the \`match dict\` only. Rare audit need and they don't pull their weight as typed fields once the predicates are out of scope.

## Backwards compatibility

- Existing \`match dict\` field preserved — audits using \`rule.match[\"srcIpRanges\"]\` keep working unchanged.
- New entries at \`13.14.2\` in \`gcp.lr.versions\`.
- No new IAM permissions.

## Test plan

- [x] \`make providers/build/gcp\` clean
- [x] Legacy firewall predicates unchanged — \`default-allow-icmp\` still returns \`openToInternet: true\` (regression check on shared helper file)
- [x] Query against \`firewallPolicies\` returns clean empty array in test project (no hierarchical firewall policy resources to exercise the new fields against — a project with \`firewallPolicy\` resources would exercise the data path fully)


tas50:
Addressed the two live review items in 21dd1070:

- **Source/destination asymmetry**: Dropped `srcFqdns` and `srcRegionCodes`. They were added originally to support the `openToInternet` predicate; with the predicate gone there's no reason to expose source-only narrowing fields without their destination counterparts. The PR now ships only the symmetric high-value pairs (`srcIpRanges`/`destIpRanges`, `srcAddressGroups`/`destAddressGroups`) plus the structurally-single fields (`layer4Configs`, `srcSecureTags`, `targetSecureTags`). Raw `srcFqdns`/`destFqdns`/`*RegionCodes`/`*ThreatIntelligences` remain queryable through the existing `match dict` field for users who need them.
- **Stray blank line** above `(*Rule).id()` — fixed.

Bot also flagged `DestSecureTags` as missing — confirmed against the SDK (`google.golang.org/api/compute/v1`); only `SrcSecureTags` exists on the matcher. No code change needed for that one.

PR #7683, opened 2026-05-13
✨ google-workspace: surface managed endpoints via Cloud Identity Devices

tas50:
## Summary

- Adds `googleworkspace.endpoints` backed by the Cloud Identity Devices API — one resource that covers Chrome OS, Android, iOS, **Windows**, macOS, Linux, and Google Sync clients, including GCPW-enrolled Windows machines and endpoint-verified desktops that the Admin SDK's `chromeosdevices` / `mobiledevices` endpoints don't return.
- Per-device user bindings exposed via `googleworkspace.endpoint.users` (Cloud Identity DeviceUsers) so BYOD multi-user devices can be audited row-per-user.
- Android-specific and Endpoint-Verification-specific payloads exposed as `dict` (`androidAttributes`, `endpointVerificationAttributes`) so platform-conditional fields stay queryable without polluting every row with nulls on the other platforms.

## Notes on the API surface

- Uses `cloudidentity.CloudIdentityDevicesReadonlyScope`; the existing service-account / domain-wide-delegation flow must have this scope added in the Admin Console (Security → API controls → Domain-wide delegation).
- The list call queries **both** `COMPANY_INVENTORY` and `USER_ASSIGNED_DEVICES` views and dedupes by device `Name`. The Devices API doesn't return both in a single call.
- The DeviceUsers list page size is capped at 20 by Google; pagination is handled.

## Test plan

- [ ] `make providers/build/google-workspace && make providers/install/google-workspace`
- [ ] `mql shell google-workspace --customer-id <id> --credentials-path <creds>`
- [ ] `googleworkspace.endpoints { deviceType model serialNumber managementState encryptionState }` returns a mix of platforms on a tenant with Windows / macOS / mobile enrollments
- [ ] `googleworkspace.endpoints.where(deviceType == "WINDOWS" && encryptionState != "ENCRYPTED")` returns expected Windows fleet results
- [ ] `googleworkspace.endpoints.first.users { userEmail managementState passwordState }` returns the per-user bindings
- [ ] `googleworkspace.endpoints.first.endpointVerificationAttributes` returns the structured signals payload (certificates, browser attributes, additional signals) on an endpoint-verified device


tas50:
Addressed the bot feedback in 8cf3e68:

- **Trailing newline on `.lr`** — fixed.
- **Spell check** — `BYOD`, `CDMA`, `GSM`, `wifi` added to `.github/actions/spelling/expect.txt` (sorted).

Two suggestions I deliberately did **not** action:

- **`parseRFC3339` "silently swallows" errors.** Kept the silent-fallback behavior because it matches the established pattern in this provider — `users.go` and `reports.go` both inline `time.Parse(time.RFC3339, …)` and discard the error the same way. Introducing a logger here only would be inconsistent; if we want logging, it should land as a sweep across the provider rather than a one-off for this file.
- **`CloudIdentityDevicesReadonlyScope` is "redundant" with `DefaultWorkspaceClientScopes`.** It isn't — `DefaultWorkspaceClientScopes` is defined but never read by `connection.Client()`. Each call to `cloudIdentityService(conn, …scope)` builds a fresh credential constrained to the scopes passed in (`conn.Client(scopes…)` → `serviceAccountAuth` → `googleoauth.CredentialsParams{Scopes: scopes}`). `groups.go` follows the exact same pattern. Removing the scope argument would broaden — or break — the credential.

PR #7750, opened 2026-05-20
🧹 ci: scope provider regen and lint to changed providers

czunker:
## Summary

CI today does heavy work for providers that didn't change. On a PR that touches only `providers/os` (e.g. #7749 the CPE cleanup):

- `generated-files` runs `make mql/generate` + `SKIP_COMPILE=yes make providers/build`, which iterates LR codegen + CLI-json gen for **every** provider (~40). 29 min.
- `golangci-lint-providers` (called from both `pr-test-lint.yml` and `pr-extended-linting.yml`) loops `providers/*/` and lints all of them. ~3 min × 2.

Each provider has its own `go.mod` and no cross-provider Go imports, so processing them independently is safe.

This PR adds `scripts/changed-providers.sh`, which diffs against `merge-base HEAD origin/main` and emits either:
- `all=true` — when a shared dep changed (`providers-sdk/`, `providers/core/`, `Makefile`, `providers.yaml`, root `*.go`, `go.{mod,sum}`, the related workflows, or the script itself), or
- `providers="os k8s"` — the list of provider dirs that actually changed.

Wired into:
- **`pr-test-generated-files.yaml`**: regenerates only changed providers (or all, when forced). Also drops the redundant `make mql/generate` pass that was already building everything before the explicit `SKIP_COMPILE=yes make providers/build` ran again. The non-provider generators (`llx/`, `shared/`, `sbom/`, `reporter/`) still run unconditionally.
- **`reusable-lint-providers.yml`**: new optional `providers` input. Empty = old behavior. Non-empty = lint only those provider dirs that have their own `go.mod`.
- **`pr-test-lint.yml`** / **`pr-extended-linting.yml`**: add a `detect-changed-providers` job and pass its output to the providers lint workflow.

## Expected impact

For a single-provider PR:
- `generated-files`: ~29 min → a few minutes (LR codegen is the dominant cost and it now runs for 1 provider instead of 40).
- Each `golangci-lint-providers`: ~3 min → under a minute.

For PRs that touch `providers-sdk/`, `providers/core/`, the Makefile, or the workflows themselves, behavior is identical to today (all providers).

## Out of scope (explicitly not changed here)

- `go-test`'s `providers/build` dependency still builds all providers. Scoping that is riskier (`providers/test` and main-module tests both feed off it), so it's a separate change.
- The duplicate `golangci-lint` between `pr-test-lint.yml` and `pr-extended-linting.yml` is still there. That's Phase 1 from the audit and was deliberately skipped here.

## Test plan

- [ ] Push to a branch and verify `generated-files` reports a single provider when only that provider changed.
- [ ] Verify a touch to `providers-sdk/` triggers `all=true` and the full provider set is processed.
- [ ] Verify `golangci-lint-providers` skips providers without `go.mod` cleanly when scoped (they're linted via the root golangci-lint job).
- [ ] Verify behavior on push events without a PR (script falls back to `merge-base HEAD origin/main`).

### DIFF
--- a/.github/workflows/pr-extended-linting.yml
+++ b/.github/workflows/pr-extended-linting.yml
@@ -20,6 +20,23 @@ permissions:
   pull-requests: read
 
 jobs:
+  detect-changed-providers:
+    runs-on: ubuntu-latest
+    outputs:
+      providers: ${{ steps.changed.outputs.providers }}
+      all: ${{ steps.changed.outputs.all }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+
+      - name: Determine changed providers
+        id: changed
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+        run: bash scripts/changed-providers.sh
+
   golangci-lint:
     permissions:
       # allow read access to the content for analysis.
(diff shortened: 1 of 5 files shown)


imilchev (approved):
good idea! let's test it out

Issue #7744, opened 2026-05-19
Windows: user.home is never populated, breaking vscode/firefox/ai-coding-tools resources


M-gre:
See https://github.com/mondoo-community/customer-issues/issues/187

## Summary

On Windows assets, the `user` resource never exposes a home directory. The `home` field is always `""` for every user. As a downstream consequence, every resource that walks `users` and reads `user.home` to locate per-user files returns nothing on Windows — most notably `vscode.extensions` / `vscode.paths`, and the equivalent Firefox and AI-coding-tool discovery paths.

This is silent: queries succeed and return `[]`, so there's no error to alert on it.

## Reproduction

On any Windows asset (with `Get-LocalUser` returning users that have profiles under `C:\Users\`):

```
mql shell <windows-asset>
> users { name home }       # every home is ""
> vscode.paths              # []
> vscode.extensions         # []
> firefox.profiles          # []  (same root cause)
```

## Root cause

1. `WindowsUserManager` (`providers/os/resources/users/ps1getlocalusers.go:91-101`) builds each `User` from `Get-LocalUser | ConvertTo-Json`. `Get-LocalUser` does **not** return `HomeDirectory`, and `winToUser` consequently never sets `User.Home`.
2. `mqlUsers` (`providers/os/resources/user.go:142`) passes `user.Home` straight to `mqlUser`, so `mqlUser.home` is `""`.
3. The `user` schema (`providers/os/resources/os.lr:743`) declares `home` as a plain field with no init or computed accessor — no other code path fills it in.
4. Consumers then filter empty homes as system accounts:
   - `providers/os/resources/vscode.go:99-115` via `isSystemHomeDir` → every Windows user skipped.
   - `providers/os/resources/firefox.go:169-176` via `isValidFirefoxUserHome` → same.
   - `providers/os/resources/ai_coding_tools.go:189` → same.

Only Linux (`etcpasswd.go:49`) and macOS (`dscl.go:117`) populate `Home`.

## Suggested fix direction

The right fix is to populate `Home` in the Windows users manager so every consumer benefits.

`Get-CimInstance Win32_UserProfile` returns `SID` → `LocalPath`, which is the actual on-disk profile path (including renamed profiles, non-default profile roots, and domain users who have logged in). Joining that with the SID-keyed output of `Get-LocalUser` in `winToUser` would close the gap. Alternative: a `Get-LocalUser | ForEach-Object` pipeline that emits both records in one PowerShell round-trip.

[...]

PR #7753, opened 2026-05-20
🐛 windows: stabilize Get-LocalUser JSON shape against locale + depth

M-gre:
## Summary

Follow-up to #7746. On real Windows hosts, the new `Get-LocalUser` script broke parsing in two ways; this PR fixes both and refreshes the testdata to match what the script actually emits.

- **`ConvertTo-Json -Depth 4` expanded `SID.AccountDomainSid` into a nested object**, but our Go struct expects a string. `cnspec run -c 'vscode.extensions'` returned `json: cannot unmarshal object into Go struct field WindowsSID.SID.AccountDomainSid of type string`, and `user` came back null-everywhere because `List()` errored out. Fixed by projecting fields explicitly via `[PSCustomObject]`, so the JSON shape no longer depends on `ConvertTo-Json` depth heuristics.
- **Default `.ToString()` produced locale-dependent dates** (e.g. `"28.01.2026 14:56:04"` on a German host). Switched to `.ToString('o')` (ISO 8601 round-trip) so dates are stable across locales. Latent footgun — nothing reads those fields in Go today, but anyone wiring up `passwordLastSet` etc. on the MQL side would have hit garbage.
- Regenerated `testdata/windows.{json,toml}` (including the new encoded PowerShell command for the mock replay) and updated parser test expectations to ISO 8601.

## Test plan

- [x] `go test ./providers/os/resources/users/...` — `TestManagerWindows` + `TestWindowsLocalUsersParser` pass
- [x] Live verification on a German-locale Windows 11 host: type-check probe script passes, `SID.AccountDomainSid` is a string, dates are ISO 8601, `LocalPath` resolves for logged-in users and stays empty for `Guest` / `DefaultAccount` / `WDAGUtilityAccount` / `Administrator`
- [x] Manual re-test of `cnspec run -c 'vscode.extensions'` and `cnspec run -c 'user'` on the affected Windows host after `make providers/build/os && make providers/install/os`

Refs: #7744, follow-up to #7746


M-gre:
Tested successfully on local Windows VM:
image

PR #7769, opened 2026-05-21
🧹 remove the manual prefixed "Deprecation" from titles+desc for resources/fields

arlimus:


tas50:
We need to get the docs page to actually use our field data for deprecation if we remove this

arlimus:
> We need to get the docs page to actually use our field data for deprecation if we remove this

Maturity addressed in docs in https://github.com/mondoohq/docs/pull/1409

PR #7786, opened 2026-05-21
✨ gcp: type four dict-style fields + add GKE Windows node config

tas50:
## Summary

Expand depth on four existing resources, no breaking changes. Three are **dict → typed sub-resource** refactors that keep the existing field as `@maturity(\"deprecated\")` alongside the new typed accessor; one is purely additive.

### Resources / fields added

- **`gcp.project.computeService.instance.confidentialCompute`** (new typed sub-resource) exposes `enabled` and `instanceType` (`SEV` / `SEV_SNP` / `TDX`). The original `confidentialInstanceConfig dict` stays as a deprecated alias so existing dict-style queries keep working. Closes the asymmetry with the already-typed `shieldedInstanceConfig`.
- **`gcp.project.cloudRunService.vpcAccessConfig`** (new shared typed sub-resource) exposes `connector`, `egress` (`ALL_TRAFFIC` / `PRIVATE_RANGES_ONLY`), and direct-VPC `networkInterfaces`. Referenced from both `service.revisionTemplate` and `job.executionTemplate.taskTemplate`. Old `vpcAccess dict` kept deprecated on both parents. `egress` is the data-exfiltration-prevention control auditors care about.
- **`gcp.project.sqlService.instance.settings.activeDirectory`** (new typed sub-resource) exposes `domain`, `mode`, `dnsServers`, `adminCredentialSecretName` for Cloud SQL for SQL Server Entra ID / AD binding audits. Old `activeDirectoryConfig dict` kept deprecated.
- **`gcp.project.gkeService.cluster.nodepool.config.windowsNodeConfig`** (new typed sub-resource) exposes `osVersion` (`LTSC2019` / `LTSC2022`). Mirrors the existing `linuxNodeConfig` for Windows GKE node pools. Purely additive — no deprecation needed.

### Backwards compatibility

The three `dict` → `typed` refactors keep both names — old field stays exactly where it is, just annotated `@maturity(\"deprecated\")`. No existing query breaks. New typed alternatives are added under new field names (`confidentialCompute`, `vpcAccessConfig`, `activeDirectory`) so the old field path stays intact.

### What was dropped from the original list

Two items from the gap analysis were removed after SDK verification:
- `AttachedDisk.sourceImageEncryptionKey` — only exists on `AttachedDiskInitializeParams`, which is documented as `[Input Only]` and isn't populated at read time.
- `Instance.secondaryBootDisk` — doesn't exist in `cloud.google.com/go/compute@v1.63.0` or the REST `compute/v1` SDK. Looks to have been hallucinated.

### Notes

- `.lr.versions` entries at `13.16.3`. No `config/config.go` version bump.
- `gcp.permissions.json` unchanged (no new API calls).
- No new SDK dependencies.

## Test plan

- [x] `make providers/build/gcp` clean; permissions JSON unchanged
- [x] `go test ./providers/gcp/resources/...` passes (includes `TestGCPPermissionsMatchValidatedList`)
- [ ] Interactive smoke against live projects:
  ```
  cnspec shell gcp
  > gcp.project.computeService.instances { confidentialCompute { enabled instanceType } }
  > gcp.project.cloudRunService.services { template.vpcAccessConfig { connector egress networkInterfaces } }
  > gcp.project.sqlService.instances { settings.activeDirectory { domain mode dnsServers } }
  > gcp.project.gkeService.clusters { nodePools.list { config.windowsNodeConfig { osVersion } } }
  ```
  Each returns null fields when the corresponding feature isn't configured on the resource — that's expected.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -551,6 +551,7 @@ serde
 serviceconnection
 servicedesk
 serviceprincipals
+SEV
 Sflags
 sfn
 signin
@@ -620,6 +621,7 @@ trainingjob
 transitgateway
 transitives
 truststore
+TDX
 TValue
 Twt
 uami
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -604,6 +604,7 @@ tanium
 targetgroup
 tcpip
 tde
+TDX
 temurin
 tensorboard
 testutils
@@ -621,7 +622,6 @@ trainingjob
 transitgateway
 transitives
 truststore
-TDX
 TValue
 Twt
 uami
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -417,6 +417,25 @@ Think of it as: MQL (like SQL or even better GraphQL) → MQLC (compiler) → LL
 - Must be unique and stable (ARN, UUID, or composite key)
 - If `__id` is empty or duplicate, caching breaks and performance degrades
 
+**Hide synthetic `__id` values; don't expose them as `id` fields.** When a sub-resource's cache key is purely internal — a parent-qualified path like `<parentId>/confidentialCompute`, not something a user would ever query by — do NOT declare `id string` in the `.lr` block. Pass the cache key directly via the magic `"__id"` argument to `CreateResource`, and omit the `id()` Go method:
+
+```go
+// schema (.lr): no `id string` field declared
+private gcp.project.computeService.instance.confidentialCompute @defaults("enabled instanceType") {
+  enabled bool
+  instanceType string
+}
+
+// creator (.go): pass "__id" directly, no id() func needed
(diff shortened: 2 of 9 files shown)


tas50 (commented):
We should consider hiding the internal only IDs here if they are truly only internal

tas50:
Good catch — done in 60b564e5b. The four new typed sub-resources (`confidentialCompute`, `activeDirectory`, `windowsNodeConfig`, `vpcAccessConfig`) had `id string` fields whose values were purely synthetic `<parent>/<leaf>` cache keys — nothing a user would ever query by. Dropped them from the `.lr` schema and pass the cache key directly via the `__id` magic arg to `CreateResource`. Mirrors the pattern already in `gcp.project.binaryAuthorizationControl.policy`. Left the pre-existing `id` fields on sibling sub-resources (`linuxNodeConfig`, `kubeletConfig`, etc.) untouched since they're out of scope for this PR; happy to do a follow-up sweep if you'd like.

Also added `SEV` and `TDX` to the spell-check allowlist (they appear in the confidential-compute doc comments and enum lists).

PR #7814, opened 2026-05-22
✨ package: add installDate (rpm + Windows registry)

czunker:
## Summary

Adds \`installDate time\` to the generic \`package\` resource.

Stacked on **#7813** (\`package.license\`); please review/merge that PR first. GitHub will auto-retarget this PR to \`main\` once #7813 lands.

## Per-backend coverage

| backend | source |
|---|---|
| **rpm** | \`%{INSTALLTIME}\` epoch seconds added to the rpm queryformat (runtime path) and \`pkg.InstallTime\` from go-rpmdb (static path). \`(none)\` and parse failures yield the zero time. |
| **Windows registry** | new \`InstallDate\` column added to the Uninstall-key \`Select-Object\`; YYYYMMDD parsed to midnight-UTC time.Time via new \`parseWinInstallDate\`. Per-user installs and many non-MSI publishers omit the field — those yield the zero time. |
| dpkg | not collected — would need \`/var/log/dpkg.log\` parsing |
| apk, pacman, macOS | not collected (no native source) |

The field is eagerly populated by both backends; zero \`time.Time\` indicates "no install time available."

## RPM regex extension

\`RPM_REGEX\` gains another capture group between LICENSE and the optional MODULARITYLABEL segment. \`providers/os/resources/reboot/rhel.go\` reuses \`ParseRpmPackages\` on the kernel package and was updated to use the new queryformat.

## Test plan

- [x] \`go test ./providers/os/resources/packages/...\` — pass
- [x] \`go test ./providers/os/resources/reboot/...\` — pass
- [x] \`golangci-lint run\` — 0 issues
- [x] \`gofmt\` — clean
- [ ] CI green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7813, opened 2026-05-22
✨ package: add license field (rpm, dpkg, apk, pacman)

czunker:
## Summary

Adds \`license() string\` to the generic \`package\` resource so software inventory consumers can expose the SPDX license expression alongside vendor/description/cpes.

Part of a series unlocking richer SBOM data for the platform's software API. Follow-ups in this series: \`package.installDate\`, \`npm.package\` description/author/license, \`sbom.Package\` proto fields.

## Per-backend coverage

| backend | source | mode |
|---|---|---|
| **rpm** | \`%{LICENSE}\` rpm queryformat (runtime); \`pkg.License\` from go-rpmdb (static) | eager |
| **dpkg** | \`/usr/share/doc/<pkg>/copyright\` first DEP-5 \`License:\` field | **lazy** |
| **apk** | \`L:\` field of APKINDEX | eager |
| **pacman** | \`%LICENSE%\` block of per-package \`desc\` (FS-fallback path) | eager |
| Windows registry | n/a — no License field on \`HKLM\\...\\Uninstall\` | empty |
| macOS | n/a | empty |

\`license()\` is a computed method, not a plain field, so the dpkg lazy fallback fires only when MQL actually asks for license — no per-package copyright reads on plain \`packages\` enumerations. The eager backends populate via \`CreateResource\` and short-circuit the method body.

## RPM regex extension

\`RPM_REGEX\` gains a new capture group between \`SUMMARY\` and the optional \`MODULARITYLABEL\` segment. The reboot detector in \`providers/os/resources/reboot/rhel.go\` reuses \`ParseRpmPackages\` and was updated to emit the new queryformat.

## Test plan

- [x] \`go test ./providers/os/resources/packages/...\` — all pass
- [x] \`go test ./providers/os/...\` — all pass (the \`device/linux\` build failure also reproduces on \`main\` — pre-existing, unrelated)
- [x] \`golangci-lint run\` — 0 issues
- [x] \`gofmt\` — clean
- [ ] CI green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7816, opened 2026-05-22
✨ npm.package: surface description, author, license

czunker:
## Summary

The \`packagejson\` parser already extracts \`description\`, \`author\`, and \`license\` from \`package.json\` — both string and object forms of author are handled. This PR wires those three fields through:

\`packageJson\` → \`languages.Package\` → \`mqlNpmPackage\` (MQL)

## What's added

Three new MQL fields on \`npm.package\`:

| Field | Source |
|---|---|
| \`description\` | package.json \`description\` |
| \`author\` | name portion of package.json \`author\` — accepts the SPDX-style \`\"Name <email> (url)\"\` string form and the \`{name, email, url}\` object form |
| \`license\` | SPDX expression from package.json \`license\` |

\`languages.Package\` gains \`Author string\` and \`License string\` alongside the existing \`Vendor\`/\`Description\`. Author and Vendor are intentionally separate concepts — some ecosystems carry only one or the other (npm has only \`author\`; dpkg has only Vendor).

## What's not in scope

- **Transitive deps**: \`Transitive()\` still emits bare \`(name, version)\` entries from the dep manifest. Surfacing description/author/license for transitive deps would require recursing into each dep's own \`package.json\` — that's what \`packagelockjson\` does, but it doesn't yet propagate these fields either. Separate follow-up.
- **Deprecated \`licenses\` array** (old packages): \`license\` is single-valued in the modern format. We surface empty when only \`licenses\` is present; can add later if a real package hits it.
- **Homepage**: explicitly out of scope.

Independent of #7813 and #7814 — touches only npm code paths.

## Test plan

- [x] \`go test ./providers/os/resources/languages/javascript/...\` — pass
- [x] \`go test ./providers/os/resources/packages/...\` — pass
- [x] \`go test ./providers/os/resources/reboot/...\` — pass
- [x] New \`TestPackageJsonExtractorAuthorForms\` covering string-form, object-form, and missing-author cases
- [x] \`golangci-lint run\` — 0 issues
- [x] \`gofmt\` — clean
- [ ] CI green

### DIFF
--- a/providers/os/resources/languages/extractor.go
+++ b/providers/os/resources/languages/extractor.go
@@ -55,6 +55,12 @@ type Package struct {
 	Origin string `json:"origin,omitempty"`
 	// Package Vendor/Publisher
 	Vendor string `json:"vendor,omitempty"`
+	// Package Author — distinct from Vendor; some ecosystems
+	// (npm/python) only carry a free-form author string.
+	Author string `json:"author,omitempty"`
+	// Package License — SPDX expression where the source can provide
+	// one (npm package.json `license`, python METADATA `License`).
+	License string `json:"license,omitempty"`
 }
 
 // SortFn is a helper function for slices.SortFunc to sort a slice of Package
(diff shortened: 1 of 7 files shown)


czunker:
Doc updated in 178f0d205 to match the implementation — the deprecated `licenses` array is explicitly called out as unsupported.

PR #7890, opened 2026-05-23
🐛 cloudflare: fix bugs surfaced by the audit pass

tas50:
## Summary

Cleans up the HIGH- and MEDIUM-severity findings from the cloudflare provider audit. The dominant pattern: the `cloudflare-go` v0.117.0 SDK is wildly inconsistent about pagination — some `List*` calls auto-paginate, several have `Page`/`PerPage` params but require the caller to loop, and many take no pagination params at all and silently return only the first page. The fixes split into three buckets.

### HIGH

- **`cloudflare.go: accounts()` infinite loop.** `cursor.Page` was fed back into the request unchanged; once `HasMorePages()` returned true the loop never advanced. Bump `info.Page + 1` per the `mtls.go` precedent.
- **`cloudflare.go: acc.Settings` nil deref.** `Account.Settings` is `*AccountSettings` and is not always populated in list responses; guarded.
- **`stream.go: fetchLiveInputs`** called `http.DefaultClient.Do` directly, swallowed the request error (`req, _ :=`), and never checked the HTTP status. A 401/403/500 parsed as an empty `result` and silently returned no records. Switched to `cf.Raw` which reuses SDK auth/retries and surfaces non-2xx as an error.
- **`stream.go: meta["name"]` panic.** Unchecked map-type-asserts on user-controlled `meta`; routed through a `metaString` helper.
- **`tunnels.go: route __id collision.** The cache key was `tunnelRoute@<network>@<tunnelId>` but the API allows the same network to be advertised through the same tunnel from multiple virtual networks. Two such routes collapsed into one cached row and `route.virtualNetwork` resolved to the wrong VNet. Include `virtual_network_id` in the cache key.
- **`tunnels.go`, `devices.go`: single-page truncation.** `ListTunnelRoutes`, `ListTunnelVirtualNetworks`, `ListTeamsDevices` all silently returned just one page. Paginated.

### MEDIUM

- New **`paginateRaw`** helper in `cloudflare.go` walks pagination on any GET endpoint via `cf.Raw` (so it reuses the same HTTP client, retries, and auth as the typed SDK methods). Used for the long list of `List*` calls whose SDK wrappers take no pagination params:
  - `workers.go` (`ListWorkers`)
  - `notifications.go` (`ListNotificationPolicies`)
  - `gateway.go` (`TeamsRules`, `ListTeamsLists`, `TeamsLocations`, `ListDLPProfiles`)
  - `devices.go` (`DevicePostureRules`, `DevicePostureIntegrations`)
- **`audit_logs.go`**: walk every page (`AuditLogFilter.Page`) instead of returning the first 100.
- **`load_balancing.go`**: paginate `ListLoadBalancers` and `ListLoadBalancerPools` (both `PaginationOptions`-aware, no `ResultInfo` → loop until short page).
- **DCL races** in `workers.go`, `r2.go`, `email_routing.go`: replaced with `sync.Once` (where the "do once" semantics fit) or always-lock (where the cache only stores success).
- **`one.go`**: stop swallowing the `corsHeaders` sub-resource creation error — propagate instead of producing a row with a nil header.

### Out of scope

- `StreamListVideos` uses cursor-based pagination via `After` timestamps; deferred — would need a different helper than `paginateRaw`.
- `waf.go` 1+N (one `GetRuleset` per ruleset) is unavoidable with the current schema; reshaping into a lazy `zone.ruleset.rules` sub-resource is a follow-up.

## Test plan

- [x] `go test ./resources/ -v` (all existing + new tests pass)
- [x] New regression tests:
  - `TestAccountsPagination` — loop terminates and consumes every page exactly once
  - `TestAccountsNilSettings` — missing `settings` in API response
  - `TestTunnelRoutesVnetIDInCacheKey` — two routes that differ only by `virtual_network_id` produce two distinct rows
  - `TestTunnelRoutesPagination` — multi-page route response walked
  - `TestDevicesPagination` — `paginateRaw` consumes every page
  - `TestStreamLiveInputsNilMeta` — nil / empty / non-string / present metadata
- [ ] Spot-check against a real Cloudflare account (recommended before merge)

### DIFF
--- a/providers/cloudflare/resources/audit_logs.go
+++ b/providers/cloudflare/resources/audit_logs.go
@@ -17,44 +17,52 @@ func (c *mqlCloudflareAccountAuditLog) id() (string, error) {
 	return "cloudflare/account/auditLog/" + c.Id.Data, nil
 }
 
-// auditLogs returns the first page of audit log entries for the account
-// (most recent first). The Cloudflare API caps a page at 100 entries; callers
-// that need deeper history should filter by `when` or page through the API
-// outside MQL.
+// auditLogs returns audit log entries for the account (most recent first).
+// Cloudflare's audit log retention is bounded server-side; this walks every
+// page of results until the API reports no more pages.
 func (c *mqlCloudflareAccount) auditLogs() ([]any, error) {
 	conn := c.MqlRuntime.Connection.(*connection.CloudflareConnection)
 
-	resp, err := conn.Cf.GetOrganizationAuditLogs(context.TODO(), c.Id.Data, cloudflare.AuditLogFilter{
+	results := []any{}
+	filter := cloudflare.AuditLogFilter{
 		PerPage:   100,
 		Direction: "desc",
-	})
-	if err != nil {
-		return nil, err
+		Page:      1,
 	}
-
-	results := []any{}
-	for i := range resp.Result {
-		entry := resp.Result[i]
-
-		res, err := NewResource(c.MqlRuntime, "cloudflare.account.auditLog", map[string]*llx.RawData{
-			"id":           llx.StringData(entry.ID),
-			"when":         llx.TimeData(entry.When),
-			"actorEmail":   llx.StringData(entry.Actor.Email),
-			"actorId":      llx.StringData(entry.Actor.ID),
-			"actorIp":      llx.StringData(entry.Actor.IP),
-			"actorType":    llx.StringData(entry.Actor.Type),
-			"actionType":   llx.StringData(entry.Action.Type),
-			"actionResult": llx.BoolData(entry.Action.Result),
-			"resourceId":   llx.StringData(entry.Resource.ID),
-			"resourceType": llx.StringData(entry.Resource.Type),
-			"ownerId":      llx.StringData(entry.Owner.ID),
-			"oldValue":     llx.DictData(anyMap(entry.OldValueJSON)),
-			"newValue":     llx.DictData(anyMap(entry.NewValueJSON)),
-		})
+	for {
+		resp, err := conn.Cf.GetOrganizationAuditLogs(context.TODO(), c.Id.Data, filter)
 		if err != nil {
 			return nil, err
 		}
-		results = append(results, res)
+
+		for i := range resp.Result {
+			entry := resp.Result[i]
+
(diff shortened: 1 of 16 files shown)


tas50:
Thanks for the review. Both findings addressed:

**1. `tunnels.go:122` (id() fallback)** — fixed in 364210bc1. Updated `id()` to call `tunnelRouteID()` so the fallback uses the same 3-component key as the explicit `__id`. The `id()` method can't be deleted because the lr code generator unconditionally calls `res.id()` when `__id` isn't supplied via args.

**2. `audit_logs.go:62` (nil `resp.ResultInfo`)** — false positive, declining. In `cloudflare-go`, `AuditLogResponse` embeds `ResultInfo` **by value**, not by pointer:

```go
type AuditLogResponse struct {
    Response   Response
    Result     []AuditLog `json:"result"`
    ResultInfo `json:"result_info"`  // embedded value, not *ResultInfo
}
```

The suggested `resp.ResultInfo == nil` check would not compile. The `paginateRaw` helper checks against nil because there it's `RawResponse.ResultInfo` which **is** `*ResultInfo` — different types, different nil semantics.

tas50:
Bot re-posted the same finding. Posting one more time with proof so we can close it.

The bot's suggested fix:
```go
if resp.ResultInfo == nil || !resp.ResultInfo.HasMorePages() {
```

does not compile — `resp.ResultInfo` is an embedded **value** of type `cloudflare.ResultInfo`, not a pointer, so comparing it to `nil` is a type error:

```
./probe.go:12:24: invalid operation: resp.ResultInfo == nil (mismatched types cloudflare.ResultInfo and untyped nil)
```

(Verified locally against `cloudflare-go v0.117.0`.)

The struct from the SDK ([auditlogs.go:55–58](https://github.com/cloudflare/cloudflare-go/blob/v0.117.0/auditlogs.go#L55-L58)):

```go
type AuditLogResponse struct {
    Response   Response
    Result     []AuditLog `json:"result"`
    ResultInfo `json:"result_info"`   // embedded value
}
```

A zero-value `ResultInfo` is fine to call `.HasMorePages()` on — it just returns `false`. No nil panic is possible at this call site.

The bot is conflating this with `RawResponse.ResultInfo` which **is** `*ResultInfo` and **does** need the nil guard (which `paginateRaw` correctly has). They're different types.

PR #7912, opened 2026-05-24
⚡ ms365: cache PowerShell reports and drop redundant MFA fetch

tas50:
## Summary

Second of three PRs landing fixes from a ms365-provider audit pass. This one is the performance cluster.

- **`ms365.exchangeonline`, `ms365.sharepointonline`, `ms365.teams`**: their `getExchangeReport` / `getSharepointOnlineReport` / `gatherTeamsReport` helpers all checked `r.fetched` but never set it. So every accessor on those resources (~25 / ~5 / ~4 fields respectively) re-ran the entire PowerShell pipeline — module install + connect + ~N cmdlets + disconnect. A query like `{ malwareFilterPolicy, transportRule, mailbox }` now runs the script once instead of 3×. Pattern matches the `security_exchange.go` helper that already does this correctly. The `errHandler` also now sets `r.fetched=true` so a failed attempt is cached instead of being retried on every field access.
- `ms365_sharepoint.go`: the `ExitStatus != 0` branch used to return `fmt.Errorf` directly, bypassing `errHandler`. Routed through `errHandler` so the failure is also cached.
- **`devices.go`**: dropped a redundant per-list fetch of `UserRegistrationDetails`. The same data is already pulled by `users.list` and lives on `microsoft.mfaResp`; the devices path was a copy-paste that overwrote that field with a separate (paginated) call, and the data isn't even read for devices.
- **`ms365_exchange.go`**: `Get-EXOMailbox` now goes through `Select-Object Identity, ExternalDirectoryObjectId`. Those are the only two fields the Go code consumes from each shared-mailbox entry, and the full mailbox blob is 10-20× larger.

No tests added here: the caching-flag fix is a one-line addition that's exercised only by a real PowerShell connection (no unit harness for that), and the other changes are mechanical refactors / template tweaks. Test additions land in PR 1 and PR 3 instead, where there are pure functions worth testing.

## Test plan

- [ ] `make providers/build/ms365 && make providers/install/ms365`
- [ ] `mql shell ms365 ...` then `ms365.exchangeonline { malwareFilterPolicy, transportRule, mailbox.length }` completes in roughly one cmdlet-pipeline pass (previously took three)
- [ ] Same for `ms365.sharepointonline { spoTenant, spoSites }`
- [ ] Same for `ms365.teams { csTeamsClientConfiguration, csTeamsMeetingPolicy }`
- [ ] `microsoft.devices { id }` no longer triggers a separate UserRegistrationDetails fetch

## Stack

Stacks on #7907. Touches `ms365_exchange.go` and `ms365_sharepoint.go` in non-overlapping hunks, so it can land independently.

### DIFF
--- a/providers/ms365/resources/devices.go
+++ b/providers/ms365/resources/devices.go
@@ -9,8 +9,6 @@ import (
 	"fmt"
 
 	abstractions "github.com/microsoft/kiota-abstractions-go"
-	betamodels "github.com/microsoftgraph/msgraph-beta-sdk-go/models"
-	"github.com/microsoftgraph/msgraph-beta-sdk-go/reports"
 	"github.com/microsoftgraph/msgraph-sdk-go/devices"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
 	"github.com/rs/zerolog/log"
@@ -52,13 +50,8 @@ func (a *mqlMicrosoftDevices) list() ([]any, error) {
 		return nil, err
 	}
 
-	betaClient, err := conn.BetaGraphClient()
-	if err != nil {
-		return nil, err
-	}
-
-	// Index of devices are stored inside the top level resource `microsoft`, just like
-	// MFA response. Here we create or get the resource to access those internals
+	// Index of devices are stored inside the top level resource `microsoft`, so
+	// we create or get the resource to access those internals.
 	mainResource, err := CreateResource(a.MqlRuntime, "microsoft", map[string]*llx.RawData{})
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


tas50:
Addressed:

- **critical (devices.go:107 → MFA regression)** — extracted the MFA registration fetch into a lazy `(*mqlMicrosoft).loadMfaResp` guarded by `sync.Once`. `users.list` still prefetches it so a batch query hits one API call for the whole map, but `mfaEnabled()` now also calls the loader, so `microsoft.user(id).mfaEnabled` works in isolation. (commit 8cea74030)
- **suggestion (ms365_exchange.go:79 → wider Select-Object)** — leaving for a follow-up. The named cmdlets (TransportRule, OwaMailboxPolicy, etc.) are audited across many fields, so a Select-Object would need to enumerate most of them and the savings would be marginal. Worth revisiting only if it shows up in profiling.
- **suggestion (ms365_sharepoint.go:176 → look for similar unwrapped errs in exchange)** — found one: `getExchangeReport` returned the raw err from `CheckAndRunPowershellScript` instead of routing through errHandler, so a transient PowerShell failure would leave `r.fetched` false and re-run the entire script on the next field access. Same fix pattern as sharepoint. (same commit)

PR #7917, opened 2026-05-24
🧹 remove outdated cnquery docs

arlimus:
- remove all stale cnquery docs from this local repo
- rename mondoo console to mondoo app


tas50:
These docs can mostly just get deleted at this point

arlimus:
updating, good catch!

PR #7960, opened 2026-05-26
✨ os: typed parent refs on nftables.chain, .rule, .set

tas50:
## Summary

Adds backwards-compatible cross-resource accessors so audits can traverse the nftables graph instead of carrying parent names around manually.

- `nftables.chain.tableRef -> nftables.table`
- `nftables.rule.tableRef  -> nftables.table`
- `nftables.rule.chainRef  -> nftables.chain`
- `nftables.set.tableRef   -> nftables.table`

The existing string fields (`table`, `chain`) are kept and re-described as the *name* of the parent, so prior queries continue to compile. New `.lr.versions` entries land at `13.18.1`.

### Why

Without typed refs, an auditor who has a `nftables.rule` cannot reach the policy of its base chain or the flags of its table without either re-fetching by name or hard-coding the parent context. The refs make queries like

```mql
nftables.rules.where(chainRef.policy == "accept" && chainRef.hook == "input")
```

natural — same shape as the typed refs in our cloud providers.

### Implementation

Each `*Ref` accessor calls `NewResource(...)` with just the parent's id fields. The runtime computes the same `__id` and returns the cached parent that `tables()` already populated when the user reached the rule/chain/set. No `init` is needed because every user-facing access path to chains/rules/sets transitively goes through `nftables.tables()` first.

## Test plan

- [ ] `make providers/build/os && make providers/install/os` succeeds
- [ ] Existing `TestParseNft*` / `TestNft*` unit tests pass (`go test ./providers/os/resources/`)
- [ ] On a Linux host with `nft` installed:
  - [ ] `mql run local -c "nftables.rules { chainRef.policy chainRef.hook tableRef.flags }"` returns data
  - [ ] `mql run local -c "nftables.chains.where(tableRef.family == \"inet\")"` filters correctly
  - [ ] Identity check: `nftables.tables[0] == nftables.chains[0].tableRef` when the chain is in the first table (i.e. the typed ref resolves to the cached parent, not a stub)
- [ ] Existing queries using the old string fields (`rule.table`, `rule.chain`, etc.) still compile and run

### DIFF
--- a/providers/os/resources/nftables.go
+++ b/providers/os/resources/nftables.go
@@ -259,6 +259,41 @@ func (s *mqlNftablesSet) id() (string, error) {
 	return s.Family.Data + "/" + s.Table.Data + "/" + s.Name.Data, nil
 }
 
+func (c *mqlNftablesChain) tableRef() (*mqlNftablesTable, error) {
+	return nftLookupTable(c.MqlRuntime, c.Family.Data, c.Table.Data)
+}
+
+func (r *mqlNftablesRule) tableRef() (*mqlNftablesTable, error) {
+	return nftLookupTable(r.MqlRuntime, r.Family.Data, r.Table.Data)
+}
+
+func (r *mqlNftablesRule) chainRef() (*mqlNftablesChain, error) {
+	raw, err := NewResource(r.MqlRuntime, "nftables.chain", map[string]*llx.RawData{
+		"family": llx.StringData(r.Family.Data),
+		"table":  llx.StringData(r.Table.Data),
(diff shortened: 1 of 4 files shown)


tas50:
Addressed: the four new `.lr.versions` entries (`nftables.chain.tableRef`, `nftables.rule.chainRef`, `nftables.rule.tableRef`, `nftables.set.tableRef`) are now at `13.17.1` — I had been reading the provider version from the `version/os-13.18.0` release branch instead of `main` (13.17.0).

PR #7987, opened 2026-05-26
✨ aws: discover SES email identities and configuration sets

tas50:
## What

Adds the `aws.ses` namespace, surfacing the account's **email-sending security posture** across all enabled regions via the SES v2 API. SES was the most conspicuous gap in AWS coverage — there was previously no way to audit DKIM, MAIL FROM, or delivery-TLS configuration.

### New resources

**`aws.ses`** — namespace with regional fan-out (mirrors `aws.mq`):
- `identities()` — every verified sending identity
- `configurationSets()` — every configuration set

**`aws.ses.identity`** — a verified domain, managed domain, or email address. Selectable by `arn`:
- `verificationStatus`, `verifiedForSending`, `identityType`
- **DKIM**: `dkimSigningEnabled`, `dkimStatus`, `dkimSigningAttributesOrigin`, `dkimSigningKeyLength`, `dkimTokens`
- **Custom MAIL FROM**: `mailFromDomain`, `mailFromDomainStatus`, `mailFromBehaviorOnMxFailure`
- `feedbackForwardingEnabled`, `policies` (sending-authorization policies), `tags`

**`aws.ses.configurationSet`** — reusable send-rule group. Selectable by `name`:
- `tlsPolicy` (REQUIRE / OPTIONAL — the minimum TLS enforced on delivery)
- `sendingEnabled`, `reputationMetricsEnabled`, `sendingPoolName`
- `suppressedReasons` (BOUNCE / COMPLAINT)
- `trackingRedirectDomain`, `trackingHttpsPolicy`, `tags`

### Notes

- Identity and configuration-set detail is read **lazily** (`GetEmailIdentity` / `GetConfigurationSet`) with double-check locking and per-resource caching, so iterating the lists doesn't trigger N+1 detail calls.
- `.lr.versions` entries are all at `13.28.1` (next patch after the current `13.28.0`).
- The permissions extractor learned the `Sesv2` connection method and a `sesv2 → ses` IAM-prefix override, so the four new actions are captured: `ses:ListEmailIdentities`, `ses:GetEmailIdentity`, `ses:ListConfigurationSets`, `ses:GetConfigurationSet`.

## Testing

Verified against a live AWS account (`us-east-1`): created a throwaway domain identity and a configuration set, confirmed every field populates correctly (DKIM tokens, `tlsPolicy: REQUIRE`, suppressed reasons, reputation/sending flags, tags), confirmed both selectors (`aws.ses.identity(arn:…)`, `aws.ses.configurationSet(name:…)`) resolve via their init functions, then tore the test resources down.


tas50:
Addressed the review feedback in 2bc1f2a:

- **Config-set `__id`** now uses the real ARN (`arn:aws:ses:<region>:<account>:configuration-set/<name>`) instead of the synthetic `region/name`, and an `arn` field was added — consistent with `aws.ses.identity` and most AWS resources. Init resolves by ARN (globally unique) or by name.
- **`@defaults`** for `aws.ses.configurationSet` is now `name region` (both eager from the listing), so iterating the collection no longer fires a `GetConfigurationSet` per row to render defaults.
- **Init disambiguation**: documented that name lookups are region-scoped (first regional match wins) while ARN lookups are unambiguous.
- **Access-denied path**: `fetchDetails()` already logs at warn level on access-denied (the suggested minimum), matching the existing `aws.mq` detail-fetch pattern.

Re-verified against live AWS: ARN-keyed listing, and both `configurationSet(arn:…)` and `configurationSet(name:…)` selectors resolve to `tlsPolicy: REQUIRE`.

tas50:
Addressed the re-flagged access-denied finding in 3a0acd6:

Both `mqlAwsSesIdentity.fetchDetails` and `mqlAwsSesConfigurationSet.fetchDetails` now mark every lazily fetched field as `StateIsSet | StateIsNull` in the access-denied branch (via `markIdentityDetailsNull` / `markConfigurationSetDetailsNull`), so the runtime treats them as known-unavailable rather than unresolved — no re-triggering, no confusing zero values.

The `.lr.versions` suggestion is already satisfied: all new SES entries are at `13.28.1` (next patch after the provider's current `13.28.0`).

PR #7993, opened 2026-05-27
✨ gcp: discover cloud domains registrations

tas50:
## Summary

Adds Google **Cloud Domains** to the GCP provider — domains registered through Cloud Domains, with the configuration that matters for domain security.

- `gcp.project.cloudDomains()` → `gcp.project.cloudDomainsService` (`enabled`, `registrations()`)
- `…cloudDomainsService.registration` — `domainName`, `state`, `expireTime`, `created`, `labels`, open `issues`, `domainProperties`, WHOIS `contactPrivacy`, and the domain-security config:
  - `transferLockState` / `effectiveTransferLockState` — protection against unauthorized domain transfers
  - `renewalMethod` / `preferredRenewalMethod`
  - `nameServers` + `dnsProvider` (`customDns` vs `googleDomainsDns`)
  - `dnssecState` and the signing `dsRecords` (`[]dict` of keyTag/algorithm/digestType/digest)

## Implementation notes

- **Listing:** uses the `locations/-` wildcard for `registrations.list` (one aggregated call; Cloud Domains registrations live in the `global` location). Each registration's `location` is parsed from its resource name.
- **Flattening:** management/DNS/contact settings are single nested structs with no IDs, so per the sub-resource guidance their scalars are flattened onto the registration; `dsRecords` is `[]dict`.
- **No `go.mod` change:** `google.golang.org/api/domains/v1` ships with the existing `google.golang.org/api` dependency.
- New permission: `domains.registrations.list`.
- One incidental churn in `gcp.permissions.json`: an `action` metadata hint flipped `Folders.Search` → `Folders.List` (same permission `resourcemanager.folders.list`). Pre-existing generator dedup nondeterminism surfaced by the new entry; harmless.

## Verification

Built and installed the provider. The target project had Cloud Domains **disabled** (no registrations to purchase for a test), so I enabled the API, confirmed the `locations/-` aggregated list call returns a clean empty result (HTTP 200, no 400s, no error spam), then disabled it again to restore state. Service group, `enabled` detection, and the list path are verified; field mapping mirrors the live-verified Dataplex pattern (#7991).

- ✅ gofmt clean · ✅ generated code consistent · ✅ `go.mod` clean · ✅ gcp module builds · ✅ `TestGCPPermissions` passes


tas50:
Rebased onto latest `main` (resolved conflicts with the merged Workflows PR #7995) and addressed the review:

- **`permissions.json` `Folders.Search` → `Folders.List`** (🟡): reverted — the manifest diff is now Cloud-Domains-only (`domains.registrations.list`). This flip is pre-existing generator nondeterminism (the dedup sorts by permission+sourceFile but ignores `action`, so an unstable sort arbitrarily picks one of the two `resourcemanager.folders.list` details). I'll open a separate PR to make the extractor's sort deterministic so this stops recurring on every GCP PR.
- **`googleDomainsDns` legacy branch** (🔵): added a comment noting it covers legacy/migrated registrations from before Google Domains moved to Squarespace.
- **`dsRecords` nil on error** (🔵): now defaults to `[]any{}` when `JsonToDictSlice` fails, matching the disabled-service pattern.

tas50:
Follow-through on the `Folders.Search`/`Folders.List` churn: opened #7996 to make the extractor's dedup deterministic (adds Action/Scope as sort tiebreakers). Once it lands, this class of unrelated `action` churn won't recur on GCP/Azure PRs.

PR #8033, opened 2026-05-27
✨ helm: render fidelity, schema enrichment, and remote chart fetch

tas50:
Expands the **Helm provider** (experimental, `13.1.0`) toward near-complete coverage so users can write complex queries against their charts in CI. New `.lr.versions` entries are `13.1.1`; provider `Version` is unchanged.

## WS1 — Render fidelity
Render config now lives on the connection and is driven by new CLI flags, reusing Helm's own value-merge semantics (`pkg/strvals`):
`-f/--values`, `--set`, `--set-string`, `--set-json`, `--set-file`, `-n/--namespace`, `--release-name`, `--kube-version`, `-a/--api-versions`, `--is-upgrade`. Rendering passes a real `chartutil.Capabilities` (target kube version + api-versions) instead of `nil`.

- `values()` is unchanged (bundled `values.yaml`) to avoid a silent behavior break; new **`renderedValues()`** exposes the coalesced defaults+overrides actually used to render.

## WS2 — Schema enrichment
- **`helm.chart`**: `renderedValues`, `notes`, `crds[]`, `valuesSchema`, `hooks[]`, `lock`, `lint`
- **New**: `helm.chart.dependencyLock` (+`dependencies`), `helm.chart.lintResult`, `helm.chart.lintMessage`
- **`helm.dependency`**: `importValues`, `resolvedVersion` (from `Chart.lock`), `chart` (link to vendored subchart)
- **`helm.resource`**: `isHook`, `hookTypes`, `hookWeight`, `hookDeletePolicies`, `isCRD`
- **`helm.template`**: `requiresCluster` (AST-detected `lookup`)
- **`helm.file`**: `size`, `isBinary`

The lock resource is named `helm.chart.dependencyLock` rather than `helm.chart.lock` — the latter would collide with the chart's `lock` field path (dotted-subresource-husk). `helm.resource` is intentionally left shaped so a typed-k8s accessor can slot in once cross-provider resource returns exist in MQL; no k8s types are re-modeled here.

## WS3 — Remote chart fetch
- `oci://` charts via `pkg/registry`; `http(s)` `.tgz`; and `chartName --repo <url>` (repo `index.yaml` resolved over `net/http`). Flags: `--repo`, `--version`, `--username`, `--password`, `--repository-config`, `--repository-cache`.
- **Dependency-stack note:** fetch deliberately avoids `pkg/getter` / `pkg/repo` / `pkg/downloader` / `pkg/cli`, which transitively pull `pkg/kube` → kubectl → `k8s.io/api/scheduling/v1alpha1` — removed in our pinned `k8s.io/api v0.36.1`, making `go mod tidy` unsatisfiable. Value/`--set` merging uses `pkg/strvals` + `net/http` for the same reason.
- **`--dependency-update` was intentionally not added** (it lives in `pkg/downloader`/`Manager` and would drag in the incompatible kube stack). Vendored `charts/` still load, so run `helm dependency build` first if you need remote deps resolved.

## Verification
- `go build`, `go vet`, full test suite, `gofmt`, `golangci-lint` (0 issues); codegen is deterministic.
- New tests: pure-function units (hook parsing, value merge, `lookup` detection, lint severity, render-option encode/decode) + provider `GetData` round-trips against a new `featurechart` fixture (hooks, CRDs, `Chart.lock`, values schema, `lookup`, NOTES, value-override and release-identity rendering).
- Live MQL confirmed `hooks`, `crds`, `lock.dependencies.resolvedVersion`, `notes`, `renderedValues` (with `--set`), `templates.where(requiresCluster)`, and an OCI pull of `bitnami nginx 18.2.5`.


tas50:
Addressed in 25290cda1:

- **HTTP timeouts + size caps** — `readValueFile` (resources/helm.go) now uses a 30s-timeout client with a 10 MiB `io.LimitReader` cap; `httpGetBytes` (connection/fetch.go) uses a 60s-timeout client with a 128 MiB cap (chart archives run larger than values files). Both stop a slow/oversized remote from hanging the query or exhausting memory.
- **Doc-comment mismatch** — moved the orphaned `newMqlHelmDependency` godoc off `renderedValues()` and back onto its own function.

Build/vet/golangci-lint (0 issues)/tests all green.

PR #8078, opened 2026-05-31
⚡ providers: defer provider schema parsing to speed up CLI startup

tas50:
## Problem

Every `mql`/`cnspec` invocation eagerly read and JSON-parsed the `.resources.json` schema of **every installed provider** before doing anything useful — just to build the CLI command tree. With ~60 providers installed that is **multiple megabytes** of JSON (and growing with every release) unmarshalled into Go structs on the cold path, even though a command only ever connects to a **single** provider.

Worse, it happened **twice**:
- `readProviderDir()` → `MustLoadSchemaFromFile()` (parse #1)
- the `ListAll()` loop → `provider.LoadResources()` (parse #2, overwriting #1)

The lazy machinery to load a schema on demand already existed — `(*coordinator).LoadSchema` and the connect path both call `LoadResources()` when `Provider.Schema == nil` — but the eager preload defeated it.

## Change

Stop pre-parsing schemas in `ListAll()`/`readProviderDir()`. Schemas now load lazily the moment a provider is actually used:
- **Connecting** to a provider already loads its schema on demand (`coordinator.go`).
- `installDependencies()` lazily loads the **one** provider's schema it needs to inspect dependencies (builtins have no file-backed schema and are skipped).
- `LoadResources()` now unmarshals into a concrete `*resources.Schema` — `json.Unmarshal` cannot target a nil `ResourcesSchema` interface field, which previously only worked because the field was pre-populated.

Only `LoadJSON()` (the small per-provider connector config, ~tens of KB total) is still read up front, since the CLI needs each connector's flags.

## Impact

Measured with ~60 providers installed (`mql run local -c 'asset.platform'`):

| Metric | Before | After | Improvement |
|---|---|---|---|
| Cold startup (median wall-clock) | ~0.71s | ~0.61s | **~14% faster** |
| Peak RSS | ~84 MB | ~70 MB | **~16% lower (multiple MB)** |

The savings **scale with the number of installed providers** — each unused provider's schema is no longer parsed on startup nor held in memory. As schemas keep growing, the absolute win grows with them.

A CPU profile of the patched binary confirms `LoadResources`/`json.Unmarshal` now runs **once** (for the active provider, which genuinely needs its schema) instead of ~60 times.

## Testing

- `go test ./providers/` passes.
- Verified real queries still resolve correctly (`asset.platform`, `command('echo hi').stdout`) — the lazy connect path loads the active provider's schema as expected.
- Verified the cross-provider fallback (`AllResources`/`Lookup` → `unsafeLoadAll`) still works for autocomplete/lookup paths.

### DIFF
--- a/providers/providers.go
+++ b/providers/providers.go
@@ -314,13 +314,13 @@ func ListAll() ([]*Provider, error) {
 			continue
 		}
 
-		if err := provider.LoadResources(); err != nil {
-			log.Error().Err(err).
-				Str("provider", provider.Name).
-				Str("path", provider.Path).
-				Msg("failed to load provider resources")
-			continue
-		}
+		// Note: we intentionally do NOT parse the provider's resource schema
+		// here. Schemas are large (multiple megabytes across all installed
+		// providers, and growing) and are loaded lazily the moment a provider is
+		// actually connected (see (*coordinator).LoadSchema / Connect) or its
+		// dependencies are inspected (installDependencies). Parsing every
+		// provider's schema up front just to build the CLI wastes startup time
+		// and memory for providers we never use.
 
 		res = append(res, provider)
 	}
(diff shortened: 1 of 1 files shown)


arlimus (approved):
Nice improvement, lazy loading for resources makes sense and decouples it in the provider loading. We indeed only need the provider conf in the bootup of the CLI for the most part, not the resources right away

PR #8126, opened 2026-06-03
🛑 openstack: typed cross-resource references; drop redundant raw ID fields

tas50:
## What

Replaces raw ID/UUID string fields with **typed resource accessors** across the OpenStack provider, so MQL users can traverse relationships directly (`openstack.ports.where(deviceOwner == /compute/) { server.name }`) instead of manually correlating IDs. Where a typed accessor fully replaces a raw string, the raw field is **removed** — hence the breaking-change marker.

## Typed accessors added (and raw fields dropped)

| Resource | Added | Dropped |
|---|---|---|
| `compute.server` | `project()`, `user()` | `projectId`, `userId` |
| `compute.serverGroup` | `project()`, `user()` | `projectId`, `userId` |
| `compute.keypair` | `user()` | `userId` |
| `network` / `subnet` / `router` / `floatingIp` / `securityGroup` / `securityGroup.rule` | `project()` | `projectId` |
| `port` | `project()`, `subnets()`, `server()`, `router()` | `projectId`, `deviceId` |
| `containerinfra.cluster` | `user()`, `stack()` | `userId`, `stackId` |
| `containerinfra.clusterTemplate` | `project()`, `user()` | `projectId`, `userId` |
| `baremetal.node` | `owner()`, `lessee()` | `owner`/`lessee` strings (string → `openstack.project`) |

### Notes
- `port.server()` / `port.router()` resolve `deviceId` by `deviceOwner` prefix (`compute:` → server, `network:router` → router). `deviceOwner` is kept as the discriminator; `port.subnets()` derives from `fixedIps[].subnet_id`.
- `securityGroup.project()` is null on the cross-project rule-reference fallback path (that init path can't populate the cache); the normal list path resolves fine.
- New accessors versioned at `13.2.1`. `baremetal.node.owner`/`lessee` bumped to `13.2.1` too, since the type changed on an existing field path. `config.go` `Version` untouched (handled by the release flow).

## Also included
A docs commit adds a **Step 1.5 typed-reference gate** to the resource-dev lifecycle in `CLAUDE.md`, between writing `.lr` and codegen — to stress catching this the first time, since fixing it later is exactly this breaking change.

## Testing
- `go build ./...`, `go vet ./resources/...`, `go test ./resources/...` all green; codegen run twice (new `Internal` structs embedded); `gofmt` applied.
- Not yet exercised against a live OpenStack cloud.


tas50:
Addressed in d1273525d:

- **`securityGroup.project()` null on cross-project Get path** (changes requested): `initOpenstackSecurityGroup`'s direct-Get fallback now creates the resource explicitly, primes `cacheProjectID` from `sg.ProjectID`, and returns it — so `project()` resolves on that path too. (Rules built in `buildSecurityGroupRules` already had their cache set, so only the parent SG needed it.)
- **Inline `project()`/`user()` duplication** (suggestion): replaced the inline bodies in `compute.go` and `network.go` with the existing `resolveProject`/`resolveUser` helpers — net −110 lines, single fix-point, consistent with `magnum.go`/`ironic.go`. `port.server()`/`router()` and `cluster.stack()` stay inline since no resolve helper exists for those types.

`go build ./...`, `go vet`, and `go test ./resources/...` all green.

PR #8135, opened 2026-06-03
✨ aws: add Bedrock AgentCore resources (MCP gateways, agents, identity)

tas50:
## What

Adds an `aws.bedrock.agentCore` namespace exposing the agentic-AI infrastructure introduced by **Amazon Bedrock AgentCore**, so users can inventory and audit the MCP servers, hosted agents, and outbound-auth configuration in their AWS accounts. This is the first of three AI-coverage PRs (AgentCore → Amazon Q → Bedrock field enrichment).

### Resources

| Resource | What it models |
|---|---|
| `gateway` | MCP gateways — endpoint URL agents connect to, protocol type, inbound authorizer, IAM role, KMS key |
| `gatewayTarget` | Tool targets federated through a gateway (Lambda / OpenAPI / Smithy) with their backing config and outbound credential-provider configs |
| `runtime` | Hosted agents — IAM role, network mode (public/VPC), inbound authorizer, environment variables |
| `runtimeEndpoint` | Addressable endpoints serving agent invocations (live/target version) |
| `memory` | Agent memory stores |
| `browser` / `codeInterpreter` | Built-in tool sandboxes |
| `oauth2CredentialProvider` / `apiKeyCredentialProvider` / `workloadIdentity` | AgentCore Identity — outbound auth |

### Why these are security-relevant

MCP gateways and credential providers are new attack surface (tool injection, over-broad outbound auth) that nothing else in the provider inventories. The gateway's inbound `authorizerType`, the runtime's `networkMode`, and the Identity credential providers are the signals an audit cares about.

## Design notes

- `aws.bedrock.agentCore` is a self-sufficient namespace reached via a new `agentCore()` accessor on `aws.bedrock`; its list accessors fan out across regions via the existing `jobpool` pattern.
- Typed references (`iamRole` → `aws.iam.role`, `kmsKey` → `aws.kms.key`) enable MQL traversal.
- Detail-only fields (gateway URL, IAM role, runtime network mode, target configuration, …) are lazy-loaded via `Get*` calls behind a `sync.Once`, mirroring the existing `aws.bedrock.agent` implementation. Each list item is keyed by its ARN (gateway/target are keyed by their AgentCore IDs, since the list APIs don't return ARNs).
- The AWS permissions extractor now knows the `bedrockagentcorecontrol` SDK package and maps it to the correct **`bedrock-agentcore`** IAM prefix. The 13 required `bedrock-agentcore:*` actions are reflected in `aws.permissions.json`.
- New `.lr.versions` entries are at `13.31.1` (next patch after the current `13.31.0`).

## Testing

- `make providers/build/aws` + `go vet ./resources/` clean.
- Queried every list accessor against a live AWS account — schema compiles, all paths execute without error (the account has no AgentCore resources, so results are empty; the lazy/typed-ref fields compile but were not exercisable against real data).


tas50:
Addressed the review in 69ccce6:

- 🔴 **Gateway cache collision** — the gateway `__id` (and the gateway-target `__id`) is now region-prefixed (`<region>/<gatewayId>[/target/<targetId>]`), so identical service-scoped IDs across regions no longer collide.
- 🔵 **Synthetic `id` field** — dropped `id string` from `aws.bedrock.agentCore.gateway`; the GatewayId is cached internally for API calls and `arn()` is the user-facing identifier.
- 🟡 **`collectJobs` name** — now a method on `aws.bedrock.agentCore`, so no package-level collision risk.
- 🟡 **Swallowed error** — `targetConfiguration()` now returns the `convert.JsonToDict` error.
- 🟡 **`sync.Once` caches transient errors** — switched the three detail fetches to the double-check `fetched`+`Mutex` pattern from CLAUDE.md so transient errors retry on the next access.

tas50:
Rebased onto latest `main` and addressed the second review round:

- 🟡 **`workloadIdentityArn` → typed reference** — added `workloadIdentity() aws.bedrock.agentCore.workloadIdentity` on the gateway as the primary accessor (raw `workloadIdentityArn` kept since the bare ARN is still useful). Added `initAwsBedrockAgentCoreWorkloadIdentity`, which resolves a workload identity from its ARN alone — region and name are both encoded in the ARN, so no extra API call is needed.
- 🟡 **`targetConfiguration()` nil dict state** — no change needed. `dict` is a plain value type, so its generated accessor goes through `GetOrCompute[any]`, which sets the `TValue` state itself after the function returns; `return nil, nil` is correctly stored as a set-null dict. The manual `StateIsNull` flag is only required on the resource-pointer path (which doesn't use `GetOrCompute`). In practice `detail` is also never nil after a successful `GetGatewayTarget`.
- 🔵 **gateway `__id` from ARN** — `ListGateways` doesn't return the ARN (it's fetched lazily via `GetGateway`), so the region-prefixed `<region>/<gatewayId>` key stays. As noted, this is the acceptable fallback.
- 🔵 **`runtimes()` `createdAt`** — the SDK `AgentRuntime` list type has no `CreatedAt`, only `LastUpdatedAt` (mapped to `updatedAt`). Correctly omitted; nothing to add.

New `.lr.versions` entry (`aws.bedrock.agentCore.gateway.workloadIdentity`) is at `13.31.2`, consistent with its AgentCore siblings. `make providers/build/aws` + `go vet ./resources/` clean; codegen produces no further diff.

PR #8176, opened 2026-06-04
🌟 ansible: project-wide static analysis (roles, inventory, galaxy, cfg, vault)

tas50:
## What

Turns the ansible provider from single-playbook analysis into whole-**project** analysis, then deepens that model. Point the provider at a directory to model the entire codebase; point it at a file and nothing changes.

A directory connects as an `ansible-project` asset with an `ansible.project` root resource.

### Project structure
| Accessor | Exposes |
|---|---|
| `playbooks` | Every playbook file (`ansible.playbook` → `plays`) |
| `roles` | Full role trees: `tasks`, `handlers`, `defaults`, `vars`, `meta`, `argumentSpecs`, `templates`, `files`, `dependencies` (typed refs) |
| `inventory` | Static INI **and** YAML inventory plus `group_vars`/`host_vars` (root **and** beside the inventory source), as typed `groups`/`hosts` |
| `requirements` | Galaxy `roles`/`collections` declared in requirements.yml |
| `collections` | Collections **vendored** under `collections/ansible_collections/` (drift vs. declared) |
| `plugins` | Custom modules/plugins (`library/`, `*_plugins/`, `plugins/<type>/`) — supply chain |
| `manifest` | The project's own `galaxy.yml` when it is a collection |
| `config` | `ansible.cfg`, with `hostKeyChecking`/`become`/`becomeUser` surfaced typed |
| `vault` | Fully-encrypted `files` (with `vaultId`) **and** inline `!vault` `variables` |
| `lintConfig`, `moleculeScenarios` | Quality/test-maturity signals |

### Deeper task & play modeling
- `ansible.task.module` / `moduleArgs` — the module a task invokes and its args, so audits select by module without parsing the `action` dict: `tasks.all(module != /command|shell/)`.
- `ansible.play.roleApplication` — how a play applies each role (name, `when`, `tags`, `vars`) plus the resolved `role`. **Fixes a parse failure**: parameterized roles (`roles: [{role: x, when: ...}]`) previously broke decoding.
- Play-level `varsFiles`, `varsPrompt`, `environment`, `collections`.

### Reference resolution
`play.roleApplications.role`, `role.dependencies`, and `task.importedTasks`/`importedPlaybook` resolve to parsed resources; Jinja2/dynamic paths fall back to the raw string fields.

## Security & robustness
- **Path traversal (fixed):** include/import resolution rejects absolute references and any path that escapes the analysis root via `..`, so a crafted playbook cannot make the provider read arbitrary host files (e.g. `import_tasks: /etc/shadow`). Resolution is confined to the project root (directory mode) or the playbook's directory (file mode). Covered by `TestResolveRefPathContainment`.
- **Symlink containment (fixed):** inventory sources and `group_vars`/`host_vars` files whose real path escapes the project root via a symlink are skipped. Covered by `TestSymlinkVarsEscapeSkipped`.
- **Integer dict serialization (fixed):** dict fields sourced from YAML stored Go `int`, which llx's dict→primitive conversion rejects (`unsupported child type: int`) — **any** integer in a `vars`/`defaults`/`action`/`moduleArgs` dict broke rendering, including the original provider's `play.vars` and `task.action`. Values are normalized to `int64`/`float64`; `uint`/`uint64` above `math.MaxInt64` widen to `float64` rather than overflowing.
- `ansible.cfg` CRLF is already handled (every line is `TrimSpace`'d before use); `.lr` gained a trailing newline.

## Backward compatibility
A bare file path still produces the `ansible-playbook` asset with the `ansible { plays }` model. Policies filtering on `asset.platform == 'ansible-playbook'` are unaffected.

## Implementation
- New `project/` package: directory walk + per-artifact parsers (roles, inventory, galaxy, collections, plugins, ansible.cfg, vault), each small and independently testable. No new dependencies (hand-rolled INI/cfg on `yaml.v3`).
- Connection branches on a `stat`: file → `play.Playbook`, directory → `project.Project`.
- New `.lr.versions` entries at `13.1.2`.

## Testing
- Unit tests over a representative fixture project (`resources/testdata/project/`) covering every artifact type, parameterized roles, module/moduleArgs, inline vault, inventory-adjacent vars, plugins, vendored collections, manifest, argument_specs, tooling signals, path-traversal containment, and symlink-escape skipping.
- Verified end-to-end against the installed binary in project and file mode.
- `golangci-lint` (0 issues), `go vet`, `gofmt`, and the full provider suite pass.


arlimus:
This is a huge addition, kudos @tas50 !!

PR #8189, opened 2026-06-05
🧹 providers: test that every local provider is registered in defaults.go

tas50:
## What

Adds `TestDefaultProvidersIncludesAllLocalProviders` to the `providers` package: it scans the `providers/` directory for every provider (identified by `config/config.go`) and asserts each one has an entry in the `DefaultProviders` map in `providers/defaults.go`.

## Why

`providers/defaults.go` is auto-generated by `make providers/defaults`, but it's committed to the repo — so it's easy to add a new provider directory and forget to regenerate it. That's exactly what happened before #8187, which had to retroactively register `bicep`, `helm`, and `kustomize`. This test is the missing tripwire that catches the omission in CI instead of after release.

## Behavior

- Discovers providers by directory scan (no hardcoded allowlist), so new providers are covered automatically.
- Extra entries for **external** providers that don't live in this repo (e.g. `ai`, `bigip`, `junos`, `fortios`, `panos`) are allowed — the test only requires that every *local* provider is present.
- A sanity assertion ensures the scan actually found providers, so a refactor can't silently turn it into a no-op.
- On failure the message points at the fix: `run \`make providers/defaults\``.

## Verification

- Passes against the current tree.
- Confirmed it fails with a clear message when a local provider is removed from the map (temporarily deleted the `bicep` entry to prove it isn't a no-op), then passes once restored.
- `gofmt`-clean; `defaults.go` is unmodified.

### DIFF
--- a/providers/defaults_test.go
+++ b/providers/defaults_test.go
@@ -1,0 +1,52 @@
+// Copyright Mondoo, Inc. 2024, 2026
+// SPDX-License-Identifier: BUSL-1.1
+
+package providers
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestDefaultProvidersIncludesAllLocalProviders guards against the recurring
+// mistake of adding a provider to ./providers without regenerating
+// defaults.go (see PR #8187, which had to retroactively register bicep, helm,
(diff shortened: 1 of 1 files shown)


arlimus (approved):
kudos for the gate!

PR #8199, opened 2026-06-05
✨ terraform: typed resources + per-argument reference provenance for HCL

tas50:
## What

A typed, cross-referenceable, value-aware layer over Terraform **HCL** — the connection type most policy is written against and previously the least typed. Three passes, each removing a class of by-hand data-walking from real `cnspec` content.

Everything is **additive**. No existing field or resource changes behavior; policies using `arguments[...]`, `.first[...]`, `blocks.where(type==...)`, `labels[...]`, `terraform.resources(...)` keep working byte-for-byte. New `.lr.versions` entries are `13.1.1`.

## Part 1 — Typed resources + per-argument reference provenance

New resources (each wraps a block with a `block()` back-link): `terraform.resource`, `terraform.datasource`, `terraform.variable`, `terraform.output`, `terraform.provider`, `terraform.reference`. Root accessors `terraform.managedResources` / `.dataResources` / `.variableDefinitions` / `.outputValues` / `.providerConfigs`.

`terraform.reference` records per-argument provenance — `argument`, `kind` (var/local/data/resource/module/output/provider/…), `target`, `name`, and a `block()` that resolves within the parsed config (handling the `data.*` vs managed-resource namespace split). The walker classifies every traversal an argument makes, including inside `for`/conditional/splat/index expressions.

Typed cross-references wherever a value names another block: `dependsOn` → blocks, `reference.block`, `resource`/`datasource`.`provider()` (matched by name **and** alias), `provider.requirement()` → version constraint.

## Part 2 — Structured data instead of walking

| Field | Replaces |
|---|---|
| `nestedBlocks` — `map[string][]terraform.block` by type | `blocks.where(type == "ingress")` ×N |
| `config` — `jsonencode` docs decoded, not list-wrapped | `arguments["policy"].first["Statement"]` |
| `preventDestroy` / `createBeforeDestroy` / `ignoreChanges` | `blocks.where(type == "lifecycle")…` |
| `referencedBy` — cached reverse index, recurses nested blocks | dead-config / dependency fan-out |

## Part 3 — Values and selection

| Field | What it does |
|---|---|
| `resolved` | **Effective values.** Evaluates `var` defaults (overridden by `tfvars`), `local`s (resolved to a fixpoint), and Terraform functions (`merge`/`concat`/`format`/`lookup`/…) into the value actually deployed. Falls back to the `config` ref form when a value depends on a data source/resource. On block/resource/datasource. |
| `tree` | The block as one walkable document — arguments plus child blocks keyed by type — so `tree["ingress"][0]["cidr_blocks"]` needs no accessor hops. |
| `terraform.module.sourceType` / `inputs` / `references` + `terraform.moduleCalls` | Module model from source HCL (no `terraform init` needed): source classification, input values, and references. |
| `terraform.resource(type:, name:)` / `terraform.datasource(type:, name:)` | Typed selection by type/name; a no-match returns a safe empty record. |

The expanded function set used by `resolved` is **separate** from the one backing `arguments`/`config`, so those keep their shape.

## Before / after on real `cnspec` content

Each pair below is taken from the queries we ship in `cnspec` content (mostly `mondoo-aws-security`) and was run against sample configs to confirm the **before and after return identical verdicts**. Examples 7–10 are newly expressible — the prior form returns the wrong answer or can't be written at all.

### Replace the cross-resource string-join

A resource that decorates another (a flow log for a VPC, an encryption/versioning/logging config for a bucket) was joined by building a CSV of references and substring-matching it against a dot-padded label (`"." + labels[1] + "."`). The reverse reference index replaces the whole dance with a real relationship.

```ruby
# 1. VPC flow-log coverage
# before
flowVpcs = terraform.resources("aws_flow_log")
  .where(arguments["traffic_type"] == "ALL" || arguments["traffic_type"] == "REJECT")
  .where(arguments["vpc_id"] != null).map(arguments["vpc_id"]).join(",")
terraform.resources("aws_vpc").all(flowVpcs.contains("." + labels[1] + "."))
# after
terraform.managedResources.where(type == "aws_vpc").all(
  referencedBy.where(resourceType == "aws_flow_log")
    .any(arguments["traffic_type"] == "ALL" || arguments["traffic_type"] == "REJECT"))
```

The same shape collapses the doubly-nested S3 encryption check (string-join **and** `blocks.where(...).all(blocks.where(...))`):

```ruby
# 2. Every S3 bucket encrypted with KMS
# before
kmsBuckets = terraform.resources("aws_s3_bucket_server_side_encryption_configuration")
  .where(blocks.where(type == "rule").all(blocks.where(type == "apply_server_side_encryption_by_default")
    .all(arguments["sse_algorithm"] == "aws:kms" || arguments["sse_algorithm"] == "aws:kms:dsse")))
  .map(arguments["bucket"]).join(",")
terraform.resources("aws_s3_bucket").all(kmsBuckets.contains("." + labels[1] + "."))
# after
terraform.managedResources.where(type == "aws_s3_bucket").all(
  referencedBy.where(resourceType == "aws_s3_bucket_server_side_encryption_configuration").any(
    nestedBlocks["rule"].all(nestedBlocks["apply_server_side_encryption_by_default"].all(
      arguments["sse_algorithm"] == "aws:kms" || arguments["sse_algorithm"] == "aws:kms:dsse"))))
```

This single pattern covers a whole family of shipping checks that use `"." + labels[1] + "."` today: S3 bucket **versioning**, **logging**, and **ownership controls**; Route 53 zone **query logging** and **DNSSEC**; Secrets Manager **rotation**; and EFS **backup policy**.

### Read policy documents without `.first`

`jsonencode({...})` arguments came back wrapped in a single-element list, so every IAM-policy check (5+ of them across the AWS pack) carried a `.first`. `config` decodes them in place.

```ruby
# 3. IAM policy: no Allow on a wildcard resource
# before
terraform.resources("aws_iam_policy").where(arguments["policy"].first["Statement"] != null).all(
  arguments["policy"].first["Statement"].all(
    _["Effect"].downcase == "deny" || [_["Resource"]].flat.any(_.contains("*")) == false))
# after
terraform.managedResources.where(type == "aws_iam_policy").all(
  config["policy"]["Statement"].all(
    _["Effect"].downcase == "deny" || [_["Resource"]].flat.any(_.contains("*")) == false))
```

### Select nested blocks by type

`blocks.where(type == "...")`, repeated and re-filtered, becomes a keyed lookup.

```ruby
# 4. Security group: SSH not open to the world
# before
terraform.resources("aws_security_group")
  .where(blocks.where(type == "ingress").contains(arguments.to_port == 22))
  .none(blocks.where(type == "ingress") { arguments.cidr_blocks.contains("0.0.0.0/0") })
# after
terraform.managedResources.where(type == "aws_security_group").all(
  nestedBlocks["ingress"].where(arguments["to_port"] == 22)
    .all(arguments["cidr_blocks"].none(_ == "0.0.0.0/0")))
```

`tree` goes further, rendering the whole block (arguments + nested blocks) as one walkable document:

```ruby
terraform.managedResources.where(type == "aws_instance")
  .all(tree["metadata_options"][0]["http_tokens"] == "required")
```

### Meta-arguments as first-class fields

```ruby
# 5. Critical resource protected from deletion
# before
terraform.resources("aws_db_instance").all(
  blocks.where(type == "lifecycle").any(arguments["prevent_destroy"] == true))
# after
terraform.managedResources.where(type == "aws_db_instance").all(preventDestroy)
```

### A cross-reference that wasn't expressible

A configured provider and its version requirement live in different blocks; `requirement` resolves the join.

```ruby
# 6. Every configured provider is pinned to a version
terraform.providerConfigs.all(requirement.version != "")
```

### Newly expressible

```ruby
# 7. Assert on the effective value behind a variable / ternary.
#    instance_type = var.environment == "prod" ? "m5.large" : "t3.micro"
terraform.managedResources.where(type == "aws_instance").all(arguments["instance_type"] == "m5.large")  # false — unresolved expression
terraform.managedResources.where(type == "aws_instance").all(values["instance_type"]    == "m5.large")  # true  — var + ternary resolved

# 8. One policy across config, plan, and state — `values` is source-independent.
#    Returns [ok] value: true on an HCL dir, a plan file, and a state file.
terraform.managedResources.where(type == "aws_s3_bucket").all(values["acl"] != "public-read")

# 9. Find dead configuration via the dependency graph.
terraform.graph.orphans.where(type == "data")             # data sources nothing consumes
terraform.dataResources.where(referencedBy.length == 0)   # equivalent

# 10. Audit resources declared inside a module (transparency + per-module value flow).
#     module "logs" { source = "./modules/bucket"  bucket_name = "logs-${var.environment}" }
terraform.moduleCalls.where(key == "logs")[0].resources.all(values["bucket"] != null)
#     -> the module's bucket surfaces as module.logs.aws_s3_bucket.this with
#        values["bucket"] == "logs-prod" — the call input flowed across the boundary and resolved
terraform.moduleCalls.all(inputs.all(variable != null))   # every module input maps to a declared variable
```

Value resolution was verified end-to-end: a `var.environment` ternary resolves through its `dev`/`prod` (incl. tfvars override) defaults, a `local` boolean resolves, and `merge(local, var, literal)` resolves to the merged map while a `data.*` reference correctly falls back.

## Testing

- Unit tests: reference classification/resolution, `depends_on` lists, `ignore_changes` normalization, `jsonencode` config-unwrap (and that a genuine single-element list is **not** unwrapped), and module source classification — plus the existing `getCtyValue` suite. All green.
- Interactive verification of every new resource/field, including both `provider()` aliases, `requirement()`, `reference.block()`, `nestedBlocks`, `config`, lifecycle flags, `referencedBy` (incl. a reference buried in a nested `ingress` block), `resolved` (ternary/local/merge/tfvars + data fallback), `tree` deep indexing, the module model, and typed selection (found, type-only, and a not-found record stressed across all accessors with no panic).
- `go test ./...`, `go vet`, `gofmt`, `go mod tidy` clean; generated code regenerated and committed.

## Part 4 — Reimagined model: infrastructure graph, not syntax tree

The earlier parts expose Terraform's syntax with semantics bolted on. This part adds a parallel **semantic** surface — resources as infrastructure objects, unified across sources, with a real dependency graph and transparent modules. The syntax surfaces (`terraform.block`, `terraform.state.*`, `terraform.plan.*`, `terraform.resources`) are untouched.

### One model across HCL / plan / state
`terraform.managedResources` / `.dataResources` are now populated from whichever source backs the connection, and every resource exposes a single `values` view plus `address` and `mode`:

```ruby
# identical query, runs on config, a plan, or state — all return [ok] value: true
terraform.managedResources.where(type == "aws_s3_bucket").all(values["acl"] != "public-read")
```
`values` = resolved config (HCL) / planned after-values (plan) / recorded values (state). Verified returning the same verdict on all three sources.

### First-class dependency graph
```ruby
resource.dependencies            # transitive: everything this resource reaches
resource.dependents              # transitive: blast radius of changing it
terraform.graph.orphans          # resources/data nothing references (dead config)
terraform.graph.edges            # { from, to, argument } across the whole config
```

### Module transparency (with value flow)
Resources inside modules now appear in `managedResources` with fully-qualified addresses and a `module` back-link, and evaluation is **per-module** — a module's variables use its own defaults overridden by the call's inputs:

```
# config: module "logs" { source="./modules/bucket"  bucket_name="logs-${var.env}" }  (var.env="prod")
#         module/bucket: resource "aws_s3_bucket" "this" { bucket = var.bucket_name }
terraform.managedResources.where(type=="aws_s3_bucket")[0]
  # address -> "module.logs.aws_s3_bucket.this"
  # module.key -> "logs"
  # values["bucket"] -> "logs-prod"   <-- call input flowed across the module boundary and resolved
```
This also fixes the prior root/module variable-name collision in `resolved`.

All additive; new `.lr.versions` entries `13.1.1`. Verified: the cross-source policy on HCL/plan/state, transitive deps, graph edges/orphans, module association + qualified addresses + cross-module value flow; full `go test ./...` / `vet` / `gofmt` / `go mod tidy` clean.


providers/terraform/resources/terraform.lr
```diff
@@ -110,6 +514,19 @@ terraform.module @defaults("key source") {
   version string
   // Path to the directory where the module is stored
   dir string
+  // Where the module is loaded from
+  //
+  // One of local (a `./` or `../` path), registry (a Terraform registry
+  // address), git (a Git or GitHub source), http (an archive URL), or
+  // empty when the source is not set.
+  sourceType() string
```
arlimus, line 522: I think this works well for now. I'd love to have the source as an asset that I can navigate long-term (which would then make this accessible via `source` eg `source.asset` for its metadata).

providers/terraform/resources/terraform.lr
```diff
@@ -110,6 +514,19 @@ terraform.module @defaults("key source") {
...
+  // Input values passed to the module
+  //
+  // The module call's arguments excluding the meta-arguments source,
+  // version, count, for_each, depends_on, and providers.
+  inputs() dict
```
arlimus: wonder if this could be typed as well

providers/terraform/resources/terraform.lr
```diff
@@ -90,10 +164,340 @@ private terraform.block @defaults("type labels") {
...
+  // instead of filtering the flat child-block list. Each value is the
+  // list of child blocks of that type.
+  nestedBlocks() map[string][]terraform.block
+  // Normalized argument values
+  //
+  // Examine the same arguments as `arguments`, but with `jsonencode`
+  // policy documents and similar function results returned as their
+  // decoded object rather than wrapped in a single-element list, so a
+  // policy reads `config["policy"]["Statement"]` directly.
+  config() dict
```
arlimus, line 308: I'm curious about arguments, resolved, and config and how they might be used.

providers/terraform/resources/terraform.lr
```diff
@@ -90,10 +164,340 @@ private terraform.block @defaults("type labels") {
...
+  // Terraform functions evaluated into the value that will actually be
+  // deployed. Arguments whose value depends on a data source or another
+  // resource fall back to the reference form used by `config`.
+  resolved() dict
+  // Full configuration as nested data
+  //
+  // Examine the resource as one walkable document: its arguments plus its
+  // child blocks keyed by type, so a policy reads
+  // `tree["ingress"][0]["cidr_blocks"]` directly.
+  tree() dict
```
arlimus, line 321: is this kind of a raw version of the resource as a JSON? (like `params` in other areas)

arlimus (changes requested):
No specific changes requested yet, just flagging this for discussion until we understand everything.

providers/terraform/resources/terraform.lr
tas50: Built out a first step toward this. A module's source is now navigable — `terraform.module` gains `resources`, `dataSources`, `variables`, `outputs`, `childModules`, and `files`, scoped to the module's own source directory. Module resources come back as `terraform.resource` records with the per-module value flow already in place, so they resolve in the module's own context (variable defaults overridden by the call's inputs):

```
terraform.moduleCalls.where(key == "logs")[0]
  .resources.all(values["server_side_encryption_configuration"] != null)
```

Works for local modules straight from source HCL, and for registry/git modules once `terraform init` has fetched them.

I kept the `source` field as the address string (back-compat). The bigger idea you describe — the module source as a first-class *asset* you can navigate via `source.asset` for its own metadata — is a discovery-layer change (registering the module as a child asset) and a clean next step on top of this; happy to take it on separately.

providers/terraform/resources/terraform.lr
tas50: Done — `inputs` is now typed. It returns `[]terraform.module.input` instead of a dict, where each input carries:
- `name` — the variable it sets
- `value` — the value passed, **resolved in the calling context** (var/local interpolations evaluated)
- `variable` — a typed link to the module's `variable` declaration it binds to

```
terraform.moduleCalls.where(key == "logs")[0].inputs
# name="bucket_name" value="logs-prod" variable -> the module's string var "bucket_name"

terraform.moduleCalls.all(inputs.all(variable != null))   # every input maps to a declared variable
```

So you can now audit module wiring directly — e.g. assert a module is called with a compliant value, or that no input sets an undeclared variable.

PR #8203, opened 2026-06-07
🐛 llx: guard bytes2time against short buffers

tas50:
## What

`bytes2time` sliced `b[0:8]` and `b[8:]` without checking the buffer length. A `Primitive` of type `time` whose `Value` is non-empty but shorter than the 12-byte encoding (8 bytes of seconds + 4 bytes of nanoseconds) therefore triggered a slice-out-of-range panic during conversion.

`ptime2raw` already special-cases an empty `Value` (returning the epoch) but does not cover a short-but-non-empty one, so the panic was reachable for any partially-formed time value.

## Change

- `bytes2time` now falls back to the zero time on a buffer shorter than 12 bytes instead of slicing out of range, matching `ptime2raw`'s handling of an empty value. The full-length path also reads `b[8:12]` explicitly to make the 12-byte contract obvious.
- Adds `byte_conversions_test.go` covering a `TimePrimitive` round-trip and every short length (0–11 bytes).

## Test

```
go test ./llx/ -run TestBytes2time -v
```

Both subtests pass.

## Note

`bytes2int` in the same file has the same shape (it panics rather than returning an error on a malformed varint). I left it out to keep this change focused; happy to follow up on it separately.

### DIFF
--- a/llx/byte_conversions.go
+++ b/llx/byte_conversions.go
@@ -48,7 +48,13 @@ func bytes2float(b []byte) float64 {
 }
 
 func bytes2time(b []byte) time.Time {
+	// A valid encoding is 12 bytes: 8 for seconds, 4 for nanoseconds.
+	// Fall back to the zero time on a short buffer instead of slicing
+	// out of range, matching ptime2raw's handling of an empty value.
+	if len(b) < 12 {
+		return time.Unix(0, 0)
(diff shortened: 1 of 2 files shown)


arlimus (approved):
Sweet fix, kudos!

PR #8206, opened 2026-06-07
🧹 providers: keep unpacked archive entries within the temp dir

tas50:
## What

`InstallIO` unpacks a provider's `tar.xz` into a temp directory:

```go
dst := filepath.Join(tmpdir, header.Name)
writer, err := os.Create(dst)
```

There was no check that an entry stays inside `tmpdir`. Because `filepath.Join` cleans the path, an archive whose entry name contains `..` segments resolves to a location outside the unpack directory. The recorded name is also reused later in the rename loop, where it is joined onto `conf.Dst`.

## Change

- Before creating the file or recording its name, verify the resolved path stays within `tmpdir` (via `filepath.Rel`); return a clear error otherwise.
- Recording the name into `files` now happens after that check, so the subsequent rename loop only ever sees in-bounds names too.
- `walkTarXz` already skips non-regular entries, so symlink entries aren't a concern here.

Uses the same `filepath.Rel` confinement idiom as the inventory key-path change (#8205).

## Test

Adds `TestInstallIO_RejectsPathTraversal`, which feeds an in-memory `.tar.xz` with a `../escapee.txt` entry and asserts the install errors and nothing is written outside the unpack directory.

```
go test ./providers/ -run TestInstallIO_RejectsPathTraversal -v
```

Passes.

### DIFF
--- a/providers/install_test.go
+++ b/providers/install_test.go
@@ -1,0 +1,52 @@
+// Copyright Mondoo, Inc. 2024, 2026
+// SPDX-License-Identifier: BUSL-1.1
+
+package providers
+
+import (
+	"archive/tar"
+	"bytes"
+	"io"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/ulikunitz/xz"
+)
+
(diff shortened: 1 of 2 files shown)


syrull:
This is legit but since we aren't doing that with the providers I see it as unnecessary complexity, I won't merge this.

PR #8210, opened 2026-06-07
✨ Expose more Azure security state: ADE, App Service Easy Auth, NSG flow logs

tas50:
## What

Improves existing **azure** provider resources to better surface security posture, targeting gaps against the CIS Microsoft Azure Foundations Benchmark. No new services — these fill in fields/accessors that audits need on resources we already model.

### Compute — observed disk encryption *state* (CIS §7)
The provider already exposes encryption *configuration* (`encryptionAtHost`, disk `encryptionType`, `diskEncryptionSetId`). These add the missing *state* signals for guest-based Azure Disk Encryption (ADE):

- `azure.subscription.computeService.vm.adeInstalled()` — whether the Azure Disk Encryption extension (BitLocker / dm-crypt) is installed, mirroring the existing `mdeInstalled` / `amaInstalled` extension-detection family.
- `azure.subscription.computeService.disk.encryptionSettingsEnabled` — whether guest-based ADE encryption settings are configured on the disk, independent of `encryptionType` (server-side SSE with a platform or customer-managed key).

### App Service — Easy Auth (CIS §9)
`appsiteauthsettings` previously only exposed a generic `properties` dict. Added typed fields so auth posture is directly assertable:

- `enabled` — whether App Service Authentication / Authorization is on.
- `unauthenticatedClientAction` — what happens to unauthenticated requests (AllowAnonymous / RedirectToLoginPage / Show401 / Show403).

Wired into both the site and deployment-slot auth-settings paths.

### Networking — NSG-to-flow-log link (CIS §6.5)
Network Watcher and flow logs are already modeled, but checking a given NSG's flow-log status meant traversing every watcher. Added:

- `azure.subscription.networkService.securityGroup.flowLog()` — resolves the Network Watcher flow log targeting this NSG (enabled + retention), or null when none. The flow-log mapping is refactored into a shared `flowLogToMql` helper reused by `watcher.flowLogs()`.

To avoid an N+1 (re-enumerating every watcher per NSG), the watcher→flow-log index is built once per subscription and cached on the `networkService` singleton (guarded by `sync.Once`); each NSG resolves that singleton and does an in-memory lookup by target resource id.

## Notes
- New `.lr.versions` entries stamp to `13.16.2` (provider is at `13.16.1`).
- `azure.permissions.json` is unchanged — no new API surface (Watchers/FlowLogs list calls were already used by `watchers()` / `watcher.flowLogs()`).

## Verification
Built, installed, and run live against the **CIS Development** subscription:
- `compute.vms { adeInstalled mdeInstalled }` → `adeInstalled` resolves; `mdeInstalled: true` cross-checks extension detection.
- `compute.disks.first { encryptionType encryptionSettingsEnabled }` → resolves (null for a CMK/SSE disk — correct, no guest ADE).
- `network.securityGroups { flowLog { ... } }` → null on every NSG, confirmed correct (the subscription has zero flow logs configured; positive path reuses the proven `flowLogToMql` helper). Re-verified after the caching change.
- `web.apps { authenticationSettings { enabled unauthenticatedClientAction } }` → `enabled: true` on the test web app.

### DIFF
--- a/providers/azure/resources/azure.lr
+++ b/providers/azure/resources/azure.lr
@@ -277,6 +277,14 @@ azure.subscription.computeService.vm @defaults("name location properties.hardwar
   omsInstalled() bool
   // Whether the Dependency Agent extension (used by VM Insights for service-map data) is installed on the VM
   dependencyAgentInstalled() bool
+  // Whether the Azure Disk Encryption (ADE) extension is installed on the VM
+  //
+  // ADE encrypts the OS and data disks from inside the guest using BitLocker
+  // (Windows) or dm-crypt (Linux). A true result indicates the encryption
+  // extension is present; it is distinct from `encryptionAtHost` (platform-side
+  // encryption of the VM host) and from server-side disk encryption set on the
+  // managed disk itself.
+  adeInstalled() bool
   // VM compute disk
   osDisk() azure.subscription.computeService.disk
   // VM compute data disk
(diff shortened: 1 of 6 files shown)


tas50:
Thanks for the review — addressed in 83fb57f3f:

- **N+1 on `flowLog()` (network.go):** fixed. The watcher→flow-log index is now built **once per subscription** and cached on the `networkService` singleton via a new Internal struct guarded by `sync.Once` (Pattern C). Each NSG resolves that singleton by its stable `__id` and does an in-memory lookup by target resource id, so `securityGroups { flowLog }` no longer re-enumerates watchers per NSG. Re-verified live against the CIS Development subscription.

- **nil `flowLog.Properties` (flowLogToMql):** intentional and consistent with `diskToMql` — the property-derived fields are simply omitted (returning zero values) when `Properties` is nil. In practice the Network Watcher API always returns `Properties` for a real flow log; the guard is purely defensive against a nil page entry.

PR #8228, opened 2026-06-07
✨ Add Power BI resources to the ms365 provider

tas50:
## What

Adds a `microsoft.powerbi` namespace to the ms365 provider, covering tenant-wide Power BI (Microsoft Fabric) governance posture.

| Resource | Exposes |
|---|---|
| `microsoft.powerbi.tenantSettings` | Tenant feature switches (Publish to web, export data, service-principal access, guest sharing, …) |
| `microsoft.powerbi.workspaces` | Workspaces (groups), their `users`/access grants, and a typed `capacity` reference |
| `microsoft.powerbi.capacities` | Premium/Fabric capacities — SKU, state, region, admins |
| `microsoft.powerbi.publishedToWeb` | Reports/dashboards published to the public web |
| `microsoft.powerbi.sharedToWholeOrganization` | Artifacts shared with the entire organization |

## How

Power BI is not a Microsoft Graph service, so collection follows the same service-token + PowerShell pattern as the existing Teams/Exchange resources:

- Acquire a Power BI–scoped token (`https://analysis.windows.net/powerbi/api/.default`) from the existing tenant credential — no new auth plumbing.
- `Connect-PowerBIServiceAccount -Token <jwt>`, then read the admin endpoints and emit one JSON document.
- **Tenant settings** come from the **Fabric** admin API (`api.fabric.microsoft.com/v1/admin/tenantsettings`) — they are not a Power BI REST endpoint (that path 404s). The remaining sections use the Power BI admin REST API (`admin/groups?$expand=users`, `admin/capacities`, `admin/widelySharedArtifacts/*`).
- Each section is collected independently (per-call try/catch returning `{data, error}`). The Fabric tenant-settings call and the Power BI `admin/*` calls have separate authorization requirements, so a permission gap in one section surfaces as that section's error instead of blanking all five.
- `workspace.capacity` is a typed reference resolved against the capacities list (raw `capacityId` is not exposed as a duplicate scalar).

## Testing status

Verified live against the `lunalectricsandbox` tenant. The full pipeline is confirmed end-to-end: cert auth, Power BI–scoped token acquisition, `MicrosoftPowerBIMgmt` install, `Connect-PowerBIServiceAccount -Token`, the REST/Fabric calls, JSON parsing, and per-section error propagation all work. `go vet`, `go test ./resources/`, and `gofmt` are clean.

**Remaining testing — needs tenant configuration, not code changes:**

- The scan service principal returns `401 Unauthorized` (Power BI admin REST) / `InsufficientScopes` (Fabric) because the tenant does not yet have **"service principals can use read-only admin APIs"** enabled, nor the SP added to the allowed security group. This is a Power BI/Fabric admin-portal change.
- Once the SP is enabled, a full data run should re-verify:
  - JSON field casing for the broadly-shared artifact sections (`ArtifactAccessEntities`, `sharer` shape).
  - The workspace `users` payload shape (`groupUserAccessRight`, `principalType`, `graphId`).
  - That `workspace.capacity` resolves against `capacities` for workspaces on dedicated capacity.
  - Single-element-array handling (PowerShell `ConvertTo-Json` collapses single-element arrays; the Go side already accepts both forms).

## Notes

- New `.lr.versions` entries are at `13.5.1`.
- `microsoft.powerbi` reuses the existing ms365 credential; the only operational prerequisite is enabling read-only admin API access for the service principal in Power BI/Fabric.


tas50:
Addressed in 77e65513b:

- 🔴 **Token injection** — single quotes in the token are now doubled (`strings.ReplaceAll(token, "'", "''")`) before interpolation into the single-quoted PowerShell string, so a token value cannot break out of the quoting.
- 🟡 **`$top=5000` truncation** — the admin groups endpoint is now paged with `$skip` (5000 per page) until a short page is returned; no silent truncation.
- 🟡 **`Install-Module` on every run** — guarded with `Get-Module -ListAvailable`, so it only installs when the module is missing.

Re-ran live against the sandbox tenant to confirm the revised PowerShell parses and executes (still gated at the same `401`/`InsufficientScopes` boundary until the SP is enabled for read-only admin APIs).

tas50:
@syrull Take a look and let me know if these meet your needs

PR #8249, opened 2026-06-08
:star: Add tls.certificateMatchesDomain field

tas50:
## What

Adds a `certificateMatchesDomain` boolean to the `tls` resource: whether the served leaf certificate covers the hostname the connection targeted.

## Why

Auditing "is the served cert valid for this host?" currently means hand-rolling RFC 6125 matching in MQL — CN match, exact SAN match, and wildcard SAN match all spelled out by hand:

```
checkA1 = tls.certificates.first.subject.commonName == asset.fqdn
if(tls.certificates.first.subject.commonName.contains(/^\*/)) {
  checkA1 == asset.fqdn.contains(tls.certificates.first.subject.commonName.split("*.")[1])
}
checkA2 = tls.certificates.first.sanExtension.dnsNames.contains(asset.fqdn)
checkA3 = tls.certificates.first.sanExtension.dnsNames.where(_ == /\*/).where(_.split(".")[-2] + "." + _.split(".")[-1]).any(asset.name.contains(_.split("*.")[1]))
checkA1 || checkA2 || checkA3
```

This is verbose, untested, and the wildcard arm is subtly wrong (non-boolean `where` body, matches `asset.name` instead of `asset.fqdn`). The `tls` resource already knows the hostname it connected to (`domainName`) and already runs full chain verification — the matching logic belongs in the provider where it can be tested, not re-derived per query.

The whole block collapses to:

```
tls.certificateMatchesDomain
```

## How

- New computed field `tls.certificateMatchesDomain() bool`.
- Backed by `x509.Certificate.VerifyHostname` against the resource's own `domainName` — SAN DNS names + RFC 6125 wildcards. **No CN fallback**, consistent with the existing chain `Verify` and with modern client/browser behavior (Go dropped CN matching in 1.15). CN-only certs report `false`, which is the correct audit outcome.
- Returns **null** when there's no domain to match (e.g. connecting directly to an IP), so an audit can distinguish "no hostname" from "hostname mismatch".
- Matching extracted into a `certMatchesDomain` helper and covered by a network-free table test (`tls_match_test.go`): exact/mismatch, wildcard covers-subdomain / not-apex / not-nested, case-insensitivity, CN-not-consulted, IP SAN match/mismatch.

## Testing

- `go test ./providers/network/resources/ -run TestCertMatchesDomain` — all 10 cases pass.
- Live: `tls("mondoo.com").certificateMatchesDomain` → `true`.

### DIFF
--- a/providers/network/resources/network.lr
+++ b/providers/network/resources/network.lr
@@ -181,6 +181,16 @@ tls @defaults("socket domainName") {
   certificates(socket, domainName) []certificate
   // Certificates provided without server name indication (SNI)
   nonSniCertificates(socket, domainName) []certificate
+  // Whether the served leaf certificate covers the connection hostname
+  //
+  // Matches the connection's domain name against the leaf certificate's
+  // Subject Alternative Name DNS entries using RFC 6125 wildcard rules,
+  // so `*.example.com` covers `api.example.com`. Unlike chain
+  // verification, this isolates hostname coverage — it is true when the
+  // certificate vouches for the host you connected to, independent of
+  // chain trust or expiry. Null when the connection has no domain name
+  // to match, such as when connecting directly to an IP address.
+  certificateMatchesDomain() bool
 }
 
 // X.509 certificate bundle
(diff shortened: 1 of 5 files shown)


arlimus (approved):
so much simpler!

PR #8255, opened 2026-06-08
✨ Add inetd config resource to simplify super-server service audits

### LINKED ISSUE #8250: ✨ Add inetd config resource to simplify super-server service audits
## Summary

Add a native `inetd` config resource to the **os** provider so that auditing which services the inetd super-server exposes — starting with the "ftp must not be enabled" check — becomes a one-liner instead of an `exists`-guarded `content.lines` regex.

## Motivation

The check **"Ensure ftp is not enabled via inetd"** currently has to be written as:

```mql
file("/etc/inetd.conf").exists == false || file("/etc/inetd.conf").content.lines.none(_ == /^ftp/)
```

Problems with the raw-file approach:

- The author has to special-case file absence (`exists == false ||`) just to express "no ftp service".
- `^ftp` regex-on-lines is brittle: it ignores leading whitespace, doesn't account for commented-out (disabled) entries vs. active ones, and would false-match a hypothetical `ftps`/`ftp-data` line depending on the anchor.
- It only inspects `/etc/inetd.conf` and misses `/etc/inetd.d/*` drop-in entries used on some systems.
- No reuse: every future inetd audit (telnet, shell/login/exec r-services, tftp, services running as root, etc.) re-implements file handling and line parsing.

This is the same shape of problem `sshd.config` solves for OpenSSH and the proposed `snmpd.config` (#8248) solves for Net-SNMP. inetd should get the same treatment, with the bonus that inetd lines are **structured records** (service / socket-type / protocol / wait / user / server), so a typed sub-resource buys real query power.

## Proposed resource

`inetd.conf` lines have the form:

```
service-name  socket-type  protocol  wait|nowait[.max]  user[.group]  server-program  server-arguments
```

so each active line is a record worth modeling as a sub-resource (this clears the sub-resource bar — multiple meaningful fields, several of which warrant their own audits, e.g. "which inetd services run as root"). Disabled entries are commented out with `#` and are excluded from `entries`.

```lr
// inetd super-server namespace
//
// Use inetd.config to examine which network services the inetd super-server exposes
inetd {}

// inetd super-server configuration
//
// Examine the services the inetd (or compatible) super-server is configured
// to launch on demand. Each active entry pairs a service with its socket
// type, protocol, wait mode, run-as user, and server program. The
// configuration is assembled from /etc/inetd.conf together with any drop-in
// files under inetd.d.
inetd.config {
  init(path? string)
  // Primary inetd configuration file
  file() file
  // All files making up the inetd configuration (main + drop-ins)
  files(file) []file
  // Merged contents of all configuration files
  content(files) string
  // Active service entries
  entries(files) []inetd.config.entry
  // Names of the active services
  serviceNames(entries) []string
}

// inetd service entry
//
// Examine a single active inetd service line: the service it exposes and how
// inetd launches it. The name field selects the service as it appears in the
// first column of inetd.conf — for example inetd.config.entries.where(name == "ftp").
private inetd.config.entry @defaults("name protocol server") @context("file.context") {
  // Service name or port (first column)
  name string
  // Socket type, such as stream, dgram, or raw
  socketType string
  // Transport protocol, such as tcp, udp, tcp6, or udp6
  protocol string
  // Wait mode (wait or nowait, with optional max-connections suffix)
  wait string
  // User the server runs as (with optional group)
  user string
  // Server program path, or "internal" for built-in services
  server string
  // Arguments passed to the server program
  arguments string
}
```

With this, the check collapses to:

```mql
inetd.config.serviceNames.none("ftp")
```

The "file doesn't exist" case is handled for free: a missing `/etc/inetd.conf` yields an empty `entries`, so `serviceNames.none("ftp")` is trivially true — no `exists ==` guard needed. Richer audits also become natural, e.g. `inetd.config.entries.where(user == "root")` or `inetd.config.entries.none(name == /^(shell|login|exec|telnet|tftp)$/)`.

## Implementation plan

1. **Schema** — add the `inetd` namespace + `inetd.config` + `inetd.config.entry` to `providers/os/resources/os.lr` with the doc-comments above. Add `.lr.versions` entries at the next os-provider patch version (current `Version` is `13.23.0` in `providers/os/config/config.go`, so `13.23.1`).

2. **Codegen** — `make providers/mqlr` then `./mqlr generate providers/os/resources/os.lr --dist providers/os/resources`.

3. **Parser** — new `providers/os/resources/inetd/parser.go`:
   - Split each line on whitespace into the 6 fixed columns + trailing arguments; skip blank lines and `#` comments (disabled entries).
   - Tolerate the `nowait.max` and `user.group` suffix forms (store the full token; consumers can split if needed).
   - Resolve `/etc/inetd.d/*` drop-in files into the ordered file set.

4. **Resource (`providers/os/resources/inetd.go`)** — follow the `sshd.go` / proposed `snmpd.go` shape:
   - `initInetdConfig` converting an optional `path` arg into the internal `file` resource (default `/etc/inetd.conf`); return a **bare resource when the file is absent** rather than erroring (this is what subsumes the `exists == false` clause).
   - `file()`, `files()`, `content()` with mutex-guarded parse + caching.
   - `entries()` building `inetd.config.entry` sub-resources (cache key = file-qualified service+protocol), `serviceNames()` projecting their names.

5. **Build & verify** — `make providers/build/os && make providers/install/os`, then test against the live Debian box (`mql shell ssh root@172.16.1.50 --identity-file ~/.ssh/id_ed25519`) with:
   - `inetd.config.file.path`
   - `inetd.config.entries { name protocol user server }`
   - `inetd.config.serviceNames`
   - the absent-file case (Debian likely has no `/etc/inetd.conf`) to confirm the bare-resource path returns an empty `entries`.

## Open questions

- **xinetd.** Modern systems use `xinetd` (per-service files under `/etc/xinetd.d/` with a block syntax) rather than the legacy single-file `inetd`. Should this resource cover only classic inetd (this issue) and xinetd land as a separate `xinetd.config` resource, or should `inetd.config` abstract over both? Leaning toward **separate resources** since the on-disk formats and per-entry fields differ substantially.
- Whether to split `wait` into `wait bool` + `maxConnections int`, and `user` into `user` + `group`, rather than storing the raw `nowait.max` / `user.group` tokens. Leaning toward raw tokens initially to match the file faithfully, with splitting deferred until an audit needs it.

tas50:
Resolves #8250.

## Summary

Adds a native `inetd.config` resource to the **os** provider so auditing which services the inetd super-server exposes becomes a one-liner instead of an `exists`-guarded `content.lines` regex.

The check **"Ensure ftp is not enabled via inetd"** collapses from:

```mql
file("/etc/inetd.conf").exists == false || file("/etc/inetd.conf").content.lines.none(_ == /^ftp/)
```

to:

```mql
inetd.config.serviceNames.none("ftp")
```

The "file doesn't exist" case is handled for free — a missing `/etc/inetd.conf` yields an empty `entries`, so `serviceNames.none("ftp")` holds trivially with no `exists ==` guard.

## What's added

- **`inetd`** namespace + **`inetd.config`** resource (`file`, `files`, `content`, `entries`, `serviceNames`).
- **`inetd.config.entry`** typed sub-resource per active service line — `name`, `socketType`, `protocol`, `wait`, `user`, `server`, `arguments` — with `file.context` for line provenance. This clears the sub-resource bar (multiple meaningful fields, several warranting their own audits, e.g. "which inetd services run as root").
- Parser package `providers/os/resources/inetd/` with unit tests: splits the fixed columns + trailing arguments, skips blank lines and `#`-commented (disabled) entries, and preserves `nowait.max` / `user.group` / `rpc/udp` tokens verbatim.
- Configuration is assembled from `/etc/inetd.conf` plus any drop-in files under `/etc/inetd.d`.

Richer audits become natural, e.g. `inetd.config.entries.where(user == "root")` or `inetd.config.entries.none(name == /^(shell|login|exec|telnet|tftp)$/)`.

## Verification

Tested against a live Debian 13 host:

- **Absent file** (host's real state): `entries`/`serviceNames` → `[]`, and `inetd.config.serviceNames.none("ftp")` → `true` with no `exists` guard.
- **Synthetic config + drop-in**: all active entries parsed with correct columns; commented `#shell` excluded; `internal` servers have empty `arguments`; multi-arg `tftp` joined; `/etc/inetd.d/echo` drop-in merged into both `entries` and `files`.
- `entries.where(user == "root")`, `files { path }`, and `entries[..].context.range` all resolve correctly.
- Parser unit tests pass; `go build ./...` and the os resources suite are green; gofmt clean; codegen idempotent.

New `.lr.versions` entries are at `13.24.1` (next os-provider patch).

## Deferred (per the issue's open questions)

- **xinetd** left to a separate future `xinetd.config` resource — its on-disk format and per-entry fields differ substantially.
- `wait` / `user` kept as raw tokens (`nowait.400`, `nobody.nogroup`) to match the file faithfully; splitting deferred until an audit needs it.


tas50:
Good catch — fixed in 5dc1eb9db. The entry `__id` now folds in the line number (`<path>/<line>/<name>/<protocol>`), so a file repeating the same service+protocol on multiple lines yields distinct entries instead of shadowing. Verified on a live Debian host: two `ftp stream tcp nowait` lines differing only in user/server now both surface (`entries.length` == 2).

PR #8262, opened 2026-06-08
✨ digitalocean: typed firewall ingress/egress rules

tas50:
## Summary

Converts DigitalOcean cloud-firewall rules from untyped `[]dict` into typed sub-resources, so security audits no longer need string-keyed dict access (`_["protocol"]`) or open-coded public-CIDR checks.

The motivating policy — `no-unrestricted-db-ports` — goes from **12 lines to 1**:

**Before**
```
digitalocean.firewalls.all(inboundRules.none(_["protocol"] == "tcp" && _["ports"] == "3306" && _["sourceAddresses"].contains("0.0.0.0/0")))
digitalocean.firewalls.all(inboundRules.none(_["protocol"] == "tcp" && _["ports"] == "3306" && _["sourceAddresses"].contains("::/0")))
... (10 more lines, one per port × per CIDR)
```

**After**
```
digitalocean.firewalls.all(ingressRules.where(openToInternet).none(ports in ["3306","5432","27017","6379","9092","9200"]))
```

## What changed

- **New sub-resources** `digitalocean.firewall.ingressRule` / `digitalocean.firewall.egressRule`, each exposing:
  - `protocol`, `ports`
  - `openToInternet` — true when sources/destinations include `0.0.0.0/0` or `::/0` (exact match, preserving the original policy's semantics)
  - `sourceAddresses`/`destinationAddresses`, `sourceTags`/`destinationTags`
  - typed refs `sourceDroplets()`/`sourceLoadBalancers()`/`sourceKubernetesClusters()` and the destination mirror
- **New accessors** `ingressRules()` / `egressRules()` on `digitalocean.firewall`.
- The existing `inboundRules`/`outboundRules` dict fields are **kept but deprecated** (`@maturity("deprecated")`) — no breakage; existing content keeps working.

## Notes

- Typed rules are built from the **persisted dict fields**, so they survive recording/replay and cross-query caching.
- **No new API permissions** — load balancer and Kubernetes refs reuse the existing list calls through shared by-id indexes (mirroring the existing `dropletByIDs` helper).
- New `.lr.versions` entries at `13.5.2`; no `config.go` version bump.

## Testing

- `go build ./resources/...` clean, `go test ./resources/...` passes, gofmt clean, two-pass codegen for the Internal structs.
- ⚠️ Not yet run against a live DigitalOcean account (no token on the dev box). Needs an interactive `mql shell digitalocean` pass over a firewall with mixed rule sources before merge.


tas50:
Addressed review feedback in fd8b313c5:

- **`kubernetesClustersByIDs` consistency** — extracted `ensureK8sClusterIndex()` and now iterate the shared index map directly, matching the `dropletByIDs` / `loadBalancerByUIDs` pattern.
- **Index-based rule `__id`** — left as-is intentionally. DigitalOcean firewall rules have no API-provided stable ID, so the `firewallId/inbound/<i>` key is the pragmatic choice; a content-derived hash would be overkill (as the review noted).

PR #8270, opened 2026-06-09
✨ os: parse Dockerfile RUN/CMD/ENTRYPOINT into typed commands

tas50:
## What

Adds a parsed command view to the `docker.file.run` resource so Dockerfile policies can audit the program invoked and its options instead of substring-matching the raw `script`.

New field on `docker.file.run`:
- `commands []docker.file.run.command`

New private sub-resource `docker.file.run.command`:

| field | meaning |
|---|---|
| `binary` | executable invoked, e.g. `apt-get`, `curl`, `npm` |
| `subcommand` | first non-flag argument (e.g. `install`), resolved independently of flag position |
| `flags` | dash-prefixed options, e.g. `["--no-install-recommends", "-y"]` |
| `args` | all arguments after the executable, in order |

The script is split on the shell operators `&&`, `||`, `;`, `|`, `&` and newlines, and each command is tokenized into argv with leading `VAR=value` assignments removed. Quotes, backslash escapes, and line continuations are honored. Exec-form instructions yield a single command from the argv. Covers RUN, CMD, and ENTRYPOINT. Parsing is best-effort and does not descend into subshells or command substitutions.

## Why

`mondoo-dockerfile-security` and `mondoo-dockerfile-best-practices` check RUN instructions with raw `script.contains("...")` matching — ~31 checks across the two files. That is fragile:

- `script.contains("-y")` matches `-y` anywhere — a path, a filename, `--yes`, a comment.
- `script.contains("apt install")` matches inside `echo`/comments and substring-matches unrelated text.
- Flag-order variations slip through (`apt-get -y install` vs `apt-get install -y`).
- `&&`-chained commands are conflated, so the cleanup in one command can satisfy a check about another.

## Result

```mql
# before (fragile)
run.where(script.contains("apt-get install")).all(script.contains("--no-install-recommends"))

# after (precise)
run.all(commands.where(binary == "apt-get" && subcommand == "install")
                .all(flags.contains("--no-install-recommends")))
```

## Tests

- **Parser unit tests** (`docker_file_shell_test.go`): table-driven coverage of chaining, pipes, `||`/`;`, env-assignment stripping, quoted operators/whitespace, line continuations, escapes, concatenated quoted/bare segments, and empty input; plus `stripEnvAssignments` edge cases.
- **Resource-level tests** (`docker_file_test.go` → `TestParseDockerfile_RunCommands`): verifies `commands` on shell-form RUN (`&&` chain → two commands), exec-form RUN, CMD, and ENTRYPOINT, with binary/subcommand/flags/args assertions.
- **End-to-end**: built+installed the provider and ran the rewritten policy checks against live Dockerfiles — passing checks return `true`, violating Dockerfiles return `false`.

## Notes

No new dependencies (the shell tokenizer is a self-contained helper). No new API calls or permissions; `permissions.json` unaffected. `.lr.versions` entries added at `13.25.1`.

### DIFF
--- a/providers/os/resources/docker_file.go
+++ b/providers/os/resources/docker_file.go
@@ -274,6 +274,10 @@ func (p *mqlDockerFile) stage2resource(stage instructions.Stage, isFinal bool) (
 				return nil, err
 			}
 			mountsSecret, mountsSsh := mountTypeFlags(mounts)
+			commands, err := p.runCommandResources(p.locationID(v.Location()), v.CmdLine, !v.PrependShell)
+			if err != nil {
+				return nil, err
+			}
 			runResource, err := CreateResource(p.MqlRuntime, ResourceDockerFileRun, map[string]*llx.RawData{
 				"__id":         llx.StringData(p.locationID(v.Location())),
 				"script":       llx.StringData(script),
@@ -284,6 +288,7 @@ func (p *mqlDockerFile) stage2resource(stage instructions.Stage, isFinal bool) (
 				"isExecForm":   llx.BoolData(!v.PrependShell),
 				"mountsSecret": llx.BoolData(mountsSecret),
 				"mountsSsh":    llx.BoolData(mountsSsh),
+				"commands":     llx.ArrayData(commands, types.Resource(ResourceDockerFileRunCommand)),
 			})
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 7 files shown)


arlimus (approved):
That makes these checks a whole lot easier!

PR #8271, opened 2026-06-09
🐳 os: add docker.file.stage.oci for quick OpenContainer label access

tas50:
## What

Adds a `docker.file.oci` sub-resource, exposed as `oci` on each `docker.file.stage`, that surfaces the standard `org.opencontainers.image.*` annotations from a stage's `LABEL` instructions as named, tab-completable fields.

Instead of indexing the raw labels map:

```
docker.file.stages.last.labels["org.opencontainers.image.source"]
```

you can now write:

```
docker.file.stages.last.oci.source
```

## Fields

Named: `created`, `authors`, `url`, `documentation`, `source`, `version`, `revision`, `vendor`, `licenses`, `refName`, `title`, `description`, `baseName`, `baseDigest`.

`all` — a `map[string]string` of every `org.opencontainers.*` label on the stage, including vendor-specific annotations with no named field.

## Quote handling

The `oci.*` fields (and `all`) strip a single matched surrounding quote pair, so a quoted `LABEL ...title="My App"` reads back as `My App`. The verbatim value (quotes included) remains available in the stage `labels` map — `oci` is the cleaned, convenience view.

Values come straight from the already-parsed `labels`, so there are no new API calls or parsing passes.

## Notes

- New `.lr.versions` entries are all at `13.25.1` (next patch after the os provider's current `13.25.0`).
- Local Dockerfile parsing only — no new IAM permissions.

## Testing

- `TestParseDockerfile_OCILabels` — multi-value fixture (unquoted, double-quoted, single-quoted); asserts named fields, `all` filtering (excludes a non-OCI `com.example.team` label), undeclared-annotation emptiness, and that the raw `labels` map retains the quotes `oci.*` strips.
- `TestTrimMatchingQuotes` — table test for the quote helper (empty, single char, matched/unmatched/mismatched pairs, nested quotes, realistic URL).
- Verified end-to-end against a real multi-stage Dockerfile with the installed provider; per-stage labels resolve independently and quotes are stripped.


arlimus (approved):
That's very cool, native suport for these labels!

PR #8273, opened 2026-06-09
🧹 vault: registry-based backends, impls out of provider SDK

chris-rock:
## What

Decouple the vault backend implementations from the provider SDK using a **registry**, and relocate the implementations out of `providers-sdk` into a top-level `vault/` tree. The provider SDK now holds only the vault **interface** + the registry. Everything stays in the **core module** — no new `go.mod`, no `replace`.

## Why

The core module force-linked every vault backend through a single edge: `vault/config/vaultconfigstore.go`'s `New()` switch statically imported all implementations, and `inventory.GetVault()` called it. So **any** consumer of `inventory` (e.g. cnspec) transitively compiled the AWS/GCP/HashiCorp/keyring SDKs whether or not it used vaults.

## How

- **Registry in the SDK** (`providers-sdk/v1/vault/registry.go`): implementations register a `Builder` for their `VaultType` via `init()`; `vault.New()` resolves them. No heavy deps in the SDK.
- **Removed `config.New`**: it was a thin wrapper that only guaranteed the in-memory backend, so a caller using it for a cloud backend would compile fine but fail at runtime. `inventory.GetVault` now calls `vault.New` directly and blank-imports the in-memory vault so memory is always available.
- **Implementations relocated to `vault/<impl>`** (`awssecretsmanager`, `awsparameterstore`, `gcpsecretmanager`, `gcpberglas`, `hashivault`, `keyring`), each self-registering. They remain ordinary packages of the core module.
- **`vault/register`** umbrella links all backends with one blank import; **`apps/mql`** imports it so the CLI keeps full vault support. The in-memory backend stays in the SDK and is always available.

## Effect

- **Library consumers that don't need cloud vaults stop compiling them** — backends are now opt-in via `vault/register` (or an individual impl package), instead of being pulled in through `inventory`.
- The provider SDK's `vault` package is interface + registry only.
- No `replace`, no separate module, no circular dependency.

The heavy cloud-SDK deps **remain in the core `go.mod`** because `apps/mql` links the backends it ships. Trimming them from the core `go.mod` itself would require moving `apps/mql` into its own module — left as a separate future step.

## Testing

- `go build ./...` and `go test ./...` (vault, SDK vault, inventory) — green; `go list ./...` now naturally covers `./vault/...`.
- End-to-end: `mql vault add-secret` with an encrypted-file keyring config stores a secret, exercising `inventory.GetVault -> vault.New -> registry -> keyring`.


providers-sdk/v1/vault/config/vaultconfigstore.go
```diff
@@ -4,81 +4,12 @@
...
-	"go.mondoo.com/mql/v13/providers-sdk/v1/vault/awsparameterstore"
-	"go.mondoo.com/mql/v13/providers-sdk/v1/vault/awssecretsmanager"
-	"go.mondoo.com/mql/v13/providers-sdk/v1/vault/gcpberglas"
-	"go.mondoo.com/mql/v13/providers-sdk/v1/vault/gcpsecretmanager"
-	"go.mondoo.com/mql/v13/providers-sdk/v1/vault/hashivault"
-	"go.mondoo.com/mql/v13/providers-sdk/v1/vault/inmemory"
-	"go.mondoo.com/mql/v13/providers-sdk/v1/vault/keyring"
 )
 
-func New(vCfg *vault.VaultConfiguration) (vault.Vault, error) {
```
chris-rock, line 23: This is intentionally removed to ensure users get compile errors.

arlimus (approved):
Love the split! this is going to further reduce dependencies/footprint

PR #8283, opened 2026-06-09
✨ os: expose Dockerfile provenance on docker.file (sha256 + source repository)

tas50:
## What

Adds four provenance fields to the `docker.file` resource so a user viewing a scanned Dockerfile in the Mondoo console can tell **where it came from** — and so policies can query it:

| Field | Description |
|---|---|
| `sha256` | Hex-encoded SHA256 of the raw Dockerfile contents — a stable fingerprint independent of path or repository. |
| `repositoryUrl` | HTTPS URL of the source repository. |
| `repositoryOrganization` | Organization / owner of the source repository. |
| `repositoryName` | Name of the source repository. |

## Why

Until now the only "where did this come from" signal on `docker.file` was `file.path` — the **absolute path on whatever machine ran the scan** (`/home/runner/work/.../Dockerfile` in CI, `/Users/.../Dockerfile` locally). That's close to useless in the console.

Meanwhile the dockerfile connection *already* parsed the git `ssh-url` to build the asset platform ID, then discarded the parsed host/org/repo. This PR plumbs that data onto the connection and surfaces it on the resource.

## Behavior

- `sha256` is always available (computed from file content).
- The `repository*` fields are populated when the Dockerfile is discovered through a source-control integration — the `ssh-url` / `http-url` connection options that GitHub and GitLab discovery already attach to IaC assets. They are **null** when the Dockerfile is scanned from a plain path with no repository context.
- `repositoryUrl` prefers an explicit `http-url`; otherwise it reconstructs `https://<host>/<org>/<repo>` from the `ssh-url`.

## Scope note (Tier 1)

This is the self-contained, high-value slice. A meaningful **repo-relative path**, **branch**, and **commit SHA** were deliberately left out: they require the scanning integration to pass the checkout root / ref, which it does not today. Those belong in a follow-up once that context is plumbed through.

## Testing

- `make providers/build/os` + interactive verification:
  - `docker.file { sha256 ... }` against a local Dockerfile → `sha256` matches `shasum -a 256`, repository fields `null`.
- Unit tests added to `docker_file_connection_test.go` covering the `ssh-url` → org/repo/URL parsing and the `http-url` preference. All pass.
- Generated code (`os.lr.go`, `os.lr.versions`) regenerated; new `.lr.versions` entries at `13.26.1`.

### DIFF
--- a/providers/os/connection/docker/docker_file_connection.go
+++ b/providers/os/connection/docker/docker_file_connection.go
@@ -29,8 +29,15 @@ type DockerfileConnection struct {
 	// FileAbsSrc must be the absolute path of the Dockerfile so
 	// that we find the file downstream
 	FileAbsSrc string
-	osFamily   shared.OSFamily
-	closer     func()
+	// RepositoryURL is the HTTPS URL of the source repository the Dockerfile
+	// was discovered in. Empty when the Dockerfile has no repository context.
+	RepositoryURL string
+	// RepositoryOrg and RepositoryName identify the source repository, parsed
+	// from the git ssh-url. Empty when the Dockerfile has no repository context.
+	RepositoryOrg  string
+	RepositoryName string
+	osFamily       shared.OSFamily
+	closer         func()
 }
 
 func NewDockerfileConnection(_ uint32,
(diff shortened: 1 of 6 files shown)


arlimus (changes requested):
Hold for one short sec, I'll push for a new RFC to handle connected assets that we don't fully understad at the time of scanning (basically discovery connectivity)

tas50:
Paging @arlimus. How do we want to handle this type of pattern going forward

tas50:
@arlimus reminder that we still need a long term plan for how to implement something like this

PR #8292, opened 2026-06-10
✨ ms365: typed Secure Score controls and unified audit log accessor

tas50:
## What

Two `mondoo-m365-security` checks reach into raw `dict` fields because the typed accessor doesn't exist yet. This adds the typed fields while keeping the dicts for backward compatibility — purely additive, no breaking type changes.

### `microsoft.security.securityscore` — new `controls`

The Secure Score control breakdown was only available as `controlScores []dict`, so five checks did:

```mql
microsoft.security.latestSecureScores.controlScores.where(controlName == 'AdminMFAV2').all(_['score'] == 10)
```

This adds a typed `controls []microsoft.security.securityscore.controlScore` (`controlCategory`, `controlName`, `description`, `score`) — the `ControlScore` struct already existed in Go and was being `JsonToDict`-ed. Checks become:

```mql
microsoft.security.latestSecureScores.controls.where(controlName == 'AdminMFAV2').all(score == 10)
```

The existing `controlScores` field is **marked `@maturity("deprecated")` and still populated** (rather than retyped) so nothing breaks.

### `ms365.exchangeonline` — new `unifiedAuditLogIngestionEnabled`

One check did dict indexing:

```mql
ms365.exchangeonline.adminAuditLogConfig['unifiedAuditLogIngestionEnabled'] == true
```

This adds a typed `unifiedAuditLogIngestionEnabled` bool:

```mql
ms365.exchangeonline.unifiedAuditLogIngestionEnabled == true
```

The accessor does a **case-insensitive** lookup over the admin-audit-log config — `Get-AdminAuditLogConfig` emits the property as `UnifiedAuditLogIngestionEnabled` (PascalCase), while the old check indexed lowercase, so the typed accessor is robust to the casing either way.

## Scope notes

I deliberately did **not** touch two things the broader analysis surfaced:
- `microsoft.devicemanagement.deviceConfigurations.settings` is a documented union-dict by design (per-platform `@odata.type` keys); dot-access is the intended interface.
- `microsoft.domain.serviceConfigurationRecords` is already typed; only the DMARC checks reach into the core `dns()` resource, which is out of scope here.

## Testing

- `go build ./...`, `go vet ./resources/`, `go test ./resources/...` all pass
- Generated files regenerated via `./lr go` + `./lr versions` + `gen/main.go`

A follow-up `cnspec` PR will migrate the six policy queries to the new fields once this ships.

### DIFF
--- a/providers/ms365/resources/ms365.lr
+++ b/providers/ms365/resources/ms365.lr
@@ -1682,8 +1682,15 @@ private microsoft.security.securityscore @defaults("id azureTenantId") {
   averageComparativeScores []dict
   // Secure Score tenant ID
   azureTenantId string
-  // Secure Score control scores
-  controlScores []dict
+  // Secure Score control scores as raw dictionaries
+  //
+  // Deprecated in favor of `controls`.
+  controlScores @maturity("deprecated") []dict
+  // Per-control score breakdown for the Secure Score
+  //
+  // Each entry exposes the control's `controlName`, `controlCategory`,
+  // `description`, and tenant-achieved `score`.
+  controls []microsoft.security.securityscore.controlScore
   // Secure Score creation time
   createdDateTime time
   // Secure Score current score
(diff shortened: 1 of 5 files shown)


tas50:
Addressed both review comments: dropped the public `id` field in favor of `__id`, and added a loop-index fallback so a nil `controlName` can't collide in the cache.

PR #8302, opened 2026-06-10
✨ gcp: typed serviceAccount/network/enableWebAccess on Vertex AI custom job

tas50:
## What

Adds typed fields to `gcp.project.vertexaiService.customJob`, replacing raw `jobSpec` dict indexing:

```mql
# before
job.jobSpec['serviceAccount']
job.jobSpec['network']
job.jobSpec['enableWebAccessPortals']
# after
job.serviceAccount        # + job.serviceAccountRef for the typed resource
job.network               # + job.networkRef for the typed resource
job.enableWebAccess
```

Because `serviceAccount` and `network` reference modeled resources, I added `serviceAccountRef()` / `networkRef()` typed accessors alongside the raw strings — mirroring the existing `pipelineJob.serviceAccountRef()` and `endpoint.networkRef()` — to satisfy the typed-reference convention up front. `jobSpec` dict retained for backward compatibility.

## Testing

Scalar extraction factored into the pure helper `customJobSpecFields` (nil-safe via protobuf getters) and unit-tested. `go build ./...`, lr regen, and `go test ./resources/...` pass.

### DIFF
--- a/providers/gcp/resources/gcp.lr
+++ b/providers/gcp/resources/gcp.lr
@@ -11317,6 +11317,16 @@ private gcp.project.vertexaiService.customJob @defaults("displayName state") {
   state string
   // Job spec (worker pool specs, scheduling, etc.)
   jobSpec dict
+  // Email of the service account the job runs as (jobSpec.serviceAccount); empty when the default is used
+  serviceAccount string
+  // The job's run-as service account, resolved to the typed resource
+  serviceAccountRef() gcp.project.iamService.serviceAccount
+  // Resource name of the VPC network the job peers with (jobSpec.network); empty when none
+  network string
+  // The job's peering VPC network, resolved to the typed resource
+  networkRef() gcp.project.computeService.network
+  // Whether interactive web-access portals are enabled for the job (jobSpec.enableWebAccessPortals)
+  enableWebAccess bool
   // User-defined key/value labels for organizing the resource
   labels map[string]string
   // Encryption spec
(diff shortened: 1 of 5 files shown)


tas50:
Thanks for the follow-up review. On the two new comments:

**`*Ref` naming** — I checked this against the codebase and the `Ref` suffix is actually the established convention for the Vertex AI resources, not unprecedented. `customJob` is a direct copy of its file-neighbors in `vertexai.go`:

- `endpoint.networkRef()` (vertexai.go:464)
- `pipelineJob.serviceAccountRef()` (659) and `pipelineJob.networkRef()` (683)
- `indexEndpoint.networkRef()` (697)

The Memorystore Redis `AuthorizedNetwork` (deprecated) + `network()` (typed) convention is a different service's choice — both patterns exist in the provider. For a new Vertex AI resource I think local consistency with the other Vertex AI resources should win over matching Memorystore; flipping just `customJob` to `network()`/`serviceAccount()` would make it the odd one out among `endpoint`/`pipelineJob`/`indexEndpoint`. (It would also contradict the first review pass on this PR, which asked to keep the `*Ref` accessors and deprecate the raw strings — done in a885b61c3.)

If the team wants to standardize the whole Vertex AI service on the Memorystore convention, that seems like a separate, deliberate refactor across all of `endpoint`/`pipelineJob`/`indexEndpoint`/`customJob` rather than a piecemeal change on one new field — happy to open that as a follow-up if desired.

**`networkRef()` returning `n, nil` after `StateIsNull`** — this mirrors `endpoint.networkRef()` exactly (it's the same defensive shape). The `n == nil` branch is indeed only reachable today via the empty-URL guard, but keeping it identical to the sibling accessor seems preferable to diverging for a nit. Happy to tighten all of them together if you'd prefer.

PR #8311, opened 2026-06-10
✨ snowflake: typed networkPolicy on account

tas50:
## What

Adds a typed convenience field `networkPolicy string` on the `snowflake.account` resource that surfaces the value of the account's `NETWORK_POLICY` parameter. It returns an empty string when the parameter is unset.

## Why

The `mondoo-snowflake-security` policy currently checks for an account-level network policy with:

```
snowflake.account.parameters.where(key == "NETWORK_POLICY").any(value != "")
```

The new typed field collapses that to a clear, readable expression:

```
snowflake.account.networkPolicy != ""
```

## How

- Factored the `SHOW PARAMETERS` (account scope) fetch into a shared `showAccountParameters()` helper so both `parameters()` and the new resolver read the same backing data.
- Added a small pure helper `findParameterValue(parameters []*sdk.Parameter, key string) string` that does a case-insensitive key lookup and returns `""` when absent.
- Implemented the lazy resolver `networkPolicy() (string, error)` on `snowflake.account` using that helper.
- Declared `networkPolicy() string` in `snowflake.lr` and regenerated the `.lr.go`, `.lr.versions`, and `.resources.json` artifacts.

## Verification

- `SKIP_COMPILE=no make providers/build/snowflake` — compiles cleanly.
- `go vet ./resources/` — clean.
- `go test ./resources/` — passes, including a new `parameters_test.go` covering the pure helper (exact match, case-insensitive match, missing key, empty value, nil entries, nil slice).


tas50:
Rebased onto latest `main` and addressed both review comments:

1. **Double API call (`networkPolicy` + `parameters`)** — the account-level `SHOW PARAMETERS` response is now cached on the resource via a `sync.Once` in `mqlSnowflakeAccountInternal`. Both `parameters()` and `networkPolicy()` route through `showAccountParameters()`, so touching either or both hits the Snowflake API at most once.

2. **String vs. typed `snowflake.networkPolicy` reference** — kept as a string, with a doc comment in `snowflake.lr` explaining why: the field reports whether an account-level policy is configured, but the named policy isn't always enumerable in `networkPolicies` (e.g. when the session role lacks the privilege to list it). A typed reference would resolve to `null` in that case and wrongly read as "no policy set", which would break the downstream presence check.

Verified: `lr` regen, `make providers/build/snowflake`, `go vet`, and `go test ./resources/` all pass.

PR #8315, opened 2026-06-10
✨ cos: emit pkg:cos PURLs for COS packages

czunker:
## Summary

- COS packages previously had no PURL — `/etc/cos-package-info.json` was parsed but `Package.PUrl` was never set, so downstream SBOM tooling synthesized a `pkg:generic/cos/...` fallback and logged warnings about packages with empty PURLs.
- Now `cos_packages.go` emits `pkg:cos/<name>@<ebuild_version>?build=<BUILD_ID>&distro=cos-<VERSION_ID>`. Shape follows [osv-scalibr](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/os/cos/cos.go) and the open [purl-spec PR #270](https://github.com/package-url/purl-spec/pull/270) — the only tools in the ecosystem that handle COS today.
- We deviate from scalibr in one place: the `@version` slot holds the **ebuild version** (the actual upstream package version, useful for CVE matching), not the COS image version. The image `BUILD_ID` is preserved as the `build` qualifier so it isn't lost.
- Older `cos-package-info.json` entries that only populate `version` (no `ebuild_version`) fall back to the COS image version and skip the redundant `build` qualifier so we don't emit `build=X&@X` duplication.

### Example output

For a typical COS 97 node (`BUILD_ID=16919.103.16`):

```
pkg:cos/zlib@1.2.11-r4?build=16919.103.16&distro=cos-97
pkg:cos/baselayout@2.2-r1?build=16919.103.16&distro=cos-97
pkg:cos/ca-certificates@20230311.3.96.1?distro=cos-97   # fallback path
```

### Files touched

- `providers/os/resources/purl/purl_types.go` — new `TypeCos` registered in `KnownTypes`.
- `providers/os/resources/purl/purl.go` — new `QualifierBuild = "build"` constant.
- `providers/os/resources/purl/platform_purl.go` — adopts the constant.
- `providers/os/resources/packages/cos_packages.go` — `CosPkgManager` now carries `*inventory.Platform`; `newCosPurl` emits the new shape.
- `providers/os/resources/packages/packages.go` — threads `asset.Platform` when constructing `CosPkgManager`.
- `providers/os/resources/packages/testdata/cos-package-info.json` — added a `ca-certificates` entry without `ebuild_version` to lock in the fallback path.

## Test plan

- [x] `go test ./providers/os/resources/packages/ -run TestParseCosPackages` — passes (covers PURL shape, build qualifier, ebuild-version fallback, no-platform path).
- [x] `go test ./providers/os/resources/purl/...` — existing purl suite still passes.
- [x] `go vet ./providers/os/resources/packages/... ./providers/os/resources/purl/...` — clean.
- [x] End-to-end smoke against a mock COS root: `mql run filesystem /tmp/cos-mock-root -c 'packages.list { name version purl }'` correctly detects `platform=cos build=16919.103.16` and emits the new PURLs for all four test packages.

### DIFF
--- a/providers/os/resources/packages/cos_packages.go
+++ b/providers/os/resources/packages/cos_packages.go
@@ -76,8 +76,7 @@ func ParseCosPackages(pf *inventory.Platform, input io.Reader) ([]Package, error
 	}
 
 	pkgs := make([]Package, len(cPkgs.InstalledPackages))
-	for i := range cPkgs.InstalledPackages {
-		src := cPkgs.InstalledPackages[i]
+	for i, src := range cPkgs.InstalledPackages {
 		// older /etc/cos-package-info.json entries only populate `version`
 		// with the COS image version; newer ones add `ebuild_version`. Prefer
 		// the ebuild version when present and fall back so we still emit a
--- a/providers/os/resources/packages/cos_packages.go
+++ b/providers/os/resources/packages/cos_packages.go
@@ -9,7 +9,9 @@ import (
 	"fmt"
 	"io"
 
+	"go.mondoo.com/mql/v13/providers-sdk/v1/inventory"
 	"go.mondoo.com/mql/v13/providers/os/connection/shared"
+	"go.mondoo.com/mql/v13/providers/os/resources/purl"
 )
 
 const (
@@ -28,7 +30,8 @@ type cosPackage struct {
 }
 
 type CosPkgManager struct {
-	conn shared.Connection
+	conn     shared.Connection
+	platform *inventory.Platform
 }
 
 func (cpm *CosPkgManager) Name() string {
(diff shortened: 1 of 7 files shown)


providers/os/resources/packages/cos_packages.go
```diff
@@ -74,14 +77,42 @@ func ParseCosPackages(input io.Reader) ([]Package, error) {
 
 	pkgs := make([]Package, len(cPkgs.InstalledPackages))
 	for i := range cPkgs.InstalledPackages {
-		pkgs[i].Name = cPkgs.InstalledPackages[i].Name
-		pkgs[i].Version = cPkgs.InstalledPackages[i].EbuildVersion
+		src := cPkgs.InstalledPackages[i]
```
czunker: This can be part of the range loop

czunker:
Addressed in 697c59eb24 — pulled `src` into the range expression.

PR #8320, opened 2026-06-10
🐛 llx: stop array contains/difference builtins panicking on a null argument

### LINKED ISSUE #8319: panic in llx.arrayContainsAll when containsAll argument evaluates to null (interface conversion: interface {} is nil)
## Summary

`llx.arrayContainsAll` panics with `interface conversion: interface {} is nil, not []interface {}` when the **argument** to `containsAll` evaluates to null. The function nil-guards the bound array (`bind.Value == nil`) but performs an unguarded type assertion on the argument:

```go
// llx/builtin_array.go:934-935
org := bind.Value.([]any)
argFilters := arg.Value.([]any)   // <-- panics when arg.Value == nil
```

Because policy content is server-served, any MQL bundle containing `<array>.containsAll(<expr that can be null>)` can panic cnspec clients fleet-wide across all versions.

## Production impact

This caused incident INC-2026-06-10-client-panic-k8s: the cnspec-enterprise-policies v5.26.0 release rewrote the `mondoo-linux-security` su-restriction check so `suRestrictedGroups` evaluates to null on hosts without `/etc/pam.d/su` (container-optimized k8s node OSes: COS, Flatcar, Bottlerocket). Result: ~10x panic surge (`panic.received`), 1,050+ panics across 19 spaces in ~3h, spanning cnspec 13.22.0, 13.6.0, 13.5.0, and 11.69.1 simultaneously — the bug is latent across major versions and content-triggered.

## Reproduction (verified with released cnspec 13.22.0, build 261db0c0)

In a rootfs **without** `/etc/pam.d/su` (e.g. `ubuntu:24.04` container after `rm /etc/pam.d/su`):

```bash
cnspec run -c '
suRestrictedPam = pam.conf.entries["/etc/pam.d/su"].where(pamType == "auth" && module == "pam_wheel.so" && params["use_uid"] != null)
suRestrictedGroups = suRestrictedPam.where(params["group"] != null).map(params["group"])
suRestrictedPam != empty
suRestrictedGroups == empty || groups.map(name).containsAll(suRestrictedGroups)
suRestrictedGroups != empty || groups.any(name.in(["wheel", "sudo"]))'
```

Panics with:

```
panic: interface conversion: interface {} is nil, not []interface {}

go.mondoo.com/mql/v13/llx.arrayContainsAll(...)
        .../llx/builtin_array.go:935
```

With `/etc/pam.d/su` present, the same query fails cleanly (no panic).

## Expected behavior

A null argument should produce a clean result or query error (e.g. treat null as the bound array's own null handling does: return `RawData{Type, Error}`), never a runtime panic.

## Scope — audit siblings

The same unguarded `arg.Value.([]any)` shape exists in neighboring builtins; at minimum:

- `arrayContainsAll` (`builtin_array.go:935`) — confirmed panicking in production
- `arrayContainsNone` (`builtin_array.go:~988`)
- `arrayContains`/eq variants around `builtin_array.go:~830`, `~873`
- dict variants (`dictContainsAll`, etc.) should be checked for the equivalent hole

tas50:
Fixes #8319

## Problem

A surge of client scan panics (`panic: interface conversion: interface {} is nil, not []interface {}`) traced to the array `contains*`/`difference` builtins.

`arrayContainsAll`, `arrayContainsNone`, and `arrayDifferenceV2` guard the **receiver** array against null but then perform an unchecked `arg.Value.([]any)` on the **argument**. When the argument resolves to a *typed null array* — e.g. a `map[string][]T` key miss like `pam.conf.services["su"]` on hosts with no `/etc/pam.d` (container-optimized k8s node OSes: COS, Flatcar, Bottlerocket) — that type assertion panics.

Because the MQL executor runs blocks in goroutines, the panic is unrecoverable by the caller and **crashes the entire scan** rather than failing the single check. The `dict` variants (`dictContainsAll`/`dictContainsNone`) already use the safe comma-ok form, so only the array variants were affected.

## Fix

Guard the argument the same way the receiver is already guarded: a null argument propagates as null (placed before the type check so an untyped null short-circuits cleanly too). The compiled `… == []` wrapper then resolves the check to a clean pass/fail.

After the fix, `realArray.containsAll(nullArg)` returns null → the check resolves to `false` (fails cleanly) instead of crashing. Whether a pam-absent host should pass or fail the check remains a policy decision; the engine's responsibility here is simply to not panic.

## Testing

- Reproduced the exact panic with `a = {a: [1,2]}; [1,2,3].containsAll(a['b'])` (typed-null arg via map key miss); confirmed crash pre-fix, clean result post-fix.
- Added regression cases for `containsAll` / `containsNone` / `difference` to `TestArray`.
- `go test ./llx/` and the full `./providers/core/resources/` suite pass.

### DIFF
--- a/llx/builtin_array.go
+++ b/llx/builtin_array.go
@@ -815,6 +815,14 @@ func arrayDifferenceV2(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64
 		return nil, rref, err
 	}
 
+	// A null argument (e.g. a missing map key resolving to a typed null
+	// array) propagates as null, mirroring the receiver-nil guard above.
+	// Without this the `arg.Value.([]any)` assertion below panics, which
+	// crashes the whole scan rather than failing the single check.
+	if arg.Value == nil {
+		return &RawData{Type: bind.Type, Error: arg.Error}, 0, nil
+	}
+
 	t := types.Type(arg.Type)
 	if t != bind.Type {
 		return nil, 0, errors.New("called `difference` with wrong type (got: " + t.Label() + ", expected:" + bind.Type.Label() + ")")
@@ -920,6 +928,14 @@ func arrayContainsAll(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64)
 		return nil, rref, err
 	}
 
+	// A null argument (e.g. a missing map key resolving to a typed null
+	// array) propagates as null, mirroring the receiver-nil guard above.
+	// Without this the `arg.Value.([]any)` assertion below panics, which
+	// crashes the whole scan rather than failing the single check.
+	if arg.Value == nil {
+		return &RawData{Type: bind.Type, Error: arg.Error}, 0, nil
+	}
+
 	t := types.Type(arg.Type)
 	if t != bind.Type {
 		return nil, 0, errors.New("called `arrayNone` with wrong type (got: " + t.Label() + ", expected:" + bind.Type.Label() + ")")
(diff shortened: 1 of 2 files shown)


jaym:
Fixes https://github.com/mondoohq/mql/issues/8319

PR #8321, opened 2026-06-10
🐛 types: stop equality comparators panicking on a null array element

tas50:
## Problem

A second latent panic in the same family as #8320: `panic: interface conversion: interface {} is nil, not string`.

The `types.Equal` comparators (`Bool`, `Int`, `Float`, `String`, `Regex`, `Score`, `Time`) perform unguarded type assertions like `left.(string)`. A **null array element** surfaces as a nil interface — e.g. `["x"].containsAll([asset.labels["nope"]])`, where the map miss yields a null string wrapped in an array — so the assertion panics.

These comparators back **every** array element-comparison builtin (`containsAll`, `containsNone`, `containsOnly`, `difference`). Because the MQL executor runs blocks in goroutines, the panic is unrecoverable and **crashes the whole scan** rather than failing the single comparison.

(`Time` already nil-checked its dereferenced pointer, but still asserted `left.(*time.Time)` on a nil *interface* first, so it was vulnerable to the same crash.)

## Fix

Guard each comparator against nil operands: two nils are equal, a nil and a value are not. This is the natural equality semantics and protects all callers at the root.

## Relationship to #8320

Independent and complementary. #8320 handles the *whole argument array* being null; this handles a null *element inside* an otherwise-present array. Both were uncovered while reproducing the `pam`-absent k8s scan crashes; either can crash a scan on its own. No overlapping lines — they can merge in any order.

## Testing

- Added `TestEqual_NilOperands` unit test in `types/` exercising every comparator with `(value, nil)`, `(nil, value)`, `(nil, nil)`; confirmed it panics pre-fix, passes post-fix.
- Added an e2e regression in `TestArray` (`['x'].containsAll([asset.labels['nope']])`) that drives the full executor path.
- `go test ./types/ ./llx/ ./providers/core/resources/` all pass.

### DIFF
--- a/providers/core/resources/mql_test.go
+++ b/providers/core/resources/mql_test.go
@@ -732,6 +732,15 @@ func TestArray(t *testing.T) {
 			Code:        "a = [1]; [2,1,2,1].containsNone(a)",
 			Expectation: []any{int64(1), int64(1)},
 		},
+		// Regression: a null array element (here a missing map key) surfaces as
+		// a nil interface. The element comparators must handle it instead of
+		// panicking with "interface conversion: interface {} is nil, not
+		// string", which would crash the whole scan. The nil filter never
+		// matches "x", so it remains in the leftover.
+		{
+			Code:        "['x'].containsAll([asset.labels['nope']])",
+			Expectation: []any{nil},
+		},
 		{
 			Code:        "['a','b'] != /c/",
 			ResultIndex: 0, Expectation: true,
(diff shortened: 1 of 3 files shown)


tas50:
Added a clarifying comment on the `Time` comparator explaining the inner nil-pointer check still covers a typed `(*time.Time)(nil)` element (the new interface-nil guard only catches an untyped nil), so it isn't redundant.

PR #8322, opened 2026-06-10
✨ os/pam: add pam.conf.exists to check whether PAM config is present

tas50:
## What

Adds a `pam.conf.exists` bool field: **true** when either the `/etc/pam.d` directory or the `/etc/pam.conf` file is present, **false** otherwise. Unlike `pam.conf.files` (and the fields derived from it), it never errors when no PAM configuration is found.

## Why

Audits that query the `pam` resource (`services`, `modules`, `entries`, …) hit an error or a typed-null on hosts that ship **no** PAM configuration — container-optimized k8s node images such as COS, Flatcar, and Bottlerocket have no `/etc/pam.d`. There was previously no clean way to guard such a check, so policies either errored or fed a null into downstream builtins.

Policies can now guard cleanly:

```mql
pam.conf.exists
# or
if (pam.conf.exists) {
  # ...PAM-based checks (su restriction, faillock, pwquality, ...)...
}
```

This is the resource-level complement to the engine hardening in #8320 / #8321 (which stop the null from crashing scans in the first place): those make the engine robust; this gives content authors an explicit signal to branch on.

## Testing

- Unit test: `pam.conf.exists` returns `false` with no error on a file-less mock.
- Live: returns `true` on a Debian 13 host (30 `/etc/pam.d` files).
- `go test ./providers/os/resources/` passes; codegen re-run is deterministic.

### DIFF
--- a/providers/os/resources/os.lr
+++ b/providers/os/resources/os.lr
@@ -954,6 +954,12 @@ packages {
 // Examine /etc/pam.d service entries with their type, control, and module
 pam.conf {
   init(path string)
+  // Whether PAM configuration is present on the system
+  //
+  // True when either the `/etc/pam.d` directory or the `/etc/pam.conf` file
+  // exists. Use it to guard PAM audits on hosts that ship no PAM
+  // configuration, such as container-optimized OS images.
+  exists() bool
   // List of files that make up the PAM configuration
   files() []file
   // The raw PAM configuration (across all files)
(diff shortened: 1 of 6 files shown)


jaym (approved):

tas50:
Addressed the changes-requested feedback: added `TestPamConfExists` covering the positive path over a mock filesystem — true for `/etc/pam.conf` present, true for the `/etc/pam.d` directory present, and false (no error) when neither exists. The version (`13.26.3`) and shared-constant sync notes were confirmation-only, no change needed.

PR #8330, opened 2026-06-10
⭐ health: attach the running query to panic reports

jaym:
## What

- `health.ReportPanicWithTags(product, version, build, tagsFn, reporters...)` — like `ReportPanic`, but attaches tags supplied at crash time. `tagsFn` is only invoked while a panic is being recovered.
- `health.QueryPanicTags(codeID, source)` + exported tag-key constants (`TagQueryCodeID`, `TagQuerySource`) — builds the query-context tags, source bounded to 2048 runes (mirrors the platform's slow-query bound).
- The exec manager's run goroutine now tracks the in-flight `*llx.CodeBundle` and reports it on panic.

## Why

During INC-2026-06-10-client-panic-k8s (#8319), the panic report told us *where* the engine died (`llx.arrayContainsAll`, `builtin_array.go:935`) but not *what it was executing*. Pinning the crash to the su-restriction check took a policy-release diff plus compiled-chunk-ref forensics in the stacktrace; with these tags the report itself would have named the query immediately. Slow-query reports already carry `CodeID` + query text — this brings panic reports to parity.

## Design notes

- The recover stays at the goroutine top: recovering at the query boundary and re-panicking would truncate the stacktrace, which is the part that identifies the engine bug. Instead the goroutine tracks the current bundle and the deferred reporter snapshots it.
- `SendErrorReq.Tags` already exists in the proto and the platform handler already forwards tags to Sentry; a companion server PR maps these two keys to the existing `mondoo.query.text` / `mondoo.query.code_id` obslog fields.
- cnspec's policy executor gets the same wiring in a companion PR (mondoohq/cnspec).

### DIFF
--- a/exec/internal/execution_manager.go
+++ b/exec/internal/execution_manager.go
@@ -100,7 +100,9 @@ func (em *executionManager) Start() {
 				}
 
 				current = item.codeBundle
-				if err := em.executeCodeBundle(item.codeBundle, props, errMsg); err != nil {
+				err := em.executeCodeBundle(item.codeBundle, props, errMsg)
+				current = nil
+				if err != nil {
 					// an error is returned if we cannot execute a query. This happens
 					// if the lumi runtime doesn't report back expected data, there is
 					// a problem with the lumi runtime, or the query is somehow invalid.
--- a/exec/internal/execution_manager.go
+++ b/exec/internal/execution_manager.go
@@ -9,8 +9,10 @@ import (
 	"time"
 
 	"github.com/rs/zerolog/log"
+	"go.mondoo.com/mql/v13"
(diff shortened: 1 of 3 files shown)


jaym:
Addressed all three review points (the two earlier "resolved" bot replies were premature — the code had not actually changed; it has now):

- `handlePanic` captures `debug.Stack()` once and reuses it for the report and all reporters.
- `panicEvent` writes platform tags last so caller tags cannot overwrite `os`/`arch` (test added).
- The exec manager clears `current` once `executeCodeBundle` returns, so a panic in non-query code between items is not misattributed to the last completed query. Same fix applied to the cnspec companion (mondoohq/cnspec#2773).

PR #8332, opened 2026-06-11
🧹 Bump ECR discovery max page size to 250.

VasilSirakov:
Reduces the chance of being throttled.


VasilSirakov:
apparently these filters are mutually exclusive

PR #8335, opened 2026-06-11
⭐ mqlx: first-class Go API for embedding MQL (ADR 027 + sample implementation)

chris-rock:
## Summary

Embedding MQL in a Go program today means wiring runtimes and providers by hand, compiling through `mqlc.CompilerConfig`, and walking checksum-keyed `map[string]*llx.RawResult` with manual label resolution. This PR proposes a first-class embedding API and ships a working sample of it:

- **ADR 027** (`docs/adr/027-mqlx-go-embedding-api.md`) — the decision record: motivation (including the ~250 lines of plumbing prism carries today), the full API contract, and the follow-up roadmap (exec context plumbing, custom in-process providers, resource expansion in Decode, discovery, prism migration).
- **`mqlx/`** — a purely additive sample implementation with tests, wrapping the existing `mqlc`/`exec`/`providers` machinery.

## Two first-class modes

**Expression mode** — MQL over your own values; no providers, no asset, no subprocesses:

```go
env, _ := mqlx.NewEnv()
q, _ := env.Compile(`props.name == /admin-.*/ && props.count > 3`,
    mqlx.WithProps(map[string]any{"name": "", "count": 0}))
res, _ := q.Eval(ctx, mqlx.WithPropValues(map[string]any{"name": "admin-x", "count": 5}))
res.Value() // true
```

**Asset mode** — connect once, compile once, evaluate against many assets, decode into structs:

```go
conn, _ := env.ConnectLocal(ctx)
res, _ := conn.Query(ctx, "users { name uid }")

var users []struct {
    Name string `mql:"name"`
    UID  int64  `mql:"uid"`
}
res.Decode(&users)
```

## Design highlights

- `Env` (process) → `Conn` (asset, optional) → `Query` (compiled, immutable, concurrent-safe) → `Result`.
- Expression mode runs on an internal runtime backed by the builtin in-process core provider; `map[string]any` props become dicts, so `props.event.process.binary.contains("/tmp")` works.
- `Result.Value()` returns label-resolved plain Go values; per-field execution errors are collected as path-carrying `FieldError`s instead of being dropped (unlike `RawData.Dereference`).
- `Result.Decode()` is a small reflection decoder (`mql` tag → `json` tag → field name) with overflow checks — no JSON round-trip, preserving int64 >2^53, time, and IP fidelity.
- Compile errors come back as `*CompileError` with `Unwrap()`, keeping typed errors like `mqlc.ErrIdentifierNotFound` reachable via `errors.As`.

## Notes for reviewers

- Found while testing: executing a bundle with zero entrypoints (e.g. `a = 1`) through `exec.ExecuteCode` intermittently panics in `llx._newBlockExecutor` ("no callback points"). The facade guards this the same way `exec.Exec` does; the underlying llx edge may deserve its own fix.
- `parse.*` lives in the os provider, so it is intentionally not part of expression mode (core resources like `regex.*`, `time.*`, and semver are).
- The expression-engine design and prism's needs are validated in the ADR; prism is the intended first migration target.

## Test plan

- `go test ./mqlx/` — expression tests against the real in-process core runtime, asset tests against the recorded Linux mock, decoder units, runnable doc example. Verified with `-count` repetition and `-race`.
- `MQLX_E2E=1 go test ./mqlx/ -run TestConnectLocalE2E` — opt-in end-to-end against the real local OS (passed locally: macOS, decoded into a struct).
- `gofmt`, `golangci-lint run mqlx/...` (0 issues), `go mod tidy` clean; ADR terms added to the spell-check expect list.

### DIFF
--- a/docs/adr/027-mqlx-go-embedding-api.md
+++ b/docs/adr/027-mqlx-go-embedding-api.md
@@ -1,0 +1,245 @@
+# ADR 027: mqlx — A First-Class Go API for Embedding MQL
+
+## Status
+
+Proposed
+
+## Context
+
+MQL is a powerful engine, but Go developers who want to embed it face a wall
+of internal machinery. To run a single query and read its results, a consumer
+today must:
+
+1. Wire up a runtime: `providers.Coordinator.NewRuntime()`, detect and start
+   the right provider, and connect to an asset.
+2. Compile: build a `mqlc.CompilerConfig` from the runtime schema and feature
+   flags, call `mqlc.Compile`, and handle a `*llx.CodeBundle`.
+3. Execute: call `exec.ExecuteCode` and receive a checksum-keyed
+   `map[string]*llx.RawResult`.
+4. Consume: walk the raw results, dereference blocks (whose keys are code
+   checksums, not field names), and map every checksum back to a
+   human-readable label through `CodeBundle.Labels` — by hand.
+
+None of these steps is documented as a public workflow; each consumer
+rediscovers it from our own CLI internals and test helpers.
+
+### Evidence: what consumers build today
+
+Prism (our Tetragon-based threat detection engine) is the most demanding MQL
+embedder and shows the real cost. It carries roughly 250 lines of plumbing:
+
+- Manual schema registration, runtime creation, and `ConnectedProvider`
+  wiring for its in-process provider (`detectionengine.go`).
+- Event-to-props conversion through `convert.JsonToDict` and
+  `llx.DictData(...).Result()`.
+- A hand-built 122-line decoder (`mqlcencoding.MQLCDecoder`) that walks
+  resources via `GetData` calls and mapstructure hooks to turn results into
+  typed structs — with open TODOs for unhandled cases.
+- Recompilation of every query for every kernel event, with the TODO
+  "pre-compile all queries passed in by the policy" marking the missing
+  compile-once API.
+
+Every Go service that embeds MQL repeats some subset of this work.
+
+### The missing use case: MQL without infrastructure
+
+Beyond infrastructure queries, there is demand for using MQL as a standalone
+expression engine: evaluating policies and filters over data the caller
+already has (an event, a request, a finding) with no asset, no provider
+install, and no subprocesses. Prism works exactly this way — every query runs
+over an event passed in as a property. The engine supports this today
+(operators, string/array/map/dict functions, and the builtin core resources
+such as `regex` and `time` all run in-process), but there is no API that
+exposes it; consumers must discover the core-provider wiring on their own.
+
+## Desired Outcome
+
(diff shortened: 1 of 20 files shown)


chris-rock:
Thanks for the review — addressed in the latest push:

- **`props.go` uint64 overflow:** `uint64` (and `uint`, which is 64-bit on most platforms) above `math.MaxInt64` now returns an error instead of silently wrapping.
- **`decode.go` float→int range:** added an explicit `math.MaxInt64`/`math.MinInt64` range check before `int64(v)`, so out-of-range floats (e.g. `1e19`) error instead of hitting implementation-defined conversion. The `||` short-circuits so `OverflowInt` only runs once the conversion is well-defined.
- **`conn.go` Connect after Close:** `trackConn` now checks `e.closed` under the lock and returns an error; `Connect` closes the freshly-created runtime and returns that error rather than tracking it on a closed Env.
- **`NewEnv` concurrent schema mutation:** documented on `NewEnv` that the schema-loading options (`WithProviders`, `WithPrivateProvider`) mutate the process-global coordinator, so concurrent `NewEnv` with those options isn't safe — construct once before sharing.

Added tests for the uint64 and float-range error paths.

On the **permanent `exprOnce` failure** suggestion: I left it as-is. The expression runtime is backed only by the builtin in-process core provider, so its initialization is deterministic — if it fails once it'll fail identically every time, so a retry path wouldn't recover anything. As you noted, a mutex there would be overkill.

chris-rock:
Thanks — second round addressed:

- **Embedded struct fields (`decode.go`):** `structFields` now recurses into anonymous (embedded) structs and promotes their fields, matching `encoding/json` — an embedded field with an explicit tag stays a normal named field, and a shallower field wins a name collision. Pointer embeds (`*Base`) are allocated on demand via a new `fieldByIndexAlloc` that walks the index path. Documented on `Decode`'s godoc and covered by a new `TestDecodeEmbedded` (value embed, pointer embed, and shadowing).
- **Core provider callbacks (`mqlx.go`):** the expression runtime no longer connects core with `nil` callbacks. Added `providers.Runtime.UseBuiltinProvider`, which connects an in-process provider with `&providerCallbacks{runtime: r}` (same as the normal `Connect` path) — this is the right place for it since `providerCallbacks` is unexported, and it now also backs the core connection that private providers may reach through the runtime.

On the **`decodeMap` vs `decodeStruct` consistency** suggestion — I left it as-is. Both paths aggregate per-element errors with `errors.Join` and surface them from `Decode`, so no failure is swallowed. The remaining difference is cosmetic: a failed map entry is omitted (a zero-valued entry would be misleading for a map), while a struct field is left at its zero/partial value (its natural reflect default). Forcing them identical would mean either inventing zero map entries or resetting partially-decoded struct fields, neither of which improves the contract. Happy to revisit if you feel strongly.

PR #8348, opened 2026-06-13
🐛 report effective version when auto-update staged a newer binary (#7751)

### LINKED ISSUE #7751: 'version' command reports old version when auto-update is enabled
`mql version` and `cnspec version` report the old (installed) version when auto-update is enabled and a newer version has been fetched
```
-> % cnspec version
cnspec 13.9.0 (99ea09e9, 2026-05-12T09:00:42Z)
```

```
-> % cnspec status
→ auto-update: using the latest installed version installed=13.10.0
→ loaded configuration from $HOME/.config/mondoo/mondoo.yml using source default
→ Platform:		macos
→ Version:		26.4.1
→ Hostname:		infra-macm1pro
→ IP:			10.2.0.206
→ Time:			2026-05-19T14:06:00+02:00
→ Version:		13.10.0 (API Version: 13)
→ Latest Version:	13.10.0
→ Updates URL:		https://releases.mondoo.com/
→ Providers URL:	https://releases.mondoo.com/providers
→ Installed Providers:	os | aws | azure | gitlab | recording | terraform | gcp | github | slack | core | sbom | network | cloudformation | k8s | mock | unifi
→ Outdated Providers:	azure | gcp | unifi
→ API ConnectionConfig:	https://api.mondoo.com/
→ API Status:		SERVING
→ API Time:		2026-05-19T12:06:00Z
→ API Version:		12
! API versions do not match, please update the client
```

tas50:
## Problem

`mql version` / `cnspec version` report the stale **compiled-in** version when auto-update is enabled and a newer binary has already been staged, even though every other command transparently runs the newer one. This makes `version` disagree with `status`:

```
-> % cnspec version
cnspec 13.9.0 (99ea09e9, 2026-05-12T09:00:42Z)

-> % cnspec status
→ auto-update: using the latest installed version installed=13.10.0
→ Version:        13.10.0 (API Version: 13)
```

Fixes #7751.

## Root cause

In `apps/mql/mql.go`, `shouldTrySelfUpdate()` short-circuited the **entire** self-update path for `version` (and `help`) so it would "run fast":

```go
case "version", "help", "--help", "-h", "--version":
    return false
```

But the part that switches into a staged newer binary — `selfupdate.execLocalIfNewer` — is **local-only** (a `stat` + a 5s local `version` call); the network round-trip happens later. By skipping the whole path, `version` never switched into the staged binary and just printed its own old compiled-in version. `status` (and everything else) went through the full path, exec'd into the staged binary, and reported the new version — hence the mismatch.

## Fix

Split the self-update entry point so `version` does a **local-only** check:

- `selfupdate.CheckLocalAndUpdate` — new exported entry point that runs the shared guards and `execLocalIfNewer` only (no network). Common guards are extracted into `updatePreflight`, shared with the existing `CheckAndUpdate` (no behavior change there).
- `selfUpdateMode()` (replaces `shouldTrySelfUpdate()`) returns `(run, localOnly)`. `version`/`--version` → `localOnly` (switch into a staged newer binary, no network); `help`/`-h` → skip entirely. The `MONDOO_AUTO_UPDATE` / `MONDOO_AUTO_UPDATE_ENGINE` env opt-outs, the `AutoUpdateEngine` feature flag, the `auto_update` config setting, and the `--auto-update=` flag are all still honored.

`version` stays network-free; in the common case (nothing staged) it's a single `os.Stat` that returns immediately.

## Verification

Built two binaries stamped `13.9.0` (one from `main`, one from this branch) with a fake newer `13.10.0` staged in an isolated `HOME`:

| Scenario | `main` | this PR |
|---|---|---|
| `version` (13.10.0 staged) | `mql 13.9.0` ❌ | `mql 13.10.0` ✅ |
| `help` | banner | banner, no switch ✅ |
| `version` w/ `MONDOO_AUTO_UPDATE=false` | — | `mql 13.9.0` (opt-out honored) ✅ |

Plus unit tests for `CheckLocalAndUpdate` (guards + network-free path). `go build`, `go vet`, and `gofmt` all clean. (`TestSwapBinaryInPlace/no-op_when_staged_equals_original` fails on `main` too — a pre-existing `/var`→`/private/var` symlink quirk in that test, unrelated to this change.)

## cnspec follow-up

The shared `selfupdate` change lives here and is reused by cnspec. cnspec needs the same one-line wiring in its own `main` (call `CheckLocalAndUpdate` for `version`) once it picks up this release — its `version` command has the identical short-circuit today.

### DIFF
--- a/apps/mql/mql.go
+++ b/apps/mql/mql.go
@@ -30,7 +30,7 @@ func main() {
 	normalizeAutoUpdateFlag()
 
 	// Check for self-update before anything else
-	if shouldTrySelfUpdate() {
+	if run, localOnly := selfUpdateMode(); run {
 		releaseURL := selfupdate.DefaultReleaseURL
 		if updatesURL := config.GetUpdatesURL(); updatesURL != "" {
 			releaseURL = updatesURL + "/mql/latest.json"
@@ -42,7 +42,13 @@ func main() {
 			BinaryName:      "mql",
 			CurrentVersion:  mql.GetVersion(),
 		}
-		if updated, err := selfupdate.CheckAndUpdate(cfg); err != nil {
+		// "version" does only the local switch (no network) so it stays fast while
+		// still reporting the version that other commands transparently exec into.
+		check := selfupdate.CheckAndUpdate
+		if localOnly {
+			check = selfupdate.CheckLocalAndUpdate
+		}
+		if updated, err := check(cfg); err != nil {
 			// Log warning but don't block - only show in debug mode
 			if os.Getenv("DEBUG") != "" {
 				os.Stderr.WriteString("self-update check failed: " + err.Error() + "\n")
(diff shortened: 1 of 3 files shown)


arlimus (approved):
Nice cleanup, this is much cleaner UX

PR #8351, opened 2026-06-13
🐛 status: add an implicit timeout to network calls

### LINKED ISSUE #7357: status command should have an implicit timeout
## Problem

The `status` command can hang indefinitely when the Mondoo API is unreachable — for example, when running behind a corporate proxy without proxy settings in `mondoo.yml`.

This was observed in the installer (mondoohq/installer#694): when installing without a registration token behind a proxy, no `mondoo.yml` is created, so `cnspec status` has no proxy config and blocks forever trying to reach the API. The install script appeared to hang after a successful installation.

We've added a 10-second timeout wrapper in the install script as a workaround (mondoohq/installer#696), but the root fix belongs here — the `status` command itself should not block indefinitely on network calls.

## Expected behavior

The `status` command should have a reasonable implicit timeout (e.g. 10-30 seconds) on any network calls. If the API is unreachable within that window, it should return an error/non-zero exit code rather than hanging.

tas50:
## Problem

The `status` command can hang indefinitely when the Mondoo API is unreachable — for example, behind a corporate proxy with no proxy settings in `mondoo.yml`. This was observed in the installer (mondoohq/installer#694), where `cnspec status` blocked forever after a successful install. A 10s timeout was added in the install script as a workaround (mondoohq/installer#696), but the root fix belongs in the `status` command itself.

Fixes #7357.

## Root cause

`status` makes several network calls with no overall deadline:

- `health.CheckApiHealth`, `mql.GetLatestReleaseName`, and `PingPong` each rely only on the HTTP client's 30s timeout, so their sum can run well past what callers expect.
- The worst offender is the provider-version check (`getProviders` → `providers.LatestVersion`): it retries each provider (`RetryMax=3`) and only circuit-broke on `ECONNREFUSED`. Behind a blackholed proxy you get *timeouts*, not connection-refused, so it looped over every installed provider — effectively hanging.

## Fix

Bound the entire command with a single **30s deadline** and thread a `context.Context` through every network call so they all share it:

- `checkStatus`, `RenderCliStatus`, and `getProviders` now take a `context.Context`.
- The health check, latest-release fetch, and `PingPong` use the shared deadline.
- `providers.GetLatestVersion` now honors its `context` (it previously ignored it) via `http.NewRequestWithContext`.
- `getProviders` also circuit-breaks on `context.DeadlineExceeded`/`Canceled`, so it stops checking the moment the deadline expires.

To avoid breaking external callers (e.g. cnspec), the SDK-exported helpers gained backward-compatible context variants: `health.CheckApiHealthContext` and `mql.GetLatestReleaseNameContext`. The original `CheckApiHealth` / `GetLatestReleaseName` now delegate with `context.Background()`.

When the API is unreachable, `status` now returns within ~30s with a non-zero exit code instead of hanging.

## Testing

- `go build` of the affected packages passes.
- `go vet` clean on changed packages (the pre-existing `providers/*_test.go` mockgen failures are unrelated).

### DIFF
--- a/apps/mql/cmd/status.go
+++ b/apps/mql/cmd/status.go
@@ -33,6 +33,10 @@ func init() {
 	rootCmd.AddCommand(StatusCmd)
 }
 
+// statusTimeout bounds the total time the status command spends on network
+// calls so it cannot hang indefinitely when the API is unreachable.
+const statusTimeout = 30 * time.Second
+
 // StatusCmd represents the version command
 var StatusCmd = &cobra.Command{
 	Use:   "status",
@@ -48,7 +52,12 @@ Status sends a ping to Mondoo Platform to verify the credentials.
 
 		config.DisplayUsedConfig()
 
-		s, err := checkStatus()
+		// Bound all network calls with a deadline so the command cannot hang
+		// indefinitely when the API is unreachable.
+		ctx, cancel := context.WithTimeout(cmd.Context(), statusTimeout)
+		defer cancel()
+
+		s, err := checkStatus(ctx)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


arlimus (approved):
yes please! nice

PR #8358, opened 2026-06-13
✨ aws: expose object provenance across EC2, databases, CloudFormation, Backup, ECS & Signer

tas50:
## What

Adds **provenance / lineage** data across the AWS provider so audits can trace where an object came from, what it derives from, and how it was signed. Pure additions (no breaking changes); the only new dependency is `service/signer`.

### Tier 1 — derivation lineage
**EC2**
- `aws.ec2.image`: `sourceInstanceId` + `sourceInstance()`, `sourceImage()`, `productCodes`
- `aws.ec2.volume`: `snapshotId` + `snapshot()`, `fastRestored`
- `aws.ec2.snapshot`: `volume()`, `dataEncryptionKeyId`, `ownerAlias`, `outpostArn`

**Databases** (raw lineage strings — sources are often cross-region/cross-account ARNs)
- `aws.rds.snapshot`: `sourceSnapshot`, `sourceRegion`, `originalCreatedAt`
- `aws.rds.dbinstance`: `readReplicaSourceInstanceId`, `readReplicaSourceClusterId`
- `aws.neptune.snapshot` / `aws.documentdb.snapshot`: `sourceSnapshot`
- `aws.redshift.snapshot`: `sourceRegion`

### Tier 2 — IaC ownership + backup lineage
- `aws.cloudformation.stack.resources()` → new `aws.cloudformation.stack.resource` (logical-to-physical mapping + drift) via `ListStackResources` — the authoritative "this resource was created by this stack" link.
- `aws.backup.vaultRecoveryPoint`: `sourceResourceArn` (what it backs up), `sourceBackupVaultArn` (what it was copied from)

### Tier 3 — supply chain
- `aws.ecs.container`: `imageDigest` (resolved digest of the *running* image)
- New `aws.signer` namespace + `aws.signer.signingProfile`

## Design note: best-effort typed source refs

The typed source references (`sourceImage`, `sourceInstance`, `volume`, `snapshot`) are **best-effort**. Provenance sources are frequently foreign-owned (e.g. a root volume's snapshot is Amazon-owned) or deleted (snapshots outlive their volumes; AMIs outlive their source instances). So:

- the **raw `*Id` field is always populated** and is the durable lineage signal;
- the **typed ref resolves to the live resource when it exists in the account, and degrades to `null` otherwise** — it never errors.

This is a deliberate change from the initial "deprecate the raw `*Id` in favor of the typed ref" plan: live testing showed an erroring ref would break the *entire* parent query (e.g. `aws.ec2.volumes { snapshot }` fails if any volume references an inaccessible snapshot). Keeping the raw id is what actually preserves provenance.

## Permissions
New IAM permissions, auto-extracted into `aws.permissions.json`: `cloudformation:ListStackResources`, `signer:ListSigningProfiles`.

## Verification (live, against a real account)
- `aws.ec2.volumes { snapshotId snapshot { id } }` — `snapshotId` populated, `snapshot` degrades to `null` for Amazon-owned source snapshots, **no error**.
- `aws.ec2.snapshots.first { volumeId volume { id } ownerAlias }` — `volumeId` populated, `volume` null (source deleted), graceful.
- `aws.cloudformation.stacks.first { resources { logicalId resourceType status driftStatus } }` — returns real stack-resource mappings.
- `aws.signer.signingProfiles` — `[]`, gracefully skipping regions without a Signer endpoint (`IsServiceNotAvailableInRegionError`).
- `aws.ec2.images { sourceImageId productCodes }` — returns live data.
- ECS `imageDigest` compiles/resolves (account has no running tasks to populate a value); DB-snapshot source fields compile and resolve (need a snapshot copy to show a non-null value).
- Provider builds clean; `go vet` clean; `go mod tidy` clean; generated `aws.lr.go` / `aws.lr.versions` (@ `13.35.2`) / `aws.permissions.json` regenerated.


tas50:
Thanks for the review — addressed the typed-reference feedback, with one principled exception.

**Changed — RDS read-replica sources are now typed (best-effort).** Added `readReplicaSourceInstance() aws.rds.dbinstance` and `readReplicaSourceCluster() aws.rds.dbcluster` alongside the raw ids. These resolve cleanly because the source primary is usually live and in-account (`aws.rds.dbinstance`/`dbcluster` have init resolvers). Verified live: a read replica resolves `readReplicaSourceInstance` to its actual source instance; the raw `readReplicaSourceInstanceId` is retained as the durable signal.

**Kept raw — the DB snapshot `sourceSnapshot` fields and `sourceBackupVaultArn`.** These point at copy sources, and I deliberately left them as raw ARN/identifier strings for two reasons:

1. **No resolver on the target.** `aws.rds.snapshot`, `aws.neptune.snapshot`, `aws.documentdb.snapshot`, and `aws.backup.vault` have no init-by-ARN function. A typed accessor would `NewResource(...)` a husk whose only populated field is the ARN — `sourceSnapshot { status }` would be null because nothing fetches it. That's strictly worse than a raw ARN.
2. **Sources cross region/account.** Snapshot copies are overwhelmingly cross-region (DR) or cross-account (shared), and a copied recovery point's source vault is typically in another vault/account. Even *with* a resolver these would resolve to null in the common case, because the source isn't in this account's inventory.

The raw ARN is the correct representation here: it's the durable lineage signal you can record even when the source lives outside your inventory — which is the whole point of provenance. This is the same resolvability-based rule applied to the EC2 refs in this PR (`sourceImage`/`sourceInstance`/`sourceVolume`/`volume.snapshot` *are* typed+graceful, because their targets have resolvers and same-account ARN construction works).

Happy to add init-by-ARN resolvers to the snapshot resources + typed `sourceSnapshot()` accessors in a follow-up if the team wants standalone snapshot-by-ARN resolution — it's just a larger change for references that still resolve to null whenever the source is cross-boundary.

Issue #5803, opened 2025-07-21
Adding empty Arrays giving unexpected behaviour


AdamVB:
**Describe the bug**
1. Adding an empty Array to a non-empty array deletes all the content of the array.
2. adding a non-empty array to an empty array leads to emoji in the output.

**To Reproduce**
```

cnspec> ["a"] + ["b"]
+: [
  0: "a"
  1: "b"
]
cnspec> ["a"] + ["b"] + ["c"]
+.+: [
  0: "a"
  1: "b"
  2: "c"
]
cnspec> ["a"] + ["b"] + ["c"] + [""]
+.+.+: [
  0: "a"
  1: "b"
  2: "c"
  3: ""
]
cnspec> ["a"] + ["b"] + ["c"] + []
+.+.+: []
cnspec> [] + ["a"] + ["b"] + ["c"]
+.+.+: [
  0:   🤷 "a"
  1:   🤷 "b"
  2:   🤷 "c"
]

```
**Expected behavior**
Adding an empty array should not affect the content of the existing array. Also, the order shouldn't make a difference.

**Screenshots or CLI Output**

<img width="822" height="1136" alt="Image" src="https://github.com/user-attachments/assets/eacc2450-03b4-45c5-9be5-4b581ed4a1bb" />

cnspec 11.63.0

tas50:
Reviewed as part of a full backlog triage — **confirmed still open** in the current codebase.

llx/builtin_array.go:1096-1110 tarrayConcatTarrayV2: when the right operand is an empty array (items.Value==nil) it returns &RawData{Type: items.Type} (lines 1096-1098), discarding the bound left array; and when the left (bind) is empty, bind.Value.([]any) is nil so line 1105 returns 'cannot add arrays to null' (the 🤷 error rendering). No test covers concat-with-empty (llx/builtin_array_test.go has no such case).

**Suggested fix approach:** In tarrayConcatTarrayV2 (llx/builtin_array.go), treat an empty/nil right operand as a no-op returning bind, and treat an empty/nil left (bind) operand as the right operand instead of erroring — i.e. coalesce nil slices to []any{} before concatenating so order doesn't matter.

PR #8368, opened 2026-06-13
🐛 MQL: fix empty-left array concat losing element type (#5803)

tas50:
## What

Fixes the remaining half of #5803. Concatenating an empty array on the **left** rendered every element as `🤷`:

```
cnspec> [] + ["a"] + ["b"] + ["c"]
+.+.+: [
  0:   🤷 "a"
  1:   🤷 "b"
  2:   🤷 "c"
]
```

## Root cause

This was a **type** bug, not a value bug — the values were always correct, which is why it slipped past existing value-only tests. In `tarrayConcatTarrayV2` (`llx/builtin_array.go`) the result always took its element type from the left (`bind`) operand. An empty literal `[]` carries no element type (`[]unset`), so the concatenated result was typed `[]unset` even though it held strings, and the reporter renders untyped elements as `🤷`. The function also still carried a dead `"cannot add arrays to null"` error branch for a nil left operand.

The earlier fix (#7196) only covered empty arrays on the **right** (`["a"] + []`), by making builtin-produced empties non-nil. This handles the left side.

## Fix

Coalesce both operands and treat an empty side as a no-op that returns the *other* side untouched, so the non-empty operand supplies both the values and the element type — and order no longer matters:

```go
if len(right) == 0 { return bind, 0, nil }
if len(left)  == 0 { return items, 0, nil }
// else copy left ++ right
```

## Verification

- Built from source and confirmed all five repro lines from the issue now behave correctly; `🤷` is gone.
- Added a **type-asserting** regression test (`[] + ["a","b"]` must be typed `Array(String)`) — it fails without the fix (`Array(unset)` vs `Array(String)`) and passes with it. A value-only test would not have caught this.
- Added value-level regression cases for both operand orders and chained `+`.
- `llx` and `providers/core/resources` suites pass; `gofmt` clean.

### DIFF
--- a/llx/builtin_array.go
+++ b/llx/builtin_array.go
@@ -1132,33 +1132,23 @@ func tarrayConcatTarrayV2(e *blockExecutor, bind *RawData, chunk *Chunk, ref uin
 		return nil, rref, err
 	}
 
-	if items.Value == nil {
-		return &RawData{Type: items.Type}, 0, nil
-	}
+	left, _ := bind.Value.([]any)
+	right, _ := items.Value.([]any)
 
-	v, _ := bind.Value.([]any)
-	if v == nil {
-		if items.Value == nil {
-			return &RawData{Type: bind.Type}, 0, nil
-		}
-		return nil, 0, errors.New("cannot add arrays to null")
-	}
-
-	list := items.Value.([]any)
-	if len(list) == 0 {
+	// Concatenating an empty array is a no-op, and the result must not depend on
+	// operand order. Returning the non-empty side as-is also lets it supply the
+	// element type: an empty literal `[]` carries no element type, so deriving
+	// the result type from it would render the other side's items as unknown.
+	if len(right) == 0 {
 		return bind, 0, nil
 	}
+	if len(left) == 0 {
+		return items, 0, nil
+	}
 
-	res := make([]any, len(v)+len(list))
-	var idx int
-	for i := range v {
-		res[idx] = v[i]
-		idx++
-	}
-	for i := range list {
-		res[idx] = list[i]
-		idx++
-	}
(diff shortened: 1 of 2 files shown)


tas50:
Closing as superseded by #9749 (`547304342`, "fix array concat with empty and null operands"), which landed on `main` and covers this fix as a strict superset.

**Why this is redundant rather than just conflicting.** Rebasing onto `main` conflicts inside `arrayConcat` because both sides rewrote the same function for the same reason. `main` resolves the element-type loss with `types.CoalesceArrays(bind.Type, items.Type)`; this branch resolved it with early returns on the empty side. Both fix `[] + ["a"]` yielding `[]unset` — but #9749 additionally handles the null-operand cases this branch never did:

```
["a"] + nullArray  ->  null, silently dropping "a"
nullArray + ["a"]  ->  error "cannot add arrays to null"
```

**Verified rather than assumed.** I lifted this PR's own distinctive assertions — the `types.Array(types.String)` element-type checks, not just the values — into a throwaway test and ran them against pristine `main`. All pass unmodified:

```
--- PASS: .../[]_+_["a","b"]
--- PASS: .../["a","b"]_+_[]
--- PASS: .../[1,2]_+_[]_+_[3]
--- PASS: .../[]_+_["a"]_+_["b"]_+_["c"]
--- PASS: .../["a"]_+_["b"]_+_["c"]_+_[]
```

`main`'s test suite also already covers the ordering cases (`[1] + [] + [2]`, `[] + [1] + [2]`), the downstream-type case (`paths = [] + ['a','bb']` then `.map(_.length)`) and the null-operand cases.

So there is nothing left to merge: resolving the conflict either way would revert `main`'s better implementation or land a no-op. The branch is left in place if the commentary is worth salvaging.

PR #8371, opened 2026-06-13
🐛 os: recover macOS package versions from Info.plist when system_profiler omits them

### LINKED ISSUE #4738: We are missing versions for some MacOS packages
**Describe the bug**
```
pkg: {
  Arch: ""
  Format: "macos"
  Name: "liquiddetectiond"
  Origin: ""
  Purl: ""
  Status: ""
  Version: ""
  cpes.map: null
  files.map: null
}
```

**To Reproduce**

1. Run a `packages` query on a Mac
4. Note the missing data

**Expected behavior**
We should see a verison.

**Additional context**
Same on Windows: https://github.com/mondoohq/cnquery/issues/4737

tas50:
## What

Fixes #4738 — some macOS packages report an empty `Version`.

## Root cause

`system_profiler SPApplicationsDataType -xml` only surfaces **`CFBundleShortVersionString`** as a package's `version`, and omits the key entirely when that string is absent. On a test Mac, 116 of 503 app entries had no version. They split into two groups:

- **Genuinely versionless** (~111): bare `.app` directories with no `Contents/Info.plist` at all (e.g. `liquiddetectiond.app` — just an executable + a PDF), system daemons on the sealed system volume, and non-app entries like `com.apple.ctcategories` (an `~/Library/HTTPStorages/…service` dir) and Firefox profile storage dirs (`https+++….app`). These carry no version *anywhere* — nothing can recover one.
- **Recoverable**: real app bundles that ship a version only in **`CFBundleVersion`** (e.g. a Microsoft Teams PWA at `7778.181`, with no `CFBundleShortVersionString`). `system_profiler` reports these blank even though the version is right there in the bundle.

## Fix

When `system_profiler` reports no version, fall back to reading the bundle's `Contents/Info.plist` via the connection filesystem (works over SSH too — same pattern as the existing Firefox `application.ini` read), preferring `CFBundleShortVersionString`, then `CFBundleVersion`. The resolved version feeds both `pkg.Version` and the PURL.

## Tests

- New Teams-PWA-style fixture (blank `version`, Info.plist with only `CFBundleVersion`) — version now recovered to `7778.181`.
- New bare-daemon `liquiddetectiond` fixture (no Info.plist) — version correctly stays empty.
- Full `./resources/packages` suite + `go vet` pass.

## Note

The two examples in the issue (`liquiddetectiond`, `com.apple.ctcategories`) are unfixable by definition — they have no version metadata at all. A related observation: a chunk of those 116 entries aren't applications at all (HTTPStorages, browser-profile storage dirs ending in `.app`). Filtering that noise out of the package list would be a reasonable separate follow-up if desired.

### DIFF
--- a/providers/os/resources/packages/macos_packages.go
+++ b/providers/os/resources/packages/macos_packages.go
@@ -32,6 +32,13 @@ type sysProfiler struct {
 	Items []sysProfilerItem `plist:"_items"`
 }
 
+// infoPlist holds the version keys we care about from an app bundle's
+// Contents/Info.plist.
+type infoPlist struct {
+	ShortVersion  string `plist:"CFBundleShortVersionString"`
+	BundleVersion string `plist:"CFBundleVersion"`
+}
+
 // parse macos system version property list
 func ParseMacOSPackages(conn shared.Connection, platform *inventory.Platform, input io.Reader) ([]Package, error) {
 	var r io.ReadSeeker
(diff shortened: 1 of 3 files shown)


tas50:
@czunker This might be the fix for the issue you hit

czunker (approved):
Thanks @tas50

PR #8378, opened 2026-06-13
🐛 mqlc/llx: fix switch with constant-boolean case conditions

### LINKED ISSUE #1174: switch statement produces incorrect results
```
cnquery run local -c '
switch {
  case true:
  1;
  case false:
  2;
}
'
```
```
→ loaded configuration from /home/jaym/.config/mondoo/mondoo.yml using source default
switch: 2
```

```
cnquery run local -c '
switch {
  case true:
  1;
  default:
  2;
}
```
```
→ loaded configuration from /home/jaym/.config/mondoo/mondoo.yml using source default
switch: 2
```

tas50:
## What

`switch` returned the wrong result whenever a case condition was a constant boolean:

```
switch { case true: 1; case false: 2; }   → 2   (should be 1)
switch { case true: 1; default: 2; }       → 2   (should be 1)
```

Reported in #1174.

## Root cause

The compiler encodes the `default` case with a synthetic condition of `BoolPrimitive(true)` (`mqlc/mqlc.go`), and the runtime identifies the default case by checking whether a case's condition arg is `Bool`-typed (`llx/builtin_global.go`). A real `case true:` / `case false:` *also* compiles to a `Bool` primitive, so the runtime misidentified those cases as the default, skipped them in the match loop, and fell through to the last default-looking arg.

Non-constant conditions (`case _ > 0:`, `case 1 == 1:`) compile to a `Ref`, not a `Bool` primitive, so everyday `switch` usage was unaffected — only literal-boolean cases broke.

## Fix

- Compiler now marks the `default` case with `UnsetPrimitive` instead of `BoolPrimitive(true)`. A real case condition never compiles to an unset type, so the marker is unambiguous.
- Runtime detects the default case by that unset sentinel rather than by `== types.Bool`. Constant-boolean cases now flow through the normal truthy evaluation like any other case.

## Tests

- Added `TestMqlSwitch` (`exec/exec_test.go`) covering literal-bool cases, default fall-through, and non-constant conditions — fails before the fix, passes after.
- Updated `TestCompiler_Switch` to expect the new `UnsetPrimitive` default encoding.
- `mqlc`, `llx`, and `exec` suites all pass.

Fixes #1174

### DIFF
--- a/exec/exec_test.go
+++ b/exec/exec_test.go
@@ -72,6 +72,36 @@ func TestMqlSimple(t *testing.T) {
 	}
 }
 
+// Regression test for https://github.com/mondoohq/mql/issues/1174: a case whose
+// condition is a constant boolean (`case true:` / `case false:`) used to compile
+// to the same Bool primitive the compiler emits for `default`, so the runtime
+// misidentified it as the default case and fell through to the last default-looking
+// arg. These must now match by their actual truthiness.
+func TestMqlSwitch(t *testing.T) {
+	tests := []struct {
+		query     string
+		assertion any
+	}{
+		{"switch { case true: 1; case false: 2; }", int64(1)},
+		{"switch { case false: 1; case true: 2; }", int64(2)},
+		{"switch { case true: 1; default: 2; }", int64(1)},
+		{"switch { case false: 1; default: 2; }", int64(2)},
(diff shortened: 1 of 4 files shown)


tas50:
@jaym This fixes your old issue

arlimus (approved):
Nice fix, kudos!

PR #8384, opened 2026-06-14
⭐ os: detect platform arch from ELF header on command-less fs scans

### LINKED ISSUE #983: fs scan cannot determine platform arch
The mondoo-operator is using filesystem scan to scan k8s nodes. In that case the `Platform Arch` is `null`

**Describe the solution you'd like**
Figure out a way to determine the platform architecture from a filesystem scan.  I believe at least in some cases we should be able to reliably determine that. In the case of the operator we could even provide the arch from the operator since the architecture is available in k8s itself.

tas50:
## Problem

Fixes #983.

A pure filesystem scan (e.g. the mondoo-operator scanning k8s nodes) has no command capability — `RunCommand` returns `ErrRunCommandNotImplemented` — so every `pf.Arch = unamem` detector path silently leaves `Platform Arch` empty/`null`. There was no filesystem-based fallback.

## Fix

- **`providers/os/detector/arch.go`** (new): `archFromELF()` opens the first existing well-known binary (`/bin/ls`, `/usr/bin/ls`, `/bin/sh`, `/sbin/init`, `/usr/lib/systemd/systemd`, `/bin/busybox`), parses its ELF header via `debug/elf`, and maps `e_machine` (with endianness for ppc64/ppc64le) to the uname-style arch string the detector already uses (`x86_64`, `aarch64`, `i386`, `arm`, `ppc64le`, `ppc64`, `s390x`, `riscv64`, `mips`, `ppc`). Unknown machines return `""` so we never write garbage.
- **`providers/os/detector/platform_resolver.go`**: the fallback is wired into the single chokepoint in `Resolve()` — when detection succeeds but `Arch` is still empty, it tries `archFromELF(conn.FileSystem())`. Command-based detection (`uname -m`) still wins whenever available; this only fills the gap. The 6 existing `pf.Arch = unamem` sites are untouched.

This fixes the operator's k8s-node case and every other command-less filesystem scan with **zero operator-side changes**.

## Tests

- `arch_test.go`: table tests for the machine→arch mapping plus `archFromELF` (candidate fall-through, missing binaries, nil FS) using in-memory ELF headers.
- Full `providers/os/detector` suite passes; `gofmt` and `go vet` clean.

### DIFF
--- a/providers/os/detector/arch.go
+++ b/providers/os/detector/arch.go
@@ -1,0 +1,96 @@
+// Copyright Mondoo, Inc. 2024, 2026
+// SPDX-License-Identifier: BUSL-1.1
+
+package detector
+
+import (
+	"debug/elf"
+
+	"github.com/rs/zerolog/log"
+	"github.com/spf13/afero"
+)
+
+// archBinaryCandidates is an ordered list of well-known binaries that are
+// present on virtually every Linux installation. The first one that exists and
+// parses as an ELF file is used to determine the machine architecture.
+var archBinaryCandidates = []string{
+	"/bin/ls",
+	"/usr/bin/ls",
+	"/bin/sh",
+	"/sbin/init",
+	"/usr/lib/systemd/systemd",
+	"/bin/busybox",
+}
+
+// archFromELF determines the platform architecture by inspecting the ELF header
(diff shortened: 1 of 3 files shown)


slntopp (approved):
Confirmed fixed with local container and GKE

Image

PR #8437, opened 2026-06-14
🐛 llx: guard Type.Underlying against empty type

tas50:
## The crash

A full scan can be killed by an unrecoverable panic:

```
index out of range [0] with length 0
```

The panic originates in `types.Type.Underlying()` (`types/types.go`) and is reached via `llx.isSuccess` → `arrayBlockCallResult.isSuccess`.

## Root cause

`Underlying()` indexes the first byte of the type slice with no length guard:

```go
func (typ Type) Underlying() Type {
	return Type(typ[0]) // panics when typ == ""
}
```

The guard at `llx/rawdata.go:387-390` only protects the `data == nil` case. A `RawData` with a **non-nil value but an empty `Type`** slips past that guard and reaches `Underlying()`, which then panics on `typ[0]`. Because the executor runs blocks in goroutines, this panic is unrecoverable and takes down the whole scan.

## The fix

Guard the empty case and return the empty type:

```go
func (typ Type) Underlying() Type {
	if len(typ) == 0 {
		return typ
	}
	return Type(typ[0])
}
```

Returning the empty type is the correct, safe behavior: callers such as `isSuccess`/`isTruthy` switch on `Underlying()`, and an empty value falls through to their `default` branch, which returns `(false, false)` — "undetermined". No panic, no crash.

## Test

Added `TestUnderlying_EmptyType` in `types/types_test.go`, which asserts that `Type("").Underlying()` does not panic and returns `Type("")`.

Verified locally: `gofmt -l` clean, `go test ./types/` passes, `go build ./types/... ./llx/...` succeeds.

Fixes #8432

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -216,6 +216,9 @@ func (typ Type) IsFunction() bool {
 
 // Underlying returns the basic type, e.g. types.MapLike instead of types.Map(..)
 func (typ Type) Underlying() Type {
+	if len(typ) == 0 {
+		return typ
+	}
 	return Type(typ[0])
 }
 
--- a/types/types_test.go
+++ b/types/types_test.go
@@ -69,3 +69,15 @@ func TestEqual_NilOperands(t *testing.T) {
 		assert.True(t, eq(nil, nil), "%s: nil == nil", c.typ.Label())
 	}
 }
+
+func TestUnderlying_EmptyType(t *testing.T) {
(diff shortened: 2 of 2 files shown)


arlimus (changes requested):
This one I think we need to solve differently. If we return the base type it can cause side-effects since some of the code requires child types to exist to process things. Let's rewrite this to potentially error out instead

PR #8438, opened 2026-06-14
🐛 llx: guard isTruthy type assertions against mismatched values

tas50:
## Problem

Scanning certain assets (e.g. a GitHub repo with workflows) could crash the entire scan with a panic like:

```
interface conversion: string is not int64
```

The panic originated in `isTruthy` / `IsTruthy` in `llx/rawdata.go`, reached via `arrayBlockCallResult.isTruthy`. Because the executor runs blocks in goroutines, this panic is unrecoverable and kills the whole scan.

## Root cause

`isTruthy` performed **unchecked** single-value type assertions keyed off the declared `types.Type` — e.g. `data.(int64)` for `types.Int`, `data.(string)` for `types.String`, `data.(map[string]any)` for `types.Block`, etc. When a value's declared type disagreed with its actual Go runtime type, the assertion panicked instead of returning a value.

This is the same class of bug already fixed for other llx builtins in #8320 / #8321; the established fix is comma-ok / guarded assertions.

## Fix

Convert the unchecked concrete-type assertions in `isTruthy` to comma-ok form. On a type mismatch we return `(false, false)` — the second bool is the "valid/determined" flag, and `false` means "could not determine", the safe non-panicking outcome. The happy-path semantics (including the empty-array-is-false rule) are unchanged.

Guarded cases in `isTruthy`:
- `types.Bool` (`bool`)
- `types.Int` (`int64`)
- `types.Float` (`float64`)
- `types.String` (`string`)
- `types.Regex` (`string`)
- `types.Version` (`string`)
- `types.Time` (`*time.Time`)
- `types.IP` (`RawIP`)
- `types.Block` (`map[string]any`)
- `types.ArrayLike` (`[]any`)
- `types.MapLike` (`map[string]any` / `map[int64]any`)

The `types.Any` / `types.Nil` cases already use comma-ok or are inherently safe and were left unchanged.

`isSuccess` in the same file had the identical unchecked pattern (`data.(bool)`, `data.(map[string]any)`, `data.([]any)`), so it was guarded the same way.

## Test

Added `TestTruthyTypeMismatch` to `llx/rawdata_test.go`, covering each guarded type with a deliberately mismatched value (e.g. `RawData{Type: types.Int, Value: "not-an-int"}`, `RawData{Type: types.String, Value: int64(1)}`). Each case asserts `IsTruthy()` does **not** panic and returns `(false, false)`.

Verified: `gofmt` clean, `go build ./llx/...`, and `go test ./llx/` all pass.

Fixes #8430

### DIFF
--- a/llx/rawdata.go
+++ b/llx/rawdata.go
@@ -265,22 +265,45 @@ func isTruthy(data any, typ types.Type) (bool, bool) {
 		return false, true
 
 	case types.Bool:
-		return data.(bool), true
+		b, ok := data.(bool)
+		if !ok {
+			return false, false
+		}
+		return b, true
 
 	case types.Int:
-		return data.(int64) != 0, true
+		i, ok := data.(int64)
+		if !ok {
+			return false, false
+		}
+		return i != 0, true
 
 	case types.Float:
-		return data.(float64) != 0, true
+		f, ok := data.(float64)
+		if !ok {
+			return false, false
+		}
+		return f != 0, true
 
 	case types.String:
-		return data.(string) != "", true
+		s, ok := data.(string)
+		if !ok {
+			return false, false
+		}
+		return s != "", true
 
 	case types.Regex:
-		return data.(string) != "", true
+		s, ok := data.(string)
+		if !ok {
+			return false, false
+		}
+		return s != "", true
 
 	case types.Time:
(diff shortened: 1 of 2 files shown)


arlimus (approved):
Thanks for guarding!

PR #8440, opened 2026-06-14
🐛 gcp: sanitize and length-cap snapshot scanner disk names

tas50:
## Problem

When the GCP snapshot scanner clones a boot disk or materializes a snapshot, it builds the temporary disk name by raw concatenation:

```go
"cnspec-"+target.InstanceName+"-snapshot-"+time.Now().Format("2006-01-02t15-04-05z00-00")
```

For a long instance name (e.g. `super-important-server`) this overflows GCP's 63-character limit, and the instance name itself is never sanitized — uppercase letters, underscores, or dots produce a name that violates GCP's rule `[a-z]([-a-z0-9]{0,61}[a-z0-9])?`. The scanner then fails with `RESOURCE name ... is not a valid disk name`.

## Fix

Add a small, unit-testable helper in `snapshot.go`:

- `newDiskName(instanceName string)` — thin wrapper calling `buildDiskName(instanceName, time.Now())`.
- `buildDiskName(instanceName string, t time.Time)` — lowercases the instance name, replaces any character outside `[a-z0-9-]` with `-`, reserves room for the `cnspec-` prefix and the `-<timestamp>` suffix, truncates the instance portion to fit, and trims trailing dashes.

Invariants guaranteed (and covered by tests):
- result length ≤ 63
- result fully matches `^[a-z]([-a-z0-9]{0,61}[a-z0-9])?$`
- result always begins with `cnspec-`
- empty / all-invalid instance names are handled without panic and still produce a valid name

The timestamp uses the layout `20060102t150405` (15 chars, all valid).

## Call sites

All three raw-concatenation expressions in `provider.go` (the snapshot-disk-from-recent-snapshot path, the clone-disk path, and the snapshot-target path) now call `newDiskName(target.InstanceName)`. Surrounding logic is unchanged.

## Test

New `snapshot_name_test.go` uses a fixed `time.Time` and the GCP name regexp, asserting the length and regex invariants across: short name, very long (60-char) name, name with uppercase/underscores/dots, empty name, all-invalid chars, and a name that would leave trailing dashes after truncation.

> Note: `snapshot.go` may lightly conflict with sibling PR #8416 — that's fine, resolved at merge.

Fixes #8417

### DIFF
--- a/providers/gcp/connection/gcpinstancesnapshot/snapshot.go
+++ b/providers/gcp/connection/gcpinstancesnapshot/snapshot.go
@@ -39,6 +39,61 @@ func NewSourceDiskUrl(projectID, zone, diskName string) string {
 	)
 }
 
+// newDiskName builds a GCP-compliant temporary disk name for the scanner's
+// clone/snapshot disk. See buildDiskName.
+func newDiskName(instanceName string) string {
+	return buildDiskName(instanceName, time.Now())
+}
+
+// buildDiskName produces a name matching GCP's rule
+// [a-z]([-a-z0-9]{0,61}[a-z0-9])? with a max length of 63. It lowercases and
+// replaces invalid characters in the instance name and truncates it so the
+// full "cnspec-<instance>-<timestamp>" name fits.
+func buildDiskName(instanceName string, t time.Time) string {
+	const (
+		prefix     = "cnspec-"
+		maxLen     = 63
+		timeLayout = "20060102t150405" // 15 chars, all valid characters
+	)
+	timestamp := t.Format(timeLayout)
+
+	// sanitize the instance name: lowercase, and replace any character that is
+	// not in [a-z0-9-] with '-'
+	sanitized := strings.Map(func(r rune) rune {
+		switch {
+		case r >= 'a' && r <= 'z':
+			return r
+		case r >= '0' && r <= '9':
+			return r
+		case r == '-':
+			return r
+		case r >= 'A' && r <= 'Z':
+			return r + ('a' - 'A')
+		default:
+			return '-'
+		}
+	}, instanceName)
+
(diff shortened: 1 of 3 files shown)


providers/gcp/connection/gcpinstancesnapshot/snapshot.go
```diff
@@ -39,6 +39,61 @@ func NewSourceDiskUrl(projectID, zone, diskName string) string {
...
+func newDiskName(instanceName string) string {
+	return buildDiskName(instanceName, time.Now())
+}
+
+// buildDiskName produces a name matching GCP's rule
+// [a-z]([-a-z0-9]{0,61}[a-z0-9])? with a max length of 63. It lowercases and
+// replaces invalid characters in the instance name and truncates it so the
+// full "cnspec-<instance>-<timestamp>" name fits.
+func buildDiskName(instanceName string, t time.Time) string {
+	const (
```
preslavgerchev: dont define as consts for a single fn usage

providers/gcp/connection/gcpinstancesnapshot/snapshot.go
```diff
@@ -39,6 +39,61 @@ func NewSourceDiskUrl(projectID, zone, diskName string) string {
...
+		prefix     = "cnspec-"
+		maxLen     = 63
+		timeLayout = "20060102t150405" // 15 chars, all valid characters
+	)
+	timestamp := t.Format(timeLayout)
+
+	// sanitize the instance name: lowercase, and replace any character that is
+	// not in [a-z0-9-] with '-'
+	sanitized := strings.Map(func(r rune) rune {
```
preslavgerchev: cant we juse do replace all combined with tolowercase

PR #8447, opened 2026-06-15
:star: add `mql inventory validate` to catch unknown connection options

chris-rock:
## Problem

A connection's `options` in an inventory file are a free-form `map[string]string`. A mistyped key — `namespace` instead of `namespaces`, `tenantId` instead of `tenant-id`, `subscription-id` (no such flag) instead of `subscription` — is **silently ignored**. There is no way to check an inventory before a scan, so the mistake only surfaces as a confusing connect-time failure, or as a setting that quietly never took effect.

## What this adds

A new `mql inventory validate <file>` command. It parses the inventory and checks each asset connection against the providers installed on the system:

- **Unknown connection type** — a `type` no installed provider provides is reported.
- **Unknown option key** — an `options` key not declared by the resolving provider's connector flags is reported. The flag `Long` names are the authoritative set of option keys, since providers read `conf.Options[flag.Long]`.

```
$ mql inventory validate prod.yaml
warning: unknown option "namespace" for connection type "k8s" (asset cluster, connection 0)
warning: unknown option "tenntid" for connection type "azure" (asset cluster, connection 1)
inventory "prod.yaml" passed with 2 warning(s); use --strict to fail on warnings
```

Findings are **warnings by default** and become **errors under `--strict`** (non-zero exit), so the command can gate inventories in CI without false-failing on a provider that simply isn't installed on the local machine. Validation is **offline** — it reads each provider's static plugin metadata via `providers.ListAll()` and never connects to a target.

## Design

- The check logic lives in a new `cli/inventoryvalidate` package, driven by a `Schema` built from provider metadata (`BuildSchema([]*plugin.Provider)`). Keeping the schema as an input makes `Check` unit-testable with hand-built providers — **no provider needs to be installed to run the tests**.
- A type's allowed option keys are the **union** of its provider's connector flags, which avoids false positives for providers that expose several connectors. The default warning severity is a deliberate hedge: if a provider ever reads an option it doesn't declare as a flag, the user sees a warning rather than a hard failure.

## Tests

`cli/inventoryvalidate/validate_test.go` covers schema resolution (type, connector name, alias), valid options, unknown options, `--strict` promotion, unknown types, nil-safety, and asset-label fallback. `go vet` and `golangci-lint` are clean; manually smoke-tested against good/typo'd/unknown-type inventories and a missing file.

### DIFF
--- a/apps/mql/cmd/inventory.go
+++ b/apps/mql/cmd/inventory.go
@@ -1,0 +1,91 @@
+// Copyright Mondoo, Inc. 2024, 2026
+// SPDX-License-Identifier: BUSL-1.1
+
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"go.mondoo.com/mql/v13/cli/inventoryvalidate"
+	"go.mondoo.com/mql/v13/cli/theme"
+	"go.mondoo.com/mql/v13/providers"
+	"go.mondoo.com/mql/v13/providers-sdk/v1/inventory"
+	"go.mondoo.com/mql/v13/providers-sdk/v1/plugin"
+)
+
+func init() {
+	rootCmd.AddCommand(inventoryCmd)
+	inventoryCmd.AddCommand(inventoryValidateCmd)
+	inventoryValidateCmd.Flags().Bool("strict", false,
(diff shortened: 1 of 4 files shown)


tas50:
Nice new feature @chris-rock

PR #8448, opened 2026-06-15
:star: k8s: filter discovery by label selectors

MaxRink:
## Summary
- add k8s provider flags for namespace and object label selectors
- apply namespace selectors before namespace assets are emitted and before staged namespace discovery continues
- apply object selectors to discovered workload assets, admission-review objects, and pod-backed container image discovery
- avoid emitting the cluster root asset when discovery is scoped by label selectors

## Review fixes
- admission-review discovery now honors namespace label selectors for Namespace objects
- admission-review discovery fails closed for namespaced non-Namespace objects when namespace labels are required but unavailable
- log a warning when a namespaced admission-review object is skipped because namespace labels are unavailable for selector evaluation

## Tests
- `git diff --check`
- `cd providers/k8s && go test ./resources -run 'TestAssetFromAdmissionReview|TestLabelSelectorFilters'`
- `cd providers/k8s && go test ./provider ./resources`
- `go test ./provider ./resources -count=1`
- `go test ./... -count=1`
- `git -c core.fsmonitor=false diff --check`

### DIFF
--- a/providers/k8s/config/config.go
+++ b/providers/k8s/config/config.go
@@ -69,6 +69,18 @@ Examples:
 					Desc:    "Only include Kubernetes objects in the matching namespaces",
 				},
 				{
+					Long:    "namespace-label-selector",
+					Type:    plugin.FlagType_String,
+					Default: "",
+					Desc:    "Only include Kubernetes namespaces matching the label selector, along with all objects discovered within those namespaces",
+				},
+				{
+					Long:    "object-label-selector",
+					Type:    plugin.FlagType_String,
+					Default: "",
+					Desc:    "Only include Kubernetes objects matching the label selector; for container image discovery this matches the pod that references the image",
+				},
+				{
 					Long:    "container-proxy",
 					Type:    plugin.FlagType_String,
 					Default: "",
--- a/providers/k8s/resources/admission.go
+++ b/providers/k8s/resources/admission.go
@@ -73,19 +73,19 @@ func (k *mqlK8sAdmissionreview) request() (*mqlK8sAdmissionrequest, error) {
 		args["object"] = llx.NilData
 	}
 
-	oldObj, err := resources.ResourcesFromManifest(bytes.NewReader(aRequest.OldObject.Raw))
-	if err != nil {
-		return nil, err
-	}
-
-	if len(oldObj) == 1 {
-		oldObjDict, err := convert.JsonToDict(oldObj[0])
+	args["oldObject"] = llx.NilData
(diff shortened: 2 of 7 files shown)


MaxRink:

MaxRink:
recheck

providers/k8s/config/config.go
```diff
@@ -67,6 +67,18 @@ Examples:
 					Default: "",
 					Desc:    "Only include Kubernetes objects in the matching namespaces",
 				},
+				{
+					Long:    "namespace-label-selector",
+					Type:    plugin.FlagType_String,
+					Default: "",
+					Desc:    "Only include Kubernetes namespaces matching the label selector",
```
imilchev: this is also discovering all the objects under the selected namespace. I think it's worth mentioning it explicitly in the docs.

imilchev (commented):
Overall, looks good. Added 1 small comment + your PR needs a rebase. Once these are resolved, we can merge it

tas50:
@MaxRink any chance you can rebase this one and look at the review comments?

providers/k8s/resources/admission.go
```diff
@@ -98,6 +98,23 @@ func (k *mqlK8sAdmissionreview) request() (*mqlK8sAdmissionrequest, error) {
...
+	}
+	obj, err := resources.ResourcesFromManifest(bytes.NewReader(aRequest.Object.Raw))
+	if err != nil {
+		return nil, err
+	}
+	if len(obj) == 0 {
+		return nil, errors.New("admission review request object did not contain any resources")
+	}
+	return convert.JsonToDict(obj[0])
+}
```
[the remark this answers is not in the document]
MaxRink: Good catch — removed `admissionRequestObjectDict` and its lone test. The validation it duplicated already lives inline in `assetFromAdmissionReview`, so there's no production caller to keep it around for.

providers/k8s/config/config.go
MaxRink: Updated the flag description to make this explicit: `Only include Kubernetes namespaces matching the label selector, along with all objects discovered within those namespaces`.

MaxRink:
@tas50 rebased onto latest `main` (resolved conflicts with the new image-filter feature — both `--images`/`--images-exclude` and the label selectors now coexist) and addressed the open review comments:
- removed the dead `admissionRequestObjectDict` helper + its orphaned test
- clarified the `--namespace-label-selector` flag description per @imilchev's note

`go test ./providers/k8s/...` is green.

PR #8475, opened 2026-06-18
🐛 mqlc: keep resource type through map(...).flat()

### LINKED ISSUE #8474: Nested .map(files.find(...).list.where(...)).flat() errors with 'cannot find functions for type block'
### Summary

A check of the form `list.map(files.find(from: _, type: "file").list.where(path == /.../)).flat()` **errors at scan time** with:

```
cannot find functions for type 'block' (called 'size')
cannot find functions for type 'block' (called 'path')
cannot find functions for type 'block' (called 'permissions')
```

The nested `.map(files.find(...).list.where(...)).flat()` loses the `file` type — the flattened elements are typed as anonymous `block`, so the `.where(path == ...)` (and any later `.path`/`.size`/`.permissions` access) fails.

### Where it bites

On a host where the searched tree is large and contains many files — e.g. `/usr/lib/modules` on a system with kernel headers under `<kver>/build/` (Ubuntu 24.04) — the pattern errors. On hosts where that tree is small/absent, the same query runs fine, so it presents as host-dependent.

Concrete pattern (kernel-module-availability style check):

```mql
moduleRoots = ["/usr/lib/modules", "/lib/modules"].where(file(_).exists)
moduleFiles = moduleRoots.map(files.find(from: _, type: "file").list.where(path == /\/kernel\/fs\/cramfs\//)).flat()
moduleFiles == empty || kernel.module("cramfs").disabled
```

The `moduleFiles` line is what errors. (The same `.map(files.find(...).list...).flat()` shape is also used to enumerate sysctl `*.conf` files.)

### Impact

Any policy check using this shape errors on affected hosts, which surfaces as failed/errored compliance checks even though the underlying state is fine.

### Repro

Run, on an Ubuntu 24.04 host with kernel headers installed:
`cnspec run local -c 'moduleRoots = ["/usr/lib/modules","/lib/modules"].where(file(_).exists); moduleRoots.map(files.find(from: _, type: "file").list.where(path == /\/kernel\/fs\/cramfs\//)).flat()'`

### Suggested direction

Preserve the `file` type through `.map(...).flat()` so `.where(path == …)`/`.path`/`.size`/`.permissions` resolve, or special-case flattening lists of resources.

syrull:
## Summary

Fixes #8474.

A query of the form `list.map(files.find(...).list.where(...)).flat()` errors at scan time with:

```
cannot find functions for type 'block' (called 'size')
cannot find functions for type 'block' (called 'path')
cannot find functions for type 'block' (called 'permissions')
```

The nested `.map(...).flat()` loses the `file` type — the flattened elements come out typed as an anonymous `block`, so `.where(path == …)` / `.path` / `.size` / `.permissions` can't resolve.

## Root cause

The compiled chunk types are actually correct (`map` → `[][]file`, `flat` → `[]file`). The bug is that **`postCompile` rewrites the entrypoint of the `map` body block**.

The body of `.map(...)` returns a value the map *collects* — it is not rendered on its own. But `postCompile` runs `expandResourceFields` over the entrypoints of *every* block, and when an entrypoint is a resource(-list) it appends an auto-expand `{}` display block and **replaces the entrypoint with it** (`mqlc.go: expandResourceFields`). For the map body that flips the entrypoint from `.list` (`[]file`) to `{}` (`[]block`).

At runtime `arrayMapV2` derives the mapped element type (and value) from that entrypoint, so it collects `block`-typed elements. `flat` then yields `[]block`, and the trailing display block binds each element as `block` → the field lookups fail.

## Fix

Mark `.map(...)` body blocks as value-producing (`valueBodyBlocks`, shared across the compiler tree) and skip display expansion for them in `postCompile`. The body keeps its real entrypoint, so the map collects the actual `file` resources and the type survives `.flat()`.

Display expansion for the *result* still happens normally at the outer level, so report output is unchanged.

## Test

`mqlc_test.go: TestCompiler_MapBodyKeepsResourceType` compiles the repro and asserts the map body's entrypoint stays `[]file` (not an auto-expanded `{}`/`[]block`). Verified it fails without the fix (`got []block`) and passes with it.

Manually verified end-to-end with `mql run local`:

- `["/etc/pam.d"].map(files.find(from: _, type: "file").list).flat() { path }` — now returns paths (was erroring)
- the original issue's `moduleFiles == empty || kernel.module("cramfs").disabled` shape — now resolves `path`/`size`/`permissions`
- single-level `files.find(...).list { path }`, scalar maps, explicit inner blocks, and nested `map().flat()` all still work

`go test ./mqlc/... ./llx/...` passes.

### DIFF
--- a/mqlc/builtin_array.go
+++ b/mqlc/builtin_array.go
@@ -520,9 +520,7 @@ func compileArrayMap(c *compiler, typ types.Type, ref uint64, id string, call *p
 	// display-expand a resource(-list) entrypoint into an anonymous `{}` block,
 	// which would otherwise make the mapped elements lose their resource type.
 	// See https://github.com/mondoohq/mql/issues/8474
-	if c.valueBodyBlocks != nil {
-		c.valueBodyBlocks[refs.block] = struct{}{}
-	}
+	c.valueBodyBlocks[refs.block] = struct{}{}
 
 	args := []*llx.Primitive{
 		llx.RefPrimitiveV2(ref),
--- a/mqlc/builtin_array.go
+++ b/mqlc/builtin_array.go
@@ -515,6 +515,15 @@ func compileArrayMap(c *compiler, typ types.Type, ref uint64, id string, call *p
 	}
 	mappedType := c.Result.CodeV2.DereferencedBlockType(block)
 
+	// The map body's entrypoint is the value this map collects, not something
+	// that gets rendered on its own. Mark the block so postCompile won't
+	// display-expand a resource(-list) entrypoint into an anonymous `{}` block,
+	// which would otherwise make the mapped elements lose their resource type.
+	// See https://github.com/mondoohq/mql/issues/8474
+	if c.valueBodyBlocks != nil {
+		c.valueBodyBlocks[refs.block] = struct{}{}
+	}
+
 	args := []*llx.Primitive{
 		llx.RefPrimitiveV2(ref),
 		argExpectation,
--- a/mqlc/builtin_resource.go
+++ b/mqlc/builtin_resource.go
@@ -281,6 +281,13 @@ func compileResourceMap(c *compiler, typ types.Type, ref uint64, id string, call
 		return types.Nil, multierr.Wrap(err, "called '"+id+"' with a bad function block, types don't match")
 	}
 
+	// The map body's entrypoint is the value this map collects, not something
+	// that gets rendered on its own. Mark the block so postCompile won't
+	// display-expand a resource(-list) entrypoint into an anonymous `{}` block,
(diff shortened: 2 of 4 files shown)


syrull:
Thanks for the review — both addressed in 657963011:

- 🟡 `compileResourceMap` had the same bug; it now marks its body block in `valueBodyBlocks` too. Confirmed `users.map(file("/etc/hosts")).flat() { path }` errored with `cannot find functions for type 'block'` before and resolves now.
- 🔵 Dropped the `valueBodyBlocks != nil` guard in both call sites (root compiler always initializes the map; `newBlockCompiler` propagates it).

Regression test is now table-driven and covers both the array and resource-list map paths.

PR #8476, opened 2026-06-18
🐛 os: make native files.find match `find -L` (symlinks + perm 0)

### LINKED ISSUE #8467: pam.conf / pam.module misses authselect-managed /etc/pam.d service files (symlinks) on RHEL
### Summary

On RHEL systems managed by **authselect**, `pam.module("<name>")` (and `pam.conf`) can report a module as **not present / `enabled = false`** even though it is configured, because the key `/etc/pam.d` service files are **symlinks** that the PAM file discovery appears to skip.

authselect manages `/etc/pam.d/system-auth`, `password-auth`, `postlogin`, `smartcard-auth`, and `fingerprint-auth` as symlinks into `/etc/authselect/`. These files hold the core stack (`pam_unix`, `pam_pwquality`, `pam_faillock`, …). If discovery excludes symlinks, those modules are never parsed.

### Evidence

On a hardened RHEL 10 image with authselect active (`/etc/authselect/authselect.conf` = `custom/custom-profile … without-nullok`) and `pam_unix` present in `system-auth`/`password-auth` with **no `nullok`** anywhere:

- The config is compliant (authselect `without-nullok`, and no `pam_unix … nullok` line exists).
- Yet a check of the form `pam.module("pam_unix").enabled && pam.module("pam_unix").entries.none(options.contains("nullok"))` evaluates to **fail** — consistent with `pam.module("pam_unix")` resolving to the empty husk (`enabled = false`) rather than the real, compliant module.

### Suspected root cause

`providers/os/resources/pam.go` `files()` → `getSortedPathFiles()` (`utils.go`) → `files.find(from: "/etc/pam.d", type: "file")`. A `type: "file"` enumeration (`find -type f`) excludes symlinks, so the authselect-managed symlinked service files under `/etc/pam.d` are skipped, and modules defined only there (e.g. `pam_unix`) are missed. `pam.module()` then returns the empty husk with `enabled = false`.

### Suggested fix

Have the PAM file discovery resolve/include symlinks under `/etc/pam.d` (authselect relies on them) — e.g. follow symlinks when enumerating, or include symlinked regular files.

### Note

I could not fully confirm the mechanism from a recording (the directory/permission datapoints weren't captured) and couldn't reproduce it in this environment. A live RHEL + authselect target is needed to confirm and to verify a fix (a unit test with a symlinked `system-auth` fixture would also reproduce it).

syrull:
## Summary

Fixes #8467.

On RHEL systems managed by **authselect**, `pam.module("pam_unix")` (and `pam.conf`) report the module as not present / `enabled = false` even though it is configured, because the key `/etc/pam.d` service files (`system-auth`, `password-auth`, …) are **symlinks** into `/etc/authselect` and PAM file discovery skipped them.

While verifying the fix on a filesystem scan I found the native `files.find` walker (used by the **filesystem** and **tar** connections) diverges from the command-based `find -L` backend in *two* ways, which together break PAM discovery on those connections. Both are needed for the authselect case to work.

## Root cause

`pam.conf` discovers service files via `getSortedPathFiles` → `files.find(from: "/etc/pam.d", type: "file")`. On filesystem/tar connections that runs the native Go walker (`fsutil.FindFiles`), which differs from `find -L` (what the local/SSH command backend runs):

1. **Symlinks are skipped.** `fs.WalkDir` reports lstat info, so a symlink keeps `fs.ModeSymlink` and never satisfies `type:"file"` (`IsRegular()` is false). The authselect-managed service files are symlinks, so `pam_unix` (defined only there) is never parsed and `pam.module("pam_unix")` resolves to the empty husk.

2. **A `0` permission mask matches nothing.** `files.find` built via `CreateResource` (as pam, sudoers, modprobe, rsyslog, logrotate, limits, apt, … all do) never runs `initFilesFind`, so `permissions` stays at the Go zero `0`. The walker then filtered out *every* file (`mode & 0 == 0`). The command backend treats the same value as `-perm -0`, which matches everything — hence local/SSH worked but filesystem/tar returned nothing.

## Fix (`providers/os/fsutil/find_files.go`)

Make the native walker mirror `find -L`:

- Resolve a symlink to its target type for type matching, so a symlink to a regular file matches `type:"file"` (and a symlink to a dir matches `type:"directory"`). This is **additive** — a symlink still matches `type:"link"`, and a broken symlink keeps its symlink type. A small guard skips the extra `stat` for plain `type:"link"` lookups.
- Treat a `0` permission mask as "no filter" (like `find -perm -0`).

No call sites change, so this fixes the `*.d` discovery for pam **and** sudoers/modprobe/rsyslog/… on filesystem/tar connections in one place.

## Tests

`providers/os/fsutil/find_files_unix_test.go`:
- `TestFindFilesFollowsSymlinks` — a symlink to a file matches `type:"file"`, a symlink to a dir matches `type:"directory"`, all symlinks (incl. broken) still match `type:"link"`. Verified it fails without the fix.
- `TestFindFilesZeroPermissionMatchesAll` — a `0` mask matches all files instead of none.

End-to-end against an authselect-style rootfs (`system-auth`/`password-auth` as symlinks into `/etc/authselect`, `pam_unix` without `nullok`) via `mql run filesystem --path …`:

- before: `pam.conf.files` empty → `pam.module("pam_unix").enabled` = `false`
- after: `pam.conf.files` lists the symlinked services → the issue's exact check `pam.module("pam_unix").enabled && pam.module("pam_unix").entries.none(options.contains("nullok"))` returns `true`

Confirmed each fix is independently required: with only the perm fix, `pam.conf.files` still drops the symlinks and `pam_unix` stays a husk.

`go test ./providers/os/fsutil/... ./providers/os/resources/...` passes.

### DIFF
--- a/providers/os/fsutil/find_files.go
+++ b/providers/os/fsutil/find_files.go
@@ -13,6 +13,12 @@ import (
 func FindFiles(iofs fs.FS, from string, r *regexp.Regexp, typ string, perm *uint32, depth *int) ([]string, error) {
 	matcher := createFindFilesMatcher(iofs, typ, from, r, perm, depth)
 	matchedPaths := []string{}
+	// Note: matcher.Match resolves a symlink to its target type so a symlink is
+	// reported under the type it points at (mirroring `find -L`), but WalkDir
+	// itself does not follow symlinks during traversal. So a symlinked directory
+	// is reported for type:"directory" yet its contents are not walked. Full
+	// `find -L` recursion would need explicit symlink following with loop
+	// detection; it isn't required for the config-file discovery this powers.
 	err := fs.WalkDir(iofs, from, func(p string, d fs.DirEntry, err error) error {
 		if d != nil && d.IsDir() && matcher.DepthReached(p) {
 			return fs.SkipDir
--- a/providers/os/fsutil/find_files.go
+++ b/providers/os/fsutil/find_files.go
@@ -62,7 +62,7 @@ func (m findFilesMatcher) DepthReached(p string) bool {
 }
 
 func (m findFilesMatcher) Match(path string, t fs.FileMode) bool {
-	matchesType := m.matchesType(t)
+	matchesType := m.matchesType(path, t)
 	matchesRegex := m.matchesRegex(path)
 	matchesPerm := m.matchesPerm(path)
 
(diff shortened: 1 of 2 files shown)


syrull:
Good catch — added a comment in 7b9b61678 noting that the walker resolves a symlink's target type for matching (like `find -L`) but does not descend into symlinked directories during traversal, and that full `-L` recursion would need explicit symlink following with loop detection (not needed for this config-file discovery).

PR #8481, opened 2026-06-18
🐛 gcp: fix nil-deref and panic crashes in resource collection

tas50:
## What

Fixes several GCP resource accessors that could panic and crash the **entire** scan. The executor runs blocks in goroutines, so an unhandled panic is unrecoverable and takes down the whole run — not just the one field.

All issues were found by a static audit of the provider and verified against the source.

### Fixes
- **cloudrun `operations()`** — on a non-`Done` iterator error the loop logged but did not `break`, then dereferenced a nil `t` (`t.Name`) → panic. (`services()`/`jobs()` already handle this correctly.)
- **cloudrun service/job revision templates** — dereferenced the nullable `Timeout.Seconds` directly; a template without an explicit timeout (common) panicked. Now `GetTimeout().GetSeconds()`.
- **cloudrun container env** — dereferenced `valueSource.SecretKeyRef` which is a nullable oneof; guarded with `GetSecretKeyRef()`.
- **recommendations** — indexed `values[7]/[5]/[3]/[1]` after splitting the resource name with no bounds check, inside a bare goroutine. Now guards the length.
- **cloudTasks queue / cloudScheduler job `retryConfig`** — a queue/job without a retry config stored a typed-nil `*mqlGcpRetryConfig` into a non-nullable field; querying it dereferenced a nil receiver. Now omits the arg and marks the field null (matches the pubsub `schemaSettings` pattern).
- **asset service init** — wrote to a possibly-nil `args` map; now guarded like the other init functions.
- **logging `alertPolicies`** — `parseAlertPolicyConditionFilterMetricName` did an unchecked `condition["filter"].(string)`, but `monitoringQueryLanguage` conditions have no `filter` key → panic. Now comma-ok.

## Test
`go build ./resources/...` passes. Behavioral changes are defensive guards on existing code paths.

### DIFF
--- a/providers/gcp/resources/asset.go
+++ b/providers/gcp/resources/asset.go
@@ -61,6 +61,9 @@ func initGcpProjectAssetService(runtime *plugin.Runtime, args map[string]*llx.Ra
 		return nil, nil, errors.New("invalid connection provided, it is not a GCP connection")
 	}
 
+	if args == nil {
+		args = make(map[string]*llx.RawData)
+	}
 	args["projectId"] = llx.StringData(conn.ResourceID())
 	return args, nil, nil
 }
--- a/providers/gcp/resources/cloudrun.go
+++ b/providers/gcp/resources/cloudrun.go
@@ -311,6 +311,7 @@ func (g *mqlGcpProjectCloudRunService) operations() ([]any, error) {
 				}
 				if err != nil {
 					log.Error().Err(err).Send()
+					break
 				}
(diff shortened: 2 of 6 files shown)


tas50:
Addressed review feedback in 2ec9d8470: cloudrun `operations()` now `continue`s on a CreateResource error instead of appending a nil entry, and the `len(values) < 8` recommendation bound has an explanatory comment.

PR #8483, opened 2026-06-18
🐛 gcp: request IAM policy schema v3 so conditional bindings are returned

tas50:
## What

Several IAM-policy reads used the default policy schema **v1**, which silently strips any binding carrying an IAM condition. Where the code then inspects bindings for public members or reads the condition fields, conditional grants (including conditional `allUsers`/`allAuthenticatedUsers`) were invisible — so e.g. `secret.public()` / `bucket.public()` could falsely report **not-public**.

### Fixes
- **secretmanager secret `iamPolicy()`** — request `RequestedPolicyVersion: 3`, and tolerate `PermissionDenied` on a single secret (return empty) so one inaccessible secret doesn't fail the whole collection, mirroring the KMS accessors.
- **storage bucket `iamPolicy()`** — add `OptionsRequestedPolicyVersion(3)`.
- **dataproc cluster `iamPolicy()`** — set `GetPolicyOptions.RequestedPolicyVersion: 3`.

The KMS keyring/cryptokey accessors already request v3; this brings the remaining IAM-policy readers in line.

## Test
`go build ./resources/...` passes.

---
**Note (review follow-up):** the secretmanager change also adds `PermissionDenied` tolerance (returns an empty binding list for a single inaccessible secret) so one inaccessible secret doesn't fail the whole collection, matching the KMS accessors. This is a small additional behavior change beyond the v3 schema fix. This PR now also extends `RequestedPolicyVersion: 3` to **all** remaining `GetIamPolicy` call sites in the provider (compute, cloudrun, dns, pubsub, spanner, cloud functions, artifact registry, cloud tasks, iap, folder/org/project), per review feedback.

### DIFF
--- a/providers/gcp/resources/artifactregistry.go
+++ b/providers/gcp/resources/artifactregistry.go
@@ -260,7 +260,7 @@ func (g *mqlGcpProjectArtifactRegistryServiceRepository) iamPolicy() ([]any, err
 	}
 	defer client.Close()
 
-	policy, err := client.GetIamPolicy(ctx, &iampb.GetIamPolicyRequest{Resource: repoPath})
+	policy, err := client.GetIamPolicy(ctx, &iampb.GetIamPolicyRequest{Resource: repoPath, Options: &iampb.GetPolicyOptions{RequestedPolicyVersion: 3}})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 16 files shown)


tas50:
Addressed review feedback: extended `RequestedPolicyVersion: 3` to all remaining GetIamPolicy call sites in the provider (gRPC + compute/cloudrun/dns/cloudresourcemanager REST), and called out the secretmanager PermissionDenied tolerance in the PR description.

PR #8486, opened 2026-06-18
🐛 azure: fix nil-deref and panic crashes in resource collection

tas50:
## What

Fixes accessors that dereferenced nullable Azure SDK pointers unconditionally. Because the executor runs blocks in goroutines, an unhandled panic is unrecoverable and crashes the **entire** scan. Found by a static audit and verified against the SDK pointer shapes.

### Fixes
- **network `virtualNetworkGateways()`** — `Properties` and the nested `SKU` are nullable (transient/failed gateways); normalized to empty structs.
- **network `connections()`** — guard `c.Properties` and the gateway-ref `.ID` before dereferencing.
- **sql `vulnerabilityAssessmentSettings()`** — `RecurringScans` (and `Properties`) are nullable; a VA config without recurring scans panicked.
- **sql `databases()`** — guard nullable `entry.Properties`.
- **storage queue/table/blob service properties** — `Logging`/`Metrics` and their `RetentionPolicy` are nullable; added `normalizeServiceProperties()`.
- **storage `dataProtection()`** — guard the nullable inner `BlobServiceProperties`.
- **monitor `logProfiles()` / `alerts()`** — guard nullable `entry.Properties` and nested `Actions`/`Condition`.
- **automation `accounts()`** — read SKU only after guarding `acct.Properties`.
- **iam `roleAssignment` / `managedIdentity`** — guard nullable `Properties` (and the raw `*TenantID` deref, now `StringDataPtr`).
- **redis `instances()`** — guard nullable `cache.Properties`.
- **connection `GetSubscriptions()`** — the `NewClient` error was checked only after the returned (nil) client was already used to build the pager; moved the check before first use.

## Test
`go build ./...` passes.

### DIFF
--- a/providers/azure/connection/client_subscriptions.go
+++ b/providers/azure/connection/client_subscriptions.go
@@ -33,13 +33,13 @@ func (client *subscriptionsClient) GetSubscriptions(filter SubscriptionsFilter) 
 	subscriptionsC, err := subscriptions.NewClient(client.token, &arm.ClientOptions{
 		ClientOptions: client.clientOptions,
 	})
+	if err != nil {
+		return nil, err
+	}
 
 	ctx := context.Background()
 	subs := []subscriptions.Subscription{}
 	res := subscriptionsC.NewListPager(&subscriptions.ClientListOptions{})
-	if err != nil {
-		return nil, err
-	}
 	for res.More() {
 		page, err := res.NextPage(ctx)
 		if err != nil {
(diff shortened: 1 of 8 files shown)


tas50:
Addressed the change request: `createRedisInstanceRawData` now normalizes a nil `cache.Properties` to an empty struct at the top, so the panic can no longer occur before the call-site guard (both list and init paths are covered). Also dropped the stray blank line in `GetSubscriptions`.

PR #8488, opened 2026-06-18
🐛 azure: fix security-posture detections returning wrong values

tas50:
## What

Three security signals could report the insecure/wrong value:

- **`defenderForServers()`** — the Qualys VA policy check did not scope the assignment to the subscription, so a Qualys policy inherited from a parent management group made `enabled` report true even when Defender-for-Servers VA was not configured here. Added the same `Scope == /subscriptions/{subId}` filter the Containers detection uses.
- **`defenderForContainers()`** — `arcClusterPolicyExtensionDefinitionId` and `kubernetesClusterPolicyExtensionDefinitionId` were **both** set to the Arc GUID (`0adc5395-…`), so `azurePolicyForKubernetes` never reflected the AKS Azure Policy Add-on. Pointed the AKS constant at its real definition (`a8eff44f-…`, "Azure Policy Add-on for Kubernetes service (AKS)…"). *Reviewers: please sanity-check this GUID.*
- **sql `azureAdOnlyAuthentication()`** — the accessor swallowed **every** error as `false` (Azure-AD-only-auth disabled — the insecure value). Now only tolerates access-denied (403) and surfaces other errors.

## Test
`go build ./...` passes.

### DIFF
--- a/providers/azure/resources/cloud_defender.go
+++ b/providers/azure/resources/cloud_defender.go
@@ -31,8 +31,12 @@ const (
 	arcClusterDefenderExtensionDefinitionId        string = "/providers/Microsoft.Authorization/policyDefinitions/708b60a6-d253-4fe0-9114-4be4c00f012c"
 	kubernetesClusterDefenderExtensionDefinitionId string = "/providers/Microsoft.Authorization/policyDefinitions/64def556-fbad-4622-930e-72d1d5589bf5"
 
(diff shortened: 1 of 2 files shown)


tas50:
Addressed: `azureAdOnlyAuthentication` now also tolerates 404 (some server SKUs return 404 instead of 403), while still surfacing other errors.

PR #8489, opened 2026-06-18
🐛 azure: fix swallowed errors and a discovery-target loop bug

tas50:
## What

- **network `virtualNetworkGateways` / `…ipConfigurations()`** — two error paths returned the (already-nil) `err` variable instead of the sub-resource's `rgs.Error` / `gateways.Error`, silently swallowing a real failure as success.
- **postgresql `databases()`** (single-server and flexible) — on a `NewDatabasesClient` failure both returned `&json.MarshalerError{}`, an empty struct literal that discards the real error and whose `Error()` can itself nil-deref. Now returns the actual error.
- **advisor `averageScore()`** — divided by `len(scores)` with no zero guard (NaN on a subscription with no advisor scores) and dereferenced `CurrentScore.Data` without a nil check. Guards both.
- **discovery `getDiscoveryTargets()`** — removed the "auto" keyword with `slices.Delete` inside a `range` over the same slice, which skips the element after a deletion. Now uses `slices.DeleteFunc`.

## Test
`go build ./...` passes.

### DIFF
--- a/providers/azure/resources/advisor.go
+++ b/providers/azure/resources/advisor.go
@@ -180,9 +180,15 @@ func (a *mqlAzureSubscriptionAdvisorService) averageScore() (float64, error) {
 	if scores.Error != nil {
 		return 0, scores.Error
 	}
+	if len(scores.Data) == 0 {
+		return 0, nil
+	}
 	avg := float64(0)
 	for _, s := range scores.Data {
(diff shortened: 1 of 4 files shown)


tas50:
Good catch — fixed: averageScore now divides by the count of scores actually summed (skipping nil currentScore), not the full list length.

PR #8495, opened 2026-06-18
🐛 vsphere: populate vulnmgmt.packages and fix disk-group id collision

tas50:
## What

vsphere is otherwise very clean. Three real fixes:

- **vulnmgmt `populateData()`** — the success path built advisories/cves/stats but **never** read `vulnReport.Packages` or set `v.Packages`, so `vulnmgmt.packages` silently resolved empty even when the report listed affected packages (only the unavailable/error path set it, to an empty slice). Now builds `vuln.package` resources like the os provider does.
- **vsan `diskGroups()`** — the `__id` used the cache SSD UUID, which is empty on all-flash/ESA topologies with no dedicated cache disk, collapsing all such groups onto one cache key (silent data loss). Fall back to the mapping index.
- **vulnmgmt `lastAssessment()`** — drop a dead `else` branch (the empty-`lastUpdate` case already returned earlier).

## Test
`go build ./...` passes.

### DIFF
--- a/providers/vsphere/resources/vulnmgmt.go
+++ b/providers/vsphere/resources/vulnmgmt.go
@@ -54,18 +54,13 @@ func (v *mqlVulnmgmt) lastAssessment() (*time.Time, error) {
 		return v.setUnavailableLastAssessment(errors.New("no update time available"))
 	}
 
-	var lastUpdateTime *time.Time
-	if lastUpdate != "" {
-		parsedLastUpdateTime, err := time.Parse(time.RFC3339, lastUpdate)
-		if err != nil {
-			return nil, errors.New("could not parse last update time: " + lastUpdate)
-		}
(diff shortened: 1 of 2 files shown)


providers/vsphere/resources/vulnmgmt.go
```diff
@@ -178,6 +173,20 @@ func (v *mqlVulnmgmt) populateData() error {
 		mqlVulnCves[i] = mqlVulnCve
 	}
 
+	mqlVulnPackages := make([]any, len(vulnReport.Packages))
+	for i, p := range vulnReport.Packages {
```
czunker, line 177: Should be fine to add it here, but we do not have packages here.
VMware is only on the platform level.

czunker (approved):

PR #8497, opened 2026-06-18
🐛 proxmox: fix cluster storage enabled=false and vm.updates agent failure

tas50:
## What

- **`proxmox.storages` enabled always false** — the cluster `/storage` endpoint reports config (a `disable` key), not the runtime `enabled` key that `/nodes/<n>/storage` returns. `StorageInfo.Enabled` was therefore always `0`, so `proxmox.storages` (and every storage resolved via `initProxmoxStorage` typed refs) reported `enabled=false` for all storages — and `enabled` is a `@defaults` field. Added a `Disable` field and normalize `Enabled` from it in `GetStorages`; the node path keeps its genuine `enabled` value, so `proxmox.node.storages` is unchanged.
- **`vm.updates()` fails on agent-less VMs** — a running VM without the QEMU guest agent returned a hard error, failing the field for the whole list during `vms { updates }`. Now returns an empty list, consistent with the Windows-update path.

## Test
`go build ./...` passes.

### DIFF
--- a/providers/proxmox/connection/storage.go
+++ b/providers/proxmox/connection/storage.go
@@ -15,10 +15,8 @@ type StorageInfo struct {
 	Content string `json:"content"`
 	Path    string `json:"path"`
 	Enabled int    `json:"enabled"`
-	// Disable is the cluster /storage config key (1 = disabled). The
-	// cluster endpoint reports config (disable), while /nodes/<n>/storage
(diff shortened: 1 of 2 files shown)


tas50:
Addressed review: GetStorages now only infers `Enabled` from `Disable` when no explicit `enabled` value was returned (forward-safe), `vm.updates()` logs a warning for agent-less VMs before returning empty, and tidied the struct field alignment.

PR #8505, opened 2026-06-18
🐛 cloudformation: guard section bodies against non-mapping YAML (panic fix)

tas50:
## What

`resources()` / `outputs()` / `parameterList()` / `extractDict()` / `types()` iterate a template section body two-at-a-time (`body.Content[i+1]`) but only validated the **parent** document node, not the section body returned by `gatherMapValue`. A section written as a YAML **sequence** (or with odd content) — e.g.

```yaml
Resources:
  - a
  - b
  - c
```

makes the body a sequence node, so the stride-2 loop indexes out of range and **panics the whole scan** on one malformed template. `types()` additionally lacked the degenerate-template guard its siblings already had (and the upstream `GetTypes` dereferences `Content[0]` / iterates Resources assuming an even mapping).

## Fix

- Added `isEvenMappingNode` and guard every stride-2 section loop with it (returns empty/nil for a non-mapping body instead of panicking).
- `types()` now skips `GetTypes` when the template/Resources body is degenerate or malformed.

## Test

`TestIsEvenMappingNode_SectionBodyGuard` reproduces the sequence-bodied `Resources:` input via `parse.String` and asserts the guard rejects it, and accepts a well-formed mapping body. `go test ./resources/` and `go build ./...` pass.

### DIFF
--- a/providers/cloudformation/resources/template.go
+++ b/providers/cloudformation/resources/template.go
@@ -435,8 +435,17 @@ func (r *mqlCloudformationTemplate) types() ([]any, error) {
 	}
 	// GetTypes iterates the Resources section assuming an even mapping and
 	// dereferences Content[0]; skip when the template/Resources body is
-	// degenerate or malformed so the upstream stride-2 access can't panic.
-	if _, body, err := gatherMapValue(template.Node.Content[0], string(cft.Resources)); err != nil || !isEvenMappingNode(body) {
+	// degenerate or malformed so the upstream stride-2 access can't panic. A
+	// missing Resources section is a valid empty state, but any other
+	// gatherMapValue error is a real parse failure and must surface rather than
+	// be swallowed as "no types."
+	_, body, err := gatherMapValue(template.Node.Content[0], string(cft.Resources))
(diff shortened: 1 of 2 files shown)


tas50:
Thanks — addressed the `types()` error-swallowing (line 442) in d0c3814cc: it now treats a missing Resources section (NotFound) as the empty state, propagates any other `gatherMapValue` error, then applies the even-mapping guard — matching the pattern the other section readers already use.

On the other two:
- **`isEvenMappingNode` parity check (line 21):** keeping it. The whole purpose of this PR is panic-safety — the stride-2 loops dereference `Content[i+1]`, so an odd-content mapping (however unlikely from yaml.v3) would index out of range and crash the entire scan. The kind-only check would reintroduce that risk.
- **`extractDict` returning `nil, nil` on a non-mapping section (line 86):** intentional, and consistent with how a NotFound section is handled two lines up. A single malformed section shouldn't fail the whole template scan; it surfaces as an empty/null field rather than aborting every other section. Happy to revisit if you'd prefer a hard error here.

PR #8508, opened 2026-06-18
🐛 atlassian: paginate admin users/policies/domains and guard empty org

tas50:
## What

The admin organization accessors read only the **first page**:

- `managedUsers()`, `policies()`, and `domains()` each called the cursor-paginated SDK method once with an empty cursor and **never followed `Links.Next`**, silently dropping every record past the first page (~100 items). For an access review this under-reports managed users and org security policies.
- `initAtlassianAdminOrganization` checked `len(Data) > 1` but then indexed `Data[0]`, **panicking** the connection init when zero organizations are returned (revoked / mis-scoped API token).

## Fix

- Loop each accessor until `Links.Next` is empty, extracting the **raw cursor value** from the Next URL via `extractAtlassianCursor` (the SDK re-encodes it as `?cursor=`, so the full URL must not be passed back).
- Guard the empty-organization case before indexing.

## Test

`TestExtractAtlassianCursor` covers raw/encoded/empty/no-cursor URLs. `go build ./...` and `go test ./resources/` pass.

### DIFF
--- a/providers/atlassian/resources/atlassian_admin.go
+++ b/providers/atlassian/resources/atlassian_admin.go
@@ -45,6 +45,13 @@ func parseAtlassianTime(s string) *time.Time {
 	return &t
 }
 
+// productAccess is one entry of a managed user's product-access list,
+// flattened to the name and the (nullable) last-active timestamp.
+type productAccess struct {
+	Name       string
+	LastActive *time.Time
+}
+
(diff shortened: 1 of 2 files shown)


tas50:
Addressed in 386373594: hoisted the per-user `productAccess` struct out of the loop to a package-level type, and lowercased the `unexpectedly received more than 1 organization` error to match Go convention and the adjacent message.

PR #8510, opened 2026-06-19
🐛 arista: build STP interface command from arguments

tas50:
## Summary

`Eos.StpInterfaceDetails(mstInstanceID, iface)` accepts an MST instance ID and an interface name, but the request type's `GetCmd()` returned a **hardcoded** command:

```go
func (s *showSpanningTreeMstInstanceDetail) GetCmd() string {
	return "show spanning-tree mst 0 interface Ethernet1 detail"
}
```

So no matter which instance/interface a caller asked for, EOS was always queried for `mst 0 interface Ethernet1`, and the returned details described that one interface. Any query iterating STP interface details silently got the wrong data.

## Fix

- Add a `cmd` field to `showSpanningTreeMstInstanceDetail`; `GetCmd()` returns it.
- Extract a pure helper `stpInterfaceDetailCmd(mstInstanceID, iface)` that builds the command via `fmt.Sprintf`.
- `StpInterfaceDetails` sets `cmd: stpInterfaceDetailCmd(mstInstanceID, iface)` so the actual arguments drive the command.

## Test

`stp_test.go` (`TestStpInterfaceDetailCmd`) asserts the helper builds the command from arbitrary instance/interface arguments and that `GetCmd()` returns the command set on the request.

### DIFF
--- a/providers/arista/resources/eos/arista.go
+++ b/providers/arista/resources/eos/arista.go
@@ -4,6 +4,7 @@
 package eos
 
 import (
+	"fmt"
 	"regexp"
 
 	"github.com/aristanetworks/goeapi"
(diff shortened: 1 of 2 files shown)


tas50:
Addressed in 9045bd26a: both `mstInstanceID` and `iface` are now validated against an allowlist (digits for the instance id; alphanumerics with `/`, `-`, `.` for the interface name — EOS interface names never contain spaces) and `stpInterfaceDetailCmd` returns an error on a malformed identifier instead of building a corrupt command. Added test cases covering injected/space/empty inputs and the real EOS interface-name shapes (`Ethernet1/1`, `Port-Channel3`, `Ethernet1.100`, …).

PR #8513, opened 2026-06-19
⭐ aws: add cross-resource edges for security policy authoring

tas50:
## Problem

AWS security checks today operate on **one resource type in isolation**. Policy authors can ask *"does a security group allow `0.0.0.0/0`?"* but not the question that actually matters and avoids false positives: *"is there an **instance** genuinely exposed?"* — because the reverse edges to traverse it (SG → instances, subnet → instances) don't exist in the schema. The result is high-false-positive checks: a security group flagged for `0.0.0.0/0` even when it's attached to nothing.

## What this adds

Backreference and convenience edges so authors can traverse between resources without manual joins:

| Resource | New field | Purpose |
|---|---|---|
| `aws.ec2.securitygroup` | `networkInterfaces()`, `instances()` | *"is this open SG attached to anything live?"* |
| `aws.vpc.subnet` | `networkInterfaces()`, `instances()` | *"compute sitting in a public subnet"* |
| `aws.ec2.instance` | `subnet()` | convenience forward edge |
| `aws.iam.role` | `usedByInstances()` | *"what compute can assume this role"* |
| `aws.kms.key` | `encryptedVolumes()` | *"what does this key protect"* (EBS-scoped start) |

Now expressible: `aws.ec2.securityGroups.where(instances.length > 0)` — only flag SGs actually in use.

## Implementation

- **Backrefs resolve via server-side `DescribeNetworkInterfaces` filters** (`group-id` / `subnet-id`) where AWS supports them — no client-side scans. ENIs are the authoritative attachment point, so SG/subnet → instances projects through `networkinterface.instance` (correctly distinguishing service ENIs like endpoints/NAT, which have no instance).
- **Role / KMS edges** scan the cached cross-region instance / volume lists (AWS has no "find by role/key" API).
- Adds one IAM permission: `iam:ListInstanceProfilesForRole`.

## Verification

All seven edges verified against a live AWS account:
- `instance.subnet` → resolves subnet + `mapPublicIpOnLaunch`
- `securitygroup.networkInterfaces` / `.instances` → ENI + attached instance
- `subnet.networkInterfaces` / `.instances` → correctly returns service ENIs with 0 instances
- `iam.role.usedByInstances` → role `EC2-SSM` → its instance
- `kms.key.encryptedVolumes` → key → its encrypted volumes

Design doc: `docs/superpowers/specs/2026-06-18-aws-cross-resource-edges-design.md`


tas50:
Addressed the review in cb7c97686:

- **`networkInterfacesByFilter` access-denied** (🟡): now checks `Is400AccessDeniedError` and returns the partial list with a warning, matching every other `DescribeNetworkInterfaces` call site — restricted ENI permissions no longer hard-fail the SG/subnet `networkInterfaces()`/`instances()` fields.
- **`instancesFromNetworkInterfaces` graceful skip** (🟡): an ENI whose instance can't be resolved (e.g. terminated) is now logged and skipped instead of failing the whole backref scan. Confirmed `initAwsEc2Instance` returns `"ec2 instance does not exist"` in that case, so the hard-fail was real.
- **`instance.subnet` semantics** (🔵): added a doc-comment noting it's the primary network interface's subnet (per `ec2types.Instance.SubnetId`).
- **`kms.key.encryptedVolumes` enumeration note** (🔵): added a doc note on the `.lr` field that first access may enumerate all EBS volumes across regions (cached after).

Also removed the design spec doc from the PR.

PR #8516, opened 2026-06-19
⭐ aws: CloudFormation provenance, instance→LB backref, faster role usage

tas50:
## What

Three follow-ups to the cross-resource edges (#8513), all in the AWS provider.

### 1. Provenance — `cloudformationStack()`
Adds `cloudformationStack() aws.cloudformation.stack` to `aws.ec2.instance`, `aws.ec2.securitygroup`, `aws.vpc`, and `aws.s3.bucket`. Resolved from the AWS-injected `aws:cloudformation:stack-name` tag via a shared helper that scans the cached cross-region stack list. Enables **"unmanaged / drifted resource"** checks (resource has no managing stack → not IaC-managed).

### 2. Reverse edge — `aws.ec2.instance.loadBalancers()`
The load balancers that route traffic to an instance (the forward `loadbalancer.instances()` already existed). Lets a policy express *"this instance is internet-reachable via an internet-facing ALB/NLB even though it has no public IP"* — the exposure case the public-IP checks miss.

### 3. Perf — `aws.iam.role.usedByInstances`
Reworked to derive the role's instance profiles from the account-wide `ListInstanceProfiles` list (one call, **cached and shared across all role evaluations**) instead of a per-role `ListInstanceProfilesForRole` call. The `aws.iam.roles.where(usedByInstances.length > 0)` pattern now makes a single IAM list call total instead of one per role. Drops the `iam:ListInstanceProfilesForRole` permission (permissions regenerated: 913 → 912).

## Notes
- The `aws:cloudformation:stack-name` correlation is a runtime tag relationship with no static-IaC analog, consistent with the other cross-resource edges.

## Verification
Built and run against a live account:
- All six new fields resolve without error and return correct negatives for resources not managed by CloudFormation / not behind a load balancer.
- `usedByInstances` returns the same correct result (`EC2-SSM`) via the faster path.
- `go build ./...` clean; codegen idempotent; permissions regenerated.

## Follow-up
The `#6 hardening` item also called for dedicated unit tests on these edges. They aren't included here: the resolvers call into the MQL runtime (`GetInstances`, `GetLoadBalancers`, `GetStacks`), and the provider's existing tests cover only pure helpers (e.g. `TestShouldExcludeInstance`) — there's no mock-runtime harness for resolver-level tests. Verified interactively instead, per the repo's stated norm. Happy to add a mock-runtime test scaffold as a separate change if wanted.

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -371,6 +371,8 @@ aws.vpc @defaults("id isDefault cidrBlock region") {
   cidrBlockAssociations() []dict
   // IPv6 CIDR block associations for the VPC
   ipv6CidrBlockAssociations() []dict
+  // CloudFormation stack that manages this VPC, if any
+  cloudformationStack() aws.cloudformation.stack
 }
 
 // Amazon Virtual Private Cloud (VPC) route table
@@ -10097,6 +10099,8 @@ private aws.s3.bucket @defaults("name location public") {
   macieCoverage() aws.macie.bucket
   // S3 access points attached to the bucket
   accessPoints() []aws.s3.bucket.accessPoint
+  // CloudFormation stack that manages this bucket, if any
+  cloudformationStack() aws.cloudformation.stack
 }
 
 // Amazon S3 bucket event notification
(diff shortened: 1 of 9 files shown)


tas50:
Addressed in 2491425ec: `instance.loadBalancers` now traverses target groups too, so it covers ALB/NLB (instance registered via target group) in addition to classic ELB (direct instance registration). Verified against a live NLB whose instance-type target group resolves through the new path (0 registered instances in the test account, so no false positive).

PR #8520, opened 2026-06-19
🐛 llx: array/map all/any/none/one fail gracefully on a null receiver

### LINKED ISSUE #8519: 🐛 llx: array/map all/any/none/one error on a null receiver (missing map key) instead of failing gracefully
## Summary

Calling the collection-assertion builtins `all()`, `any()`, `none()`, or `one()` on a **null array or map** raises a hard **error** instead of failing the check gracefully. A null receiver most commonly arises from a **missing key in a typed `map[string][]T` field** — e.g. `secpol.privilegerights["SeMissing"]` resolves to a typed null `[]string`.

Because the engine runs query blocks in goroutines, an erroring check surfaces as an evaluation error on the whole datapoint rather than a clean pass/fail — which forces policy authors into ugly `switch` workarounds.

## Reproduction

Against the Windows secpol mock (a privilege right that is not assigned to anyone simply does not appear in the policy, so the key is absent):

```mql
secpol.privilegerights['SeMissingRight'].all(_ == 'S-1-5-32-544')   # ERROR
secpol.privilegerights['SeMissingRight'].any(_ == 'S-1-5-32-544')   # ERROR
secpol.privilegerights['SeMissingRight'].none(_ == 'S-1-5-32-544')  # ERROR
secpol.privilegerights['SeMissingRight'].one(_ == 'S-1-5-32-544')   # ERROR
```

All four return:

```
failed to validate all entries (list is null)
```

This is exactly the case the following Windows CIS-style pattern works around. The `switch(...) { case _ != empty: ... default: false }` exists **only** to stop the assertion from erroring when the privilege right is absent:

```mql
interactiveLogonRight = secpol.privilegerights['SeInteractiveLogonRight'];
switch(interactiveLogonRight) {
  case _ != empty:
      interactiveLogonRight.contains('S-1-5-32-544');
      interactiveLogonRight.contains('S-1-5-32-545');
  default:
      false;
}
```

## Root cause

`arrayAllV2` / `arrayAnyV2` / `arrayNoneV2` / `arrayOneV2` (`llx/builtin_array.go`) and `mapAll` / `mapAny` / `mapNone` / `mapOne` (`llx/builtin_map.go`) handle a `nil` receiver by returning a `RawData` with a hard `errors.New("failed to validate all entries (… is null)")`.

This is inconsistent with:

- the **dict** variants `dictAllV2` / `dictAnyV2` / `dictNoneV2` / `dictOneV2`, which already return a graceful null bool on a nil receiver;
- `arrayLengthV2`, `arrayContainsAll` / `arrayContainsNone`, `nonNilDataOpV2`, and bracket access, which were converted to null-propagation / graceful-false in the ongoing campaign of #7036, #7079, and #8320.

`contains()`, `length`, `where`, `== empty`, `== [...]`, and `containsAll/Only/None` on a missing key already fail gracefully today. Only `all` / `any` / `none` / `one` still error — they were missed by the earlier fixes.

## Proposed fix

Make the nil-receiver branch of the array and map `all` / `any` / `none` / `one` builtins return a graceful null bool (preserving any genuine upstream `bind.Error`), matching the dict variants:

```go
if bind.Value == nil {
    return &RawData{Type: types.Bool, Error: bind.Error}, 0, nil
}
```

A null bool scores as a clean `false` (via `isSuccess`), so the check **fails cleanly** instead of erroring — which is precisely what the `default: false` branch of the `switch` workaround produces. Policies can then drop the `switch` wrapper entirely.

## Related (out of scope here, follow-up)

A few dict builtins still error on a null receiver and could be aligned for full consistency: `dictContainsArray{String,Dict,Int,Regex}`, `dict{Camelcase,Downcase,Upcase,Lines,Split,Trim}V2`, and `{map,dict}{Keys,Values}V2`. These are not on the `map[string][]T`-key-miss path that bites policy authors today, so they are noted here rather than fixed in the same PR.

syrull:
Fixes #8519

## Problem

The collection-assertion builtins `all()` / `any()` / `none()` / `one()` raise a hard **error** when called on a **null array or map**, instead of failing the check gracefully:

```
failed to validate all entries (list is null)
```

A null receiver most commonly comes from a **missing key in a typed `map[string][]T` field**. The canonical case is `secpol.privilegerights["SeMissing"]` — a privilege right assigned to nobody simply does not appear in the Windows policy, so the key is absent and resolves to a typed null `[]string`. Because query blocks run in goroutines, the error surfaces on the whole datapoint rather than as a clean pass/fail.

This is the reason Windows CIS-style policies wrap privilege-rights checks in a `switch`:

```mql
interactiveLogonRight = secpol.privilegerights['SeInteractiveLogonRight'];
switch(interactiveLogonRight) {
  case _ != empty:
      interactiveLogonRight.contains('S-1-5-32-544');
  default:
      false;   # <- only here to stop the assertion erroring when the key is absent
}
```

`contains()`, `length`, `where`, `== empty`, `== [...]`, and `containsAll/Only/None` on a missing key already fail gracefully (via #7036 / #7079 / #8320). Only `all` / `any` / `none` / `one` were missed.

## Fix

Make the nil-receiver branch of the array (`llx/builtin_array.go`) and map (`llx/builtin_map.go`) `all` / `any` / `none` / `one` builtins return a graceful null bool, preserving any genuine upstream `bind.Error`:

```go
if bind.Value == nil {
    return &RawData{Type: types.Bool, Error: bind.Error}, 0, nil
}
```

This matches the **dict** variants (`dictAllV2` etc.), which already do this, and `arrayLengthV2` / `arrayContainsAll` / `arrayContainsNone`, which preserve `bind.Error` the same way. A null bool scores as a clean `false` via `isSuccess`, so the check **fails cleanly** instead of erroring — exactly what the `default: false` branch of the `switch` workaround produced. Policies can drop the wrapper.

A genuine upstream error still propagates, because `bind.Error` is preserved (covered by a regression test).

## Testing

- **Reproduced** the error against the Windows secpol mock for `all` / `any` / `none` / `one` on a missing privilege right; confirmed they erred pre-fix and return a graceful `false` post-fix.
- Added `TestArrayAssertions_NullReceiver` and `TestMapAssertions_NullReceiver` in `llx/` (graceful null bool + upstream-error-preserved cases).
- Added an end-to-end regression sub-test to `TestResource_Secpol` covering `secpol.privilegerights['SeMissingRight'].{contains,any,all,none,one}(…)`.
- `go test ./llx/`, `go test ./providers/os/resources/`, and `go test ./providers/core/resources/` all pass.

## Out of scope (follow-up)

A handful of dict builtins still error on a null receiver and could be aligned for full consistency: `dictContainsArray{String,Dict,Int,Regex}`, `dict{Camelcase,Downcase,Upcase,Lines,Split,Trim}V2`, and `{map,dict}{Keys,Values}V2`. They are not on the `map[string][]T`-key-miss path that bites policy authors today, so they are left out of this PR and noted in #8519.

### DIFF
--- a/llx/builtin_array.go
+++ b/llx/builtin_array.go
@@ -340,7 +340,13 @@ func arraySample(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64) (*Ra
 
 func arrayAllV2(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64) (*RawData, uint64, error) {
 	if bind.Value == nil {
-		return &RawData{Type: types.Bool, Error: errors.New("failed to validate all entries (list is null)")}, 0, nil
+		// A null receiver (e.g. a missing map key resolving to a typed null
+		// array, like secpol.privilegerights["SeMissing"]) propagates as a
+		// null bool rather than erroring. The check then resolves to a clean
+		// pass/fail instead of crashing the whole scan with an error. This
+		// mirrors the dict variants (dictAllV2 etc.) and arrayLengthV2, which
+		// already null-propagate. A genuine upstream error is preserved.
+		return &RawData{Type: types.Bool, Error: bind.Error}, 0, nil
 	}
 
 	filteredList := bind.Value.([]any)
@@ -353,7 +359,9 @@ func arrayAllV2(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64) (*Raw
 
 func arrayNoneV2(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64) (*RawData, uint64, error) {
 	if bind.Value == nil {
-		return &RawData{Type: types.Bool, Error: errors.New("failed to validate all entries (list is null)")}, 0, nil
+		// See arrayAllV2: a null receiver propagates as a null bool so the
+		// check fails cleanly instead of erroring the whole scan.
+		return &RawData{Type: types.Bool, Error: bind.Error}, 0, nil
 	}
 
 	filteredList := bind.Value.([]any)
@@ -366,7 +374,9 @@ func arrayNoneV2(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64) (*Ra
 
 func arrayAnyV2(e *blockExecutor, bind *RawData, chunk *Chunk, ref uint64) (*RawData, uint64, error) {
 	if bind.Value == nil {
-		return &RawData{Type: types.Bool, Error: errors.New("failed to validate all entries (list is null)")}, 0, nil
+		// See arrayAllV2: a null receiver propagates as a null bool so the
+		// check fails cleanly instead of erroring the whole scan.
+		return &RawData{Type: types.Bool, Error: bind.Error}, 0, nil
 	}
 
 	filteredList := bind.Value.([]any)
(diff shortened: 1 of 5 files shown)


syrull:
Thanks for the review! Addressed both suggestions in d24f2a1:

1. **`bind.Error` in the dict variants** — went with the stronger option: `dictAllV2`/`dictAnyV2`/`dictNoneV2`/`dictOneV2` now preserve `bind.Error` on a null receiver too (they previously dropped it), so the "matches the dict variants" comment is now literally accurate and the dict path also surfaces genuine upstream errors.
2. **Map test parity** — added the "preserves a genuine upstream error" sub-test to the map test, plus a new `TestDictAssertions_NullReceiver` covering the dict variants the same way.

PR #8526, opened 2026-06-19
⭐ aws: consistent internet-exposure predicates (isPublic, rds internetReachable)

tas50:
## What

"Internet exposed" in AWS is **two distinct mechanisms**. The EC2 `internetReachable` work covered network reachability; this adds the policy-based mechanism and extends reachability to RDS, with consistent predicate names so policy authors can sweep uniformly.

### Resource-policy / public access
- **`aws.iam.policyStatement.hasPublicPrincipal()`** — structural predicate: an `Allow` statement grants to a wildcard principal `"*"`. Mirrors the existing `hasWildcardAction` / `hasWildcardResource`.
- **`isPublic()`** on `aws.kms.key`, `aws.sns.topic`, `aws.sqs.queue`, `aws.ecr.repository`, `aws.lambda.function` — the resource policy grants to a wildcard principal **with no scoping conditions**. A wildcard narrowed by an org-id / source-account / source-ARN condition is the standard cross-account pattern, not public exposure, so it's excluded (false-positive control). `isPublic` is already an established field name in the schema (snapshots, AMIs), so this is consistent.

### Network reachability
- **`aws.rds.dbinstance.internetReachable()`** — publicly accessible **and** an attached security group allows inbound from `0.0.0.0/0`/`::/0`. The raw `publiclyAccessible` flag is the on-paper setting; this gates on actual reachability, mirroring the EC2 model. Factors the SG-ingress check into a shared `securityGroupsAllowPublicIngress` helper (now used by both RDS and `ec2.instance.internetReachable`).

Result — a uniform exposure sweep:
```
aws.kms.keys.where(isPublic)
aws.sqs.queues.where(isPublic)
aws.rds.instances.where(internetReachable)
```

## Tests

This PR ships **unit tests** — `aws_iam_policy_statement_test.go` — covering `principalsIncludeWildcard`, `dictIsEmpty`, `hasPublicPrincipal`, and `statementsAllowPublic` (Allow/Deny, wildcard vs specific, scoped-by-condition vs unconditional, mixed). They run **without a mock runtime** by constructing policy statements with pre-set `TValue` fields — a pattern that generalizes to the other cross-resource resolvers and starts to close the long-standing provider test-harness gap.

## Notes
- Reads cached policy/SG data — **no new API calls or permissions** (912, unchanged).

## Verification
- `go test ./resources/ -run 'PublicPrincipal|StatementsAllowPublic|Wildcard|DictIsEmpty'` → all pass.
- Live: an RDS instance with `publiclyAccessible=true` but a locked-down security group correctly reports `internetReachable=false` — the on-paper flag would false-positive; this doesn't.

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -1716,6 +1716,8 @@ private aws.kms.key @defaults("id region aliases metadata.Description") {
   policy() string
   // Parsed statements from the key policy
   policyStatements() []aws.iam.policyStatement
+  // Whether the resource policy grants public access (a wildcard principal with no scoping conditions)
+  isPublic() bool
   // Describes the type of key material in the key
   //
   // One of: SYMMETRIC_DEFAULT, RSA_2048, RSA_3072, RSA_4096, ECC_NIST_P256, ECC_NIST_P384, ECC_NIST_P521, ECC_SECG_P256K1, HMAC_224, HMAC_256, HMAC_384, HMAC_512, SM2.
@@ -2127,6 +2129,8 @@ private aws.iam.policyStatement @defaults("effect") {
   // True when the statement includes the global `*` action or any service-wide
   // wildcard such as `s3:*` or `ec2:*`.
   hasWildcardAction() bool
+  // Whether the statement grants access to a wildcard principal ("*")
+  hasPublicPrincipal() bool
 }
 
 // AWS IAM policy
@@ -4657,6 +4661,8 @@ private aws.sns.topic @defaults("arn") {
   policy() dict
   // Parsed statements from the topic access policy
   policyStatements() []aws.iam.policyStatement
+  // Whether the resource policy grants public access (a wildcard principal with no scoping conditions)
+  isPublic() bool
   // Whether the topic is a FIFO topic
   fifoTopic() bool
   // Whether content-based deduplication is enabled (FIFO topics only)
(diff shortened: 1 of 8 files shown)


tas50:
Addressed in a59d83977.

- **`isPublic` vs `allowsPublicAccess` (comment 1) + naive condition handling (comment 2):** You were right that they could contradict, and that `dictIsEmpty` was the inferior implementation. Fixed by deleting my duplicate `principalsIncludeWildcard` and the naive `dictIsEmpty`, and reusing the existing `hasWildcardPrincipal` / `hasSourceScopingCondition`. `statementsAllowPublic` now applies the same source-scoping rule (only `aws:SourceArn` / `aws:SourceAccount` / `aws:PrincipalOrgID` count), so a wildcard grant narrowed only by e.g. `aws:RequestedRegion` is correctly still public. And `lambda.function.allowsPublicAccess` now delegates to `statementsAllowPublic`, so the two predicates share one implementation and can't diverge. Added a test for the non-scoping region-condition case.

- **`securityGroupsAllowPublicIngress` error propagation (comment 3):** Kept the propagation (surfacing a real `GetIncludesPublicSource` error beats silently treating the rule as closed) — in practice it only reads already-resolved `ipRanges`/`ipv6Ranges`, so it won't error on well-formed data. Flagging the behaviour change here as you suggested; happy to swallow-to-`false` instead if you'd prefer the old semantics.

PR #8527, opened 2026-06-20
⭐ aws: NACL-aware reachability, Lambda Function URL + API Gateway exposure

tas50:
Stacked on #8526 (bases on `aws-internet-exposure-predicates`; rebase to main once that merges).

Sharpens and extends the internet-exposure analysis along three vectors, each with unit tests.

### #3 — NACL-aware reachability
`aws.ec2.instance.internetReachable` now also requires the subnet's **network ACL** to permit inbound from the internet. A restrictive NACL blocks reachability even when the security group is open, so this removes a false-positive class. Adds `naclAllowsPublicIngress` (rules evaluated in rule-number order, first match wins) + `networkAclAllowsPublicIngress`.

### #4 — Lambda Function URL
`aws.lambda.function.isPublic` now also returns true when the function has a **Function URL with auth type `NONE`** (unauthenticated public invocation). Previously it inspected only the resource policy, missing the most common Lambda exposure vector. Adds `functionUrlIsPublic`.

### #5 — API Gateway
`aws.apigatewayv2.route.isPublic` — the route's authorization type is `NONE` (or unset), so it is invokable without IAM / JWT / custom auth. Adds `routeAuthIsPublic`.

## Tests
`aws_internet_exposure_test.go` — pure unit tests (no runtime):
- `naclAllowsPublicIngress`: default-allow, shadowing deny, allow-wins-over-later-deny, specific-cidr-only, ordering, empty.
- `functionUrlIsPublic`, `routeAuthIsPublic`: NONE/unset vs AWS_IAM/JWT/CUSTOM.

## Notes
- Reads cached data — **no new API calls or permissions** (912, unchanged). Versions `13.37.1`.
- **Data gap flagged:** REST (v1) API Gateway *method-level* authorization isn't modeled in the provider, so v1 method exposure can't be computed yet — a follow-up.

## Verification
- `go test ./resources/ -run 'Nacl|FunctionUrl|RouteAuth'` → pass.
- Live: `internetReachable` positive case preserved under the new NACL gate (reachable instance has a default-allow NACL); `route.isPublic` resolves (no v2 APIs in the test account — unit tests cover the logic).

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -13907,7 +13907,7 @@ private aws.apigatewayv2.route @defaults("routeKey apiId authorizationType") {
   requestModels map[string]string
   // Whether the route is owned and managed by another AWS service
   apiGatewayManaged bool
-  // Whether the route is publicly invokable (no authorizer — authorization type NONE)
+  // Whether the route is publicly invocable (no authorizer — authorization type NONE)
   isPublic() bool
 }
 
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -13907,6 +13907,8 @@ private aws.apigatewayv2.route @defaults("routeKey apiId authorizationType") {
   requestModels map[string]string
   // Whether the route is owned and managed by another AWS service
   apiGatewayManaged bool
+  // Whether the route is publicly invokable (no authorizer — authorization type NONE)
+  isPublic() bool
 }
 
 // Authorizer in an API Gateway v2 API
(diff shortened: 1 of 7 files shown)


tas50:
Good catch on the NACL port/protocol limitation. That code is merged, and I'm building the next exposure PR (network reachability fan-out) on top of it now — I'll add the clarifying code comment ("port/protocol-aware evaluation not yet implemented — conservative approximation") to `naclAllowsPublicIngress` there rather than spin a standalone PR for a one-line comment.

PR #8531, opened 2026-06-20
🐛 aws: detect wide-CIDR and prefix-list public sources (fix false negatives)

tas50:
First of the exposure-analysis follow-ups (foundational — every network predicate inherits this).

## Problem
`includesPublicSource` (security-group rules) and `cidrEntryIsPublic` (NACL entries) only matched the **literal** `0.0.0.0/0` / `::/0`. They missed:
- wide-but-not-`/0` public blocks (`0.0.0.0/1`, a public `/8`), and
- security-group rules that reference a **managed prefix list** containing public ranges (the rule named no CIDR at all).

That's a false negative inherited by every network-exposure predicate shipped so far — `instance.internetReachable`, `rds.internetReachable`, and the NACL evaluation.

## Fix
New `cidrIsPublic` helper: a CIDR is public when it is `0.0.0.0/0`/`::/0`, **or** a broad block (prefix length ≤ `/8` IPv4, ≤ `/32` IPv6) that is **not** contained within a private/reserved range (RFC1918, loopback, link-local, CGNAT, documentation, reserved-for-future). Narrow or specific ranges (a `/32` host, a corporate `/24`) stay scoped — a deliberately conservative threshold to avoid false positives on large legitimate allow-lists.

`includesPublicSource` now also expands referenced managed prefix lists and checks their entries.

## Notes
- The `/8` (IPv4) / `/32` (IPv6) broad-public threshold is the one judgment knob; documented in `aws_cidr.go` and easy to tune.
- Pure helpers — **no schema change, no new API calls or permissions.**

## Tests
`aws_cidr_test.go` — table-driven `cidrIsPublic` (all-/0, broad public, every private/reserved range, narrow/specific, malformed), `anyCidrPublic`. The pre-existing `TestCidrEntryIsPublic` still passes (documentation range `2001:db8::/32` correctly excluded).

## Verification
Live: public SGs still flagged (no regression on the literal `0.0.0.0/0` case), now with wide-CIDR + prefix-list coverage added.

### DIFF
--- a/providers/aws/resources/aws_cidr.go
+++ b/providers/aws/resources/aws_cidr.go
@@ -1,0 +1,92 @@
+// Copyright Mondoo, Inc. 2024, 2026
+// SPDX-License-Identifier: BUSL-1.1
+
+package resources
+
+import "net"
+
+// privateOrReservedCIDRs are address ranges that are not routable from the
+// public internet. A security-group or network-ACL rule whose source is wholly
+// inside one of these does not constitute internet exposure.
+var privateOrReservedCIDRs = mustParseCIDRs(
+	"10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16", // RFC1918
+	"127.0.0.0/8",    // loopback
+	"169.254.0.0/16", // link-local
+	"100.64.0.0/10",  // carrier-grade NAT
+	"0.0.0.0/8",      // "this host on this network" (RFC1122)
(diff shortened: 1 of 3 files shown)


tas50:
Addressed in bb6fc40ba — `mustParseCIDRs` now panics on a malformed literal (init-time, hardcoded inputs), matching the `must*` convention so a typo in the private/reserved list fails loudly.

PR #8532, opened 2026-06-20
⭐ aws: mitigation-context fields (exposed-and-undefended detection)

tas50:
Exposure dimension A — "exposed *and undefended*". A public resource with a WAF and enforced TLS is far less of a finding than one without; these fields let policies make that distinction instead of treating every exposure equally.

## Fields
- **`aws.cloudfront.distribution.protectedByWaf`** — a WAF web ACL is associated with the distribution.
- **`aws.cloudfront.distribution.enforcesHttps`** — viewer protocol policy is not `allow-all` (plaintext HTTP rejected/redirected).
- **`aws.elb.loadbalancer.enforcesTls`** — no plaintext HTTP/TCP/UDP listener accepts traffic.

Enables findings like:
```
aws.cloudfront.distributions.where(enabled).all(protectedByWaf && enforcesHttps)
aws.elb.loadBalancers.where(scheme == "internet-facing").all(enforcesTls)
```

## Notes
- Pure helpers (`viewerPolicyEnforcesHttps`, `listenerProtocolIsPlaintext`) with table-driven tests.
- Reads cached `webAclId` / `viewerProtocolPolicy` / `listeners` — **no new API calls or permissions** (912, unchanged). Versions `13.37.1`.

## Verification
Live: an NLB with a plaintext TCP listener reports `enforcesTls=false`; an all-encrypted load balancer reports `true`.

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -5307,6 +5307,8 @@ private aws.elb.loadbalancer @defaults("name region elbType scheme dnsName") {
   instances() []aws.ec2.instance
   // List of listeners for ALB/NLB load balancers (typed)
   listeners() []aws.elb.listener
+  // Whether all listeners use an encrypted protocol (no plaintext HTTP, TCP, or UDP listener)
+  enforcesTls() bool
   // Typed attributes for ALB/NLB load balancers
   attribute() aws.elb.loadbalancer.attribute
   // Health check configuration (classic load balancers only; null for ALB/NLB)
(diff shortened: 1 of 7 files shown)


tas50:
All three addressed in 76180fb3a.

- **🔴 `enforcesHttps` cache-behavior gap:** now checks the default behavior *and* every entry in `cacheBehaviors`, returning false if any uses allow-all. One correction to the suggested fix: the dict key is **`ViewerProtocolPolicy`** (PascalCase), not lowercase — aws-sdk-go-v2 types carry no json tags, so `JsonToDict` (a `json.Marshal`→unmarshal) preserves the Go field name. The lowercase key would have silently never matched, so the field would still always pass. Verified against the SDK type (0 json tags in `cloudfront/types`).
- **🟡 `enforcesTls` classic-ELB miss:** confirmed `listeners()` is ALB/NLB-only (elbv2 `DescribeListeners`), while classic ELBs populate `listenerDescriptions` at creation. Switched `enforcesTls` to read `listenerDescriptions`, which covers both — classic (protocol nested under `Listener`) and ALB/NLB (top-level `Protocol`) — via a dual-shape extractor with its own unit test. A classic HTTP ELB now correctly reports `enforcesTls=false`.
- **🔵 doc comment:** updated to "Whether every cache behavior requires viewers to use HTTPS…".

PR #8536, opened 2026-06-20
⭐ aws: ec2.instance.exposure sub-resource (internet-exposure breakdown)

tas50:
Reshapes the `internetReachable` boolean into an explainable **sub-resource** so a finding is triageable, not just true/false.

## `aws.ec2.instance.exposure`
| field | meaning |
|---|---|
| `internetReachable` | the overall verdict |
| `hasPublicIp` | instance has a public IP |
| `inPublicSubnet` | subnet routes to an internet gateway |
| `securityGroupAllowsIngress` | an attached SG permits public ingress |
| `networkAclAllowsIngress` | the subnet NACL permits public ingress |
| `openIngressRules` | the SG ingress rules that open it, **with port ranges** |
| `internetFacingLoadBalancers` | internet-facing LBs that front the instance |

`instance.internetReachable` stays as a convenience shortcut delegating to `exposure.internetReachable`, so `instances.where(internetReachable)` still works while `{ exposure { ... } }` gives the full breakdown.

## Notes
- Reads cached data — **no new API calls or permissions** (912, unchanged). Versions `13.37.1`.

## Verification
Live: a public-IP instance in a public subnet behind an SG opening 80/443 reports `internetReachable=true` with both rules listed in `openIngressRules`, and all contributing signals true.

## Follow-up
Per the request to apply exposure analysis broadly, a companion PR will add exposure breakdowns to the other network-reachable resources (RDS, ELB, and the `publiclyAccessible` services) via a shared exposure shape.

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -15853,8 +15853,12 @@ private aws.ec2.instance @defaults("instanceId region state instanceType archite
   inPublicSubnet() bool
   // Whether the instance is directly reachable from the internet
   //
-  // True when the instance has a public IP, sits in a public subnet (its subnet's route table has an internet gateway route), and an attached security group permits inbound traffic from 0.0.0.0/0 or ::/0. Exposure through an internet-facing load balancer is a separate path — query `loadBalancers` and their `scheme` for that.
+  // True when the instance has a public IP, sits in a public subnet (its subnet's route table has an internet gateway route), and an attached security group permits inbound traffic from 0.0.0.0/0 or ::/0. A convenience shortcut for `exposure.internetReachable`.
   internetReachable() bool
+  // Internet-exposure analysis for the instance
+  //
+  // Breaks down whether and how the instance is reachable from the internet — the contributing signals, the security-group ingress rules that open it, and any internet-facing load balancers that front it.
+  exposure() aws.ec2.instance.exposure
   // CloudFormation stack that manages this instance, if any
   cloudformationStack() aws.cloudformation.stack
   // Infrastructure-management system that owns this instance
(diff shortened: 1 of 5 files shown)


tas50:
- **🔴 `Arn.Error` unguarded:** fixed in 6eed05449 — `exposure()` now reads `GetArn()` and returns on error before building the `__id`.
- **🟡 eager evaluation / short-circuit:** I kept the full computation deliberately. Short-circuiting on `hasPublicIp == false` would hide the two signals the breakdown exists for: an **open security group on a private instance** (a real misconfiguration finding) and exposure **via an internet-facing load balancer** (independent of the instance's own public IP). The cost is also modest in practice — the load-balancer list is materialized once and cached across all instances, and the SG scan iterates already-resolved attached SGs — and the field is lazy (only computed when `exposure` is queried). Happy to revisit if profiling shows it's hot.

tas50:
Rebased onto latest `main` and addressed review feedback:

- **Stale doc comments removed** — the duplicate `inPublicSubnet` block above `instanceInPublicSubnet`, and the orphaned `internetReachable` comment fragment left between `instanceInPublicSubnet` and `instanceSubnetNaclAllowsPublicIngress`.
- **Failing `go-test` fixed** — `TestIpPermissionIncludesPublicSource/private_only` panicked with a nil-pointer deref. The test builds a synthetic `ippermission` with no runtime, and the private-only case (no public CIDR to short-circuit on) reached `includesPublicSource`'s lazy `prefixLists()` resolution, which dereferenced `MqlRuntime.HasRecording`. Fixed by pre-resolving the `prefixLists` field in the test helper so the prefix-list path runs with empty data instead of touching a nil runtime. Production resources always carry a runtime, so this was a test-setup gap only.

The ARN-error guard (critical) and eager-evaluation note were already handled in the prior commits. Full `providers/aws/resources` test suite passes locally; `gofmt` clean.

PR #8544, opened 2026-06-20
⭐ gcp: compute instance network-exposure breakdown

tas50:
Cross-provider exposure (one PR per provider) — GCP. Mirrors `aws.network.exposure`.

## `gcp.project.computeService.network.exposure`
Added `gcp.project.computeService.instance.exposure`:
| field | meaning |
|---|---|
| `internetReachable` | an external IP **and** an applicable firewall rule admitting any address |
| `hasPublicIp` | the instance has an external IP (reuses existing `hasPublicIp`) |
| `firewallAllowsIngress` | a firewall rule that applies to this instance admits `0.0.0.0/0`/`::/0` |
| `openIngressFirewalls` | the applicable internet-open INGRESS firewall rules |

GCP firewalls are VPC-level and apply by **network + target tags / target service accounts**, so the resolver correlates each firewall's network and targeting against the instance (a rule with no targets applies to all instances in its network).

## Tests
The non-trivial matching logic is in pure, unit-tested helpers:
- `firewallRuleOpenIngress` (enabled INGRESS + any-address source)
- `networkNameFromUrl` (so full URLs and short names compare equal)
- `firewallTargetsInstance` (no-targets = all; else tag / service-account match)

## Notes
- Reads cached instance/firewall data — **no new API calls**. Versions `13.24.1`.
- The firewall-matching helpers are unit-tested; no GCP account available for live verification, so the resolver wiring (network correlation across the project's firewalls) is verified by build + the helper tests rather than a live scan.

### DIFF
--- a/providers/gcp/resources/gcp.lr
+++ b/providers/gcp/resources/gcp.lr
@@ -1383,8 +1383,11 @@ gcp.project.computeService.instance @defaults("name id") {
   networkStackTypes []string
   // Whether the instance has at least one external IP attached via a network interface accessConfig
   hasPublicIp() bool
-  // Internet-exposure breakdown (public IP combined with applicable firewall ingress)
-  exposure() gcp.project.computeService.network.exposure
+  // Internet-exposure breakdown
+  //
+  // Explains whether the instance is reachable from the internet by combining
+  // external-IP presence with the VPC firewall ingress rules that apply to it.
+  exposure() gcp.project.computeService.instance.exposure
   // Whether the instance runs as the default Compute Engine service account (<projectNumber>-compute@developer.gserviceaccount.com)
   usesDefaultServiceAccount() bool
   // Whether any attached service account has the broad cloud-platform OAuth scope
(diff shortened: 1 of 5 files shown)


tas50:
Addressed in a5969f40e:
- 🟡 Renamed the resource `gcp.project.computeService.network.exposure` → `gcp.project.computeService.instance.exposure` so it lives under the instance namespace instead of colliding with the VPC-network resource. This follows the established `instance.confidentialCompute` pattern, where the `exposure()` field accessor and its same-named sub-resource share a single `.lr.versions` entry. Updated `.lr`, regenerated `.lr.go`/`.lr.versions`, and updated the Go resolver type `mqlGcpProjectComputeServiceInstanceExposure`.
- Also promoted the `exposure()` field doc-comment to the two-part form per the earlier suggestion.

Build + unit tests pass.

tas50:
Rebased onto latest `main`. Note: #8546 (the broader GCP exposure PR) merged in the meantime and also adds `gcp_exposure.go`, so the rebase required a union merge — both this PR's compute-instance `exposure()`/firewall helpers and main's GKE `controlPlaneInternetReachable` now coexist in the file. Regenerated codegen (no drift), build + full GCP resources tests pass.

PR #8545, opened 2026-06-20
⭐ proxmox: add internet-exposure analysis across resources

tas50:
## What

Adds internet-exposure analysis to the Proxmox provider via a new `allowsPublicIngress()` predicate on `proxmox.firewall.rule`.

## Why this mechanism

Proxmox VE is on-prem virtualization. There is no cloud-style public-IP / security-group / authorized-networks model to map onto network-reachability exposure — VMs and containers have no `publiclyAccessible`/`publicIp` signal, and reachability is governed entirely by the PVE firewall. The only place where "internet exposure" maps cleanly and meaningfully is the firewall rule itself.

## Resource covered

- **`proxmox.firewall.rule.allowsPublicIngress` (bool)** — firewall-rule-level public-ingress predicate. The same `proxmox.firewall.rule` resource is shared across cluster, node, VM, and container scopes (`firewallRules` on each), so this single predicate covers every firewall scope.

`allowsPublicIngress` is true only when the rule is:
- enabled, and
- inbound (`type == "in"`, or empty which PVE defaults to inbound), and
- `ACCEPT` (case-insensitive), and
- sourced from anywhere — empty source (PVE treats absent as "any"), `0.0.0.0/0`, `::/0`, or a bare all-zero address.

Outbound/`DROP`/`REJECT`/disabled rules, group references, and rules restricted to a specific CIDR/IP/alias/IPset all return false. Alias/IPset references are deliberately treated as not-anywhere to avoid over-reporting (their breadth is not resolved here).

## Implementation notes

- Pure helpers `firewallRuleAllowsPublicIngress` and `sourceIsAnywhere` in `providers/proxmox/resources/proxmox_exposure.go`, with table-driven unit tests (`proxmox_exposure_test.go`, plain `testing`, no new deps).
- The resolver reads only fields already cached on the rule resource (`type`, `action`, `source`, `enable`) — **no new API calls and no new IAM/token permissions**.
- New `.lr.versions` entry `proxmox.firewall.rule.allowsPublicIngress` at `0.1.13` (next patch after the provider's current `0.1.12`).

## Tests

`go test ./resources/ -run 'TestSourceIsAnywhere|TestFirewallRuleAllowsPublicIngress'` passes; `go build ./...` is clean.

### DIFF
--- a/providers/proxmox/resources/proxmox.lr
+++ b/providers/proxmox/resources/proxmox.lr
@@ -810,6 +810,17 @@ proxmox.firewall.rule @defaults("pos action") {
   // than a verb like ACCEPT/DROP, and apply that named rule set in
   // place. This resolves the reference; null for non-group rules.
   group() proxmox.firewall.group
+  // Whether this rule permits inbound traffic from any source
+  //
+  // True when the rule is enabled, an inbound (`type == "in"`) ACCEPT
+  // rule, and its `source` is unrestricted: empty (Proxmox treats an
+  // empty source as "any"), `0.0.0.0/0`, `::/0`, or an equivalent
+  // all-addresses match. Such a rule opens the guarded VM, container,
+  // node, or cluster to every network the host can reach, including the
+  // internet when the bridge is routed publicly. False for outbound
+  // rules, DROP/REJECT rules, disabled rules, and rules restricted to a
+  // specific CIDR, IP, alias, or IPset reference.
+  allowsPublicIngress() bool
 }
 
 // Proxmox VE user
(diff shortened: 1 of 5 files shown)


tas50:
Addressed:
- 🔵 Expanded the code comment on `firewallRuleAllowsPublicIngress` to spell out the over-report rationale: a misconfigured rule with `type=""` + `action=ACCEPT` is intentionally flagged as exposed, since over-reporting is the safer direction for an exposure check than silently missing a real exposure. The behavior is already covered by the `empty type defaults inbound` test case.
- 🟡 The version note is informational — `0.1.13` is correct as the next patch after the provider's current `0.1.12`; no change needed.

Also rebased onto latest main. Build + tests pass.

PR #8550, opened 2026-06-20
⭐ stackit: add internet-exposure analysis across resources

tas50:
Extends the STACKIT provider with internet-exposure predicates on the managed-database and load-balancer resources, reusing the shared `stackit.network.exposure` sub-resource already established for servers (server exposure was done on another branch and is left untouched here).

### Network reachability — `exposure() stackit.network.exposure`
- `stackit.loadBalancer`
- `stackit.alb.loadBalancer`

A load balancer is internet-reachable when it carries a public external address (and is not `privateNetworkOnly`) **and** the access-control allow-list (`options.accessControl.allowedSourceRanges`) is empty (no restriction) or admits any address (`0.0.0.0/0` or `::/0`). The allow-list takes the place of a security group, so `openIngressRules` is empty for these.

### Authorized-network reachability — `internetReachable() bool`
Managed databases have no security group; they gate access via an ACL / parameters allow-list, so they get the simpler predicate:

- **Flex databases (ACL-based):** `stackit.postgresFlex.instance`, `stackit.mongoDbFlex.instance`, `stackit.sqlServerFlex.instance` — reachable when the connection ACL is empty or admits a default route.
- **DBaaS instances (parameters-based):** `stackit.openSearch.instance`, `stackit.mariaDb.instance`, `stackit.redis.instance`, `stackit.rabbitMq.instance`, `stackit.logMe.instance` — reachable when `enable_public_access` is on and the `sgw_acl` source allow-list is empty or admits any address.

## Notes
- All predicates **read already-cached data** (ACL / options / parameters) — no new API calls and no new IAM permissions.
- Rule/CIDR/dict parsing lives in pure helpers (`stackit_exposure.go`) with table-driven unit tests (`stackit_exposure_test.go`), using plain `testing` (no new dependencies).
- New `.lr.versions` entries land at `13.0.6` (next patch after the provider's `13.0.5`).
- `go build ./...` and `go test ./resources/` both pass.

### DIFF
--- a/providers/stackit/resources/stackit.lr
+++ b/providers/stackit/resources/stackit.lr
@@ -934,9 +934,10 @@ private stackit.postgresFlex.instance @defaults("id name version status") {
   options() map[string]string
   // Whether the instance accepts connections from any address
   //
-  // True when the connection ACL is empty or admits any address (0.0.0.0/0
-  // or ::/0); a managed Flex instance with a public endpoint and such an ACL
-  // is reachable from the internet.
+  // True only when the connection ACL explicitly admits a default route
+  // (0.0.0.0/0 or ::/0). An empty ACL is not flagged: the Flex API exposes no
+  // public-endpoint indicator, so an unpopulated ACL cannot be assumed to be
+  // internet-reachable.
   internetReachable() bool
 }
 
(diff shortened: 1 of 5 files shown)


tas50:
Addressed in e04b13fd6:
- 🟡 `flexInstanceReachable` no longer reports `internetReachable: true` for an empty ACL. The Flex SDK (`postgresflex.Instance`) exposes no public-endpoint flag, so an unpopulated/VPC-only ACL can't be assumed reachable — the Flex path now flags only when the ACL *explicitly* admits a default route (`0.0.0.0/0`/`::/0`). The DBaaS and LB paths keep empty-list semantics since they gate on `enable_public_access`/public-IP first. Updated `.lr` docs + tests.
- 🔵 Added a comment noting the ALB reads `privateNetworkOnly` from the `options` blob (unlike the legacy LB's top-level field).

Build + unit tests pass.

PR #8556, opened 2026-06-20
Add windows.update.policy resource for Windows Update for Business GPO settings

tas50:
## What

Extends the existing `windows.update` resource with a `policy()` accessor backed by a new `windows.update.policy` resource. It exposes the **Windows Update for Business** and **Automatic Updates** group policy settings, read directly from the registry under `HKLM\SOFTWARE\Policies\Microsoft\Windows\WindowsUpdate` and its `AU` subkey.

## Why

The CIS Windows 10/11 and Server 2016/2019/2022 policies currently audit Windows Update via raw `registrykey.property(...)` lookups. This resource lets those checks become simple, human-readable fields. It needs no command execution — it works wherever the registry is readable (including filesystem/snapshot scans).

### Before → after

| Current registry check | With this resource |
|---|---|
| `registrykey.property(path: '...\WindowsUpdate\AU', name: 'NoAutoUpdate').data == 0` | `windows.update.policy.automaticUpdatesEnabled` |
| `...\AU` `NoAutoUpdate` | `windows.update.policy.noAutoUpdate == 0` |
| `...\AU` `ScheduledInstallDay` | `windows.update.policy.scheduledInstallDay == 0` |
| `...\AU` `NoAutoRebootWithLoggedOnUsers` | `windows.update.policy.noAutoRebootWithLoggedOnUsers` |
| `...\WindowsUpdate` `ManagePreviewBuildsPolicyValue` | `windows.update.policy.managePreviewBuilds == 0` |
| `DeferFeatureUpdates` / `DeferFeatureUpdatesPeriodInDays` | `deferFeatureUpdates` / `deferFeatureUpdatesPeriodInDays` |
| `DeferQualityUpdates` / `DeferQualityUpdatesPeriodInDays` | `deferQualityUpdates` / `deferQualityUpdatesPeriodInDays` |
| `AllowTemporaryEnterpriseFeatureControl` | `allowTemporaryEnterpriseFeatureControl` |
| `SetAllowOptionalContent` | `allowOptionalContent` |
| `SetDisablePauseUXAccess` | `disablePauseUXAccess` |

This covers **every** `WindowsUpdate`/`AU` registry value checked across those five policies (plus `scheduledInstallTime` for completeness).

## Design notes

- Integer fields mirror the underlying `REG_DWORD` values and are **null when the policy is not configured**, so checks can distinguish "explicitly set to 0" from "left at the default". This matters for values whose compliant setting is `0` (e.g. `ScheduledInstallDay = 0` = "every day", `DeferQualityUpdatesPeriodInDays = 0`) — an absent value must not silently read as a passing `0`.
- `automaticUpdatesEnabled` is a convenience bool (`NoAutoUpdate` present and `== 0`).
- Reuses the existing `readRegistryKey` / `regInt` helpers and the `wuPolicyKey` / `wuPolicyAUKey` path constants already in `windows_update.go`.
- Extraction logic is factored into a pure `computeWindowsUpdatePolicy` function with unit tests.

## Testing

- `go test ./resources/ -run WindowsUpdate` — new `TestComputeWindowsUpdatePolicy` covers not-configured (all null), `NoAutoUpdate=0/1`, the absent-vs-explicit-0 distinction, case-insensitive value names, and the WUfB deferral values.
- `go build ./...` and `go vet ./resources/` pass.
- Schema/codegen regenerated via `lr go` + `lr versions` + `gen/main.go`.


tas50:
Addressed both review comments in 0b0edcedb:

- **`policy()` masking registry failures** — `policy()` now checks the `ok` booleans from `readRegistryKey` (mirroring `config()`) and returns an error when *neither* the WindowsUpdate policy key nor its AU subkey could be read. Absent keys on an unmanaged host still read successfully (`ok=true` via the `NotFound` path) and report "not configured", so this only fires on a genuinely unreadable registry.
- **Doc nit** — the `windows.update.policy` field doc now notes each value is nullable, where null signals "not configured" (distinct from an explicit 0).

PR #8557, opened 2026-06-20
🌟 Add windows.smartScreen resource for Microsoft Defender SmartScreen

tas50:
## Summary

Adds a **`windows.smartScreen`** resource exposing the Microsoft Defender SmartScreen policy configuration across its three surfaces. SmartScreen is a separate feature from Defender Antivirus (it has no `Get-Mp*` cmdlet), so it gets its own resource backed by the well-known Group Policy registry values.

### Fields
| Field | Surface | Backing policy value |
|---|---|---|
| `explorerEnabled` | Windows / File Explorer | `…\Windows\System:EnableSmartScreen` |
| `explorerLevel` | Windows / File Explorer | `…\Windows\System:ShellSmartScreenLevel` (`Block`/`Warn`) |
| `edgeEnabled` | Microsoft Edge | `…\Edge:SmartScreenEnabled` |
| `edgePuaEnabled` | Microsoft Edge | `…\Edge:SmartScreenPuaEnabled` |
| `edgePreventOverride` | Microsoft Edge | `…\Edge:PreventSmartScreenPromptOverride` |
| `edgePreventOverrideForFiles` | Microsoft Edge | `…\Edge:PreventSmartScreenPromptOverrideForFiles` |
| `storeAppsEnabled` | Microsoft Store apps | `…\AppHost:EnableWebContentEvaluation` |

### Notes
- All values are read in a **single PowerShell call** and emitted as normalized JSON. Values are numeric DWORDs / strings — **locale-invariant**, so the resource works on non-English Windows.
- An **unconfigured** value resolves to its disabled default (`false`, or `""` for `explorerLevel`) rather than erroring. DWORD toggles are pointer-backed so an unconfigured value is distinguishable from an explicit `0`.

### Why
Enables the enterprise Windows policies (windows-10/11, windows-server-2016/2019/2022) to express the SmartScreen CIS controls through a typed resource instead of raw `registrykey.property(...)` reads.

### Tests
`ParseSmartScreenSettings` is covered for configured, not-configured (null), and explicit-zero inputs.

### DIFF
--- a/providers/os/resources/os.lr
+++ b/providers/os/resources/os.lr
@@ -6359,6 +6359,9 @@ windows {
   deviceGuard() windows.deviceGuard
   // System-wide Windows Exploit Protection configuration
   exploitProtection() windows.exploitProtection
+
+  // Microsoft Defender SmartScreen policy
+  smartScreen() windows.smartScreen
 }
 
 // Windows Exploit Protection
--- a/providers/os/resources/os.lr
+++ b/providers/os/resources/os.lr
@@ -6453,6 +6453,34 @@ private windows.exploitProtection.heap @defaults("terminateOnError") {
   terminateOnError string
 }
 
+// Microsoft Defender SmartScreen
+//
+// Examine the Microsoft Defender SmartScreen policy configuration across its
+// three surfaces: Windows / File Explorer (which checks apps and files), the
+// Microsoft Edge browser, and Microsoft Store apps. SmartScreen is configured
+// through Group Policy registry values rather than a dedicated cmdlet; a value
+// that has not been configured resolves to its disabled default (false, or an
+// empty string for `explorerLevel`).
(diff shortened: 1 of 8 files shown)


tas50:
Addressed in 5363a0e15:
- **`enabledFlag` doc comment** — fixed ("Enabled reports…" → "enabledFlag reports…").
- **`smartScreen()` field on `windows`** — leaving this out intentionally, for consistency with the other top-level security resources (`windows.defender`, `windows.exploitProtection`, `windows.bitlocker`, `windows.firewall`, `windows.update`), which are all reached directly as `windows.<x>` rather than as fields on the `windows {}` grouping. Happy to revisit if we decide to expose all of them under `windows {}` together.
- Also expanded tests (mixed per-surface config, malformed JSON, `enabledFlag` helper).

PR #8559, opened 2026-06-20
🌟 Add windows.exploitProtection resource for system-wide exploit mitigations

tas50:
## Summary

Adds a **`windows.exploitProtection`** resource exposing the system-wide Windows Exploit Protection (the EMET successor) configuration from `Get-ProcessMitigation -System`, plus the `DisallowExploitProtectionOverride` policy value. This is a separate feature from Defender Antivirus and SmartScreen, so it gets its own resource.

### Structure
| Accessor | Backing |
|---|---|
| `available` | whether `Get-ProcessMitigation` exists on the target |
| `disallowOverride` | `…\Windows Defender Security Center\App and Browser protection:DisallowExploitProtectionOverride` |
| `dep` | `.Dep` → `enable`, `emulateAtlThunks` |
| `aslr` | `.Aslr` → `bottomUp`, `forceRelocateImages`, `highEntropy`, `requireInfo` |
| `controlFlowGuard` | `.Cfg` → `enable`, `suppressExports`, `strictControlFlowGuard` |
| `sehop` | `.Sehop` → `enable`, `telemetryOnly` |
| `heap` | `.Heap` → `terminateOnError` |

Mitigation values are `"ON"`, `"OFF"`, or `"NOTSET"`.

### Reliability / non-English systems
- The `ON/OFF/NOTSET` values are an enum that **serializes inconsistently** across PowerShell versions (integer vs. name), so the script casts each to a string. The enum names are **locale-invariant** — the resource behaves identically on non-English Windows.
- Cmdlet availability is detected with **`Get-Command`** (not by matching a localized error string). When the ProcessMitigations module is absent, the mitigation accessors resolve to **null** while `disallowOverride` still resolves from the registry.
- `disallowOverride` is pointer-backed so an unconfigured value is distinguishable from an explicit `0`.

### Why
Enables the enterprise Windows policies to express the exploit-protection and `DisallowExploitProtectionOverride` CIS/STIG controls through a typed resource instead of raw registry reads. (`DisallowExploitProtectionOverride` was the only Defender-area check that couldn't move to `windows.defender`.)

### Tests
`ParseExploitProtectionSettings` is covered for the configured system case, the cmdlet-unavailable case, and explicit-zero override.


tas50:
Addressed in a324868a9: `GetExploitProtectionSettings` now checks the `RunCommand` exit status and surfaces a non-zero result (execution policy, access denied, etc.) as a clear `failed to query exploit protection: <stderr>` error instead of letting empty/error-text stdout become an opaque JSON-unmarshal failure. The exit-status handling is factored into a pure `exploitProtectionResult` helper with tests covering: stderr surfaced, status-only fallback when stderr is empty, the exit code winning even when stdout looks like JSON, and partial/malformed JSON.

PR #8562, opened 2026-06-20
Harden windows.defender availability detection for non-English systems

tas50:
## Summary

Follow-up to #8555. Makes the `windows.defender` "is Defender installed?" detection reliable on **non-English Windows**.

## Problem

When the Defender cmdlets are absent (e.g. a server running third-party AV), PowerShell writes a *localized* "command not recognized" error to stderr. The original detection matched English phrases (`"is not recognized"`), which don't appear on German/French/Japanese/etc. hosts. It had partial fallbacks (the `.NET` exception type and the cmdlet name), but those depend on stderr formatting and aren't guaranteed.

## Fix

Each Defender script is now wrapped with a **`Get-Command` guard** that exits with a fixed **sentinel code (`2`)** when the cmdlet is missing — *before* any localized text is produced:

```powershell
if (-not (Get-Command Get-MpComputerStatus -ErrorAction SilentlyContinue)) { exit 2 }; Get-MpComputerStatus | ConvertTo-Json -Compress
```

`runDefenderCommand` treats that exit code as the primary, **language-independent** `ErrDefenderUnavailable` signal. The stderr heuristic is kept only as a secondary fallback.

## Tests

New `TestDefenderUnavailable` and `TestDefenderScriptsHaveAvailabilityGuard` cover:
- sentinel exit code with no stderr → unavailable
- sentinel exit code with **fully localized German stderr the heuristic cannot match** → unavailable (the key non-English case)
- English stderr fallback without the sentinel → unavailable
- genuine errors (`Access is denied` / `Zugriff verweigert`) on a non-sentinel code → **not** misclassified
- every script carries the `Get-Command` guard and `exit 2`

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -75,6 +75,9 @@ reciever = "reciever"
 # that must stay verbatim.
 yor = "yor"
 yoman = "yoman"
+# German word ("as") in a localized PowerShell error string used to test
+# non-English Defender availability detection.
+als = "als"
 # Pre-existing misspellings in MQL schema fields / exported Go identifiers.
 # Renaming these would break the schema or API, so they are baselined.
 appliable = "appliable"
--- a/providers/os/resources/windows/defender.go
+++ b/providers/os/resources/windows/defender.go
@@ -38,8 +38,10 @@ import (
 // https://learn.microsoft.com/en-us/powershell/module/defender/
 
 // defenderUnavailableExitCode is the process exit code our scripts use to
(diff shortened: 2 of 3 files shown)


tas50:
Addressed in 5b74c82ee: changed the sentinel from `2` to a distinctive `200` (with a comment explaining why) to avoid colliding with common PowerShell/Windows exit codes. Test updated to assert `exit 200`.

PR #8565, opened 2026-06-20
🌟 Add windows.telemetry resource for diagnostic-data and consumer cloud-content policy

tas50:
## Summary

Adds a `windows.telemetry` MQL resource exposing the Windows diagnostic-data (telemetry) and consumer cloud-content Group Policy settings, so the CIS Windows privacy/telemetry checks can move off raw `registrykey.property(...)` lookups.

- `DataCollection` values (`HKLM\SOFTWARE\Policies\Microsoft\Windows\DataCollection`) live directly on `windows.telemetry`.
- `CloudContent` values (`HKLM\SOFTWARE\Policies\Microsoft\Windows\CloudContent`) are grouped under the `windows.telemetry.consumerContent()` sub-resource for clarity.
- Every field is a **nullable int**, so an absent value (policy *not configured*) stays distinguishable from an explicit `0` — a security-relevant distinction for these controls (e.g. `AllowTelemetry` `0` = Security is a compliant value, not "unset").

## Before → After

| Setting | Before (raw registry) | After |
|---|---|---|
| `AllowTelemetry` | `registrykey('...\DataCollection').property('AllowTelemetry')['value']` | `windows.telemetry.allowTelemetry` |
| `DisableEnterpriseAuthProxy` | raw property lookup | `windows.telemetry.disableEnterpriseAuthProxy` |
| `DisableOneSettingsDownloads` | raw property lookup | `windows.telemetry.disableOneSettingsDownloads` |
| `DoNotShowFeedbackNotifications` | raw property lookup | `windows.telemetry.doNotShowFeedbackNotifications` |
| `EnableOneSettingsAuditing` | raw property lookup | `windows.telemetry.enableOneSettingsAuditing` |
| `LimitDiagnosticLogCollection` | raw property lookup | `windows.telemetry.limitDiagnosticLogCollection` |
| `LimitDumpCollection` | raw property lookup | `windows.telemetry.limitDumpCollection` |
| `DisableCloudOptimizedContent` | `registrykey('...\CloudContent').property(...)` | `windows.telemetry.consumerContent.disableCloudOptimizedContent` |
| `DisableConsumerAccountStateContent` | raw property lookup | `windows.telemetry.consumerContent.disableConsumerAccountStateContent` |
| `DisableWindowsConsumerFeatures` | raw property lookup | `windows.telemetry.consumerContent.disableWindowsConsumerFeatures` |

`AllowTelemetry` levels: `0`=Security, `1`=Basic/Required, `2`=Enhanced, `3`=Full.

## Design notes

- **Nullable correctness.** Extraction goes through `regIntPtr`, which returns `*int64` — `nil` when the value name is absent, a pointer (including to `0`) when present. Sub-resource args are built with `llx.IntDataPtr`, which yields a null field for `nil`.
- **Pure functions for testing.** `computeTelemetry(items)` and `computeConsumerContent(items)` take a `map[string]registry.RegistryKeyItem` and do the extraction with no live registry, so they are fully unit-testable.
- **Registry-only.** Reads via the `registrykey` resource using `getEntries()`; a missing key (`codes.NotFound`) is treated as "not configured" (all-null) rather than an error, following the `windows_winrm.go` / `windows_update.go` conventions. Works on connections that cannot run commands.
- **Sub-resource grouping.** CloudContent is grouped under `consumerContent` because it is a distinct policy key with a distinct purpose (consumer experiences vs. diagnostic-data collection); the flat DataCollection fields stay on the parent.

## Testing

- `./lr go providers/os/resources/os.lr --dist providers/os/dist` — no fatal/error
- `./lr versions providers/os/resources/os.lr` — OK
- `cd providers/os && go run ./gen/main.go .` — OK
- `cd providers/os && go build ./...` — OK
- `cd providers/os && go vet ./resources/` — clean
- `cd providers/os && go test ./resources/ -run 'Telemetry|RegIntPtr|ConsumerContent' -v` — all pass (`TestRegIntPtr`, `TestComputeTelemetry`, `TestComputeConsumerContent`, covering absent / explicit-0 / set / partial cases)
- `gofmt -l` on both new Go files — clean

### DIFF
--- a/providers/os/resources/os.lr
+++ b/providers/os/resources/os.lr
@@ -7433,6 +7433,60 @@ private windows.spooler.ipp @defaults("requireIpps") {
   blockCertCNInvalid bool
 }
 
+// Windows diagnostic data (telemetry) and consumer cloud-content policy
+//
+// Examine the effective Windows privacy and telemetry posture read from the
+// Group Policy keys HKLM\SOFTWARE\Policies\Microsoft\Windows\DataCollection and
+// HKLM\SOFTWARE\Policies\Microsoft\Windows\CloudContent. These are GPO-only
+// DWORD values; every field is nullable so that an absent value (the policy is
+// not configured) is distinguishable from an explicit 0. Read from the
+// registry, so it is available even on connections that cannot run commands.
+windows.telemetry {
+  // Diagnostic data level (DataCollection\AllowTelemetry): 0=Security, 1=Basic/Required, 2=Enhanced, 3=Full. Null when not configured.
+  allowTelemetry() int
+  // Whether the telemetry service may not use an authenticated proxy (DataCollection\DisableEnterpriseAuthProxy)
+  //
+  // Applies to the Connected User Experiences and Telemetry service. Null when
+  // the policy is not configured.
+  disableEnterpriseAuthProxy() int
+  // Whether downloading configuration from the OneSettings service is disabled (DataCollection\DisableOneSettingsDownloads). Null when not configured.
+  disableOneSettingsDownloads() int
+  // Whether feedback notifications are suppressed (DataCollection\DoNotShowFeedbackNotifications)
+  //
+  // When set, the user does not receive telemetry feedback notifications. Null
+  // when the policy is not configured.
+  doNotShowFeedbackNotifications() int
+  // Whether auditing of OneSettings usage is enabled (DataCollection\EnableOneSettingsAuditing). Null when not configured.
+  enableOneSettingsAuditing() int
+  // Whether collection of diagnostic logs is limited (DataCollection\LimitDiagnosticLogCollection). Null when not configured.
+  limitDiagnosticLogCollection() int
+  // Whether collection of crash dumps is limited (DataCollection\LimitDumpCollection). Null when not configured.
+  limitDumpCollection() int
+  // Consumer cloud-content (CloudContent) policy settings
+  consumerContent() windows.telemetry.consumerContent
+}
+
+// Windows consumer cloud-content policy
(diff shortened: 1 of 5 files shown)


tas50:
Rebased onto main. Addressed review:
- **Critical (fields returned 0):** the seven DataCollection fields are now computed methods (`allowTelemetry() int`, etc.) in os.lr, so codegen wraps them in `GetOrCompute` and invokes the Go accessors. Implemented with the established `populate()` pattern (same as windows.lsa): each accessor delegates to a single `populate()` that reads the key once and sets every field's TValue directly, preserving nil-vs-0 (null) semantics.
- **Caching warning:** `populate()` runs exactly once (subsequent accessors see the fields already set), so the registry items are parsed a single time across all seven fields with one error path — no repeated `computeTelemetry` passes.
- **consumerContent sub-resource:** kept as-is. It groups values from a distinct registry key (`CloudContent`) separate from the parent's `DataCollection` key, which is the structural rationale; the doc-comment documents this. Happy to flatten it into the parent if you'd prefer.

tas50:
Update: rather than keep the sub-resource, I **flattened** `consumerContent` into `windows.telemetry` per the project's sub-resource guidelines (three scalar CloudContent DWORDs with a synthetic `__id` don't meet the bar for a sub-resource). The three values are now `disableCloudOptimizedContent`, `disableConsumerAccountStateContent`, and `disableWindowsConsumerFeatures` directly on `windows.telemetry`; `populate()` reads both the DataCollection and CloudContent keys once and sets all ten fields. (Supersedes my earlier comment about keeping it grouped.)

PR #8594, opened 2026-06-22
✨ k8s: effective-access RBAC (k8s.rbacSubjects + k8s.rbac.whoCan)

tas50:
Adds cluster-wide RBAC subject reachability to the k8s provider so "who can do X" is a single query instead of a manual join across roles and bindings. This is the effective-access theme: the biggest "simpler queries" win for RBAC, and it works on every cluster (live or manifest).

## What it adds

**`k8s.rbacSubjects`** — every distinct `User`, `Group`, and `ServiceAccount` named in any RoleBinding or ClusterRoleBinding, deduplicated, each carrying the effective permissions folded across all bound roles:
- `isClusterAdmin`, `canEscalatePrivileges`, `canReadSecrets`, `hasWildcardPermissions`
- `roleBindings` / `clusterRoleBindings` that grant the subject
- `serviceAccount` resolving a ServiceAccount subject to its object (null for users/groups)

**`k8s.rbac.whoCan(verb, resource, group, namespace, name)`** — the subjects whose bound roles allow a given action. It honors the namespace-scope difference between ClusterRoleBindings (cluster-wide) and RoleBindings (per-namespace); an empty `namespace` considers every namespace.

## Design

Resolution uses the local RBAC graph via the existing `rbac.go` helpers (`rbacRuleGrants`, `rbacGrantsClusterAdmin`, etc.), so it works on both live clusters and manifest scans with no per-subject API calls. It generalizes the per-ServiceAccount predicates that already existed to all subject kinds and adds a flat, cluster-wide enumeration that was previously impossible to express in one query. No new API permissions are required (it reuses the existing rolebinding/clusterrolebinding list calls).

It reflects RBAC grants only and does not account for the node or webhook authorizers; `k8s.accessReview` remains the authoritative per-subject decision from the API server. The two pair naturally: `whoCan` narrows the candidate set locally, `accessReview` confirms authoritatively.

## Example queries

```mql
# Any human/group cluster-admins (not service accounts)
k8s.rbacSubjects.where(isClusterAdmin && kind != "ServiceAccount") { kind name }

# Who can read secrets anywhere in the cluster?
k8s.rbac.whoCan(verb: "list", resource: "secrets").subjects { kind name namespace }

# Who can escalate via binding roles?
k8s.rbac.whoCan(verb: "bind", resource: "roles", group: "rbac.authorization.k8s.io").subjects { kind name }
```

## Testing

- Unit tests (`rbac_subject_test.go`) over a dedicated manifest fixture covering Users, Groups, ServiceAccounts, a subject named in two bindings (dedup), and the CRB-vs-RB namespace-scoping cases.
- Verified end-to-end through the real executor with `mql run k8s <manifest>` (enumeration, the `whoCan(...)` constructor path, and `.where(...)` filtering). A live-minikube smoke test was blocked only by Docker not running locally; the manifest path exercises the identical resource code.

`.lr.versions` entries are at `13.3.1` (the generator's `IncPatch` of the current `13.3.0` in `config.go`); no provider version bump.


tas50:
Addressed in 8e75fc598. `k8s.rbac.whoCan` now honors the `name` (object) filter: `rbacRuleGrants` takes a `resourceName` and checks `rule.ResourceNames` — a non-empty list matches only those named objects, an empty list matches all, and an unscoped query (empty `name`) matches any rule, mirroring the API server's RBAC authorizer and `k8s.accessReview`. Added `TestRbacWhoCan_ResourceName` covering the unscoped / matching-name / non-matching-name cases, verified end-to-end against the manifest fixture.

PR #8598, opened 2026-06-22
✨ k8s: exposure→workload attack paths + secret hygiene predicates

tas50:
Builds on the merged effective-access RBAC work (#8594) by adding the missing **network edge**: the ability to traverse from an internet-exposed Service, Ingress, or Gateway to the workloads behind it. Combined with a few secret-hygiene predicates, this turns separate signals (network posture, RBAC, workload hardening) into single-line attack-path queries.

## Exposure → workload edges

Reuse the existing `podsMatchingSelector` helper that already powers `deployment.pods`, etc.:

- **`k8s.service.pods`** — pods selected by the service's label selector (selectorless and ExternalName services select nothing)
- **`k8s.pod.services`** — reverse: the services whose selector matches the pod
- **`k8s.ingress.pods`** — pods behind the ingress, via its default backend and rule backend services
- **`k8s.gateway.pods`** — pods behind the gateway, via the HTTPRoutes and GRPCRoutes attached to it and their backend services (L4 TCP/TLS/UDP routes are not folded in)
- **`k8s.networkExposure.pods`** — resolves the exposure's source (Service, Ingress, or Gateway) to the backing pods; HBN/non-Kubernetes sources resolve to empty

## Secret hygiene (`k8s.secret`)

Building on the existing `type` + `usedBy`:

- **`isServiceAccountToken`** — `type == kubernetes.io/service-account-token`
- **`isImagePullSecret`** — `dockercfg` / `dockerconfigjson`
- **`isUnused`** — no pod **and** no ServiceAccount references the secret (the SA check keeps mounted token secrets from being flagged as orphaned)

## Why

The network source objects already linked *to* exposures, but there was no typed traversal from an exposure (or a Service) *to* the pods behind it — `sourceRef` was just a string and Service selectors lived only inside the `spec` dict. These edges close that gap.

```mql
# internet-exposed workloads that run privileged
k8s.networkExposures.where(internetExposed).all(pods.none(runsPrivileged))

# LoadBalancer services fronting a cluster-admin ServiceAccount
k8s.services.where(type == "LoadBalancer").pods { serviceAccount.isClusterAdmin }

# orphaned credentials
k8s.secrets.where(isUnused)
```

## Testing

New `testdata/exposure_targets.yaml` fixture and tests cover selector matching, namespace scoping, selectorless services, ingress/gateway/route resolution with dedup across HTTPRoute + GRPCRoute, the `networkExposure` dispatch (Service-sourced public LoadBalancer resolved end-to-end), and the three secret predicates. Verified end-to-end through the real executor with `mql run k8s <manifest>` for every edge, including the headline `networkExposures.where(internetExposed) { pods { serviceAccountName } }` attack-path query.

No schema-permission changes (all source lists were already required); `.lr.versions` entries at `13.3.1`; no provider version bump.

### DIFF
--- a/providers/k8s/resources/exposure_targets.go
+++ b/providers/k8s/resources/exposure_targets.go
@@ -1,0 +1,297 @@
+// Copyright Mondoo, Inc. 2024, 2026
+// SPDX-License-Identifier: BUSL-1.1
+
+package resources
+
+import (
+	"go.mondoo.com/mql/v13/llx"
+	"go.mondoo.com/mql/v13/providers-sdk/v1/plugin"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+// serviceKey identifies a Service by namespace and name, used to deduplicate
+// backend references when resolving ingress and gateway routing targets.
+type serviceKey struct {
+	namespace string
+	name      string
+}
+
+func k8sCluster(runtime *plugin.Runtime) (*mqlK8s, error) {
+	o, err := CreateResource(runtime, "k8s", map[string]*llx.RawData{})
+	if err != nil {
+		return nil, err
+	}
+	return o.(*mqlK8s), nil
+}
+
+// pods returns the pods selected by the service's label selector. Selectorless
+// services (including ExternalName) select no pods.
+func (k *mqlK8sService) pods() ([]any, error) {
+	svc, err := k.getService()
+	if err != nil {
+		return nil, err
+	}
+	if len(svc.Spec.Selector) == 0 {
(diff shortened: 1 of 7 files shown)


tas50:
Addressed in 3174ffb37:

- **Type assertions** — added comma-ok guards to the `*mqlK8sService`/`*mqlK8sIngress`/`*mqlK8sGateway` assertions in `networkExposure.pods` and the `*mqlK8sPod` assertion in `podsForServiceKeys`, matching the guarded pattern already used elsewhere in the file.
- **Versions** — bumped the new fields to `13.3.3`. Good catch: `13.3.1` (the generator's `IncPatch` of `config.go` `13.3.0`) would have put `k8s.networkExposure.pods` below its own resource `k8s.networkExposure` (`13.3.2`), implying the field predates the resource. `13.3.3` is the next unused patch and matches merge order. Confirmed a regen preserves it.

PR #8601, opened 2026-06-22
✨ k8s: host-network/hostPort ingress exposure + egress blind spot

tas50:
Closes the ingress/egress coverage gap for pods that bind directly to the node network — the one exposure vector `networkExposures()` did not capture. It iterated Services, Ingresses, Gateways, and HBN sources, but **never pods**, so a `hostNetwork` or `hostPort` pod (CNI agents, ingress-controller daemonsets, monitoring) was a silent blind spot even though it can be reachable straight from the node IP.

## Ingress: host-level pod exposure

Pods with `hostNetwork` or any `hostPort` now produce a `k8s.networkExposure` with `sourceKind: "Pod"`:

- Reachability mirrors the existing NodePort model: the pod is `internetExposed` when **the node it runs on** has a public address (resolved per-node, not cluster-wide, to avoid over-reporting).
- `hostNetwork` promotes every `containerPort` to a node port; explicit `hostPort`s are captured for any pod.
- Classification reasons: `hostPortPublicNode` / `hostPortPrivateNode` / `hostPortNodeUnknown` (and `hostNetwork*`), with `medium` confidence for unscheduled pods.
- `networkExposure.pods` resolves a `Pod`-sourced exposure back to the pod, so `pod.exposures` (from #8599) links both directions automatically.

## Egress blind spot

- **`k8s.pod.escapesNetworkPolicy`** — true for host-network pods, whose ingress *and* egress traffic Kubernetes NetworkPolicy cannot govern. A default-deny posture has a hole for exactly these pods, and this makes them a one-line query.

## How we know it is 100%

`TestNetworkExposureCoverage` is a **completeness guard**, not just a unit test: the fixture contains one instance of every modeled ingress vector (LoadBalancer, NodePort, externalIPs, Ingress, Gateway, hostPort, hostNetwork), and the test asserts each produces an exposure record. If a future change drops a vector — or a new exposure-capable field is added without wiring it into `networkExposures()` — the missing `sourceRef` fails CI.

## Testing (extensive, per request)

- **Completeness guard** over all ingress vectors.
- **Classification** across public / private / unscheduled nodes (internet-exposed vs not, exposure reason, addresses, confidence).
- **Round-trip**: `networkExposure.pods` ↔ `pod.exposures` for a Pod-sourced exposure.
- **`escapesNetworkPolicy`** across hostNetwork / hostPort / ordinary pods.
- **Pure-Go unit tests** for the `hostExposurePorts` and `podHostExposureArgs` helpers (init-container hostPorts, protocol defaulting, hostNetwork-with-no-ports still exposed, plain pod produces nothing).
- Verified end-to-end with `mql run k8s <fixture>`.

`.lr.versions` at `13.3.5` (next unused patch above the in-flight `13.3.4`); no new API permissions (reuses the already-required pod/node reads); no provider version bump.

```mql
# every internet-reachable workload, now including host-network/hostPort pods
k8s.networkExposures.where(internetExposed) { sourceKind name }

# host-network pods that escape NetworkPolicy
k8s.pods.where(escapesNetworkPolicy) { name namespace }
```

### DIFF
--- a/providers/k8s/resources/exposure_targets.go
+++ b/providers/k8s/resources/exposure_targets.go
@@ -245,6 +245,21 @@ func (k *mqlK8sNetworkExposure) pods() ([]any, error) {
 				return gw.pods()
 			}
 		}
+	case "Pod":
+		// Host-network and hostPort exposures are sourced from the pod itself.
+		k8sPods := cluster.GetPods()
+		if k8sPods.Error != nil {
+			return nil, k8sPods.Error
+		}
+		for i := range k8sPods.Data {
+			p, ok := k8sPods.Data[i].(*mqlK8sPod)
+			if !ok {
+				continue
+			}
+			if p.Namespace.Data == ns && p.Name.Data == name {
+				return []any{p}, nil
+			}
+		}
 	}
 
 	// HBN and other non-Kubernetes sources have no in-cluster workload to
--- a/providers/k8s/resources/host_network_exposure.go
+++ b/providers/k8s/resources/host_network_exposure.go
@@ -1,0 +1,167 @@
+// Copyright Mondoo, Inc. 2024, 2026
+// SPDX-License-Identifier: BUSL-1.1
+
+package resources
+
+import (
+	"fmt"
+
+	"go.mondoo.com/mql/v13/llx"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// escapesNetworkPolicy reports whether the pod uses host networking, in which
+// case Kubernetes NetworkPolicy (both ingress and egress) does not apply to its
+// traffic. Such pods are a coverage gap for any default-deny posture.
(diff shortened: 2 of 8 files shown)


tas50:
Rebased on main and addressed:

- **Version** — main now includes the `13.4.0` release (#8600), so the new `k8s.pod.escapesNetworkPolicy` field is re-versioned `13.4.1` (next patch after the release), replacing the pre-release `13.3.5`.
- **Unscheduled classification** (review suggestion) — the `NodeName == ""` branch now sets `networkClassifications` to `["unknown"]` instead of letting `classifyAddresses([])` report `["internalOnly"]`, which was misleading for a pod whose node is not yet known. Added an assertion in `TestHostExposureClassification` to cover it.

Full suite green after the rebase (new deps from #8595 included); no go.mod/go.sum churn.

PR #8660, opened 2026-06-23
🐛 packages: drop empty-name appx entries instead of leaving holes

tas50:
## Bug

`ParseWindowsAppxPackages` in `windows_packages.go`:

```go
pkgs := make([]Package, len(appxPackages))
for i, p := range appxPackages {
    if p.Name == "" {
        continue
    }
    pkgs[i] = p.toPackage(platform)
}
return pkgs, nil
```

The slice is pre-sized to `len(appxPackages)` and written by index `i`. When an entry has an empty `Name`, the loop `continue`s but never assigns `pkgs[i]`, leaving a zero-value `Package{}` hole. Any `Get-AppxPackage` entry with a missing `Name` produces a bogus empty package (no name/version/purl) in the output / SBOM instead of being omitted.

The sibling `ParseWindowsAppPackages` does this correctly with `[]Package{}` + `append`.

## Fix

Use the append pattern so skipped entries are truly omitted. Added a regression test with an empty-name entry between two valid ones.

https://claude.ai/code/session_015U1ocAxfcBhVYi3f9JnyZZ

---
_Generated by [Claude Code](https://claude.ai/code/session_015U1ocAxfcBhVYi3f9JnyZZ)_

### DIFF
--- a/providers/os/resources/packages/windows_packages.go
+++ b/providers/os/resources/packages/windows_packages.go
@@ -164,7 +164,7 @@ func ParseWindowsAppxPackages(platform *inventory.Platform, input io.Reader) ([]
 		return nil, err
 	}
 
-	pkgs := []Package{}
+	pkgs := make([]Package, 0, len(appxPackages))
 	for _, p := range appxPackages {
 		if p.Name == "" {
 			continue
--- a/providers/os/resources/packages/windows_packages.go
+++ b/providers/os/resources/packages/windows_packages.go
@@ -164,12 +164,12 @@ func ParseWindowsAppxPackages(platform *inventory.Platform, input io.Reader) ([]
 		return nil, err
 	}
 
-	pkgs := make([]Package, len(appxPackages))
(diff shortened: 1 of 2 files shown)


providers/os/resources/packages/windows_packages.go
```diff
@@ -164,12 +164,12 @@ func ParseWindowsAppxPackages(platform *inventory.Platform, input io.Reader) ([]
 		return nil, err
 	}
 
-	pkgs := make([]Package, len(appxPackages))
```
czunker, line 167: We should keep the preallocation.

IIRC, this should work with append:
```
pkgs := make([]Package, 0, len(appxPackages))
```

tas50: Good call, done — switched to `make([]Package, 0, len(appxPackages))` so we keep the preallocation while appending. Also rebased onto latest main.

czunker (approved):

PR #8677, opened 2026-06-23
🐛 aws: fix cloudtrail trail lookup by name (used S3 ARN pattern)

tas50:
## Bug

`aws_cloudtrail.go`, `initAwsCloudtrailTrail` (name-only lookup path):

```go
nameVal := args["name"].Value.(string)
arn = fmt.Sprintf(s3ArnPattern, nameVal)   // s3ArnPattern = "arn:aws:s3:::%s"
...
if trail.Arn.Data == arn { return args, trail, nil }
```

It synthesizes the lookup key from the trail name using `s3ArnPattern` — an **S3 bucket** ARN (the function was cloned from the S3 bucket init; note the stale `// load all s3 buckets` comment). It then compares that S3-shaped string against each trail's real CloudTrail ARN (`arn:aws:cloudtrail:<region>:<account>:trail/<name>`), which can never match. So `aws.cloudtrail.trail(name: "my-trail")` always fails with "cloudtrail trail does not exist". (The common path supplying both `arn` and `name` is unaffected.)

## Fix

A CloudTrail ARN can't be synthesized from the name alone (it needs region + account), so match the trail's `name` field directly when no ARN is supplied, and match by ARN when one is. Also corrected the stale comment.

https://claude.ai/code/session_015U1ocAxfcBhVYi3f9JnyZZ

---
_Generated by [Claude Code](https://claude.ai/code/session_015U1ocAxfcBhVYi3f9JnyZZ)_

### DIFF
--- a/providers/aws/resources/aws_cloudtrail.go
+++ b/providers/aws/resources/aws_cloudtrail.go
@@ -60,53 +60,59 @@ func initAwsCloudtrailTrail(runtime *plugin.Runtime, args map[string]*llx.RawDat
 		return nil, nil, errors.New("arn or name required to fetch aws cloudtrail trail")
 	}
 
-	// construct arn of cloudtrail if missing
-	var arnVal string
+	// We match the requested trail by ARN when supplied, otherwise by name.
+	// A CloudTrail ARN cannot be synthesized from the name alone (it needs the
+	// region and account), so never fabricate one — match the name directly.
+	var arnVal, nameVal string
 	if args["arn"] != nil {
 		arnVal = args["arn"].Value.(string)
-	} else {
-		nameVal := args["name"].Value.(string)
-		arnVal = fmt.Sprintf(s3ArnPattern, nameVal)
+	}
+	if args["name"] != nil {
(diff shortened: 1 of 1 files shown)


tas50:
Rebased onto latest `main`. Heads-up on a **semantic conflict** during the rebase worth a reviewer's eye:

While this PR was open, a perf-optimization landed on `main` (targeted single-resource init, #8749) that **reintroduced the exact bug this PR fixes** — it synthesized a CloudTrail ARN from the trail name using `s3ArnPattern` (`arn:aws:s3:::%s`) and then matched on that fabricated ARN.

I resolved it to keep **both** behaviors rather than picking a side:

- **Real ARN supplied** → keep `main`'s targeted lookup (derive region from the ARN, single `GetTrail` instead of describing every trail in every region).
- **Name only** → apply this PR's fix: never fabricate an ARN; fall through to the trail scan and match on `trail.Name`.

Net: the optimization is preserved for the ARN path, and the name-lookup bug stays fixed. Builds and `go vet` clean locally; full CI re-running.

PR #8689, opened 2026-06-23
⭐ oci: internet-exposure analysis (instance/load-balancer exposure, ADB internetReachable)

tas50:
## What

Ports the established `<provider>.network.exposure` pattern (aws/gcp/azure/digitalocean/hetzner/openstack/stackit) to the **OCI** provider, which already models every building block (VNIC public IPs, security lists, NSGs, subnets with internet-ingress policy, load balancers, autonomous databases) but exposed no reachability predicate.

## New surface

A shared private `oci.network.exposure` sub-resource:

| field | meaning |
|---|---|
| `internetReachable` | overall verdict |
| `hasPublicIp` | resource has a public IP |
| `securityGroupAllowsIngress` | an attached NSG (or, for compute, an unrestricted subnet with no NSG) admits inbound from any address |
| `openIngressRules` | `[]dict` of the ingress rules / listeners that open it |

Wired onto:

- **`oci.compute.instance.exposure()`** — a VNIC with a public IP, on a subnet that does **not** set `prohibitInternetIngress`, whose attached NSGs have an INGRESS rule admitting `0.0.0.0/0`/`::/0`. An NSG-less VNIC on an unrestricted subnet is treated as open, since OCI's default security list admits SSH from the internet (the "no firewall == open" convention used by the other providers).
- **`oci.loadBalancer.loadBalancer.exposure()`** — not `isPrivate`, carries a public `ipAddresses[].isPublic`, and has ≥1 listener; listeners surface as `openIngressRules` (name/port/protocol).
- **`oci.database.autonomousDatabase.internetReachable()`** — public endpoint (empty `privateEndpointIp`) and either `isAccessControlEnabled == false` or a `whitelistedIps` allow-list that admits any address. (Unlike a security-group set, an *empty* ADB allow-list with access control on denies everyone, so only a `0.0.0.0/0`/`0.0.0.0` entry counts as open.)

## Notes

- Reuses cached VNIC/subnet/load-balancer data. NSG `ingressSecurityRules()` and instance `vnics()` resolve via the existing on-demand resolvers; like the Azure VM `exposure()`, that cost is only paid when the field is queried. **Errors are propagated, not swallowed**, so a permissions failure surfaces rather than reading as "not exposed".
- Pure CIDR/rule helpers (`ociCidrIsAny`, `ociNsgRuleOpensIngress`, `ociWhitelistOpensInternet`) are table-unit-tested in `oci_exposure_test.go` (no mock runtime), matching the hetzner/stackit test style.
- `.lr.versions` entries land at **13.11.4**; provider version bumped to match.

## Follow-up (not in this PR)

- Subnet-level **security lists** aren't directly linked to subnets in the provider model, so compute exposure relies on the subnet's `prohibitInternetIngress` flag + per-VNIC NSGs. A `subnet.securityLists()` resolver (filtered by VCN) would let us fold security-list ingress in too.
- `oci.database.dbSystem` has no explicit public flag; an `internetReachable()` for it would infer from subnet + NSG (lower confidence) — deferred.

## Verification
- `go build ./...` and `go vet ./resources/` clean.
- `go test ./resources/ -run 'Oci(CidrIsAny|NsgRuleOpensIngress|WhitelistOpensInternet)'` → all pass.

### DIFF
--- a/providers/oci/config/config.go
+++ b/providers/oci/config/config.go
@@ -13,7 +13,7 @@ import (
 var Config = plugin.Provider{
 	Name:            "oci",
 	ID:              "go.mondoo.com/cnquery/v9/providers/oci",
-	Version:         "13.11.4",
+	Version:         "13.12.0",
 	ConnectionTypes: []string{provider.ConnectionType},
 	Connectors: []plugin.Connector{
 		{
--- a/providers/oci/config/config.go
+++ b/providers/oci/config/config.go
@@ -13,7 +13,7 @@ import (
 var Config = plugin.Provider{
 	Name:            "oci",
 	ID:              "go.mondoo.com/cnquery/v9/providers/oci",
-	Version:         "13.11.3",
+	Version:         "13.11.4",
 	ConnectionTypes: []string{provider.ConnectionType},
 	Connectors: []plugin.Connector{
 		{
--- a/providers/oci/resources/loadbalancer.go
+++ b/providers/oci/resources/loadbalancer.go
@@ -122,6 +122,7 @@ func (o *mqlOciLoadBalancer) getLoadBalancers(conn *connection.OciConnection, re
 					"shape":                     llx.StringDataPtr(lb.ShapeName),
 					"isPrivate":                 llx.BoolDataPtr(lb.IsPrivate),
 					"ipAddresses":               llx.ArrayData(ipAddresses, types.Dict),
+					"nsgIds":                    llx.ArrayData(convert.SliceAnyToInterface(lb.NetworkSecurityGroupIds), types.String),
 					"isDeleteProtectionEnabled": llx.BoolDataPtr(lb.IsDeleteProtectionEnabled),
 					"state":                     llx.StringData(string(lb.LifecycleState)),
 					"created":                   llx.TimeDataPtr(created),
--- a/providers/oci/resources/oci.lr
+++ b/providers/oci/resources/oci.lr
@@ -426,7 +426,11 @@ private oci.network.exposure @defaults("internetReachable hasPublicIp") {
   internetReachable bool
   // Whether the resource has a public IP
   hasPublicIp bool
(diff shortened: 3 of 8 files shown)


tas50:
Thanks — both load-balancer comments addressed in the latest commit by giving the LB path real NSG analysis (full parity with the instance path), rather than documenting the gap:

- Added `oci.loadBalancer.loadBalancer.nsgIds` + `securityGroups()` (from the SDK's `NetworkSecurityGroupIds`), mirroring `oci.compute.vnic`.
- Factored the NSG ingress scan into a shared `ociCollectOpenNsgRules` helper used by **both** the instance and LB paths.
- `securityGroupAllowsIngress` now reflects actual NSG ingress (no longer `hasListener`). `internetReachable` for an LB now requires **public IP AND a listener AND an NSG set that admits ingress** (or no NSG attached — OCI's default open posture). A public LB fronted by a restrictive NSG now correctly reports `internetReachable = false`.

Helpers stay table-unit-tested; `go build`/`go vet`/tests green. Version of the new fields is 13.12.0.

tas50:
Both addressed in the latest commit:

1. **`securityGroupAllowsIngress` conflation (instances)** — it now reflects the **NSG verdict alone** (`nsgAllows`), so it no longer reads `false` just because the subnet blocks ingress. The subnet's `prohibitInternetIngress` gate is applied only to `internetReachable`. Field doc updated to say so. The LB path was already NSG-only (no subnet gate), so it's unchanged.

2. **Empty-NSG test case** — extracted the pure `ociNsgIngressVerdict([][]map[string]any)` helper out of `ociCollectOpenNsgRules` and added `TestOciNsgIngressVerdict`, which pins: no NSG attached → open; **NSG attached with an empty rule list → closed**; NSG with only specific CIDRs → closed; any-address rule → open; mixed sets. No mock runtime needed.

`go build`/`go vet`/tests green.

PR #8726, opened 2026-06-25
🧹 Broaden permission generation script for AWS.

VasilSirakov:
Some permissions were missed because we weren't crawling through all provider directories. And also, because we weren't considering helper functions that accepted an AWS client as a parameter as opposed to an assignment from the AwsConnection struct.

#### In this PR:
- Crawl through all provider dirs for go files.
- Also consider clients passed in as fn parameters.
- Regenrate aws.permissions.json file.


vjeffrey:
thanks @VasilSirakov looks great!

PR #8740, opened 2026-06-25
✨ depsdev: expose version licenses, links, registries, SLSA provenance, and related projects

tas50:
## What

Expands `depsdev.packageVersion` with five new fields, all sourced from the deps.dev `GetVersion` endpoint through a single shared lazy fetch (one HTTP call populates all five):

| Field | Type | Notes |
|---|---|---|
| `licenses` | `[]string` | **Bug fix** — was always empty before (see below) |
| `links` | `map[string]string` | Keyed by label (`SOURCE_REPO`, `ISSUE_TRACKER`, `ORIGIN`, ...) |
| `registries` | `[]string` | Registries the version is published to |
| `slsaProvenances` | `[]dict` | Build provenance attestations (`sourceRepository`, `commit`, `url`, `verified`) |
| `relatedProjects` | `[]depsdev.relatedProject` | New sub-resource: `relationType`, `relationProvenance`, typed `project()` |

The new `depsdev.relatedProject` resource exists to hold a typed `project()` reference (meeting the nested-typed-ref bar for a sub-resource), so you can traverse from a version straight into the upstream project:

```mql
depsdev.packages {
  name
  versions.where(isDefault).first {
    licenses
    links
    relatedProjects { relationType project { id starsCount } }
  }
}
```

## Bugs fixed

1. **`depsdev.packageVersion.licenses` was always empty.** The field was set to `[]` at creation (`// licenses are not in the package endpoint`) and `GetVersion` — which does return licenses — was never consulted. It is now a computed field backed by the version-detail fetch. Verified live: `github.com/rs/zerolog@v1.31.0` now reports `["MIT"]`.

2. **`depsdev.package.project()` ignored the relation type.** It returned the first related project of any type, so a package whose first relation was an issue tracker would surface that instead of the source repo. It now prefers the `SOURCE_REPO` relation, falling back to the first non-empty entry.

3. **Cache-key collisions from `id()` reading fields set after `CreateResource`** (two resources, same root cause):
   - `depsdev.relatedProject` (new code, caught in review): `CreateResource` computes `__id` via `id()` before the `projectID`/`versionID` Internal fields are assigned.
   - `depsdev.packageVersion` (pre-existing, surfaced while verifying the above): `packageName` is assigned after `CreateResource`, so `id()` produced `depsdev.packageVersion/@<version>` with an empty package name. Any two modules sharing a version string collided — e.g. `cockroachdb/errors` and `rs/zerolog` both have `v1.14.0`, so in a multi-package query `zerolog.versions.where(isDefault).first` silently returned cockroachdb's version object (and `stretchr/testify` came back `null`).

   Both are fixed by passing `__id` explicitly in the `CreateResource` args so the generated code never calls `id()` with unset fields. Note: this class of bug is invisible in single-resource queries (separate queries have separate caches) and only manifests when multiple instances are materialized in one query.

## Not changed (deliberately)

- `slsaProvenances` is modeled as `[]dict` rather than a sub-resource: it has no natural ID, and its only typed-ref candidate (`sourceRepository`) is redundant with `relatedProjects`. It's also empty in practice for Go modules (the provider is Go-only today). Revisit if/when the provider covers npm/PyPI, where provenance is populated.
- `depsdev.project.archived()` keeps a manual double-check that is effectively dead code — `GetOrCompute` already caches and honors the proactive `StateIsNull` for non-GitHub projects. Harmless; left to avoid churn.
- Larger missing data (not bugs, future work): per-version `isDeprecated`/`deprecatedReason`, vulnerability `advisoryKeys`, the resolved dependency graph, CAPSLOCK capabilities.

## Testing

- `go build ./...`, `go vet`, and `go test ./...` pass for the provider.
- Generated code (`.lr.go`, `.lr.versions`) regenerated with `mqlr generate`; new `.lr.versions` entries are at `13.0.20` (next patch after the provider's current `13.0.19`). The changed `licenses` entry keeps its original version (stored → computed is not a version bump).
- Verified end-to-end against the live deps.dev API with the installed `mql` binary. A single query over all `go.mod` deps now resolves each package to its own data: `zerolog → rs/zerolog` (v1.35.1), `testify → stretchr/testify` (v1.11.1), `mql → mondoohq/cnquery`, instead of collapsing to the first package. `licenses`, `links`, `registries`, `slsaProvenances`, and `relatedProjects` all populate, and `relatedProject.project()` resolves to a real `depsdev.project`.

### DIFF
--- a/providers/depsdev/resources/api.go
+++ b/providers/depsdev/resources/api.go
@@ -43,9 +43,20 @@ type depsDevVersionResponse struct {
 		Name    string `json:"name"`
 		Version string `json:"version"`
 	} `json:"versionKey"`
-	IsDefault       bool      `json:"isDefault"`
-	PublishedAt     time.Time `json:"publishedAt"`
-	Licenses        []string  `json:"licenses"`
+	IsDefault   bool      `json:"isDefault"`
+	PublishedAt time.Time `json:"publishedAt"`
+	Licenses    []string  `json:"licenses"`
+	Links       []struct {
+		Label string `json:"label"`
+		URL   string `json:"url"`
+	} `json:"links"`
+	Registries      []string `json:"registries"`
+	SlsaProvenances []struct {
+		SourceRepository string `json:"sourceRepository"`
+		Commit           string `json:"commit"`
+		URL              string `json:"url"`
+		Verified         bool   `json:"verified"`
+	} `json:"slsaProvenances"`
 	RelatedProjects []struct {
 		ProjectKey struct {
 			ID string `json:"id"`
--- a/providers/depsdev/resources/depsdev.lr
+++ b/providers/depsdev/resources/depsdev.lr
@@ -46,7 +46,28 @@ private depsdev.packageVersion @defaults("version publishedAt") {
   // Whether this is the default version
   isDefault bool
   // SPDX license identifiers
-  licenses []string
+  licenses() []string
+  // Source and metadata links keyed by label (e.g. SOURCE_REPO, ISSUE_TRACKER, ORIGIN)
+  links() map[string]string
+  // Package registries where this version is published
+  registries() []string
+  // SLSA build provenance attestations
+  //
+  // Each entry records the sourceRepository, commit, url, and verified
+  // flag of a build provenance attestation recorded for this version.
+  slsaProvenances() []dict
+  // Projects related to this version (e.g. source repository, issue tracker)
+  relatedProjects() []depsdev.relatedProject
+}
+
+// deps.dev related project
+private depsdev.relatedProject @defaults("relationType") {
+  // Relationship type (e.g. SOURCE_REPO, ISSUE_TRACKER, ORIGIN)
+  relationType string
+  // How the relationship was determined (e.g. GO_ORIGIN, SLSA_PROVENANCE)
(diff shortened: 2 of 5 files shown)


tas50:
Both review comments addressed in 688ac4cbf:

- **🔴 `relatedProject` cache key** — now passes `__id` explicitly in `CreateResource` (exactly as suggested), so `id()` is never called before the Internal fields are set.
- **🟡 doc-comment** — retitled to `// deps.dev package version related project`.

Verifying that fix surfaced a **pre-existing bug of the same class on `depsdev.packageVersion`**: its `packageName` is also assigned after `CreateResource`, so `id()` produced `depsdev.packageVersion/@<version>` with an empty package name. Any two modules sharing a version string collided — e.g. `cockroachdb/errors` and `rs/zerolog` both have `v1.14.0`, so `zerolog.versions.where(isDefault).first` silently returned cockroachdb's version object. Fixed the same way (explicit `__id`).

Verified across all `go.mod` deps in a single query: each package now resolves to its own project (`zerolog → rs/zerolog`, `mql → mondoohq/cnquery`) instead of collapsing to the first.

czunker (approved):
Thanks @tas50

PR #8745, opened 2026-06-26
🐛 avoid incorrect error-propagation on empty rawdata

arlimus:
1. Empty rawdata due to the way the conversion works was getting returned and then culled by the array-handling. This caused erroneous behavior in e.g. @context blocks to be culled from the results. This change makes it loud (with error) while retaining the nil-behavior that this entry has for the user.
2. The issue that sparked this was the incorrect handling of @context data on failures, i.e. the positioning of the codeblock in the compiler. fixed

### DIFF
--- a/llx/data_conversions.go
+++ b/llx/data_conversions.go
@@ -11,6 +11,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/rs/zerolog/log"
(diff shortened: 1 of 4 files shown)


arlimus:
Fixed the initial broad nature and adjusted the title/desc

PR #8800, opened 2026-06-30
Gcp storage tag

gerysz:
**✨ GCP: add `tags()` to storage buckets (Resource Manager Tags)**

## What

Adds a `tags()` field to `gcp.project.storageService.bucket` that returns the GCP Resource Manager Tags bound to a bucket, including tags inherited from parent resources (project, folder, organization).

```mql
gcp.project.storageService.bucket(name: "my-bucket") {
  tags
  tags.keys.any(_ == /glz-resource-optout/)
}
```

## How

GCS buckets require the **regional Cloud Resource Manager endpoint** (`https://{location}-cloudresourcemanager.googleapis.com/`) — the global endpoint rejects bucket resource names. The implementation calls `EffectiveTagBindingCollections.Get` (not `TagBindings.List`) so that tags inherited from parent folders and the organization are included, not just directly-attached bindings.

Resource name format used: `//storage.googleapis.com/projects/_/buckets/{bucketName}` (the `projects/_` wildcard is the documented format for GCS).

**Note on key format:** Org-managed tag keys are namespaced with the org number (e.g. `815471563813/glz-resource-optout`). Use a regex match in MQL to find them without knowing the org number: `tags.keys.any(_ == /glz-resource-optout/)`.

## Files changed

| File | Change |
|---|---|
| `gcp.lr` | Added `tags() map[string]string` field + updated bucket doc comment |
| `gcp.lr.go` | Regenerated |
| `gcp.lr.versions` | New entry `gcp.project.storageService.bucket.tags 13.26.1` |
| `gcp.permissions.json` | Added `resourcemanager.effectiveTagBindingCollections.get` |
| `storage.go` | `tags()` method + `effectiveTagsToInterface()` helper |
| `storage_test.go` | `TestEffectiveTagsToInterface` (6 cases) + `TestBucketTagKeyMatching` (4 cases) |

### DIFF
--- a/providers/gcp/resources/gcp.lr
+++ b/providers/gcp/resources/gcp.lr
@@ -2398,7 +2398,8 @@ private gcp.project.storageService {
 //
 // Examine a Cloud Storage bucket's configuration, access controls, and
 // data-protection settings. Surfaces the `storageClass`, `location` and
-// `locationType`, `labels`, IAM policy (including `public()` which flags
+// `locationType`, `labels`, `tags` (Resource Manager Tags bound to the bucket),
+// IAM policy (including `public()` which flags
 // any `allUsers` / `allAuthenticatedUsers` grant), `iamConfiguration`
 // (uniform bucket-level access, public access prevention), `retentionPolicy`
 // and `retentionPolicyLocked`, object `versioningEnabled`, default CMEK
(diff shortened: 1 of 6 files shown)


tas50:
Thanks @gerysz

PR #8802, opened 2026-06-30
feat(logger): add loggerconf as single logging config entry point

preslavgerchev:
## What

Adds a new `logger/loggerconf` package with a **single function** that decides how logging is configured, so cnspec and the cnspec-runner (and any future consumer) set up logging identically.

- `loggerconf.Options{Writer, Level, Options}` — declarative logging config with `json` + `mapstructure` tags.
- `loggerconf.Configure(*Options) error` — uplifted from the cnspec-runner's `ConfigureLogger`. Supports the `cli` writer (`json` / `gcp-json` / plain formats) and `stackdriver`, sets the level, and lets `DEBUG`/`TRACE` env vars override the configured level. Returns an error (instead of `log.Fatal`) so it's safe to use as a library.
- `loggerconf.Load(path)` — reads `Options` from a YAML or JSON file.

It lives in its own subpackage (not the base `logger` package) so the GCP/stackdriver dependency stays opt-in.

## Why

Logging setup was duplicated in the cnspec-runner and only configurable there. This centralizes it so cnspec can expose it via a new `--logging-config` flag and the runner can reuse the exact same logic.

## Follow-ups (separate PRs)

- **cnspec**: new `--logging-config` flag that calls `loggerconf.Load` + `Configure`.
- **server/cnspec-runner**: `ConfigureLogger` delegates to `loggerconf.Configure`.

Both depend on this PR merging and an mql release.

## Test

`go test ./logger/loggerconf/` — covers level setting, nil/empty-writer defaults, env override, unknown-writer + stackdriver validation errors, and YAML/JSON loading.

### DIFF
--- a/logger/log.go
+++ b/logger/log.go
@@ -93,11 +93,11 @@ func GetEnvLogLevel() (string, bool) {
 		return v, true
 	}
 
-	if os.Getenv("TRACE") == "true" || os.Getenv("TRACE") == "1" {
+	if v := os.Getenv("TRACE"); v == "true" || v == "1" {
 		return "trace", true
 	}
 
-	if os.Getenv("DEBUG") == "true" || os.Getenv("DEBUG") == "1" {
+	if v := os.Getenv("DEBUG"); v == "true" || v == "1" {
 		return "debug", true
 	}
 
--- a/logger/log.go
+++ b/logger/log.go
@@ -27,7 +27,7 @@ func SetWriter(w io.Writer) {
 
 // UseJSONLogging for global logger
 func UseJSONLogging(out io.Writer) {
-	log.Logger = zerolog.New(out).With().Timestamp().Logger()
+	log.Logger = zerolog.New(out).With().Caller().Timestamp().Logger()
 }
 
 // UseGCPJSONLogging for global logger. This is a JSON logger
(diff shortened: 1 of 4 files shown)


logger/loggerconf/loggerconf.go
```diff
@@ -0,0 +1,95 @@
...
+			projectID := opts.Options["project-id"]
+			if projectID == "" {
+				return fmt.Errorf("stackdriver logging requires a `project-id` option")
+			}
+			logID := opts.Options["log-id"]
+			if logID == "" {
+				return fmt.Errorf("stackdriver logging requires a `log-id` option")
+			}
+
+			w, err := stackdriver.NewStackdriverWriter(projectID, logID)
```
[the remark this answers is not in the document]
preslavgerchev: Good catch in principle, but I'm leaving the signature as-is for this PR. The `stackdriver` writer (`logger/stackdriver/stackdriver.go`) doesn't retain the underlying `*logging.Client` and exposes no `Close`/`Flush` (it only flushes synchronously on Fatal entries), so a `cleanup func` returned from `Configure` couldn't actually close the client without first enhancing the stackdriver package. That's a broader change, and it also matches the existing cnspec-runner behavior we're consolidating here (no cleanup handle today). The writer lives for the process lifetime in all current consumers (cnspec, cnspec-runner). Happy to do the stackdriver Close/Flush + cleanup-func wiring as a dedicated follow-up.

logger/log.go
```diff
@@ -99,5 +101,10 @@ func GetEnvLogLevel() (string, bool) {
 		ok = true
 	}
 
+	if v := os.Getenv("MONDOO_LOG_LEVEL"); v != "" {
+		level = v
```
preslavgerchev: use returns directly

preslavgerchev: same for the ones above.

preslavgerchev: Done in 1f7467b52 — `GetEnvLogLevel` now returns directly. `MONDOO_LOG_LEVEL` is checked first so its precedence over `TRACE`/`DEBUG` is preserved.

PR #8884, opened 2026-07-01
⭐ os: add file context to Dockerfile instructions

tas50:
Exposes a `context` accessor on every Dockerfile instruction resource, mirroring the `terraform.block` file context added in #8703. Each instruction now surfaces the file, the line range it spans, and the raw source text within that range, so a reviewer can be pointed at the exact lines that triggered a policy.

## Example

```
mql run docker file Dockerfile -c 'docker.file.stages.first { run { script context.range context.content } }'
```

```
docker.file.stages.first: {
  run: [
    0: {
      context.range:   2-3
      context.content: "2:  RUN apk add --no-cache curl \
3:      && rm -rf /var/cache/apk/*
"
      script: "apk add --no-cache curl     && rm -rf /var/cache/apk/*"
    }
  ]
}
```

## What changed

- `@context("file.context")` on the 14 instruction resources: `docker.file.run`, `.copy`, `.add`, `.from`, `.env`, `.arg`, `.expose`, `.user`, `.healthcheck`, `.volume`, `.workdir`, `.onbuild`, `.shell`, `.stopsignal`.
- Reuses the generic `file.context` resource (already used by `sshd.config.matchBlock` and `inetd.config.entry`) rather than a new context type — `docker.file` is backed by a real `file`, so `file.context`'s content extraction works directly.
- The range is built from the buildkit parser's per-instruction `Location`. The parser tracks line numbers only (no columns), so the range spans whole lines, from the first through the last reported range — this captures multi-line `RUN` continuations in full (see the 2-3 span above).

## Testing

- `go test ./resources/...` passes; the Dockerfile parse test now also asserts the instruction `context` is populated.
- Verified interactively against a sample Dockerfile that FROM/RUN/COPY/USER/EXPOSE all report accurate line ranges and source snippets.


arlimus (approved):
Woooot!! more context yay

PR #8891, opened 2026-07-02
⭐ gitlab: deepen pipeline, runner, approval, package & token resources

tas50:
## What

Fills the remaining depth gaps on GitLab resources the provider already fetches (Tier-3 of the resource-depth audit). All additive and non-breaking; new `.lr.versions` entries at `13.3.9`.

### `gitlab.project.pipeline` — largest gap
The list endpoint (`ListProjectPipelines`) only returns a subset of fields. This adds the richer provenance fields via a **per-pipeline `GetPipeline`**, fetched once and shared across accessors through an Internal-struct cache (mirroring the existing runner-details pattern):

- `user()` — typed reference to the user who triggered the pipeline (accountability)
- `detailedStatus` — dict (icon/text/label/group/tooltip) richer than the plain `status`
- `beforeSha`, `tag`, `yamlErrors`, `committedAt` — commit/config provenance
- `startedAt`, `finishedAt`, `duration`, `queuedDuration` — execution timing (anomaly detection)
- `coverage` — reported test coverage

### `gitlab.project.runner`
- `tokenExpiresAt` — registration-token expiry (already on the list payload). Audit for non-expiring/stale runner tokens.

### `gitlab.project.approvalSetting`
- `approvers` / `approverGroups` — the project's default named approvers and approver groups, as typed `gitlab.user` / `gitlab.group` references.

### `gitlab.project.package.file`
- `pipelines` — the CI pipeline(s) that produced the artifact, as typed `gitlab.project.pipeline` references (build provenance / SLSA-style link). Empty when the package was not published by a pipeline.

### `gitlab.project.accessToken` / `gitlab.group.accessToken`
- `user()` — typed reference to the bot user the token authenticates as, for correlating token activity with audit events. Resolves lazily and returns null (a bare resource on 403/404) when the bot user is not resolvable via the users API.

## Example queries
```
gitlab.project.pipelines.where(status == "failed") { ref user.username duration finishedAt }
gitlab.project.runners.where(tokenExpiresAt == null)
gitlab.project.approvalSettings { approvers { username } approverGroups { fullPath } }
gitlab.project.packages.flatMap(files).flatMap(pipelines) { webURL sha }
gitlab.project.accessTokens { name scopes user.username }
```

## Notes
- `pipeline` detail fields resolve lazily — a query that only reads list-level fields (`status`, `ref`, `sha`) issues no extra API calls; the `GetPipeline` fetch fires only when a detail field is accessed, and is cached per pipeline.
- Not in scope (separate follow-ups from the audit): `mergeRequest` accountability fields, `release` build-out, `user`/`group`/`project` governance scalars, and the deprecated `*Enabled`→`*AccessLevel` migration.

## Test
- `go build ./...` (gitlab module) — clean
- `go test ./resources/` — pass


tas50:
Thanks for the review — addressed all three:

- 🟡 **`userId` → typed `user()`** (fixed): both `gitlab.project.accessToken` and `gitlab.group.accessToken` now expose `user() gitlab.user` instead of a raw `userId int`, with the raw id cached on a `cacheUserID` Internal field and resolved via `NewResource`. `initGitlabUser` degrades to a bare resource on 403/404, so unresolvable access-token bot users return a bare `gitlab.user` (or null when there's no id) rather than erroring.

- 🔵 **package-file pipeline loop** (fixed): now ranges directly (`for _, pl := range *f.Pipeline { ... &pl }`) instead of index-and-address.

- 🔴 **approver-group nil check** (no change — false positive): `MergeRequestApproverGroup.Group` is a value type (`MergeRequestApproverNestedGroup`), not a pointer, so `g.Group == nil` doesn't compile and `g.Group.ID` can't panic. The existing `if g == nil` guard already covers the only nil-able value (the `*MergeRequestApproverGroup` slice element), matching the `approverUsers` loop.

tas50:
Addressed the two follow-up suggestions:

- 🔵 **package-file pipeline re-fetch** (fixed): `newMqlGitlabPipelineFromDetail` now seeds the cache with the detail payload it already holds — `pipeline.details = pl` and `pipeline.detailsOnce.Do(func() {})` — so lazy accessors (`user()`, `duration()`, …) on a package-file pipeline reuse it instead of issuing a redundant `GetPipeline`.

- 🔵 **`tokenExpiresAt` nullability** (no change — premise doesn't hold): `llx.TimeDataPtr(nil)` returns `NilData`, not a zero-time (see `llx/rawdata.go`), so when the SDK's `TokenExpiresAt` is nil the field resolves to MQL null exactly as the comment states. This is the same plain-`time` + `TimeDataPtr` pattern the provider already uses for every other nullable timestamp (`expiresAt`, `lastUsedAt`, `contactedAt`), so no computed accessor is needed.

PR #8894, opened 2026-07-02
⭐ gcp: add Identity Platform config + tenants

tas50:
## What

Adds the `gcp.project.identityPlatformService` resource tree, exposing **Google Cloud Identity Platform** (GCIP, `identitytoolkit.googleapis.com`) — customer identity and access management. This was the top gap from a GCP security-coverage analysis: Identity Platform governs how a project's customer-facing users authenticate, and none of its posture was queryable before.

### Resources

- **`gcp.project.identityPlatformService`** — service wrapper, gated on `isServiceEnabled`. Alias `gcp.identityplatform`.
- **`gcp.project.identityPlatformService.config`** — project-level authentication settings:
  - MFA state (`DISABLED`/`ENABLED`/`MANDATORY`) + enabled/configured second-factor providers
  - Sign-in methods: email/password (and whether a password is required), phone, anonymous, allow-duplicate-emails
  - **Improved email privacy** (blocks user enumeration)
  - Password policy enforcement state + strength constraints
  - reCAPTCHA enforcement (email/password + phone) and account-defender use
  - Blocking-function triggers (`beforeCreate`/`beforeSignIn`)
  - Self-service account restrictions (disabled user signup / deletion)
  - Request logging, multi-tenancy enablement, authorized OAuth redirect domains, SMS region config
- **`gcp.project.identityPlatformService.tenant`** — per-tenant auth config for multi-tenant projects (sign-in methods, MFA, email privacy, password policy, reCAPTCHA, test phone numbers).

### Why these fields

Every field maps to a concrete auth-posture check (MFA enforced? anonymous sign-in on? email enumeration blocked? password policy enforced? bot protection on?). Complex nested structures (sign-in sub-configs, blocking functions, SMS regions, password strength) are exposed as scalars where they're the primary signal and as `dict` where the shape is heterogeneous, per the sub-resource guidance.

## Verification

Built and installed the provider; ran against a live GCP test project (`--discover projects`):

- **Service-disabled path**: `config` resolves to `null`, no error.
- **Service-enabled, uninitialized project**: debug log confirms the provider issues the real Admin API call (`GET identitytoolkit.googleapis.com/v2/.../config`) and maps the resulting `404` (GCIP not yet initialized) to `null` gracefully.

The end-to-end path — service gate → authenticated client → correct Admin API URL → response handling — is verified live. Field population against a fully **initialized** GCIP project (with a Config + tenants) hasn't been exercised, since initializing Identity Platform is a persistent provisioning step on the target project; happy to do a full data-population pass if we stand up an initialized project.

Unit tests (`resources` permissions + discovery) pass; generated code, `.lr.versions` (all new entries at `13.28.1`), and `go.mod` are clean.

## Follow-ups

- Identity federation trust: `defaultSupportedIdpConfigs`, `inboundSamlConfigs`, `oauthIdpConfigs` (which external IdPs can authenticate users) — the natural next slice of GCIP security coverage.


tas50:
Addressed both review suggestions in 84ebafee0 — and it turned out the package already had shared equivalents, so rather than rename/move I reused them:

- `isNotFoundOrDenied` → removed; now uses the existing `isHTTPSkippable` (`common.go`), which is a superset (also matches "not enabled" / "has not been used" messages).
- `jsonSliceToDict` → removed; now uses `convert.JsonToDictSlice`.

Build, vet, and the resources unit tests pass.

PR #8896, opened 2026-07-02
⭐ gitlab: add project & group governance scalars; migrate deprecated *Enabled

tas50:
## What

Second half of the depth-audit **Tier 1** (part 1 was #8895, protected-branch access levels). This adds governance/compliance visibility to `gitlab.project` and `gitlab.group`, and migrates the SDK-deprecated feature toggles.

### `gitlab.project`
- **Deprecated-field migration:** adds the modern `*AccessLevel` fields — `issuesAccessLevel`, `mergeRequestsAccessLevel`, `wikiAccessLevel`, `snippetsAccessLevel`, `containerRegistryAccessLevel`, `buildsAccessLevel`, `repositoryAccessLevel`, `forkingAccessLevel`, `securityAndComplianceAccessLevel` — and marks the matching `*Enabled` bools `@maturity("deprecated")`. The SDK deprecated those bools; they can return stale/`false` values on modern instances because the data moved to the access-level fields. Old fields are kept for back-compat (versions unchanged).
- `emailsEnabled` (deprecates `emailsDisabled`).
- `complianceFrameworks`, `sharedWithGroups`, `ciJobTokenScopeEnabled`, `publicJobs`, `preReceiveSecretDetectionEnabled`, `securityAndComplianceEnabled`, and the mirror security posture (`mirrorTriggerBuilds`, `onlyMirrorProtectedBranches`).

### `gitlab.group`
- `ipRestrictionRanges` (network perimeter), `sharedWithGroups`, `shareWithGroupLock`, `sharedRunnersSetting`, `projectCreationLevel`, `subGroupCreationLevel`, `autoDevopsEnabled`, `wikiAccessLevel`, `defaultBranchProtection` (dict), and LDAP governance (`ldapCn`, `ldapAccess`, `ldapGroupLinks`).
- `emailsEnabled` (deprecates `emailsDisabled`).
- **Cleanup:** `subgroups()` now reuses `populateGroupArgs` instead of a duplicated inline field map, so subgroups expose the same field set as top-level groups and inherit these new fields automatically.

## Example queries
```
gitlab.projects.where(issuesAccessLevel == "enabled" && repositoryAccessLevel == "private")
gitlab.projects.where(complianceFrameworks.length == 0)
gitlab.projects.where(ciJobTokenScopeEnabled == false)   # unscoped CI tokens
gitlab.groups.where(ipRestrictionRanges == "")           # no IP allowlist
gitlab.groups { fullPath sharedRunnersSetting projectCreationLevel ldapGroupLinks }
```

## Notes
- All additions are additive and non-breaking; new `.lr.versions` entries at `13.3.9`. Deprecations do **not** bump versions.
- `sharedWithGroups`, `ldapGroupLinks`, and `defaultBranchProtection` are modeled as `dict`/`[]dict` (following the existing `approvalRule.groups` precedent) rather than new sub-resources, since they're data containers without a standalone identity.

## Test
- `go build ./...` (gitlab module) — clean
- `go test ./resources/` — pass

### DIFF
--- a/providers/gitlab/resources/gitlab.lr
+++ b/providers/gitlab/resources/gitlab.lr
@@ -307,7 +307,11 @@ gitlab.group @defaults("name visibility webURL") {
   // Whether forking projects outside this group is forbidden
   preventForkingOutsideGroup bool
   // Whether group email notifications are disabled
-  emailsDisabled bool
+  //
+  // Deprecated in favor of emailsEnabled.
+  emailsDisabled @maturity("deprecated") bool
+  // Whether group email notifications are enabled
+  emailsEnabled bool
   // Whether group mentions within issues and merge requests are disabled
   mentionsDisabled bool
   // Whether users can request membership to the group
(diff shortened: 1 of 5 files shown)


tas50:
Added the explicit `int64(g.GroupID)` cast in both `projectSharedGroupsToDicts` and `groupSharedGroupsToDicts`.

For the record: `GroupID` is already `int64` in both SDK structs (`ProjectSharedWithGroup` and `SharedWithGroup`), so there was no actual runtime type mismatch — the value reaching the MQL dict was already `int64`. I've added the explicit cast anyway so the numeric dict values are uniformly `int64` and match `groupAccessLevel`.

PR #8961, opened 2026-07-06
🧹 os: migrate off deprecated docker/docker to moby/moby

czunker:
## Summary

Starting with Docker v29, `github.com/docker/docker` is deprecated and no
longer updated in favor of two purpose-built modules: `github.com/moby/moby/client`
and `github.com/moby/moby/api`. See:

- https://github.com/moby/moby#go-modules
- https://github.com/moby/moby/releases/tag/docker-v29.0.0

This migrates mql's Docker Engine connection/discovery code off the
deprecated module before it goes unmaintained:

- Rewrites every `providers/os/connection/docker/*` and
  `providers/os/resources/{docker.go,discovery/docker_engine/*,processes/dockertop.go}`
  call site for the v29 client's `<Method>Options`/`<Method>Result` wrapper
  pattern and method renames (`ContainerExecCreate`→`ExecCreate`, etc.)
- Rebuilds `docker_engine.FromDockerEnv`'s DOCKER_HOST parsing, since
  `client.Opt` can no longer be re-applied to an already-constructed
  `*client.Client` in the new API
- Drops `github.com/docker/docker` from the root `go.mod` and all 6 provider
  `go.mod`s (aws, azure, gcp, github, google-workspace, k8s), which only ever
  carried it indirectly
- Bumps `github.com/google/go-containerregistry` to v0.21.7 so its
  `pkg/v1/daemon` package (used for local image loading) drops its own last
  transitive reference to `docker/docker` — fully removing the deprecated
  module from the dependency graph, not just moving it to indirect
- Switches remaining `client.NewClientWithOpts` calls to the non-deprecated
  `client.New`

No `.lr` schema changes — this is a pure internal client-library swap with
no intended behavior change.

## Test plan

- [x] `go build ./...` at root and in all 6 provider modules
- [x] `go vet` / `gofmt` clean on all changed files
- [x] `go test` on all docker-related packages against a live Docker Desktop
      daemon, including the normally-excluded `debugtest`-tagged test
- [x] Interactively verified `docker.images`, `docker.containers`,
      `docker.container.hostConfig`, the exec path, and the file-read path
      via the installed `mql` binary against real containers


providers/os/connection/docker/command.go
```diff
@@ -27,10 +26,8 @@ func (c *Command) Exec(command string) (*shared.Command, error) {
 	c.Stats.Start = time.Now()
 
 	ctx := context.Background()
-	res, err := c.Client.ContainerExecCreate(ctx, c.Container, container.ExecOptions{
+	res, err := c.Client.ExecCreate(ctx, c.Container, client.ExecCreateOptions{
 		Cmd:          []string{"/bin/sh", "-c", c.Command.Command},
-		Detach:       true,
-		Tty:          false,
```
czunker, line 33: Why are these removed?

Even when they are the new default, we should keep them for clarity.

providers/os/connection/docker/command.go
```diff
@@ -39,10 +36,7 @@ func (c *Command) Exec(command string) (*shared.Command, error) {
 		return nil, err
 	}
 
-	resp, err := c.Client.ContainerExecAttach(ctx, res.ID, container.ExecStartOptions{
-		Detach: false,
-		Tty:    false,
```
czunker, line 44: Same here.

czunker:
## What I tested

- \`go build ./...\` at root and in all 6 provider modules (aws, azure, gcp, github, google-workspace, k8s); \`go vet\` / \`gofmt\` clean on every changed file
- \`go test\` on all 4 docker-related packages (\`providers/os/connection/docker\`, \`providers/os/resources/discovery/docker_engine\`, \`providers/os/resources\`, \`providers/os/resources/processes\`) against a **live Docker Desktop daemon** — real image pulls, container create/start/exec/remove, all passing
- The \`debugtest\`-tagged \`TestDockerCommand\` (excluded from default \`go test\` runs) — ran explicitly, passing
- Interactive verification via the installed \`mql\` binary:
  - \`docker.images\`, \`docker.containers\`, \`docker.container.hostConfig\` against real running containers (read-only)
  - Exec path (\`command('whoami')\`) and file-read path (\`file('/etc/os-release').content\`) against a disposable container I created and removed for this purpose, to avoid running commands against pre-existing containers

No \`.lr\` schema changes, so no schema/version bump was needed.

providers/os/connection/docker/command.go
czunker: Fixed in 77d73251e5, but with a caveat: `Tty` still exists on the new API as `TTY` and I've restored it explicitly. `Detach` is gone entirely though — it's not just defaulted, the field was removed from `client.ExecCreateOptions`/`client.ExecAttachOptions` in moby/moby's v29 client. It was actually already inert in the old docker/docker client too (`// Deprecated: the Detach field is not used, and will be removed in a future release.` on `container.ExecOptions.Detach`), so nothing behavioral changes here — there's just no field left to set.

providers/os/connection/docker/command.go
czunker: Same fix as above (77d73251e5): `TTY: false` restored explicitly on `ExecAttachOptions`. `Detach` has no equivalent field on `ExecAttachOptions` in the new API — attach is inherently non-detached now (that's handled by the separate `ExecStart` call, which we don't use here), so there's nothing to keep.

PR #8966, opened 2026-07-06
🐛 streamline TF HCL var resolution

arlimus:
Given the previous behavior was buggy, we are now streamlining this capability because it matches the users' intent. Testing for variables instead of testing for contents is still possible, but it no longer confuses users who just want to see what the value is.

### DIFF
--- a/features.go
+++ b/features.go
@@ -94,7 +94,7 @@ const (
 
 	// Resolve var.* and local.* references in Terraform HCL block arguments to their effective values (variable defaults overridden by .tfvars, locals evaluated from those). Unresolvable references fall back to their reference string; the unresolved view stays available via terraform.block.argumentReferences.
 	// start:  v13.x
-	// status: new
+	// status: default
 	TerraformResolveVars Feature = 17
 
 	// Placeholder to indicate how many feature flags exist. This number
@@ -128,6 +128,7 @@ var DefaultFeatures = Features{
 	byte(SerialNumberAsID),
 	byte(ForceShellCompletion),
 	byte(ResourceContext),
+	byte(TerraformResolveVars),
 }
 
 // AvailableFeatures are a set of flags that can be activated
@@ -137,5 +138,4 @@ var AvailableFeatures = Features{
 	byte(AutoUpdateEngine),
 	byte(BiosUUIDAsID),
 	byte(ExchangeTokenForToken),
-	byte(TerraformResolveVars),
 }
--- a/features.yaml
+++ b/features.yaml
@@ -97,4 +97,4 @@ Features:
   desc: "Resolve var.* and local.* references in Terraform HCL block arguments to their effective values (variable defaults overridden by .tfvars, locals evaluated from those). Unresolvable references fall back to their reference string; the unresolved view stays available via terraform.block.argumentReferences."
   idx: 17
   start: v13.x
-  status: new
+  status: default


vjeffrey:
Tested this branch against an **authenticated** Terraform scan (service account → platform, using a space-assigned policy) and it **worked 1 of 5 times** — non-deterministic, ~3/14 across all my runs. So this doesn't reliably fix the authenticated case yet. Details + repro below.

## What I think is going on

This PR moves `TerraformResolveVars` into `DefaultFeatures` (always-on). But in the authenticated failures the feature isn't the problem — it's already active on the connection, and the provider is already returning the **resolved** value. The check passes because **`terraform.resources` comes back empty (`length == 0`) in authenticated scans**, so `terraform.resources.where(...).all(...)` is evaluating `.all()` over an empty list → vacuously `true` → passes. There's simply nothing to test.

I confirmed this on a passing run with `DEBUG=1`:

- `terraform.resources.length` datapoint decodes to `0`
- the check's `.all(...)` boolean datapoint decodes to `true` (vacuous pass)
- yet `arguments()` on the block still returns the correctly-resolved value (`[0.0.0.0/0]`) — resolution works; the resource *list* is what's empty

Incognito scans of the same module are always correct (CRITICAL). Only the authenticated / server-resolved-policy path shows the empty list. Making the feature default seems to just perturb the timing enough that it *occasionally* wins the race (hence 1/5), which is why a single run can look like a fix.

So I think the real fix needs to target why `terraform.resources` (the list, via `refreshCache`) collects empty in the authenticated execution path, rather than the feature gating. Happy to share the full instrumented evidence + `DEBUG=1` dumps.

## How to test it (important: it's a race — run it many times)

**Build cnspec against this branch:**
```bash
cd mql && git fetch origin dom/tf-vars && git checkout dom/tf-vars
cd ../cnspec
go mod edit -replace go.mondoo.com/mql/v13=../mql
go build -o /tmp/cnspec-test ./apps/cnspec
go list -m go.mondoo.com/mql/v13   # confirm "=> ../mql"
# cleanup later: git checkout origin/main -- go.mod
```
(The released terraform provider already honors the feature; no need to rebuild it.)

**⚠️ Do NOT use `-f`.** `cnspec scan -f <bundle>` forces `--incognito`, which bypasses the authenticated path and always works. Use a **bare** `cnspec scan terraform <dir>` with a service-account config so it runs the space-assigned policy. The check (upload + enable it in the space):
```
terraform.resources.where( nameLabel == "aws_security_group_rule" ).all(
  arguments["cidr_blocks"].none( _ == "0.0.0.0/0" )
)
```

**Run it many times** (a single run proves nothing):
```bash
CN=/tmp/cnspec-test
crit=0; n=8
for i in $(seq 1 $n); do
  D=$(mktemp -d); U="u$(date +%s%N)$i"
  printf "# %s\nresource \"aws_security_group_rule\" \"%s\" {\n  cidr_blocks = [var.ingress_cidr]\n}\n" "$U" "$U" > "$D/main.tf"
  printf 'variable "ingress_cidr" { default = "10.0.0.0/8" }\n'  > "$D/variables.tf"
  printf 'ingress_cidr = "0.0.0.0/0"\n'                          > "$D/terraform.tfvars"
  out="$("$CN" scan terraform "$D" --auto-update=false 2>&1)"
  grep -qiE "Security group.*CRITICAL|CRITICAL.*Security group" <<<"$out" \
      && { echo "run $i: CRITICAL"; crit=$((crit+1)); } || echo "run $i: PASS"
  rm -rf "$D"
done
echo "CRITICAL=$crit / $n"
```
Fixed = CRITICAL on **every** run. Unique content per run avoids stored-data caching.

**Confirm the mechanism on a PASS run:**
```bash
mkdir /tmp/dbg && cd /tmp/dbg
DEBUG=1 $CN scan terraform <dir> --auto-update=false   # writes report.json here
```
```python
import json, base64
d = json.load(open("report.json"))   # under /tmp/dbg/cnspec-debug-*/
data = list(d["reports"].values())[0]["data"]
LEN = "5A5PSZ5gn4eHAqVkBiajrnDzaPEezD4MIMYwkDVMtYUfsUm+8wPLRFGcuZa48Uvg0QbM9andr/82AfKRH25CUw=="
print("terraform.resources.length =", base64.b64decode(data[LEN]["data"]["value"]))  # 0 on a broken run
```

The fix is verified only when the loop is CRITICAL every run **and** `terraform.resources.length > 0` in an authenticated scan.

arlimus:
FYI: Fixed two race-conditions since, thank you so much for finding this before it went out and raising it. Should be good to go now :crossed_fingers:

PR #8973, opened 2026-07-06
⭐ aws: resolve Route 53 alias records to load balancer and CloudFront

tas50:
## What

Adds typed accessors on `aws.route53.record` that resolve an alias record's target to the fronting resource:

| Accessor | Resolves to | Match |
|---|---|---|
| `aliasLoadBalancer()` | `aws.elb.loadbalancer` | alias target DNS name == load balancer DNS name |
| `aliasCloudFrontDistribution()` | `aws.cloudfront.distribution` | alias hosted zone == `Z2FDTNDATAQYW2` and DNS name == distribution domain name |

## Why

An A/AAAA alias record previously exposed only `aliasTargetDnsName` and `aliasTargetHostedZoneId` as raw strings, so "what does this DNS name point at" could not be followed to the actual resource. The entry point of most external attack paths is a DNS name — making the alias target a typed edge lets a query walk `record -> load balancer -> target group -> backend`, composing with the rest of the network graph:

```mql
// public DNS names and the internet-facing load balancer they front
aws.route53.hostedZones {
  records.where(isAlias) {
    name
    aliasLoadBalancer { dnsName scheme exposure { internetReachable } }
    aliasCloudFrontDistribution { domainName }
  }
}
```

## Notes

- DNS matching normalizes case, the trailing `.` of the FQDN, and the `dualstack.` prefix Route 53 prepends to load balancer alias targets.
- Resource enumeration reuses the cached-list pattern from `aws.ec2.instance.loadBalancers()`, so **no new IAM permissions** (`permissions.json` change is version/timestamp metadata only).
- Scope is the internet-facing ingress targets (ELB, CloudFront). S3 website endpoints and API Gateway aliases are natural follow-ups.

## Testing

- `TestNormalizeAliasDNSName` covers the DNS normalization (dualstack prefix, trailing dot, case).
- `TestRecordAliasTargetNullState` covers the null-dispatch branches (empty target, non-CloudFront hosted zone, CloudFront zone with no DNS name).
- Full `providers/aws/resources` test package passes.
- Note: the test account has no Route 53 hosted zones, so the positive resolution path was not exercised live; the enumeration logic is copied verbatim from the live-verified `instance.loadBalancers()` accessor.

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -13181,6 +13181,20 @@ private aws.route53.record @defaults("name type") {
   aliasTargetDnsName string
   // Hosted zone ID of the alias target
   aliasTargetHostedZoneId string
+  // Load balancer this record's alias points to
+  //
+  // Resolves an A or AAAA alias record whose target is an Application, Network,
+  // or Classic Load Balancer, matched on the alias target DNS name. Null when
+  // the record is not an alias to a load balancer. Follows the DNS front door of
+  // an ingress path to the load balancer that fronts the workload.
+  aliasLoadBalancer() aws.elb.loadbalancer
+  // CloudFront distribution this record's alias points to
+  //
+  // Resolves an A or AAAA alias record whose target is a CloudFront distribution
+  // (identified by the fixed alias hosted zone Z2FDTNDATAQYW2), matched on the
+  // distribution domain name. Null when the record is not an alias to a
+  // distribution.
+  aliasCloudFrontDistribution() aws.cloudfront.distribution
   // Whether to evaluate target health for alias records
   aliasEvaluateTargetHealth bool
   // Routing policy identifier (for weighted, latency, failover, geolocation routing)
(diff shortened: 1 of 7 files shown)


tas50:
Addressed in the latest push. I built the normalized-DNS-name index on the runtime-cached `aws.elb` / `aws.cloudfront` list resource (mutex-guarded, built once) rather than on the record's `Internal` struct — since each `aws.route53.record` is a separate instance, a per-record cache wouldn't be shared across records, whereas the list singleton is. Resolving alias targets across many records is now a map lookup each instead of an O(n×m) rescan.

tas50:
Fixed — `loadBalancerByDNSName` now indexes both `GetLoadBalancers()` (v2) and `GetClassicLoadBalancers()`, so a Route 53 alias record pointing at a Classic ELB resolves too. Since DNS names are unique across LB types there's no collision in the shared index.

On the `.lr.versions` note: `13.43.1` is correct (provider is at `13.43.0`), which the inline comment confirms — no change needed there.

PR #8981, opened 2026-07-07
⭐ digitalocean: App Platform depth, Functions listing, and typed refs

tas50:
## Summary

The DigitalOcean provider is already one of the most complete providers in the repo, so this PR is about **depth, schema consistency, and docs** rather than new product areas.

### Tier 1 — schema correctness
- **`database.project()`** typed accessor (the cluster's `projectId` had no ref, unlike sibling resources).
- **Deprecated stray `vpcUuid` fields** on `gradientai.agent` and `gradientai.dedicatedInferenceEndpoint` (they already expose `vpc()`), matching the existing droplet/k8s/load-balancer convention.
- **Typed URN/ID references** via a shared `parseDropletID` helper (handles both bare numeric IDs and `do:droplet:N` URNs):
  - `snapshot.droplet()` / `snapshot.volume()` (discriminated by `resourceType`)
  - `alertPolicy.droplets()` (resolves droplet-targeting policies)
  - `byoipPrefix.resource.droplet()`

### Tier 3 — depth
- **App Platform**: new `digitalocean.app.deployment` and `digitalocean.app.alert` sub-resources, plus `app.deployments()`, `app.activeDeployment()`, `app.alerts()`, and `app.instances()`. `instances` is `[]dict` (ephemeral, no stable ID and no typed refs → fails the sub-resource bar).
- **Functions**: godo has no endpoint to list the functions (actions) deployed in a namespace — they live behind each namespace's Apache OpenWhisk API host. This adds a minimal read-only OpenWhisk client (HTTP Basic `UUID:Key` against `{ApiHost}/api/v1/namespaces/_/actions`) backing `function.namespace.functions()` → `digitalocean.function.action` (runtime, web-export posture, limits). The OpenWhisk API is officially documented (it's what `doctl serverless` drives); it's just not wrapped by godo.

### Tier 4 — docs
- New `providers/digitalocean/README.md` (auth, Spaces credentials, discovery flags, example queries).
- Filled thin two-part doc-comments on private sub-resources (`database.user`/`replica`/`pool`, `domain.record`, `kubernetes.nodePool`, `function.trigger`, `byoipPrefix.resource`).

## Versioning
New schema entries are at `13.10.2` (provider currently `13.10.1`). `config.go` `Version` intentionally left unchanged.

## Testing
- `go build ./...`, full `make providers/build/digitalocean`, `go test ./resources/`, `gofmt`, and `go vet` all pass.
- Codegen is drift-free (`mqlr generate` idempotent).
- ⚠️ **Not yet verified against a live account** — no DigitalOcean token was available in the dev environment. The OpenWhisk `functions()` path and App Platform depth (`deployments`/`alerts`/`instances`) should be smoke-tested against a real account with a deployed function before relying on them.


tas50:
Thanks for the review — addressed in 2f4891da7:

**🟡 `loadBalancerId` should be a typed accessor** — Fixed. Added `loadBalancer() digitalocean.loadBalancer` on `app.deployment`, resolved through the existing cached load-balancer index. Also added `previousDeployment() digitalocean.app.deployment` for the sibling `previousDeploymentId` pointer (the second raw-ID field). I kept the raw `*Id` fields public alongside the accessors, matching the provider's established convention (`reservedIp.dropletId` + `droplet()`, `loadBalancer.projectId` + `project()`).

**🔵 single-`/` doc-comment typo on `domains`** — This looks like a false positive from a `.lr.go` ↔ `.lr` line-number mismatch. `digitalocean.lr:1192` is a correct `// Custom domains attached to the app ...` (double-slash), and there are no single-slash comment lines in the `.lr` file. The referenced `.lr.go:1192` is an unrelated generated `loadBalancer.projectId` accessor. No change needed — happy to be corrected if I'm missing something.

Still outstanding: this branch hasn't been verified against a live DigitalOcean account (no token in the dev env), so the OpenWhisk `functions()` path and App Platform depth should be smoke-tested before merge.

PR #8988, opened 2026-07-07
⭐ aws: detect Nitro instances via instance-type hypervisor

tas50:
## What

Adds an `instanceTypeHypervisor` computed field to `aws.ec2.instance` reporting the instance type's hypervisor (`nitro` or `xen`), so MQL can tell whether an EC2 instance runs on the AWS Nitro System.

## Why

The instance-level `hypervisor` field (from `DescribeInstances`) only ever returns `xen` or `ovm` — it **never** returns `nitro`. So there was no way in MQL to distinguish a Nitro-based instance from a genuine Xen one. Nitro is only exposed on the *instance type's* hypervisor, returned by `DescribeInstanceTypes`.

This is the one unmet goal of the "Support AWS Nitro Enclaves in MQL" work (mondoohq/planning#1079), whose demo flow calls for Mondoo to "surface whether each runs on Nitro or a previous generation hypervisor." The Enclaves/NitroTPM/ENA fields it lists already shipped (`enclaveEnabled`, `tpmSupport`, `enaSupported`); this fills the remaining Nitro-detection gap.

## Implementation

- Lazy computed method on `aws.ec2.instance`, backed by a single `DescribeInstanceTypes` call.
- The instance-type → hypervisor mapping is an immutable, region-independent AWS fact, so it's memoized process-wide to avoid an API call per instance (instances of the same type share one lookup).
- Access-denied is logged rather than silently returning an empty value, and empty API results are not cached (so a type missing from one region can't poison the shared cache).
- Adds `ec2:DescribeInstanceTypes` to the permissions manifest.

## Verification

Ran against live EC2. The instance-level field misreports Nitro instances; the new field gets it right:

```
instanceType: "t3.small"   hypervisor: "xen"   instanceTypeHypervisor: "nitro"
instanceType: "m5.2xlarge" hypervisor: "xen"   instanceTypeHypervisor: "nitro"
instanceType: "t2.large"   hypervisor: "xen"   instanceTypeHypervisor: "xen"
```

Example queries:

```mql
// Flag any instance still on a previous-generation (non-Nitro) type
aws.ec2.instances.where( instanceTypeHypervisor == "xen" )

// Group the fleet by underlying hypervisor
aws.ec2.instances { instanceType instanceTypeHypervisor }
```

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -16678,6 +16678,14 @@ private aws.ec2.instance @defaults("instanceId region state instanceType archite
   vpcArn @maturity("deprecated") string
   // Hypervisor type of the instance: ovm or xen
   hypervisor string
+  // Hypervisor of the instance type: nitro or xen
+  //
+  // Reports whether the instance type is built on the AWS Nitro System
+  // (nitro) or the Xen hypervisor (xen). Unlike the instance-level
+  // hypervisor field, which only reports ovm or xen, this reflects the
+  // underlying platform and is how you tell whether an instance is
+  // Nitro-based.
+  instanceTypeHypervisor() string
   // Whether this is a Spot Instance or a Scheduled Instance: spot, scheduled, or capacity-block
   instanceLifecycle string
   // Root device type used by the AMI: ebs or instance-store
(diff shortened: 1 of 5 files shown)


tas50:
Addressed review feedback:

- **Access-denied no longer silently returns an empty value** — it now logs a warning (`no permission for ec2:DescribeInstanceTypes`) so the missing permission is surfaced rather than looking like a real "not Nitro" answer.
- **Empty API results are no longer cached** — a `len(resp.InstanceTypes) == 0` result returns without storing, so a type absent from one region can't poison the process-wide cache for regions where it exists.
- **Dropped the `nitro` bool field** (per follow-up request) — the raw `instanceTypeHypervisor` string is queryable as-is (`where(instanceTypeHypervisor == "nitro")`), so the separate boolean was redundant. This also resolves the third suggestion (the duplicate resolution path between `nitro()` and `instanceTypeHypervisor()` is gone).

PR #8994, opened 2026-07-07
🐛 fix CodeQL allocation-size-overflow bounds

tas50:
## What

Resolves four CodeQL `go-allocation-size-overflow` alerts by adding explicit upper-bound validation before allocations whose size is derived from externally-controlled data.

| Alert | Site | Change |
|---|---|---|
| [#49](https://github.com/mondoohq/mql/security/code-scanning/49) | `providers/activedirectory/resources/privileged.go` | Reject SIDs with more than 255 sub-authorities before sizing the buffer |
| [#40](https://github.com/mondoohq/mql/security/code-scanning/40) | `mqlc/mqlc.go` (`resourceArgs`) | Guard `len(args)` before `make(…, len(args)*2)` |
| [#39](https://github.com/mondoohq/mql/security/code-scanning/39) | `mqlc/mqlc.go` (`unnamedArgs`) | Same guard |
| [#100](https://github.com/mondoohq/mql/security/code-scanning/100) | `providers/os/resources/shadow.go` | Bound the entry count before sizing the result slice |

## Why

The rule fires when a `make([]T, n)` size flows from data an attacker might influence, covering both integer overflow (wrapping to an undersized buffer) and unbounded allocation.

The Active Directory case is the substantive one: `subAuthCount` comes from splitting a SID string returned by the directory server, and it feeds both the allocation size `8+4*subAuthCount` and `buf[1] = byte(subAuthCount)`. That second assignment silently truncates any value over 255, so a malformed SID would have produced an incorrect binary encoding. The binary SID format defines `SubAuthorityCount` as a single byte, so the `<= 255` check is a correct spec bound and fixes that latent encoding bug as well.

The other three take `len()` of an already-materialized in-memory slice, so they cannot realistically overflow, but the size still flows from parsed/file input. An explicit bound in the same function as the allocation clears the alert and doubles as honest input validation.

## Testing

- `go build` passes for `mqlc`, `providers/os`, and `providers/activedirectory`.
- `mqlc` test suite passes.
- `gofmt` clean.

### DIFF
--- a/mqlc/mqlc.go
+++ b/mqlc/mqlc.go
@@ -6,6 +6,7 @@ package mqlc
 import (
 	"errors"
 	"fmt"
+	"math"
 	"regexp"
 	"sort"
 	"strconv"
@@ -835,7 +836,12 @@ func (c *compiler) unnamedArgs(callerLabel string, init *resources.Init, args []
 	}
 
 	// add all calls to the chunk stack
-	// collect all their types and call references
+	// collect all their types and call references.
+	// len(args) is bounded by the parsed query, but guard the size computation
+	// explicitly so len(args)*2 can never overflow on pathological input.
+	if len(args) > math.MaxInt/2 {
+		return nil, errors.New("Called " + callerLabel + " with too many arguments")
+	}
 	res := make([]*llx.Primitive, len(args)*2)
 
 	for idx := range args {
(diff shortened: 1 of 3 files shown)


arlimus (commented):
Great findings xgrep 😁  Thank you @tas50

PR #9007, opened 2026-07-08
✨ gitlab: expose GitLab Duo AI indicators on settings and projects

tas50:
## What

Exposes the GitLab Duo / AI governance indicators that the SDK returns, so audits can assert on AI-feature posture. All additive; new `.lr.versions` entries at `13.4.2`.

**`gitlab.settings` (instance-wide):**
| Field | Meaning |
|---|---|
| `duoFeaturesEnabled` | Whether GitLab Duo AI features are enabled for the instance |
| `lockDuoFeaturesEnabled` | Whether the Duo toggle is locked so groups/projects can't override it |

**`gitlab.project`:**
| Field | Meaning |
|---|---|
| `autoDuoCodeReviewEnabled` | Whether Duo automatically reviews merge requests |
| `modelExperimentsAccessLevel` | Visibility of the ML model-experiments feature (`disabled`/`private`/`enabled`) |
| `modelRegistryAccessLevel` | Visibility of the ML model-registry feature (same enum) |

## Why

AI-feature governance is increasingly audit-relevant (data leaving the org boundary to an LLM, autonomous MR review, ML model hosting). These are the read-side signals for "is Duo/AI on, and is it locked down" at the instance and project tiers.

## Implementation notes

- Every field is read from an existing GET payload (`Settings.GetSettings`, `Projects.GetProject` / project list), so there are **no new API calls** — just extra fields mapped in `initGitlabSettings` and `getGitlabProjectArgs`.
- The model-* access levels use GitLab's uniform `AccessControlValue` enum, mapped to `string` to match how every other `*AccessLevel` field in the provider is already modeled.

## Scope note — group-level Duo

The most-wanted tier for Duo governance is the **group**, but in `client-go` v1.46.0 the group Duo fields (`duo_availability`, `duo_features_enabled`, `lock_duo_features_enabled`, `experiment_features_enabled`) exist **only in `CreateGroupOptions`/`UpdateGroupOptions`** — they are write-only and absent from the `Group` GET response struct, so they can't be read through the typed SDK. There is also no dedicated Duo/AI service in the SDK. Exposing group Duo settings needs either an SDK bump (once the `Group` response carries them) or a supplementary raw GET; tracked as follow-up.

## Follow-up to #9004

Builds on #9004 (now merged); rebased onto `main` so this PR contains only the AI-indicator diff.

## Verification

- `go build ./...` — clean

⚠️ **Live `mql shell gitlab` verification was not run** — no GitLab token in this environment. Suggested smoke test:

```
gitlab.settings { duoFeaturesEnabled lockDuoFeaturesEnabled }
gitlab.project { autoDuoCodeReviewEnabled modelExperimentsAccessLevel modelRegistryAccessLevel }
```

### DIFF
--- a/providers/gitlab/resources/gitlab.lr
+++ b/providers/gitlab/resources/gitlab.lr
@@ -77,6 +77,10 @@ gitlab.settings @defaults("requireTwoFactorAuthentication") {
   sessionExpireDelay int
   // Maximum web-terminal session time, in seconds
   terminalMaxSessionTime int
+  // Whether GitLab Duo AI features are enabled for the instance
+  duoFeaturesEnabled bool
+  // Whether the GitLab Duo features toggle is locked so groups and projects cannot override it
+  lockDuoFeaturesEnabled bool
 }
 
 // GitLab user
@@ -716,6 +720,12 @@ gitlab.project @defaults("fullName visibility webURL") {
   mirrorTriggerBuilds bool
   // Whether mirroring is limited to protected branches
   onlyMirrorProtectedBranches bool
+  // Whether GitLab Duo automatically reviews merge requests
+  autoDuoCodeReviewEnabled bool
+  // Visibility of the machine-learning model experiments feature (disabled, private, enabled)
+  modelExperimentsAccessLevel string
+  // Visibility of the machine-learning model registry feature (disabled, private, enabled)
+  modelRegistryAccessLevel string
   // Approval rules for the project
   approvalRules() []gitlab.project.approvalRule
   // Merge methods for the project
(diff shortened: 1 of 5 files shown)


arlimus (approved):
And another super interesting expansion to a git repo provider - thank you!

tas50:
Thanks for the review. On the "may silently return false on older instances" note:

The SDK models these as non-pointer types (`DuoFeaturesEnabled bool`, `AutoDuoCodeReviewEnabled bool`, `ModelExperimentsAccessLevel string`), so a missing JSON field is unmarshalled to the zero value (`false` / `""`) inside client-go before the provider sees it. That means the absent-vs-disabled distinction is already lost upstream — making the `.lr` fields nullable wouldn't recover it, since the SDK still hands us `false`.

Rather than leave it silent, I documented the behavior in the field descriptions (ed7d6d4): each field now states it reads `false`/empty on instances that don't support or report the setting. Doc-only change, no schema or version bump.

If we later want a true null here, it needs an SDK change to pointer types (or a raw GET) — happy to file that as a follow-up.

PR #9030, opened 2026-07-10
🐎 recording: fix concurrent-map panic in finalize during parallel scans

tas50:
## Problem

A parallel scan panics and aborts:

```
panic: runtime error: index out of range [5547] with length 5547
recording.(*Asset).finalize(...) asset_recording.go:61
recording.(*recording).Save(...)  recording.go:163
providers.(*Runtime).Close.func1() runtime.go:97
discovery.(*AssetExplorer).CloseAsset(...)
policy/scan.(*scanDispatcher).scanSingleAsset(...)
```

## Root cause

This is a **data race on the per-asset recording maps**, not a logic error.

In a parallel scan all assets share one `recording`. When the *first* asset's `Runtime.Close()` fires, it calls the shared recording's `Save()` → `finalize()`, which iterates **every** asset's `resources` map. Meanwhile other assets are still fetching data via `AddData()`, which inserts into those same maps.

Go maps are not safe for concurrent use. `finalize()` sizes a slice with `make([]Resource, len(asset.resources))` and then range-iterates the map to fill it:

```go
asset.Resources = make([]Resource, len(asset.resources)) // len == N
i := 0
for _, v := range asset.resources {                      // grows to N+1 concurrently
    asset.Resources[i] = *v                              // i == N -> out of range
    i++
}
```

If the map grows between the `len()` sizing and the range completing, the loop overruns the pre-sized slice and panics with `index out of range [N] with length N`.

## Fix

Add a per-asset `sync.Mutex` guarding all access to the `resources`, `connections`, and `IdsLookup` maps:

- `asset_recording.go`: `finalize`, `GetResource`, `RefreshCache`
- `recording.go`: `AddData`, `EnsureAsset` (connection write), `resyncAsset`, `resolveResource`, `GetAssetData`

The lock lives on the `Asset` (not the recording), so distinct assets still scan and finalize in parallel. `EnsureAsset` scopes its lock to the connection write so it does not re-enter through `resyncAsset`. The mutex is unexported and untagged, so `encoding/json` skips it and the recording file format is unchanged.

## Testing

- Added `TestSaveWhileAddingData`, which hammers `AddData` while repeatedly calling `Save()`. It reproduces the exact production panic (same `finalize` line) **before** the fix and passes **after** — plainly and under `-race`.
- Full `providers-sdk/v1/recording` package passes under `-race`.
- `go vet` clean (confirms `Asset` is never copied by value now that it holds a lock).

## Notes

- `merge.go` is intentionally left unguarded: it operates on completed recordings during post-scan merge/clone, not concurrently with `AddData`.

### DIFF
--- a/providers-sdk/v1/recording/asset_recording.go
+++ b/providers-sdk/v1/recording/asset_recording.go
@@ -5,6 +5,7 @@ package recording
 
 import (
 	"fmt"
+	"maps"
 	"sort"
 	"sync"
 
@@ -94,7 +95,19 @@ func (asset *Asset) GetResource(name string, id string) (*Resource, bool) {
 	defer asset.mu.Unlock()
 
 	r, ok := asset.resources[name+keySep+id]
-	return r, ok
+	if !ok {
+		return nil, false
+	}
+
+	// Return a snapshot: callers iterate the Fields map after we release the
+	// lock, while AddData may still be inserting into the live map.
+	clone := &Resource{
+		Resource: r.Resource,
+		ID:       r.ID,
+		Fields:   make(map[string]*llx.RawData, len(r.Fields)),
+	}
+	maps.Copy(clone.Fields, r.Fields)
+	return clone, true
 }
 
 func (asset *Asset) RefreshCache() {
--- a/providers-sdk/v1/recording/asset_recording.go
+++ b/providers-sdk/v1/recording/asset_recording.go
@@ -6,6 +6,7 @@ package recording
 import (
 	"fmt"
 	"sort"
+	"sync"
 
 	"go.mondoo.com/mql/v13/llx"
 	"go.mondoo.com/mql/v13/providers-sdk/v1/inventory"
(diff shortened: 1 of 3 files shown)


tas50:
Good catch — fixed in 03399253a. The asset lock was released inside `resolveResource` before `GetData`/`GetResource` touched `obj.Fields`, leaving the field access unsynchronized against `AddData`.

Changes:
- `resolveResource` now returns the owning `*Asset`, and `GetData` holds `asset.mu` while reading `obj.Fields[field]`.
- `GetResource` returns a **snapshot copy** of the fields under the lock, because callers iterate the returned map after the call returns (`providers/runtime.go` `GetRecording`, `providers/sbom.go`) — handing back the live map would race even if we locked only during the lookup.
- Applied the same snapshot fix to the sibling `Asset.GetResource` (2-arg) path, which had the identical issue at its call sites but wasn't in the original diff.

I also made the race test meaningful for this specific bug: it now hammers a single "hot" resource's fields with concurrent writes and reads. Verified the reader goroutine reports `DATA RACE` under `-race` *without* the fix and passes *with* it; full package is green under `-race`.

tas50:
Addressed the `defer` suggestion in 4d8201b7f — wrapped the connection-map write in a closure with `defer recordingAsset.mu.Unlock()`, keeping the lock scoped away from `resyncAsset` (which takes the same mutex, so a function-level `defer` would deadlock). `gofmt`/`vet` clean, package green under `-race`.

syrull (approved):

PR #9034, opened 2026-07-10
✨ Add support for Azure JWT fetching from a compute environment.

VasilSirakov:
Relies on 2 custom env vars to be injected into the workload that would select the exact identity to use when minting the token and the target audience for which the token applies.

In the future, we can modify this to maybe use parameters, or extend the WIF config json, but for now environment variables should suffice.

### DIFF
--- a/providers-sdk/v1/upstream/tokenauth/azure.go
+++ b/providers-sdk/v1/upstream/tokenauth/azure.go
@@ -17,8 +17,8 @@ import (
 )
 
 const (
-	// envClientID and envTokenAudience are variables this package defines. When set,
-	// they select the identity to mint a token with, and the audience to mint it for:
+	// envClientID and envTokenAudience are custom environment variables used to
+	// select the identity to mint a token with, and the audience to mint it for:
 	//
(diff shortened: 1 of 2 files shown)


providers-sdk/v1/upstream/tokenauth/azure.go
```diff
@@ -6,37 +6,241 @@ package tokenauth
...
+			clientID: os.Getenv(envClientID),
+		})
+	}
+
+	// Cloud Shell predates the app registration flow and mints for the exchange
+	// audience directly, so it needs nothing from the environment. It is always
+	// tried, with whatever audience the caller passed — this is the exact request
+	// this provider sent before App Service support existed, and an empty audience
+	// is preserved rather than guarded so that behavior stays byte-for-byte intact.
+	reqs = append(reqs, cloudShellTokenRequest{
```
preslavgerchev, line 119: why do we try this, isnt the identity endpoint enough?

VasilSirakov: this cloud shell token request was here prior to me adding the identity endpoint support. i've kept it for backwards compatibility

preslavgerchev (approved):

PR #9049, opened 2026-07-12
✨ aws: inspector account status and scan configuration

tas50:
## What

Adds two **posture-plane** accessors to `aws.inspector`. Today the resource only exposes the *findings* plane (`coverages()`, `findings()` — "what did Inspector find"). This adds the *configuration* plane — "is scanning even on, for which scan types, with what settings":

- **`accountStatuses() []aws.inspector.accountStatus`** — per-account, per-region enablement broken out by scan type (EC2, ECR, Lambda standard, Lambda code, code repository) plus the overall activation status. Backed by `BatchGetAccountStatus`.
- **`configurations() []aws.inspector.configuration`** — per-region ECR automated re-scan settings and EC2 scan mode. Backed by `GetConfiguration`. Fields: `ecrRescanDurationDays`, `ecrPullDateRescanDurationDays`, `ecrPullDateRescanMode`, `ecrRescanDurationStatus`, `ecrRescanUpdatedAt`, `ec2ScanMode`, `ec2ScanModeStatus`.

Both are modeled as per-region lists with a `region` field, matching how `aws.guardduty.detectors` / `aws.securityhub.hubs` model regional security-service settings.

## Why

This is the settings surface CIS/posture policies assert on — e.g.:

```mql
# Inspector on for every scan type, every region
aws.inspector.accountStatuses.all(ec2Status == "ENABLED" && ecrStatus == "ENABLED")

# ECR keeps rescanning old images rather than aging them out
aws.inspector.configurations.all(ecrRescanDurationDays == "LIFETIME")
```

The `ecrRescanDurationDays` / `ecrPullDateRescanDurationDays` enums include the recently added `DAYS_3` / `DAYS_7` values.

## Notes

- Both accessors fan out over all regions via the existing jobpool pattern.
- `GetConfiguration` returns `ResourceNotFoundException` ("No CustomerRecord exists") in regions where Inspector was never enabled; this is handled as an empty result (via typed `errors.As`) rather than a hard error, alongside the usual access-denied degradation. Without this, the resource errors out in any account not fully enabled in every region.
- Regenerated `aws.lr.go`, `aws.lr.versions` (new entries at 13.45.1), and `aws.permissions.json` (adds `inspector2:BatchGetAccountStatus`, `inspector2:GetConfiguration`).

## Verification

Tested live against a real account:
- `accountStatuses`: 17 regions, all scan-type statuses populated.
- `configurations`: 2 regions with a persisted customer record returned full field values (`LIFETIME` push-date, `DAYS_90` pull-date, `EC2_SSM_AGENT_BASED`), with clean skips in the ~15 regions that 404.


tas50:
Thanks for the review — addressed in f2c0b962b:

1. **`ecr*Duration` naming** — renamed `ecrRescanDurationDays` → `ecrRescanDuration` and `ecrPullDateRescanDurationDays` → `ecrPullDateRescanDuration` to match the SDK (`RescanDuration` / `PullDateRescanDuration`). Agreed the `Days` suffix was misleading given the values are duration enums (`LIFETIME`, `DAYS_30`, …). Done before the version ships, so no migration.
2. **`accountStatus.id()` uniqueness** — the creator now falls back to the connection account id when the API response's `AccountId` is empty, so the `__id` stays unique. (`AccountId` is a required field on the response, so this is belt-and-suspenders, but cheap.)
3. **Empty-string defaults in `@defaults`** — acknowledged as an FYI; no change. When `GetConfiguration` succeeds but a sub-config is absent the field shows empty, which is the accurate representation. Regions where Inspector was never enabled are already filtered by the `ResourceNotFoundException` guard.

Re-verified live: renamed fields return correctly (`ecrRescanDuration: "LIFETIME"`, `ecrPullDateRescanDuration: "DAYS_90"`).

PR #9063, opened 2026-07-13
✨ Expose SSH key & IPSec tunnel crypto for PQC readiness (azure, gcp, oci, digitalocean)

tas50:
## Summary

Surfaces the classical cryptographic posture needed to assess post-quantum-cryptography (PQC) readiness across four providers. This is **Tier 1** of a cross-provider PQC-visibility gap analysis — the highest-value, lowest-cost wins, each mirroring the pattern already shipped in the `hetzner` and `os` providers.

Two crypto families are covered:

### 1. SSH key algorithm & size

The crux of *"which keys are RSA-2048 vs Ed25519 vs a PQC signature?"*. In every case the key material was already being fetched; this just parses the algorithm and size out of it (no new API calls).

| Provider | Field | Shape |
|---|---|---|
| digitalocean | `digitalocean.sshKey.algorithm`, `.bits` | scalars on the dedicated key resource |
| azure | `...computeService.vm.sshPublicKeys[]` | each dict entry gains `algorithm` + `bits` |
| gcp | `...computeService.instance.instanceSshKeys()` | new `[]dict` from the `ssh-keys` metadata blob |
| oci | `oci.compute.instance.sshAuthorizedKeys()` | new `[]dict` from the `ssh_authorized_keys` metadata |

### 2. OCI IPSec VPN tunnel negotiated crypto

The OCI analog of the AWS Site-to-Site VPN tunnel work — the key-exchange dimension where PQC hybrids plug in. Adds to `oci.network.ipsecConnectionTunnel`:

- `phase1EncryptionAlgorithm()`, `phase1AuthenticationAlgorithm()`, `phase1DhGroup()`
- `phase2EncryptionAlgorithm()`, `phase2AuthenticationAlgorithm()`, `phase2DhGroup()`, `phase2PfsEnabled()`

Negotiated crypto is not populated by the tunnel *list* call, so these are lazily loaded via `GetIPSecConnectionTunnel` and cached (one API call shared across the seven accessors).

## Implementation notes

- SSH parsing uses `golang.org/x/crypto/ssh` (already an indirect dep in each module); unparseable keys yield empty algorithm / `0` bits so weak-key audits can distinguish "unknown" from a real value.
- The gcp `ssh-keys` metadata is `username:<openssh-key>` per line; oci `ssh_authorized_keys` has no username prefix. Both parsers are unit-tested.
- Additive only. Version entries: digitalocean `13.12.1`, gcp `13.30.2`, oci `13.15.1`. Azure's dict enrichment adds no schema entries.

## Testing

- `go build ./...` clean for all four providers.
- `go test` passes for the new gcp (`TestParseInstanceSSHKeys`) and oci (`TestParseAuthorizedKeys`) parser tests.
- Interactive/live verification against real cloud infra still recommended before merge (SSH-key fleets on gcp/oci, an IPSec tunnel on oci).

## Follow-ups (deferred)

- `azure ...networkService.virtualAppliance.sshPublicKey` scalar decomposition (vendor-managed appliance login key, low audit value).
- Tier 2/3 items from the gap analysis (deployed-certificate signature/key algorithm parsing; OCI KMS key length/curve; Azure App Gateway backend TLS & Managed HSM keys) — separate PRs.


tas50:
Addressed in ada62ca85: extracted the shared parser into `providers-sdk/v1/util/sshutil` (`ParsePublicKey` + `ParseAuthorizedKey`). azure/digitalocean/hetzner now call `ParsePublicKey`; the gcp and oci metadata parsers keep their provider-specific username-split/multi-line wrapping and delegate the per-key parse to `ParseAuthorizedKey`. The duplicated copies and their tests are removed, the parser unit tests moved to the shared package, and hetzner's `golang.org/x/crypto` dependency is now indirect.

tas50:
Rebased onto latest main and addressed the review findings in c5e2af383:

- **`sync.Once` caching transient errors** — replaced with a mutex-guarded `fetched` flag set only on success, so a transient `GetIPSecConnectionTunnel` failure now retries on the next access instead of being cached forever. The lock is always taken (no unlocked fast-path read), so the flag isn't a data race.
- **`phase2PfsEnabled` false-vs-unknown** — documented in the `.lr` that `false` also means the tunnel's crypto is unavailable (tunnel down), with a note to combine with `status` when auditing; the negotiated string fields are likewise documented as empty in that case.

Also reworded the tunnel crypto field descriptions to name what each secures (key-exchange / data-channel) rather than protocol-phase jargon. The DigitalOcean `13.12.1` version was already correct. Build + unit tests green.

PR #9065, opened 2026-07-13
✨ Expose KMS key size, firewall TLS-inspection & App Gateway/HSM/VPN crypto for PQC (oci, azure)

tas50:
## Summary

**Tier 3** of the cross-provider PQC-visibility work — additive cryptographic-posture fields that the OCI and Azure SDKs already expose but that MQL did not surface. Follows Tier 1 (#9063).

### OCI

- **`oci.kms.key`** — adds `length` (bits) and `curveId`. The `ListKeys` summary omits the key shape, so both are lazily loaded via a single cached `GetKey`. Lets audits distinguish RSA-2048 from RSA-4096 and read the EC curve.
- **`oci.networkFirewall.policy.decryptionProfile`** (new sub-resource) — models SSL forward-proxy and inbound-inspection profiles: whether unsupported TLS versions/ciphers and expired/untrusted certificates are **blocked in transit**. The certificate-validation controls are forward-proxy-only and surface as null on inbound-inspection profiles. Fetched via `ListDecryptionProfiles` + per-profile `GetDecryptionProfile` (the list carries only name/type).

### Azure

- **`applicationGateway`** — typed `backendHttpSettings` (TLS to the backend + cert validation), `sslProfiles` (per-listener TLS policy + mutual-TLS client auth), `trustedRootCertificates` (with a `keyVaultSecret` ref), and `trustedClientCertificates`; plus `sslProfileId`/`sslProfileName` on the listener. Previously only the frontend SSL policy was typed; the rest lived in a raw `properties` dict.
- **`keyVaultService.managedHsm.keys()`** — enumerates HSM-held keys (kty/keySize/curve) via the `azkeys` data plane, reusing the existing `keyVaultService.key` resource. Degrades gracefully (returns partial + logs) when the data-plane RBAC role is absent. Broadens the shared key-vault-ID regex to accept `*.managedhsm.azure.net`.
- **`virtualNetworkGateway.vpnClientIpsecPolicies()`** — point-to-site client IPsec/IKE crypto, reusing the existing `connection.ipsecPolicy` sub-resource.

## Testing

Per request, all **pure Go logic** is unit-tested (no live cloud needed):
- `TestDecryptionProfileFields` (oci) — the forward-proxy-vs-inbound null-field invariant.
- `TestAzureKeyvaultIdParser` (azure) — extended with `managedhsm.azure.net` cases, locking in the shared-regex change against vault-side regression.
- `TestAzureAppGatewaySSLPolicyFields` (azure) — the SSL-policy flattening helper (nil-safe, skips nil cipher entries).

`go build ./...` passes for both providers; `go.mod`/`go.sum` clean; generated code deterministic; `azure.permissions.json` unchanged (no new ARM SDK client — managed-HSM keys use the existing `azkeys` data plane; the rest reuse already-fetched properties). New `.lr.versions` entries at oci `13.15.1`, azure `13.29.1`.

## Verification status / caveats

- OCI KMS length/curve, OCI decryption profiles, Azure App Gateway, and Azure P2S VPN are all verifiable on standard infra and should be smoke-tested with `mql shell` before merge (no cloud creds were available in this environment, so resources were validated structurally + by unit test + build).
- **Azure Managed HSM keys**: end-to-end verification requires a live Managed HSM (billed hourly, no cheap tier, security-domain activation), so this path ships on code-review + build confidence. The code is a faithful reuse of the verified vault `keys()` path against the API-compatible `azkeys` client. Listing HSM keys also needs a Managed HSM **local RBAC** role (e.g. "Managed HSM Crypto User") distinct from the management-plane read that lists the pool — hence the graceful 401/403 degradation.

## Follow-ups (deferred)

- Deployed-certificate signature/key-algorithm parsing (the Tier 2 cross-provider item — Azure/GCP leaf certs expose thumbprint/expiry but not sig/key algorithm).
- Azure App Gateway L4 `BackendSettingsCollection` (TCP/TLS) and surfacing `sslPolicyName` on the parent gateway.


tas50:
Addressed in 3fb7da362: the managed HSM `keys()` now guards `entry.Attributes` for nil, populating the attribute-derived fields (enabled/created/updated/expires/notBefore/recoveryLevel) only when present and leaving them null otherwise. Good catch on the panic risk for sparse key summaries.

PR #9084, opened 2026-07-16
🐛 terraform: resolve jsonencode([...]) list arguments

### LINKED ISSUE #9079: terraform provider: `jsonencode([...])` list argument evaluates to empty
## Summary

A Terraform argument assigned with `jsonencode([ ... ])` (a JSON **list**) evaluates to an empty value in MQL, so a check that iterates the encoded list sees nothing and passes vacuously.

## Reproduction

```hcl
resource "aws_ecs_task_definition" "ex" {
  family                = "ex"
  container_definitions = jsonencode([
    { name = "app", image = "x", privileged = true }
  ])
}
```

Query:

```coffee
terraform.resources("aws_ecs_task_definition").all(
  arguments.container_definitions.all(_["privileged"] != true)
)
```

**Expected:** the check fails (a privileged container is defined).
**Actual:** `arguments.container_definitions` is empty, so the inner `.all(...)` is vacuously true and the check **passes**.

Note: `jsonencode({ ...map... })` with literal values does resolve (list-wrapped, indexable via `.first`); the empty result is specific to a top-level `jsonencode([list])`.

## Scope

ECS `container_definitions` and other list-valued `jsonencode` arguments across the aws policy.

## Repro fixtures

`KNOWN_BUG.md`-marked `.../fail/jsonencode*` scenarios in cnspec PR #3007.

tas50:
## Summary

A Terraform argument assigned with `jsonencode([ ... ])` (a JSON list) evaluated to an empty value in MQL, so a check iterating the encoded list saw nothing and passed vacuously (e.g. ECS `container_definitions` with a privileged container).

## Root cause

`getCtyValue` fully evaluates `jsonencode(...)` to a JSON string, then re-parses it. The unmarshal target was hardcoded to `map[string]any`:
- `jsonencode({map})` → JSON object → decodes fine → list-wrapped, indexable via `.first`.
- `jsonencode([list])` → JSON array → `json.Unmarshal` into `map[string]any` **fails**; the error was swallowed and the argument stayed empty.

## Fix

Decode into `any` (handles both objects and arrays) and append via the existing `appendCtyResult` helper. The object case stays list-wrapped as before; a decoded top-level list is flattened into the result, so the argument becomes an iterable list.

## Tests

Added to `hcl_test.go`:
- `jsonencode([{name="app", privileged=true}])` → iterable list whose element resolves (`name == "app"`, `privileged == true`)
- `jsonencode({...})` → still list-wrapped map (guards the preserved behavior)

Full `providers/terraform/resources` suite passes.

Fixes #9079

### DIFF
--- a/providers/terraform/resources/hcl.go
+++ b/providers/terraform/resources/hcl.go
@@ -806,10 +806,16 @@ func getCtyValue(expr hcl.Expression, ctx *hcl.EvalContext) any {
 		subVal, err := t.Value(ctx)
 		if err == nil && subVal.Type() == cty.String {
 			if t.Name == "jsonencode" {
-				res := map[string]any{}
-				err := json.Unmarshal([]byte(subVal.AsString()), &res)
-				if err == nil {
-					results = append(results, res)
+				// Decode into `any` so both a JSON object (`jsonencode({...})`)
+				// and a JSON array (`jsonencode([...])`) resolve. Unmarshalling
+				// into a map[string]any silently failed for a top-level array,
+				// leaving the argument empty so a check iterating the encoded
+				// list saw nothing and passed vacuously. appendCtyResult keeps
+				// the object case list-wrapped while flattening a decoded list
+				// into the result so it stays iterable.
+				var res any
+				if err := json.Unmarshal([]byte(subVal.AsString()), &res); err == nil {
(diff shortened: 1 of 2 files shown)


arlimus (approved):
very cool!

PR #9132, opened 2026-07-18
🐛 aws: fix init husks, cloudwatch kmsKey null-state, and S3 wrong-region calls

tas50:
## Summary
AWS Tier-2 correctness fixes from the provider logic-error audit.

- **init husks → not-found errors** (batch computeEnvironment/schedulingPolicy; ec2 internet/transit/egress-only gateways; vpc natgateway/peeringConnection; ec2 capacityReservation; ipam): a targeted Describe-by-id that returned nothing (or access-denied) returned `args, nil, nil`, letting the runtime build a blank husk that surfaces client-side as `no type information`. Converted those concrete-id-miss branches to not-found/access-denied errors (mirrors `initAwsEc2ManagedPrefixList`). The early "args already complete" fast-path and the genuine "no id supplied → bare resource" paths are unchanged.
- **cloudwatch `loggroup.kmsKey()`**: set `StateIsSet | StateIsNull` before returning nil (mirrors `logAnomalyDetector.kmsKey`), fixing null-coercion on the cross-account placeholder path.
- **s3**: ~15 region-aware accessors read raw `a.Location.Data`; a bucket reached lazily by name has empty Location, so the request hits the default region and 301s for any out-of-region bucket. Swapped to the region-aware `bucketRegion()` resolver (which also handles placeholder/cross-account buckets via its `ok` return).

## Deferred
DynamoDB `id` field is always empty, but a proper fix needs a schema change (`id string` → `id() string`) or a per-table `DescribeTable` N+1, so it's out of scope for this Go-only change.

Go-only; no schema/codegen/version changes.

### DIFF
--- a/providers/aws/resources/aws_batch.go
+++ b/providers/aws/resources/aws_batch.go
@@ -1662,8 +1662,11 @@ func initAwsBatchComputeEnvironment(runtime *plugin.Runtime, args map[string]*ll
 	if err != nil {
 		return nil, nil, err
 	}
+	// Returning (args, nil, nil) here would let the runtime create a resource
+	// whose fields are all unset, which surfaces as malformed nil data when
+	// those fields are queried.
 	if len(resp.ComputeEnvironments) == 0 {
-		return args, nil, nil
+		return nil, nil, fmt.Errorf("aws.batch.computeEnvironment with arn %q not found", arnStr)
 	}
 	ce := resp.ComputeEnvironments[0]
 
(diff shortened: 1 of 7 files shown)


tas50:
Addressed: `kmsKey()` now returns the key resolved at creation time and only marks the field null (with explicit state) when no key was set — fixing the unconditional-null regression while keeping the null-state fix for the cross-account placeholder path.

/review

tas50:
Addressed the S3 `kmsKey()` suggestion: it now checks the `ok` return from `bucketRegion()` explicitly and marks the field null (with state) instead of relying on the implicit guard in `fetchEncryptionConfig`.

/review

PR #9134, opened 2026-07-18
🐛 azure: fix policyAssignments pagination, mysql ssl error-handling, nil-derefs, elasticPoolName

tas50:
## Summary
Azure Tier-2 correctness fixes from the provider logic-error audit.

- **armsecurity**: `getPolicyAssignments` read only the first page of `Microsoft.Authorization/policyAssignments`; added `nextLink` pagination. Prevents `cloud_defender` false-negatives (a secured subscription reported as unprotected) when the qualifying assignment spills past page 1.
- **mysql**: `sslEnforcement()` swallowed every config-Get error and returned `true` (secure), masking transient/permission failures; now tolerates only 403/404 and re-raises other errors (mirrors `azureAdOnlyAuthentication`).
- **aks / resource_groups**: guard nullable `Properties` before the early derefs.
- **iam `role()`**: check `.Error` on `GetIam()`/`GetRoles()` instead of nil-deref on a denied listing.
- **sql**: `elasticPoolName` carried the full ARM ID; now the last path segment via a nil-safe helper. Field name/type unchanged.
- **web**: appslot `configuration()` now maps `minTlsCipherSuite` (previously always null on deployment slots).

## Tests
`elasticPoolNameFromID` unit test.

Go-only; no schema/codegen/version changes.

### DIFF
--- a/providers/azure/resources/aks.go
+++ b/providers/azure/resources/aks.go
@@ -140,6 +140,11 @@ func (a *mqlAzureSubscriptionAksService) clusters() ([]any, error) {
 			return nil, err
 		}
 		for _, entry := range page.Value {
+			// Normalize a nil Properties to an empty struct so the many early
+			// accesses below are nil-safe; the later blocks already guard it.
+			if entry.Properties == nil {
+				entry.Properties = &clusters.ManagedClusterProperties{}
+			}
 			storageProfile, err := convert.JsonToDict(entry.Properties.StorageProfile)
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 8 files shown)


tas50:
Addressed both pagination comments in `getPolicyAssignments`:
- **token expiry (armsecurity.go:72):** the bearer token is now fetched per page inside the loop. `conn.GetToken()` wraps an `azcore.TokenCredential`, which caches and only refreshes when the token is near expiry, so a long pagination run can't fail on an expired token and there's no extra network cost per page.
- **empty nextLink (armsecurity.go:99):** the loop now breaks on `page.NextLink == nil || *page.NextLink == ""` instead of assigning an empty URL.

/review

PR #9135, opened 2026-07-18
✨ azure: adopt Cosmos DB v4 RBAC resources + bump provider deps

tas50:
## What

Bumps the Azure **armcosmos SDK across its v3 → v4 major** (plus routine AWS and STACKIT dependency updates), handles the major cleanly, and adopts the new data-plane surface v4 unlocks.

## Dependency bumps

| Provider | Change |
|---|---|
| azure | `armcosmos` **v3.4.0 → v4.0.0** (major) |
| aws | patch/minor bumps: `backup`, `cognitoidentityprovider`, `emr`, `guardduty`, `rds`, `redshift`, `s3`, `sagemaker`, `storagegateway`, `smithy-go` |
| stackit | `kms` v1.11→v1.12, `sqlserverflex` v1.16.0→v1.16.1 |

### Breaking changes in the major

Audited the full v3.4.0→v4.0.0 changelog against our code: **no call-site breakage**. Every removed struct (`ARMProxyResource`, `TrackedResource`, `Resource`, `RestoreParametersBase`, …) is an internal codegen base type we never referenced, and the `ManagedServiceIdentity.UserAssignedIdentities` value-type change is transparent because we only iterate its keys. The import-path bump is the entire fix.

## New capabilities adopted from v4

**Data-plane RBAC for the non-SQL APIs.** v4 adds role definitions and role assignments for Cassandra, Gremlin, Table, and MongoDB managed-instance, parallel to the SQL role RBAC we already model. New accessors on `azure.subscription.cosmosDbService.account`:

- `cassandraRoleDefinitions` / `cassandraRoleAssignments`
- `gremlinRoleDefinitions` / `gremlinRoleAssignments`
- `tableRoleDefinitions` / `tableRoleAssignments`
- `mongoMIRoleDefinitions` / `mongoMIRoleAssignments`

This rounds out the "who can access this account" picture for accounts using those APIs. The four APIs share the same `RoleDefinitionType`/`Permission` SDK types, so the arg-mapping is factored into two shared builders.

**Provenance.** v4 adds `SystemData` to the Cosmos sub-resources, so `systemMetadata()` is now available on the existing SQL sub-resources (`sqlDatabase`, `container`, `sqlRoleDefinition`, `sqlRoleAssignment`) and on every new role resource.

## Notes

- **Enums:** no existing `.lr` enum needed changes — the only new v4 enum types (`AutoReplicate`, `ScheduledEventStrategy`) belong to managed Cassandra clusters, which we don't model. The new role `roleType` reuses the existing closed `BuiltInRole`/`CustomRole` set.
- **Permissions manifest:** added `Microsoft.DocumentDB/databaseAccounts/{cassandra,gremlin,table,mongoMI}Role{Definitions,Assignments}/read` overrides (mirroring the SQL entries), so the regenerated manifest reflects the correct RBAC reads instead of the coarse client-derived form. Verified against Microsoft Learn.
- `permissions` stays `[]dict` (small `{dataActions, notDataActions, id}` struct, no independent key), consistent with the existing SQL role definitions and the sub-resource bar.

## Testing

- `go build ./...`, `go vet`, and cosmos unit tests pass for the azure provider; aws and stackit build clean.
- `.lr.versions` entries land at 13.31.1; generated code and permissions manifest regenerated.


tas50:
Thanks for the review — addressed in b26010d5:

**Duplication (cosmosdb_rbac.go:36).** Collapsed the eight near-identical accessors onto two generic collectors (`collectCosmosRoleDefinitions` / `collectCosmosRoleAssignments`) that own the pager drain + resource creation + SystemData caching. Each accessor now only supplies its API-specific client, pager, item extractor, and cache setter. Also pulled the arg-mapping into pure `cosmosRoleDefinitionArgs` / `cosmosRoleAssignmentArgs` functions and added unit tests covering enum coercion, nil-field defaults, nil scope/permission dropping, and permission-to-dict conversion.

**Typed `roleDefinitionId` (cosmosdb_rbac.go:82).** Agreed in principle. Leaving it as a raw string here to stay consistent with the shipped `sqlRoleAssignment.roleDefinitionId`, and because a typed `roleDefinition()` accessor needs the role-definition resources to support lookup-by-id (an `init`, which they don't have today). Tracking this as a follow-up that converts all five variants (SQL + the four new ones) together, as suggested.

PR #9137, opened 2026-07-18
✨ aws: add Amazon Personalize resources

tas50:
### What

Adds the **Amazon Personalize** namespace to the AWS provider, covering the real-time recommendation service's core control-plane resources for inventory and security posture.

New resources:

| Resource | Highlights |
|---|---|
| `aws.personalize.datasetGroup` | `kmsKey()` + `iamRole()` typed refs, `domain`, `status`, `failureReason` |
| `aws.personalize.dataset` | `datasetType`, `schema()` typed ref |
| `aws.personalize.solution` | `recipe`, `performAutoML`, `performHPO`, `eventType` |
| `aws.personalize.campaign` | deployed inference endpoint: `minProvisionedTPS`, `solutionVersion` |
| `aws.personalize.recommender` | domain recommender: `recipe`, `status` |
| `aws.personalize.eventTracker` | `trackingId` (sensitive), `accountId` |
| `aws.personalize.filter` | `filterExpression` |
| `aws.personalize.schema` | raw Avro `schema`, `domain` |

### Why

Personalize dataset groups carry real security surface — customer KMS encryption at rest, an assumed IAM service role, and a public real-time inference endpoint (campaigns) — none of which was queryable before. This closes a gap in the AWS AI/ML coverage alongside the existing Bedrock and SageMaker namespaces.

Amazon Personalize is an **active** service (confirmed July 2026). This work deliberately excludes the AWS ML services now in maintenance mode / closed to new customers (Kendra, Fraud Detector, Forecast), so we don't model resources AWS is sunsetting.

### How

- Dataset groups and schemas enumerate across all enabled regions via the standard jobpool pattern; sub-collections (datasets, solutions, campaigns, recommenders, event trackers, filters) resolve lazily per parent.
- Detail-only fields that the `List*` summaries omit are populated from the matching `Describe*` call, best-effort on access-denied so a single permission gap doesn't fail the whole collection.
- KMS key, IAM role, and dataset schema are exposed as typed references (raw ARNs cached on the resource's `Internal` struct, resolved via `NewResource`).
- New `personalize` client wired into the connection factory and the permissions extractor; `aws.permissions.json` regenerated (16 new `personalize:*` permissions).

### Verification

- `mqlr generate` clean, provider builds and installs, `go vet` clean, generated files stable, resource registered in the provider schema.
- **Live verification against real infrastructure is pending the batched `provider-verification` run** covering this PR together with the other AI-resource PRs in this series.


tas50:
Thanks for the review — all three addressed in the latest commit:

1. **🔴 nil-schema crash** — `buildSchemaResource` no longer returns `(nil, nil)`; both call sites now guard. The dataset `schema()` accessor sets `StateIsSet | StateIsNull` when `DescribeSchema` returns a nil schema, and `getSchemas` skips nil entries before appending.
2. **🔵 `convertToStr`** — removed; replaced every call with `aws.ToString`.
3. **🔵 silent Describe swallowing** — added `logPersonalizeDescribeWarn`, which logs a warning on any non-access-denied Describe failure across solution, campaign, recommender, event tracker, filter, and dataset. Access-denied stays quiet (expected under least-privilege); everything else is now observable instead of silently defaulting.

PR #9140, opened 2026-07-18
✨ gcp: add Document AI resources

tas50:
### What

Adds the **Document AI** service to the GCP provider, covering deployed processors and their model versions for inventory and security posture.

New resources:

| Resource | Highlights |
|---|---|
| `gcp.project.documentaiService` | `processors()` across the `us` and `eu` locations |
| `gcp.project.documentaiService.processor` | `type`, `state`, `kmsKey()` (CMEK) ref, `processEndpoint`, `defaultProcessorVersion`, `satisfiesPzs` |
| `gcp.project.documentaiService.processor.version` | `state`, `modelType`, `googleManaged`, `kmsKey()` (CMEK) ref, and `deprecated` / `deprecationTime` / `replacementVersion` |

### Why

Document AI processors extract structured data from documents and can be protected with customer-managed encryption keys — none of which was queryable before. This adds to the GCP AI coverage alongside Vertex AI and Discovery Engine.

Two extra angles of value:
- **CMEK visibility** on both processors and versions (resolved to `gcp.project.kmsService.keyring.cryptokey`).
- **Deprecation surfacing.** Document AI legacy processors are being discontinued; the `deprecated` / `deprecationTime` / `replacementVersion` fields on processor versions make it possible to find versions Google has scheduled for retirement before they stop working.

Document AI is an **active** service (confirmed July 2026); only specific legacy processor *versions* are deprecating, which is exactly what this PR makes queryable.

### How

- Processors enumerate per location (`us`, `eu`) via the standard credentials + regional-endpoint pattern; versions resolve lazily per processor using the location parsed from the processor resource name.
- KMS key names are cached on the resource `Internal` struct and resolved through the shared `newKmsCryptoKeyRef` helper.
- `documentai.processors.list` and `documentai.processorVersions.list` added to `gcp.permissions.json` and the validated-permissions list; `TestGCPPermissionsMatchValidatedList` passes.

### Verification

- `mqlr generate` clean, provider builds and installs, `go build` / `go vet` clean, GCP permissions test passes, resource registered in the provider schema.
- **Live verification against real infrastructure is pending the batched `provider-verification` run** covering this PR together with the AWS Personalize PR.


tas50:
Thanks — all four addressed in the latest commit:

1. **🟡 missing doc-comment** — added a two-part title + description block on `gcp.project.documentaiService` (led with the noun per the `.lr` style rules).
2. **🔵 `deprecationTime` nil-typing** — now `llx.TimeDataPtr(nil)` (default) / `llx.TimeDataPtr(timestampAsTimePtr(...))`, matching `createdAt`; no more untyped `NilData`.
3 & 4. **🔵 typed refs for `defaultProcessorVersion` / `replacementVersion`** — done. Both are now `... processor.version` accessors resolved through a `GetProcessorVersion`-backed `init`, so a referenced version that wasn't part of a `versions()` listing populates fully rather than husking. Made the change pre-release specifically to avoid a later breaking string→ref type change.

The typed ref needs `documentai.processorVersions.get`; the extractor derived the singular `documentai.processorVersion.get`, so I added a `gcpPermissionOverrides` entry to emit the correct plural form and updated the validated-permissions list. `TestGCPPermissionsMatchValidatedList` passes.

PR #9209, opened 2026-07-19
✨ atlassian: add confluence blogpost and attachment resources

tas50:
## What

Adds two new Confluence content resources, closing a data-exposure gap: blog posts (a full content type parallel to pages, previously unaudited) and attachments (the file-exfiltration surface inside the wiki).

### `atlassian.confluence.blogpost`
Listed via `atlassian.confluence.space.blogposts()`. Same access posture as `page`: status, space, version, creator/last-editor, version metadata, labels, restrictions, and attachments. Populated eagerly from the list expand (blog posts have a single entry point, so no lazy cache is needed). Restrictions reuse the content-generic `atlassian.confluence.page.restriction` resource.

```
atlassian.confluence.spaces {
  key
  blogposts { title status createdBy { name } hasRestrictions restrictions { operation groupNames } }
}
```

### `atlassian.confluence.attachment`
Accessor `attachments()` on both `page` and `blogpost`. Exposes title, mediaType, fileSize, comment, upload version/uploader, and absolute web + download links — a direct inventory for finding files (credentials, exports, PII) that should not live in the wiki.

```
atlassian.confluence.spaces {
  pages { title attachments { title mediaType fileSize downloadLink } }
}
```

## Implementation notes
- New file `atlassian_confluence_content.go` holds both resources plus shared helpers.
- Extracted the page's inline restriction-fetch loop into a shared `confluenceContentRestrictions` helper; `page.fetchRestrictions` and `blogpost.fetchRestrictions` now both use it (no behavior change to pages).
- Attachments use the v1 `Content.Attachment.Gets` endpoint (returns `ContentScheme`), consistent with the existing v1 page plumbing; `version` is expanded for uploader/timestamp.
- Absolute links (`webUrl`, `downloadLink`) only build when the response base is present, so they stay absolute or empty, never a bare relative path.
- New `.lr.versions` entries at `13.3.10`; no provider `Version` bump. `atlassian.resources.json` is gitignored; no `permissions.json` for this API provider.

## Testing
- `make providers/build/atlassian` — builds clean; new Internal struct + init/id detected by codegen.
- Live verification against a real Confluence instance to be batched with other provider PRs.


tas50:
Addressed the review:

- **`blogpost.type` constant (suggestion)** — clarified the doc to note it's always `"blogpost"`, kept for symmetry with `page.type` (ddec56b).
- **`FileSize` cast (warning)** — verified `ContentExtensionScheme.FileSize` is `int` in the SDK, so `int64(...)` is a safe widening cast that never truncates. No change needed.
- **blogpost/page `attachments()` caching (suggestions)** — the generated `GetAttachments` uses `GetOrCompute`, which caches the first result, and nothing else calls `attachments()` on either resource, so no extra double-check-lock is needed (the lock pattern on restrictions exists only because `restrictions()` and `hasRestrictions()` share one fetch). No change needed.

PR #9216, opened 2026-07-19
🌟 alicloud: add Alibaba Cloud provider (RAM, OSS, ECS, VPC, SLB, RDS, Redis, MongoDB, PolarDB)

tas50:
## Summary

Adds a new **Alibaba Cloud (`alicloud`)** provider covering identity, storage, the network attack path, and managed databases: **35 resources / ~590 fields** across 12 Alibaba Cloud services, built on the official Alibaba Cloud Go SDK v2 (Darabonba per-service clients + the OSS v2 SDK).

This lands the first three planned tranches as a single PR: (1) identity/storage/audit core, (2) network attack-path core, (3) managed data stores.

## Services & resources

**Identity / storage (Tier 1)**
- `alicloud.ram` — users, groups, roles, policies, password policy, access keys, MFA/login-profile, security preferences (global service)
- `alicloud.oss` — buckets with ACL, encryption, versioning, logging, policy, public-access-block, transfer-acceleration, cross-region-replication

**Network attack path (Tier 2)**
- `alicloud.ecs` — instances, disks, images, key pairs, security groups (+ per-rule `permission`), with within-service typed refs (`securitygroup.instances()`, `instance.securityGroups()`) and an `internetExposed()` predicate
- `alicloud.vpc` — networks, vswitches, route tables (+ route entries), NAT gateways, EIP addresses, network ACLs
- `alicloud.slb` — CLB load balancers (+ listeners, backend servers) with an `internetFacing()` predicate

**Managed data stores (Tier 3)** — each with the security-posture accessors (TLS, TDE/encryption, IP whitelist, security groups):
- `alicloud.rds` (ApsaraDB RDS), `alicloud.redis` (ApsaraDB for Redis/Tair), `alicloud.mongodb` (ApsaraDB for MongoDB), `alicloud.polardb` (PolarDB clusters)

## Auth & connection

- Credentials resolved from CLI flags, then the standard Alibaba env vars, then `~/.alibabacloud/credentials`, then an ECS instance RAM role.
- Supports AccessKey, STS token, and **RAM role assumption** (`--role-arn`).
- Per-service, per-region client factory with client caching; region enumeration via ECS `DescribeRegions` (or an explicit `--regions` filter); account UID resolved via STS `GetCallerIdentity` for the platform ID.

## Registration

Registered in `Makefile`, `DEVELOPMENT.md`, `providers/defaults.go`, and `README.md`. New provider version `13.0.0`; all `.lr.versions` entries at `13.0.0` per the new-provider rule.

## Verification

- `go build ./...`, `go vet ./...`, and `gofmt -l` are clean; `make providers/build/alicloud` produces the plugin binary; `mqlr generate` is idempotent (no diff on re-run).
- **Not yet live-verified** against real Alibaba infrastructure (no account available in this environment). Field mappings were written directly against the SDK response structs but should be smoke-tested with the provider-verification flow once an Alibaba account + credentials are available.

## Follow-ups (next on this line of work)

- Cross-service typed edges (`ecs.instance.vpc()` / `.vswitch()`, `slb.loadBalancer.backends()` as typed ECS refs, DB `securityGroups()` as typed refs) — raw IDs are exposed today; typed accessors need lookup-by-id inits on the target resources.
- Per-resource asset discovery (turning each ECS/RDS/etc. instance into its own scannable asset); today discovery resolves the account asset.
- Additional services: ActionTrail (audit), KMS, Security Center, ACK/ACR, Function Compute, ALB/NLB.


tas50:
Thanks for the review — addressed in the latest push:

**Typed-reference gate (§1.5).** Replaced raw `vpcId` / `vSwitchId` strings with typed `vpc()` / `vswitch()` accessors on `ecs.instance`, `rds.instance`, `redis.instance`, `mongodb.instance`, `polardb.cluster`, `slb.loadBalancer`, and the VPC `vswitch` / `routeTable` / `natGateway` / `networkAcl` resources. Since the provider is unshipped, the raw fields are removed (not deprecated). New `resources/typed_refs.go` provides `resolveVpcNetwork` / `resolveVpcVswitch` plus `init`-by-id so a ref to an as-yet-unlisted VPC/vSwitch resolves and fully populates.

**Region-scoped `__id`.** All VPC item resources now key on `region/nativeId`.

**`routeEntries()`** returns an error when the region is unknown instead of a silent empty slice.

**MongoDB `sync.Once`** replaced with double-check locking, so a transient detail-fetch error no longer permanently poisons the instance.

**Page size** raised 50 → 100.

Two intentional retentions (not in the flagged consumer list): `ecs.securitygroup.vpcId` and `vpc.eipAddress.vpcId` stay raw for now — both are themselves network primitives whose `vpcId` reads as an own-attribute, and an EIP is frequently unbound. Happy to type these too if preferred; flagging as a tracked follow-up.

`securityGroupIds []string` remains a raw token list per the §1.5 exception for token-carrying lists.

PR #3626, opened 2024-03-21
⭐️ Signing of Windows Providers

benr:

PR #9305, opened 2026-07-21
🐛 os: don't nest PowerShell inside PowerShell on local Windows

### LINKED ISSUE #9112: 🧹 os: avoid double PowerShell wrap for encoded commands on local Windows
## Problem

On a **local Windows** connection the connection shell is already `powershell -c` (`providers/os/connection/local/local.go:79`). Any command run through `powershell.Encode` (`providers/os/resources/powershell/encode.go`) is prefixed with `powershell.exe -NoProfile -EncodedCommand <b64>`, so the effective invocation becomes:

```
powershell -c "powershell.exe -NoProfile -EncodedCommand <b64>"
```

i.e. **two PowerShell processes** for a single command — an outer shell that only spawns an inner encoded PowerShell. This showed up in a customer Sysmon capture during hypervisor detection (`powershell → powershell → wmic → WmiPrvSE.exe`).

This is **not** specific to hypervisor detection — it affects *every* `powershell.Encode`d command on a local Windows connection. For example `providers/os/resources/windows.go` `computerInfo()` does the same double-wrap.

## Context

The wmic/CIM hypervisor-detection CPU spike this was discovered from is being fixed separately (single CIM query + per-connection caching in `providers/os/id/hypervisor/`). That change removes the repeated WMI sweeps, which were the real CPU cost. The extra PowerShell process is comparatively cheap, so this is filed as a standalone cleanup rather than part of that fix.

## Why it's not a trivial fix

`powershell.Encode` is genuinely required for **WinRM** and **SSH-to-Windows** connections, which run the command directly with no shell wrap (`providers/os/connection/winrm/winrm.go:120`). So the encoding can't simply be removed — the fix needs to be connection-aware: skip the extra `powershell.exe -EncodedCommand` layer when the connection's own shell is already PowerShell (local Windows), while keeping it for WinRM/SSH.

## Suggested approach

- Detect when the target connection's shell is already PowerShell and, in that case, pass the raw script (still prepending `$ProgressPreference='SilentlyContinue';`) through the connection's `RunCommand` without the `powershell.exe -EncodedCommand` wrapper.
- Keep the encoded form for connection types that execute commands directly (WinRM, SSH-to-Windows).
- Audit all `powershell.Encode` call sites (`grep -rn "powershell.Encode" providers/os`) since the change benefits all of them.

M-gre:
## Problem

A customer's SentinelOne agent (26.1.2.177, ~2026-07-17) began classifying `cnspec.exe serve` as **Malware / Suspicious** on Windows 11 Enterprise multi-session hosts. The trigger is the *shape* of how the os provider shells out to PowerShell on the **local** connection.

The threat export shows the actual command lines cnspec spawned:

```
powershell.exe -c "powershell.exe -NoProfile -EncodedCommand <b64>"   ← SMBIOS Get-CimInstance, Azure IMDS Invoke-RestMethod
powershell.exe -c hostname                                            ← os.hostname (single process)
os.exe run_as_plugin --log-level info                                 ← provider plugin (reported NotSigned, see below)
```

That **double `powershell.exe`** is a textbook LOLBIN-chain + encoded/obfuscated-command + AMSI-evasion pattern, which the new S1 heuristics score on. It raised: *"An encoded PowerShell command was detected"*, *"Indirect command was executed"*, *"Lolbins were chained together"*, *"Suspicious WMI query"* (Get-CimInstance), *"PowerShell encoded command connected to the web"* (IMDS Invoke-RestMethod).

## Root cause

The local Windows connection defaults its shell to `powershell -c` (`local.go`). But every `powershell.Encode()` / `powershell.Wrap()` helper already returns a **complete** PowerShell invocation. `CommandRunner.Exec` then nests it inside a second `powershell -c "…"`, spawning a redundant powershell.exe.

The **SSH** connection runs the same command raw (one process), so it was never affected — this is unique to the local connection (i.e. `cnspec serve` scanning its own host). It's a long-standing pattern; S1's update newly flags it, so this is not a regression on our side.

## Fix

- `powershell.SplitInvocation()` — parses the output of `Encode` / `EncodeUnix` / `Wrap` back into an argv (single source of truth for the emitted format).
- `unnestPowershell()` in the local runner — when the shell is powershell/pwsh **and** the command is already a self-contained PowerShell invocation, exec it directly → **one** process: `powershell.exe -NoProfile -EncodedCommand <b64>`.

Plain commands (`hostname`) still route through `powershell -c`; unix `sh -c` is untouched.

## Testing

- Unit tests in `encode_test.go` cover Encode/EncodeUnix/Wrap round-trips + negative cases.
- `go build ./...`, `go vet`, `gofmt` clean.
- ✅ **Verified live on Windows Server 2022** that the double-PowerShell nesting is eliminated (see comment below: 2 processes → 1). Draft pending confirmation with the actual SentinelOne agent that the indicators clear (test box has no S1).

## Out of scope (separate follow-ups)

1. **Unsigned provider binary.** The trace reports `os.exe` as `NotSigned`, raising the *"signed process loaded a suspicious unsigned library"* indicator. The release pipeline *has* signed Windows providers via Azure Trusted Signing since 2024 (`scripts/provider_bundler.sh`, #3626), so the customer likely runs an os-provider version predating signing — needs its own investigation.
2. **Prefer native APIs** over spawning PowerShell for the hot paths (SMBIOS, IMDS, hostname) to reduce the PowerShell signal further.

The other S1 indicators in the export (Chrome memory access, RDP credential dumping, process hollowing, root-cert install) are storyline over-correlation with no matching command in the trace.

### DIFF
--- a/providers/os/connection/local/local.go
+++ b/providers/os/connection/local/local.go
@@ -20,6 +20,7 @@ import (
 	"go.mondoo.com/mql/v13/providers/os/connection/shared"
 	"go.mondoo.com/mql/v13/providers/os/connection/ssh/cat"
 	"go.mondoo.com/mql/v13/providers/os/registry"
+	"go.mondoo.com/mql/v13/providers/os/resources/powershell"
 )
 
 type LocalConnection struct {
@@ -204,7 +205,15 @@ func (c *CommandRunner) Exec(usercmd string, args []string) (*shared.Command, er
 	var cmd string
 	cmdArgs := []string{}
 
-	if len(c.Shell) > 0 {
+	// When the default shell is PowerShell and the command is already a
+	// self-contained PowerShell invocation (as produced by powershell.Encode /
+	// powershell.Wrap), run it directly. Otherwise it would be nested inside a
+	// second `powershell -c "..."`, spawning a redundant powershell.exe that
+	// endpoint protection flags as an encoded-command / LOLBIN-chain attack.
+	if argv, ok := unnestPowershell(c.Shell, usercmd); ok {
+		cmd = argv[0]
+		cmdArgs = append(cmdArgs, argv[1:]...)
+	} else if len(c.Shell) > 0 {
 		shellCommand, shellArgs := c.Shell[0], c.Shell[1:]
 		cmd = shellCommand
 		cmdArgs = append(cmdArgs, shellArgs...)
@@ -246,4 +255,21 @@ func (c *CommandRunner) Exec(usercmd string, args []string) (*shared.Command, er
 
 	// all other errors are real errors and not expected
 	return &c.Command, err
+}
+
+// unnestPowershell returns the argv to run usercmd directly, without wrapping
+// it in the shell, when both the shell and the command are PowerShell. This
+// avoids spawning powershell.exe inside powershell.exe. It only fires for the
+// PowerShell shell (Windows local default) so unix `sh -c` behavior and plain
+// commands (e.g. `hostname`) are untouched.
+func unnestPowershell(shell []string, usercmd string) ([]string, bool) {
+	if len(shell) == 0 {
+		return nil, false
+	}
+	switch strings.ToLower(shell[0]) {
+	case "powershell", "powershell.exe", "pwsh", "pwsh.exe":
+		return powershell.SplitInvocation(usercmd)
+	default:
+		return nil, false
+	}
 }
--- a/providers/os/resources/powershell/encode.go
+++ b/providers/os/resources/powershell/encode.go
@@ -6,10 +6,68 @@ package powershell
 import (
 	"encoding/base64"
 	"fmt"
+	"strings"
 
 	"github.com/rs/zerolog/log"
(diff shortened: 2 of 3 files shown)


M-gre:
## Verified live on Windows Server 2022

Built the identical verifier from `main` (broken) and from this branch (fixed) — each calls the real `local.NewConnection().RunCommand(powershell.Encode(script))` — and ran both on a live Windows host. The encoded script reports its own process-ancestry chain. Exit code 0 in both cases (functionality preserved):

| | Process chain | `powershell.exe` count |
|---|---|---|
| **main (broken)** | `powershell.exe(3996) ← powershell.exe(3856) ← verify_broken.exe` | **2 (nested)** |
| **this branch (fixed)** | `powershell.exe(5036) ← verify_fixed.exe` | **1** |

The broken baseline reproduces the customer's exact `powershell.exe -c "powershell.exe -EncodedCommand …"` nesting; the fix collapses it to a single `powershell.exe` whose parent is the calling process directly. This removes the double-PowerShell LOLBIN pattern on a real host.

PR #9306, opened 2026-07-21
⚡ os: use native APIs for hostname, IMDS, and SMBIOS on local connections

M-gre:
## Motivation

Follow-up to #9305 (SentinelOne flagging `cnspec.exe serve` on Windows). That PR stopped nesting PowerShell inside PowerShell; this one removes the PowerShell spawn entirely for three hot paths when scanning the **local** host, so there's less for endpoint protection to flag and it's faster.

All three gate on `conn.Type() == shared.Type_Local`. Remote connections (SSH / WinRM) keep the existing command paths, so there is **no behavior change for remote scans**. Follows the established `platformid.windowsMachineId` / `detector/windows.GetWmiInformation` pattern (local-native + PowerShell fallback, build-tag split).

## Changes

| Path | Local now uses | Fallback / remote |
|---|---|---|
| **hostname** | `os.Hostname()` in-process | `powershell -c hostname` |
| **Azure IMDS** | direct Go HTTP GET, returns raw JSON like the unix `curl` path | `Invoke-RestMethod` (win) / `curl` (unix) |
| **SMBIOS** | WMI via `StackExchange/wmi` (`Win32_Bios`/`BaseBoard`/`SystemEnclosure`/`ComputerSystemProduct`) | encoded PowerShell script, also used if the native query fails |

New files: `resources/smbios/smbios_windows.go` (native WMI), `resources/smbios/smbios_other.go` (fallback dispatch). SMBIOS `win.go` refactored so the mapping to `SmBiosInfo` is shared between both paths.

## Verified live on Windows Server 2022 (EC2)

Ran the **real** package code path (`local.NewConnection` → `smbios.ResolveManager` → `nativeWindowsSmbios`, and `hostname.Hostname`) on an actual Windows host:

- `hostname` → `EC2AMAZ-B96TDOB` ✓
- SMBIOS native → BIOS `Amazon EC2 / 1.0`, SysInfo `t3.medium` + UUID, Chassis `Type=1` — **matches the PowerShell path field-for-field**.
- IMDS parsing covered by a new httptest-backed unit test (`TestLocalHTTPMetadata`).

### Bug caught by live testing (second commit)

`Win32_SystemEnclosure.ChassisTypes` comes back from the COM SAFEARRAY as `VT_I4`, so `[]uint16` makes the `StackExchange/wmi` lib **panic** (`reflect.Value.Uint on int32`) — which would crash the scan. Fixed to `[]int32`, and the native path is now wrapped in `defer/recover` so any future WMI surprise falls back to PowerShell instead of panicking.

## Notes

- `bios.releaseDate` on local Windows is now RFC3339 (`2017-10-16T00:00:00Z`) instead of the old PowerShell `/Date(1508112000000)/`. Net improvement; local-Windows only (remote unchanged).
- Draft because IMDS could not be exercised against real Azure infra (test box is EC2); the parse path is unit-tested and the SMBIOS/hostname paths are fully verified on live Windows.

Builds + `go vet` clean on darwin and `GOOS=windows`; smbios/hostname/azcompute test suites pass.

### DIFF
--- a/providers/os/id/azcompute/azcompute.go
+++ b/providers/os/id/azcompute/azcompute.go
@@ -9,7 +9,9 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"strings"
+	"time"
 
 	"github.com/rs/zerolog/log"
 	"go.mondoo.com/mql/v13/providers-sdk/v1/inventory"
@@ -17,6 +19,9 @@ import (
 	"go.mondoo.com/mql/v13/providers/os/resources/powershell"
 	"go.mondoo.com/mql/v13/utils/multierr"
 )
+
+// var so tests can override the endpoint
+var imdsBaseURL = "http://169.254.169.254"
 
 const (
 	// https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=windows#supported-api-versions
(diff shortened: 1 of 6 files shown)


M-gre:
Addressed the review in 83aff567d:

- 🔴 **IMDS status check** — `httpMetadata` now errors on any non-2xx response (with status + body) instead of returning the error body as if it were valid metadata. Covered by a new `TestLocalHTTPMetadataErrorStatus`.
- 🔵 **HTTP client reuse** — the `http.Client`/`Transport` is now a package-level var reused across the instance + loadbalancer calls.
- 🟡 **`imdsBaseURL` global** — replaced with an `imdsBaseURL` field on `commandInstanceMetadata` (defaults to `defaultIMDSBaseURL`); the test sets the field, no shared mutable global.
- 🟡 **shadowed `err`** — renamed the inner `wmi.Query` error to `qErr` in all four queries.

Builds + `go vet` clean on darwin and `GOOS=windows`; azcompute/smbios/hostname tests pass.

PR #9308, opened 2026-07-21
🧹 Protect resource init from wrong ARN service.

VasilSirakov:
## Problem

Some cnspec runs produced malformed, unattributed errors like:

```
x provider returned no data and no error for a field; the field was never set on the resource (provider bug) field=tags id=arn:aws:logs:<region>:<acc-id>:log-group:/aws/apigateway/<resource>:* provider=aws resource=aws.msk.cluster

x llx: encountered a primitive with no type information, coercing to null (an upstream layer — a provider or the compiler — produced a malformed primitive)

! cross-account CloudWatch log group reference arn=arn:aws:route53:::hostedzone/<zone> currentAccount=<accId> logGroupAccount=
```

Root cause: a singular resource `init` that resolves itself from an ARN (e.g. `aws.msk.cluster`, `aws.es.domain`, `aws.cloudwatch.loggroup`) only nil-checked `args["arn"]` — it never verified the ARN actually belonged to that service. So an ARN from an *unrelated* service could be accepted on either path:

- **discovery adoption** — a bare singular query like `aws.msk.cluster` adopts whatever ARN the currently-scanned asset carries (a Route53 hosted zone, a CloudWatch log group, …), regardless of service;
- **explicit arg** — `aws.msk.cluster(arn: "arn:aws:logs:...")`.

The wrong ARN then builds a "husk" resource keyed on it with every field left **unset** (not null — unset). Those fields cross the plugin boundary as malformed nil data, which is exactly the errors above — and with no attribution back to the init at fault.

Example (old):
image

Example (new):
image

## Fix

Each ARN-resolving init now declares a small `arnSpec` next to it with the ARN service token(s) it accepts, and calls `resolve()`, which:

- parses the ARN and rejects it early if its service (and optional resource-segment prefix) doesn't match — on **both** the adoption and explicit-arg paths, so a foreign ARN produces a clean, attributed error instead of a husk;
- reads the value with comma-ok, so a non-string `arn` arg returns an error instead of panicking the init.

Service tokens were cross-referenced against [AWS's policies.js](https://awspolicygen.s3.amazonaws.com/js/policies.js). The only deviation is `vpc` instead of `ec2`, because we construct that ARN ourselves rather than using the real service token. (The one token that couldn't be settled from that file, `es` vs `opensearch` for OpenSearch domains, was confirmed as `es` against a real domain ARN.)

## Old vs new

Scanning a Route53 hosted-zone asset while a policy references `aws.msk.cluster`:

- **Before:** adopts `arn:aws:route53:::hostedzone/Z…` as the cluster's ARN → husk resource → querying any field yields
  `provider returned no data and no error for a field … resource=aws.msk.cluster` + `llx: encountered a primitive with no type information`.
- **After:** the foreign ARN is not adopted → clean, attributed error:
  `arn required to fetch aws.msk.cluster`

And an explicitly-passed wrong-service ARN, e.g. `aws.es.domain(arn: "arn:aws:iam::…:role/x")`:

- **Before:** slips past the nil-check into the SDK →
  `operation error Elasticsearch Service: DescribeElasticsearchDomain, failed to resolve service endpoint … Invalid Configuration: Missing Region`
- **After:**
  `arn "arn:aws:iam::…:role/x" is not an aws.es.domain arn: service is "iam", expected es`

### DIFF
--- a/providers/aws/resources/aws_apigateway.go
+++ b/providers/aws/resources/aws_apigateway.go
@@ -128,14 +128,9 @@ func initAwsApigatewayRestapi(runtime *plugin.Runtime, args map[string]*llx.RawD
 		return args, nil, nil
 	}
 
-	if len(args) == 0 {
-		if assetArn := getAssetIdentifier(runtime); assetArn != "" {
-			args["arn"] = llx.StringData(assetArn)
-		}
-	}
-
-	if args["arn"] == nil {
-		return nil, nil, errors.New("arn required to fetch gateway restapi")
+	arnVal, err := resolveArnArg(runtime, args, "gateway restapi", "apigateway")
+	if err != nil {
+		return nil, nil, err
 	}
 
 	obj, err := CreateResource(runtime, ResourceAwsApigateway, map[string]*llx.RawData{})
@@ -149,7 +144,6 @@ func initAwsApigatewayRestapi(runtime *plugin.Runtime, args map[string]*llx.RawD
 		return nil, nil, rawResources.Error
 	}
 
-	arnVal := args["arn"].Value.(string)
 	for _, rawResource := range rawResources.Data {
 		restApi := rawResource.(*mqlAwsApigatewayRestapi)
 		if restApi.Arn.Data == arnVal {
--- a/providers/aws/resources/aws_apigateway.go
+++ b/providers/aws/resources/aws_apigateway.go
@@ -123,12 +123,17 @@ func (a *mqlAwsApigateway) getRestApis(conn *connection.AwsConnection) []*jobpoo
 	return tasks
 }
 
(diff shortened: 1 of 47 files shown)


providers/aws/resources/aws_apigateway.go
```diff
@@ -123,19 +123,19 @@ func (a *mqlAwsApigateway) getRestApis(conn *connection.AwsConnection) []*jobpoo
 	return tasks
 }
 
+var apigatewayRestapiArnSpec = arnSpec{
+	resource: ResourceAwsApigatewayRestapi,
+	services: []string{"apigateway"},
```
preslavgerchev, line 128: any chance theres an aws const for that

PR #9320, opened 2026-07-21
✨ os: add file, user, network, and platform security resources

chris-rock:
Adds a batch of native MQL resources for common host security-posture data. Each is a small, self-contained addition with an ADR under `docs/adr/`.

## Resources

| Resource | What it exposes |
|---|---|
| `file.md5` / `file.sha1` / `file.sha256` | File content digests (single read → all three) |
| `file.signature` | Code signature — macOS `codesign`, Windows Authenticode; `null` on Linux |
| `user.knownHosts` (+ `knownhosts`) | Parsed SSH `known_hosts` (hashed hosts, `@cert-authority`/`@revoked` markers) |
| `user.shellHistory` | bash / zsh / fish / PowerShell PSReadLine history (timestamps where the shell records them) |
| `network.hosts` / `network.protocols` / `network.services` | Parsed hosts, protocols, and services files, as getters on the `network` resource |
| `windows.driver` | Installed device drivers with signing status (`Win32_PnPSignedDriver`) |
| `macos.screenlock` | Screen-saver password requirement + delay (fail-closed) |

## Example queries

```mql
file("/usr/bin/sudo").sha256
file("C:/Windows/System32/cmd.exe").signature { signed verified authority }
users.where(name == "root").first.knownHosts.all(isHashed)
network.hosts.where(ip == "0.0.0.0")
network.services.where(name == "https") { port protocol }
windows.drivers.where(signed == false)
macos.screenlock.askForPassword == true && macos.screenlock.askForPasswordDelay <= 5
```

## Design notes

- **Attach vs. new:** digests/signature attach to `file`; `knownHosts`/`shellHistory` attach to `user` (mirroring `authorizedkeys`/`sshkeys`); the network databases are getters on `network` (grouping with `interfaces`/`routes`/`neighbors`, and avoiding a name clash with the system `services` resource).
- `file.signature` unifies the two platform code-signing models behind one field, with a `format` discriminator (`codesign` / `authenticode`).
- The network-database list resources keep `file`/`content` as pure getters (not `list()` deps) so `.where()` works — a resource-typed field can't round-trip through the filter's `StoreData`.

## Verification

- Builds and **cross-compiles for macOS, Linux, and Windows**; `go vet` clean.
- Parser unit tests for the tricky bits (comment stripping, zsh/bash/fish history timestamps, Windows `/Date()/` driver dates).
- **Runtime-verified on all three platforms** (native macOS, Linux via a container-image scan, and native Windows + Windows-over-SSH): digests match `shasum`/`Get-FileHash`; `file.signature` matches `codesign`/`Get-AuthenticodeSignature`; `windows.drivers` count matches `Win32_PnPSignedDriver`; the services file parses on Windows (277 entries).

ADRs: `docs/adr/029`–`033`.


chris-rock:
Thanks for the review — addressed in `f6c0ea3`:

- **🔴 `file.signature` id collision (critical):** the resource id is now scoped to the file path via an explicit `__id` (`file.signature/<path>`), so files sharing an authority (e.g. every Apple binary under `Software Signing`) no longer collide in the cache. The colliding field-based `id()` was removed. Verified: `file("/bin/ls").signature.signed` and `file("/etc/hosts").signature.signed` resolve independently in one query.
- **Copyright headers:** switched all new files to the required `// Copyright Mondoo, Inc. 2024, 2026` format.
- **`readDefault` shell quoting:** added a comment that it takes trusted literals only (the two fixed `com.apple.screensaver` keys).
- **codesign timestamp locale-dependence:** documented that the parse is best-effort and leaves `timestamp` null on a non-English locale (the signed/verified/authority fields are unaffected).
- **ADR 030 vs. schema:** aligned the ADR with the implementation — `knownhosts.entry` has no `file` field (the source file lives on the parent `knownhosts` resource), `file()`/`content()` are pure getters (so `.where()` works), and `shellHistory.command` carries `line`.
- **Weak-hash scanning (md5/sha1):** dismissed as intentional — these digests are for file identification / IOC matching (osquery `hash` parity), not security-sensitive hashing; the provider already uses `crypto/md5` in `fsutil` for the same purpose. Added a code comment noting this.

PR #9336, opened 2026-07-21
fix(network): http.get.header resolves to null when the target has no HTTP

chris-rock:
## Problem

`http.get.header` returned the transport **error** whenever the request failed. So on a host that serves no HTTP — mail servers, DNS servers, infra subdomains, all commonly surfaced by attack-surface enumeration — every `http.get.header.*` access **errors**, and there's no non-erroring field to gate on. Downstream (e.g. the `mondoo-http-security` policy) this turns into ~10 "could not be evaluated" noise findings per non-web host.

## Fix

Mirror what the `tls` resource already does. `tls.params` returns **null** (not an error) when TLS isn't enabled, which is why policies can gate with `tls.params != empty` and skip cleanly. Do the same for HTTP:

- Add `httpNotReachable(err)` — classifies **connection-level** failures (connection refused/reset, dial timeout, DNS failure) via `errors.As` on `*net.OpError` / `*net.DNSError` / `net.Error.Timeout()`.
- In `header()`, on such a failure set `http.get.header` to null and return `nil` instead of the error. **Genuine** errors (e.g. redirect loops) still surface.

Policies can then gate HTTP-header checks on `http.get.header != empty` and **skip** non-HTTP hosts instead of erroring.

## Validation

Unit test (`TestHttpNotReachable`) covers refused / reset / DNS / timeout → true, and redirect-loop / generic → false.

End-to-end with a locally-built provider + `cnspec` + the `mondoo-http-security` policy, across a host matrix:

| Host | Before | After |
|---|---|---|
| a.gtld-servers.net, mx.zoho.com (no HTTP) | ~10 errored checks | **skipped** |
| example.com, github.com (web) | 9 fail / 5 pass | unchanged |
| neverssl.com (HTTP-only, reachable) | 10 fail / 4 pass | unchanged |

Confirmed directly: `http.get.header` → `null` and `http.get.header != empty` → `false` (not an error) on a no-HTTP host.

`go test ./providers/network/resources/` green; `gofmt`/`go vet` clean.

## Note

Paired with a policy change in `mondoo-http-security` that adds `http.get.header != empty` to the header group filter — that policy change depends on this provider behavior, so it should land only once a release carrying this fix is in use.

### DIFF
--- a/providers/network/resources/http.go
+++ b/providers/network/resources/http.go
@@ -122,10 +122,11 @@ func (x *mqlHttpGet) do() error {
 // httpNotReachable reports whether err means the target did not speak HTTP on
 // this endpoint — nothing was listening, the connection was refused/reset, the
 // dial timed out, or the name did not resolve — as opposed to an unexpected
-// internal error. In that case http.get.header resolves to null (like the tls
-// resource's params when TLS is not enabled), so policies can gate on
-// `http.get.header != empty` and skip instead of erroring on non-HTTP hosts
-// (mail/DNS/infra servers commonly surfaced by attack-surface enumeration).
+// internal error. In that case the http.get fields (header, statusCode, version,
+// body) all resolve to null (like the tls resource's params when TLS is not
+// enabled), so policies can gate on `http.get.header != empty` and skip instead
+// of erroring on non-HTTP hosts (mail/DNS/infra servers commonly surfaced by
+// attack-surface enumeration).
 func httpNotReachable(err error) bool {
 	if err == nil {
 		return false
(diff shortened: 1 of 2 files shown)


chris-rock:
Good catch — fixed in d1ec9da3c. `statusCode()`, `version()`, and `body()` now apply the same `httpNotReachable` classification as `header()`, so all four `http.get` fields resolve to null (not a hard error) on a target that doesn't speak HTTP.

Verified against a no-HTTP host:
```
http.get: {
  statusCode: null
  body: null
  header: null
  version: null
}
```
Genuine errors (redirect loops, etc.) still surface on all of them. `go test ./providers/network/resources/` green; gofmt/vet clean.

PR #9344, opened 2026-07-22
✨ alicloud: bind singular resources to their discovered asset scope

tas50:
## Summary

Fine-grained discovery (#9239) emits a per-object asset for each ACK cluster, ALB, NLB, VPC, WAF instance, and Cloud Firewall, stamping the object id (`cluster-id`, `alb-id`, `vpc-id`, `waf-instance-id`, …) and region on the child connection. **Nothing consumed those scope options**, so a bare `alicloud.cs.cluster` (and the ALB/NLB/VPC/WAF equivalents) could not resolve to the scanned object — a policy check on one of these assets had to enumerate the account-wide collection (`alicloud.cs.clusters.all(...)`) every time.

This wires the scope through so the singular resources bind directly to the asset they are scanned against.

## What changed

- **`AlicloudConnection.ScopedObject(option)`** returns the pinned object id + region from the child asset's connection config (`ok=false` on the account root or any asset not scoped to that object kind).
- **`scopedInitArgs`** shared helper backfills a singular resource's identifying arguments from that scope when the resource is invoked bare (no args). Wired into the `cs.cluster`, `alb.loadBalancer`, `nlb.loadBalancer`, and `vpc.network` inits.
- **New `initAlicloudWafInstance`** — scope-only, since `DescribeInstance` returns the region's single instance and has no by-id form.
- Cloud Firewall already binds bare as a service resource (`alicloud.cloudFirewall`), so no change there.
- **New `alicloud.vpc.network.flowLogs`** field, filtered server-side to the VPC (`ResourceType=VPC, ResourceId=<vpc>`), so a VPC-scoped asset asserts its own flow-log coverage instead of scanning every flow log in the account.

Behavior on the account root asset is unchanged: bare access falls through to the existing by-id lookup or required-argument error.

| Platform | Bare resource |
|---|---|
| `alicloud-ack-cluster` | `alicloud.cs.cluster` |
| `alicloud-alb-loadbalancer` | `alicloud.alb.loadBalancer` |
| `alicloud-nlb-loadbalancer` | `alicloud.nlb.loadBalancer` |
| `alicloud-vpc` | `alicloud.vpc.network` |
| `alicloud-waf-instance` | `alicloud.waf.instance` |
| `alicloud-cloud-firewall` | `alicloud.cloudFirewall` |

## Testing

- New `TestScopedObject` covers the scoped / account-root / region-fallback / nil-options paths.
- `go build ./...`, `go vet ./...`, and `go test ./...` pass for the alicloud module.
- Provider version bumped `13.2.0 → 13.2.1` to match the new field's stamped min-version.

A companion cnspec PR rewrites the Alibaba policy's per-object checks to the singular forms this enables.

### DIFF
--- a/providers/alicloud/config/config.go
+++ b/providers/alicloud/config/config.go
@@ -14,7 +14,7 @@ import (
 var Config = plugin.Provider{
 	Name:            "alicloud",
 	ID:              "go.mondoo.com/mql/providers/alicloud",
-	Version:         "13.2.0",
+	Version:         "13.2.1",
 	ConnectionTypes: []string{provider.DefaultConnectionType},
 	Connectors: []plugin.Connector{
 		{
--- a/providers/alicloud/config/config.go
+++ b/providers/alicloud/config/config.go
@@ -14,7 +14,7 @@ import (
 var Config = plugin.Provider{
 	Name:            "alicloud",
 	ID:              "go.mondoo.com/mql/providers/alicloud",
-	Version:         "13.2.1",
+	Version:         "13.2.0",
 	ConnectionTypes: []string{provider.DefaultConnectionType},
 	Connectors: []plugin.Connector{
 		{
--- a/providers/alicloud/connection/connection.go
+++ b/providers/alicloud/connection/connection.go
@@ -166,6 +166,29 @@ func (c *AlicloudConnection) AccountID() string {
 	return c.accountID
 }
 
+// ScopedObject returns the object id and region a discovered child asset is
+// pinned to for the given scope option (OptionClusterID, OptionAlbID, ...). Fine-
+// grained discovery stamps the object id under the scope key and the object's
+// region under OptionRegions, so a singular resource's init can resolve itself
+// from the connection when invoked with no identifying arguments. ok is false on
+// the account root asset (or any asset not scoped to this kind of object), where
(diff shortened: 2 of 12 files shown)


tas50:
Addressed both review comments in f0413d7f6:

- **Doc comment split** — moved the `scopedInitArgs` doc comment above its own function and re-attached `requiredStringArg`'s comment to its function, so godoc is correct for both.
- **Version bump** — reverted `config.go` `Version` back to `13.2.0` per the feature-PR convention (CLAUDE.md); the `.lr.versions` entry for `alicloud.vpc.network.flowLogs` stays at `13.2.1`.

`go build`/`go vet`/`go test` and `gofmt` all clean.

tas50:
Addressed the new review comment in 99018ab1e:

- **waf.instance scope not verified** — `DescribeInstance` returns whichever instance the region hosts rather than looking one up by id, so the scoped id went unchecked. Added a guard comparing `resp.Body.InstanceId` to the scoped `instanceID`, erroring on a mismatch (matching the by-id "not found" contract of the other singular inits).

(The two earlier comments were the re-posted originals, already resolved in f0413d7f6.) `gofmt`/`go build`/`go vet`/`go test` all clean.

PR #9393, opened 2026-07-23
🐛 aws: populate aws.iam.role.lastUsedAt/lastUsedRegion via GetRole

syrull:
## Problem

`aws.iam.role.lastUsedAt` and `aws.iam.role.lastUsedRegion` are **always `null`/`""` for every role**, so any policy relying on role last-used activity (e.g. flagging roles unused for more than N days) is silently broken.

Root cause: `aws.iam.roles` enumerates roles with the IAM `ListRoles` API, and `ListRoles` **does not populate the `RoleLastUsed` field** — it is only returned by `GetRole`. The mapping in `roles()` (`if role.RoleLastUsed != nil { … }`) was therefore always skipped.

Reproduction (any account with at least one used role):

```
ListRoles → role.RoleLastUsed == nil          # lastUsedAt/lastUsedRegion end up null/""
GetRole   → role.RoleLastUsed.LastUsedDate set # the real value
```

## Fix

Make `lastUsedAt` and `lastUsedRegion` **computed fields** backed by a memoized `GetRole`, mirroring the existing `aws.kms.key.lastUsedAt` lazy-getter pattern:

- `resources/aws.lr`: `lastUsedAt time` → `lastUsedAt() time`; `lastUsedRegion string` → `lastUsedRegion() string`.
- `resources/aws_iam.go`: drop the dead `RoleLastUsed` block from `roles()`; add a `mqlAwsIamRoleInternal` cache plus a memoized `getRoleDetails()` helper and the two getters.
- Regenerated `resources/aws.lr.go`.

The `GetRole` call is only issued when one of these two fields is actually queried, and both fields share a single cached response, so enumerating roles for other fields still costs one `ListRoles` page as before. No new IAM permission is required (`iam:GetRole` is already used by `permissionsBoundary()` / the role init path).

## Verification

Built and installed the provider, scanned a live AWS account:

- Before: `lastUsedAt` was `null` for all roles.
- After: `lastUsedAt`/`lastUsedRegion` match the values returned by `aws iam get-role` exactly, including `null` for never-used roles.
- A "role used within the last 90 days" check now evaluates per-role and correctly identifies the single stale role in the test account instead of collapsing to a scalar.

### DIFF
--- a/providers/aws/resources/aws.lr
+++ b/providers/aws/resources/aws.lr
@@ -2573,9 +2573,9 @@ private aws.iam.role @defaults("arn name") {
   // AWS account IDs outside this account that the trust policy allows to assume the role
   assumableByExternalAccounts() []string
   // Time when the role was last used to make an AWS request
-  lastUsedAt time
+  lastUsedAt() time
   // Region in which the role was last used
-  lastUsedRegion string
+  lastUsedRegion() string
   // Maximum session duration in seconds for the role (3600-43200)
   maxSessionDuration int
   // ARN of the permissions boundary policy attached to the role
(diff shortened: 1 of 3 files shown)


tas50:
Nice fix. Thanks @syrull

PR #9396, opened 2026-07-23
✨ os: detect WizOS (Alpine-derivative) and wire apk packages

tas50:
## What

Adds platform detection for **WizOS**, a hardened, Alpine-lineage container distribution from Wiz. It ships its own os-release identity (`ID=wizos`, `ID_LIKE=alpine`) and uses `apk`.

## Changes

- **`detector_all.go`** — new `wizos` leaf resolver under `linuxFamily`, matched by exact os-release `ID` (same pattern as `wolfi`). Registered **before** `alpine` so its exact-name match wins over alpine's `/etc/alpine-release` marker-file fallback (an Alpine derivative may also ship that file). `wizos` matches only `pf.Name == "wizos"`, so it can never steal genuine Alpine assets.
- **`packages.go`** — `wizos` added to the apk package-manager case alongside `alpine`/`wolfi`, so package inventory works.
- **Tests** — `detect-wizos.toml` mock fixture, `TestWizOSLinuxDetector`, and a `device_type` table entry.

## Result

A WizOS asset now resolves to platform `wizos` with family `["linux","unix","os"]` and gets apk package detection, instead of falling through to `generic-linux` with no package manager.

## Verification

```
go test ./providers/os/detector/ ./providers/os/resources/packages/   # ok
```

`ID=wizos` / `ID_LIKE=alpine` and apk usage were confirmed with the maintainer; the images are behind Wiz's customer-preview registry so they can't be pulled here for a live scan.

### DIFF
--- a/providers/os/detector/detector_all.go
+++ b/providers/os/detector/detector_all.go
@@ -110,6 +110,20 @@ var wolfi = &PlatformResolver{
 	},
 }
 
+// WizOS is an Alpine-lineage distro (ID_LIKE=alpine) that ships its own
+// ID=wizos in /etc/os-release and uses apk. It is resolved before alpine so
+// its exact-name match wins over alpine's /etc/alpine-release fallback.
+var wizos = &PlatformResolver{
+	Name:     "wizos",
+	IsFamily: false,
+	Detect: func(r *PlatformResolver, pf *inventory.Platform, conn shared.Connection) (bool, error) {
+		if pf.Name == "wizos" {
+			return true, nil
+		}
+		return false, nil
+	},
+}
+
 var arch = &PlatformResolver{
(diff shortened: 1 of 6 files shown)


tas50:
Addressed in 906c2f8: WizOS now maps to `OpenrcServiceManager` (grouped with the alpine case in `services/manager.go`), since WizOS is Alpine-based and uses OpenRC — consistent with treating it as apk-based for package selection.

Note for context: `wolfi` is intentionally *not* in that switch (it's a distroless-style base with no init, so the systemd-probe fallback returning empty is correct for it). WizOS, positioned as a fuller Alpine-based userland, tracks `alpine` here instead.

Marked the PR as draft per request.

PR #9402, opened 2026-07-23
🐛 proxmox: fix firewall-rule cache collision, apt update detection, and container mount backup default

tas50:
## Summary

A static bug-review pass over the `proxmox` provider surfaced several defects that returned **wrong data with no error**. This PR fixes the three confirmed data-correctness bugs; each ships with a regression test.

### CRITICAL — firewall rules collide in the resource cache
`firewallRulesToResources` created each `proxmox.firewall.rule` **without** an explicit `__id` and assigned the `scope` Internal-struct field *after* `CreateResource` returned. But the cache key is computed *inside* `CreateResource` via `id()`, which reads `scope` — so every rule's key was baked with an **empty scope segment** (`proxmox.firewall.rule//<pos>/<type>/<action>/<source>/<dest>`). Two rules sharing a `pos/type/action/source/dest` tuple in different scopes — cloned VMs (`vm/100` and `vm/101`), or a default `in DROP` replicated across cluster/nodes/guests — collided, and `CreateResource` returned the first, discarding the second. A firewall audit (`allowsPublicIngress`, `group()`, port checks) was then handed the wrong rule. Fixed by passing `__id` explicitly, matching the sibling `ipset`/`alias`/`options` creators.

### HIGH — `vm.updates` never reports pending updates on Debian/Ubuntu
`getAptUpdates` ran `apt list --installed` (which lists every installed package with no upgrade annotation) and the parser flagged upgrades via `"upgradable to:"` — a token apt never emits. The real marker, only produced by `apt list --upgradable`, is `"upgradable from:"`. So on any running Debian/Ubuntu guest `vm.updates` returned the full installed inventory with `upgradable=false` on every row, and an audit like `updates.where(upgradable && severity=="security").none` passed vacuously on every VM. Switched to `apt list --upgradable`, parse `"upgradable from:"` (version column is the candidate/new version, bracket carries the installed one), and degrade `ErrQGANotRunning` like the Windows path.

The existing `TestParseAptOutput_WithUpgradable` masked this by feeding a fabricated `[installed,upgradable to: X]` line that neither apt mode produces — the tests were rewritten against real `apt list --upgradable` output.

### MEDIUM — container additional mount points wrongly reported as backed up
`parseContainerMountPoint` defaulted `backup=true`. In PVE, additional LXC mount points (`mp0`..`mp254`) are **excluded** from vzdump unless `backup=1` is set explicitly; only `rootfs` is backed up by default. (The identical default in `parseVMDiskConfig` is correct — QEMU drives are included unless `backup=0` — it was just copied to containers where the semantics differ.) Now defaults to `id == "rootfs"`, with explicit `backup=` still honored.

## Testing
- `go test ./resources/ ./connection/` passes.
- Added: a real-`apt list --upgradable` parse test (incl. a no-upgrades header-only case), a cross-scope firewall-rule `__id` distinctness test (identical tuple on `vm/100` vs `vm/101` vs `ct/100`), a `helpers.go` production-format guard, and an additional-mount-point backup-default test.
- No `.lr`/schema or `*.permissions.json` changes.

## Deliberately deferred (follow-ups)
From the same review, a broader **access-denied consistency** cluster: only `user.tfaFactors` and `disk.smart` degrade 401/403/404 via `IsAccessDeniedOrNotFound`; the ~30 other per-guest/per-node accessors either propagate the raw error (one unreadable guest sinks the whole `vms { disks {...} }` query for a scoped token) or, for the scalar `cfgStr`/`cfgBool` helpers, swallow it and return `false`/`""` (a config-fetch error makes `vm.protection` read `false`). These want a focused refactor of their own. Also deferred: the firewall-rule `enable` absent-key default (needs live confirmation of PVE's response shape) and the init dangling-ref husk on typed refs.

### DIFF
--- a/providers/proxmox/connection/guest_agent.go
+++ b/providers/proxmox/connection/guest_agent.go
@@ -127,9 +127,15 @@ func (c *PveConnection) GetUpdates(node string, vmid int, osInfo *OsInfo) ([]Upd
 }
 
 func (c *PveConnection) getAptUpdates(node string, vmid int) ([]UpdateInfo, error) {
-	result, err := c.QGAExec(node, vmid, []string{"sh", "-c", "apt list --installed 2>/dev/null"})
+	// `apt list --upgradable` is the only apt mode that reports pending
+	// upgrades; `--installed` lists every installed package with no upgrade
+	// annotation, so it can never surface an actual update.
+	result, err := c.QGAExec(node, vmid, []string{"sh", "-c", "apt list --upgradable 2>/dev/null"})
 	if err != nil {
-		return nil, fmt.Errorf("apt list --installed failed: %w", err)
+		if errors.Is(err, ErrQGANotRunning) {
+			return []UpdateInfo{}, nil
+		}
+		return nil, fmt.Errorf("apt list --upgradable failed: %w", err)
 	}
 	return ParseAptOutput(result.Stdout), nil
 }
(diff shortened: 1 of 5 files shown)


tas50:
Addressed in 0b44b43 — added the same `ErrQGANotRunning` guard to `getDnfUpdates`, so a RHEL/Fedora guest without a running QEMU guest agent degrades `vm.updates` to an empty slice instead of hard-erroring, matching the apt and Windows paths.

PR #9426, opened 2026-07-24
🐛 os: only follow symlinks in files.find when searching for links

czunker:
## Summary
- `files.find`'s unix command backend always ran `find -L`, which makes `find` `stat()` (not `lstat()`) every entry and follow symlinked directories — needed only to resolve link targets for a `type: "link"` search.
- For every other search (e.g. `type: "directory"`), that's wasted work: extra syscalls per symlink, and a walk that can hang if a symlink points at a stale/unresponsive mount. It also let `xdev: false` leak results from other filesystems whenever a symlink crossed a mount boundary (a symlinked directory on a different device got reported as a match even though the walk is supposed to stay on one filesystem).
- Fix: only emit `-L` (GNU) / `-H` (BSD) when the search is actually for `type: "link"`.

## Benchmark

Ran the before/after comparison on a real Ubuntu 24.04 Linux VM (arm64), inflated with extra installed kernel versions and Python packages so the filesystem has a realistic number of entries/symlinks (root fs grew to ~300K dirents, ~16.5K symlinks).

Compiled two `mql` binaries (pre-fix and post-fix) and ran the same query end-to-end through the real MQL compiler/executor:

```
files.find(from: "/", type: "directory", xdev: false, permissions: 2).all( permissions.sticky )
```

| | Wall time (avg of 3 runs) | Result |
|---|---|---|
| Before (unpatched, `find -L`) | ~0.55s | `true` |
| After (fixed, `find`) | ~0.28s | `true` |

**~2x faster**, identical (correct) result. Isolating just the generated `find` command (no MQL/plugin overhead) showed an even larger gap, ~2.7x, confirming the difference is from symlink-follow overhead and not measurement noise. The larger the number of symlinks and mount-crossing bind points on a host, the bigger this gap gets.

## Test plan
- [x] `go test ./providers/os/resources/filesfind/...` — updated unit tests cover both link and non-link command generation
- [x] `go test ./providers/os/resources/...` — full package suite passes (fixture mocks updated for the new command string)
- [x] `go vet` / `gofmt` clean
- [x] Manually verified via a real Linux VM end-to-end run (see benchmark above) that the fixed query returns the same result, faster

### DIFF
--- a/providers/os/resources/filesfind/unix_findcmd.go
+++ b/providers/os/resources/filesfind/unix_findcmd.go
@@ -39,6 +39,12 @@ func BuildFilesFindCmd(from string, xdev bool, fileType string, regex string, pe
 		}
 	}
 
+	// -L makes find stat() (not lstat()) every entry and follow symlinked
+	// directories, which is only needed to resolve link targets for a
+	// type: "link" search. Emitting it unconditionally forces every other
+	// search to pay for symlink resolution it doesn't need, and lets the
+	// walk hang on a symlink pointing at a stale/unresponsive mount.
+	//
 	// GNU find: -L -xtype l follows all symlinks AND finds symlinks.
 	// BSD find: -xtype is absent; fall back to -H -type l which
 	// follows only the start path but still detects symlinks.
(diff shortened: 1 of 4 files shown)


czunker:
This should not contradict https://github.com/mondoohq/mql/pull/8874

It only extends it.

PR #9429, opened 2026-07-24
feat(languages): add deno.lock, vcpkg, and opam package extractors

chris-rock:
## What

Three new self-contained `languages.Extractor` implementations for package-file SBOM/SCA coverage, following the existing extractor-only precedent (`conanlock`, `conda`):

### `javascript/denolock` — `deno.lock`
Parses both the v3 (`packages.npm` / `packages.jsr`) and v4 (top-level) layouts and emits the resolved packages:
- **npm** packages carry integrity hashes (SRI) and reuse the shared `javascript` purl builder, so a package resolved via both `deno.lock` and `package-lock.json` gets an identical `pkg:npm` purl (clean dedup).
- **JSR** packages become `pkg:npm` with a `repository_url=https://jsr.io` qualifier so they aren't confused with npm-registry packages.
- Deno's `_`-suffixed peer-dependency key tags are stripped; scoped names handled.

### `cpp/vcpkg` — `vcpkg.json`
The port itself is the root package; declared dependencies (both the bare-string and object forms) are direct deps, with `overrides`-pinned versions applied (manifest deps are otherwise versioned via the registry baseline, which isn't resolvable offline). purl type `vcpkg`.

### `ocaml/opam` — `*.opam`, `*.opam.locked`
Extracts the package name (from the `name:` field or the filename) and version, plus the `depends:` list via a brace-aware tokenizer (so quoted version strings inside `{…}` filters are never mistaken for dependency names). Pinned versions come from `{= "x"}` formulas (present in lock files); `{with-test}` / `{with-doc}` dependencies are scoped `dev`. purl type `opam`.

## Tests

Each extractor ships table tests with fixtures (deno v3+v4 incl. hash decoding and peer-suffix/dedup; vcpkg string+object deps and override pinning; opam locked+plain incl. the `>=` vs `=` constraint distinction and test/doc scoping). All green, `go vet` and `gofmt` clean.

## Notes

These are extractor-only packages (no MQL resource layer), matching `conanlock`/`conda`. purl types `vcpkg` and JSR-as-`pkg:npm` follow common SBOM-tooling conventions; `opam` uses the registered purl type.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>


chris-rock:
Thanks for the review — addressed in b2f3491a6:

**Integration gap (the main point).** All three are now wired into the resource layer, so they're queryable at runtime, not just parseable:

- **deno.lock** → added to `npm.go` (`collectNpmPackages` + `hasLockfile`) alongside the other JS lockfiles, so Deno projects surface via `npm.packages`.
- **vcpkg** → new `vcpkg.packages` / `vcpkg.package` MQL resources (`vcpkg.go` + `os.lr`), discovering `vcpkg.json` in a project path or the default application paths.
- **opam** → new `opam.packages` / `opam.package` MQL resources (`opam.go` + `os.lr`), discovering `*.opam` / `*.opam.locked` (a directory may declare several packages). Regenerated `os.lr.go` + `os.lr.versions`.

**Code nits.**
- `vcpkgDependency.UnmarshalJSON` now wraps the object-parse error with context (`vcpkg dependency: expected string or object: %w`).
- The opam `depends:` tokenizer is now **quote-aware** — `dependsBlock` and the filter scanner skip over string literals, so a `[`/`]`/`{`/`}` byte inside a quoted string can't throw off the depth (this also covers the multi-byte/UTF-8 concern, since bracket counting no longer looks inside strings).
- Added **disjunction handling**: a `|` in the depends list keeps the first alternative and drops the rest, so `"a" | "b"` is one dependency, not two. Covered by a new `TestParseDependsDisjunction`.

All builds/tests/`go vet`/`gofmt` green.

PR #9441, opened 2026-07-26
🐛 os: accept every valid character in an rpm package name

tas50:
## Problem

`RPM_REGEX` enumerated the characters each field may contain:

```go
^([\w-+]*)\s(\d*|\(none\)):([\w\d-+.:]+)\s([\w\d]*|\(none\))__...
```

The name class `[\w-+]` is letters, digits, `_`, `-`, `+` — **no dot**. `ParseRpmPackages` has no else branch and no log, so a line that doesn't match is simply never appended. The package vanishes from the inventory with no error and no count discrepancy.

Run against real `rpm -qa --queryformat` output on `main`:

```
tzdata                 -> ok
java-1.8.0-openjdk     -> NO MATCH (dropped)
python3.11             -> NO MATCH (dropped)
dotnet-sdk-8.0         -> NO MATCH (dropped)
1.0~rc1-1.el8          -> NO MATCH (dropped)
vendor Hewlett-Packard -> NO MATCH (dropped)
```

On any RHEL 8/9, Fedora, Amazon Linux 2023 or Oracle Linux host, that means `java-1.8.0-openjdk`, `java-1.8.0-openjdk-headless`, `python3.9`/`3.11`/`3.12`, `dotnet-sdk-8.0` and `aspnetcore-runtime-8.0` are all absent. An asset with OpenJDK installed reports **no JDK**, so every Java CVE for that host goes unreported.

Only the runtime path is affected — the static/image path uses `go-rpmdb` — which is why container-image scans look correct while live-host scans quietly lose packages.

Two adjacent classes were too narrow for the same reason:
- the version class had no `~` or `^`; rpm accepts `~` since 4.10 and `^` since 4.15, and both ship in real releases
- the vendor class had no `-`, so anything from `Hewlett-Packard` was dropped

## Fix

Rather than widen the enumerations one character at a time, match the real invariant. `NAME`, `EPOCH:VERSION-RELEASE` and `ARCH` are space-separated and none can contain whitespace, so they're matched as runs of non-space; the remaining fields are free-form text delimited by `__`.

```go
^(\S+)\s(\d*|\(none\)):(\S+)\s(\S*?)__(.*?)__(.*?)__(.*?)__(\d+|\(none\))(?:__(.*))?$
```

**Arch is non-greedy on purpose.** Arch contains underscores (`x86_64`), so a greedy run spans the `__` separator and shifts every later field down by one — on `x86_64__Microsoft__dotnet sdk__MIT__1704067200` it reads arch as `x86_64__Microsoft` and license as `1704067200`. I hit exactly this while writing the tests; `TestRpmRegex_FieldAlignment` now pins the alignment.

Three smaller fixes ride along:

- a trailing `__` with an empty modularity label dropped the whole package, because the optional group required at least one character. Confirmed against `main`: `foo 0:1.0-1.el9 x86_64__V__S__MIT__1704067200__` parses to zero packages there, one here
- unparseable lines are counted and the total is reported at warn (the per-line detail stays at debug, which is off by default, so on its own it would leave the silently-shorter package list invisible to whoever ran the scan)
- the scanner buffer is raised past bufio's 64 KB default. A line over the buffer does not drop one package, it ends the scan and drops every package behind it, and `%{LICENSE}` is free-form text with no length bound — it already runs past 250 characters on `java-*-openjdk`. A scanner error still logs at warn

An earlier revision of this description claimed an empty epoch produced a version of `":1.0-1.el9"`. That was wrong — `normalizeRpmEpoch` plus the `if epoch != ""` guard already handled it on `main`, which returns `"1.0-1.el9"` today. No code changed for it; the new test just pins the behaviour.

## Tests

`rpm_packages_charset_test.go` — 8 tests, 39 cases. Verified against the **original** regex: 17 sub-tests fail, across every category.

- **name characters** (9): dots, dots+hyphens, `+` (`libstdc++`, `gcc-c++`), leading digits (`389-ds-base`, `2ping`), mixed case, underscore
- **version characters** (3): `~` pre-release, `^` snapshot, `_` in a modular release
- **vendor characters** (4): hyphenated vendor, `&`, bracketed URL — asserting both the raw capture *and* that `cleanupVendorName` still strips the brackets, so a regex regression stays distinguishable from a change to that helper
- **field alignment** (1): every capture group checked by position, which is what catches the greedy-arch class of bug
- **arch and sentinels** (13): all 7 real arches, `(none)` arch/epoch/license, empty epoch, trailing empty modularity label, no modularity field
- **round-trip** (3 platforms × 6 packages): renders the real `queryFormat()` output the way rpm does and parses it back, covering both the `EPOCHNUM`+modularity and `EPOCH`-only variants
- **negative** (1): blank lines, `error:`/`warning:` output and other noise must not parse as packages — widening the classes must not make the regex match arbitrary text
- **realistic listing** (1): a 6-package listing survives intact and in order

All pre-existing package tests pass, including the real-fixture parsers (redhat8, photon4, suse, oracle, almalinux, modular) and `TestRpmQueryFormatRoundTrip`.

## Found by

A static review pass over the `os` provider parsers.

## Verified against real `rpm -qa` output

Both regexes were run over live output from six images, checking match counts *and* field alignment on every matched line (installtime must be digits or `(none)`, arch must not span a `__`):

| source | lines | old regex | new regex |
|---|---|---|---|
| AlmaLinux 8 + `java-1.8.0-openjdk-headless` | 196 | 195 | **196** |
| AlmaLinux 9 + `python3.11` | 183 | 179 | **183** |
| AlmaLinux 9 (base) | 159 | 159 | 159 |
| Fedora latest | 147 | 147 | 147 |
| Amazon Linux 2023 | 106 | 106 | 106 |
| openSUSE Leap 15.6 (`%{EPOCH}` variant) | 141 | 141 | 141 |

The packages the old regex drops are exactly the ones named above: `java-1.8.0-openjdk-headless` on el8, and `python3.11`, `python3.11-libs`, `python3.11-pip-wheel`, `python3.11-setuptools-wheel` on el9. No new drops and no misalignment anywhere, including `gpg-pubkey` with its `(none)` arch and vendor.

## Same bug elsewhere

`PACMAN_REGEX` in `pacman_packages.go` has the identical too-narrow name class (`[\w-]*`, no `.` and no `+`) and the identical silent-drop loop; on a stock `archlinux:latest` container it drops `libstdc++`. Being fixed separately.

### DIFF
--- a/providers/os/resources/packages/rpm_name_charset_test.go
+++ b/providers/os/resources/packages/rpm_name_charset_test.go
@@ -1,0 +1,376 @@
+// Copyright Mondoo, Inc. 2024, 2026
+// SPDX-License-Identifier: BUSL-1.1
+
+package packages
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.mondoo.com/mql/v13/providers-sdk/v1/inventory"
+)
+
+var rpmCharsetPlatform = &inventory.Platform{
+	Name:    "redhat",
+	Version: "9",
+	Arch:    "x86_64",
+	Family:  []string{"redhat", "linux"},
+}
+
+// A package the regex cannot match is dropped from the inventory with no
+// error, and an absent package is exactly what a vulnerability scan cannot
+// report on. The name class used to be `[\w-+]`, which excluded the dot, so
+// every RPM whose name contains one went missing from live-host scans.
+func TestParseRpmPackages_NameCharacters(t *testing.T) {
+	tests := []struct {
+		name string
+		line string
+		want Package
+	}{
+		{
+			name: "dots and hyphens in the name",
+			line: `java-1.8.0-openjdk 1:1.8.0.402.b06-2.el8 x86_64__Red Hat, Inc.__OpenJDK 8 Runtime__ASL 2.0__1704067200__(none)`,
+			want: Package{Name: "java-1.8.0-openjdk", Version: "1:1.8.0.402.b06-2.el8", Epoch: "1", Arch: "x86_64", Vendor: "Red Hat, Inc.", License: "ASL 2.0"},
+		},
+		{
+			name: "dot in the name",
+			line: `python3.11 0:3.11.5-1.el9 x86_64__Red Hat, Inc.__Python 3.11__Python-2.0__1704067200__(none)`,
+			want: Package{Name: "python3.11", Version: "3.11.5-1.el9", Arch: "x86_64", Vendor: "Red Hat, Inc.", License: "Python-2.0"},
+		},
+		{
+			name: "dot in a hyphenated name",
+			line: `dotnet-sdk-8.0 0:8.0.1-1 x86_64__Microsoft__dotnet sdk__MIT__1704067200__(none)`,
+			want: Package{Name: "dotnet-sdk-8.0", Version: "8.0.1-1", Arch: "x86_64", Vendor: "Microsoft", License: "MIT"},
+		},
+		{
+			name: "plus signs in the name",
+			line: `libstdc++ 0:14.2.1-3.el9 x86_64__Red Hat, Inc.__GNU Standard C++ Library__GPLv3+__1704067200__(none)`,
+			want: Package{Name: "libstdc++", Version: "14.2.1-3.el9", Arch: "x86_64", Vendor: "Red Hat, Inc.", License: "GPLv3+"},
+		},
+		{
+			name: "plus signs after a hyphen",
+			line: `gcc-c++ 0:11.5.0-2.el9 x86_64__Red Hat, Inc.__C++ support for GCC__GPLv3+__1704067200__(none)`,
+			want: Package{Name: "gcc-c++", Version: "11.5.0-2.el9", Arch: "x86_64", Vendor: "Red Hat, Inc.", License: "GPLv3+"},
+		},
+		{
+			name: "leading digits in the name",
+			line: `389-ds-base 0:2.4.5-1.el9 x86_64__Red Hat, Inc.__389 Directory Server__GPLv3+__1704067200__(none)`,
+			want: Package{Name: "389-ds-base", Version: "2.4.5-1.el9", Arch: "x86_64", Vendor: "Red Hat, Inc.", License: "GPLv3+"},
+		},
+		{
+			name: "name that is digits then letters",
+			line: `2ping 0:4.5.1-2.el9 noarch__Fedora Project__Ping utility__GPLv2+__1704067200__(none)`,
+			want: Package{Name: "2ping", Version: "4.5.1-2.el9", Arch: "noarch", Vendor: "Fedora Project", License: "GPLv2+"},
+		},
+		{
+			name: "mixed case name",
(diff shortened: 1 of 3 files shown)


chris-rock:
Reviewed this and verified the claims against live `rpm -qa` output rather than reading the regex alone. The fix is correct — I pushed three follow-ups onto the branch (385f5da) and corrected one inaccurate bullet in the description.

## The bug reproduces on real hosts

Ran the old and the new regex over output from six images, checking match counts *and* field alignment on every matched line (installtime must be digits or `(none)`, arch must not span a `__`):

| source | lines | old regex | new regex |
|---|---|---|---|
| AlmaLinux 8 + `java-1.8.0-openjdk-headless` | 196 | 195 | **196** |
| AlmaLinux 9 + `python3.11` | 183 | 179 | **183** |
| AlmaLinux 9 (base) | 159 | 159 | 159 |
| Fedora latest | 147 | 147 | 147 |
| Amazon Linux 2023 | 106 | 106 | 106 |
| openSUSE Leap 15.6 (`%{EPOCH}` variant) | 141 | 141 | 141 |

The drops are exactly the packages named in the description: `java-1.8.0-openjdk-headless` on el8, and the four `python3.11*` packages on el9. Zero new drops, zero misalignment — `gpg-pubkey` with its `(none)` arch and vendor included. The non-greedy arch is doing real work; `(\S+)` greedy takes `x86_64__Microsoft`.

Also confirmed the trailing-empty-modularity fix independently: on `main`, `foo 0:1.0-1.el9 x86_64__V__S__MIT__1704067200__` parses to **0** packages.

## What I pushed

1. **The per-line debug log doesn't surface the failure this PR is about.** Debug is off by default, so a format drift that drops every package still looks like a clean short list. Now the drops are counted and the total goes out at warn, with the per-line detail left at debug.

2. **Raised the scanner buffer past bufio's 64 KB default.** The description flagged the truncation but only warned about it. A line over the buffer doesn't drop one package — it ends the scan and drops every package behind it. `%{LICENSE}` is free-form with no length bound and already runs past 250 characters on `java-*-openjdk`. There's a regression test now; without the buffer bump it loses both packages of a two-line listing.

3. **Nits:** exact-version assertion in the round-trip test instead of `assert.Contains` (so an epoch leaking in as `0:` or `:` is caught), and renamed the test file to `rpm_packages_charset_test.go` so it sorts next to `rpm_packages_test.go`.

## One correction to the description

> an empty epoch produced a version of `":1.0-1.el9"` — now treated like `(none)`

This one isn't right — there's no epoch handling in the diff, and `main` already returns `"1.0-1.el9"` for that input, because `normalizeRpmEpoch` collapses it and the `if epoch != ""` guard skips the prefix. The new test pins existing behaviour, which is worth having; I've amended the body so it doesn't land in the commit message as a fixed bug.

## Follow-up: the same bug is in the pacman parser

`PACMAN_REGEX` (`pacman_packages.go:32`) is `^([\w-]*)\s([\w\d-+.:]+)$` — same too-narrow name class (no `.`, no `+`), same silent-drop loop, no logging. On a stock `archlinux:latest` container it drops `libstdc++` (136 of 137 packages parse). Fixing that separately.

chris-rock:
Pacman follow-up is up: #9868 — same enumeration bug, same missing else branch. `libstdc++` and `db5.3` are dropped on a stock `archlinux:base-devel` container (168 of 170 packages parse).

PR #9443, opened 2026-07-26
🐛 MQL: bind a block to its own operand instead of the stack tail

tas50:
## The bug

Attaching a block to an array literal whose elements are resources crashes the compiler:

```
$ mql run local -c '[asset] { name }'
panic: cannot compute checksum for chunk, it doesn't seem to reference a function on the stack
  llx.(*Function).checksumV2   llx/chunk.go:118
  llx.(*Chunk).ChecksumV2      llx/chunk.go:105
  llx.(*Block).AddChunk        llx/code.go:57
  mqlc.(*compiler).addChunk    mqlc/mqlc.go:223
  mqlc.Compile.func1           mqlc/mqlc.go:2337
```

`Compile()` recovers only in order to **re-panic** with a wrapped error, so this takes the caller down — the CLI dies and `health.ReportPanic` fires — instead of returning a compile error.

Provider-independent; reproduced with `core` and `os` resources:

```
[asset] { name }        panic
[mondoo] { version }    panic
[aws.waf] { scope }     panic
[1, 2] { _ }            fine
```

## Root cause

A block binds to its operand's value, so that value has to exist as a chunk on the stack. `compileOperand` pushed it only when the stack *looked* empty:

```go
if x := c.tailRef(); (x & 0xFFFFFFFF) == 0 {
```

That is not the same question as "did this operand produce a ref?". An array literal compiles its **elements** first, so `[aws.waf]` leaves a chunk on the stack (the resource) while the array primitive itself was never pushed and the operand's own `ref` stayed `0`. The block was then bound to a chunk that is not on the stack, and checksumming panicked.

A scalar array (`[1, 2]`) pushes nothing, so the old condition happened to hold — which is why this went unnoticed.

## The fix

Test the operand's own `ref`, and push the **already-compiled** primitive instead of recompiling the value (recompiling would emit a second copy of every chunk the elements produced).

## Verification

```
[asset] { name }       ->  [ { name: "..." } ]
[mondoo] { version }   ->  [ { version: "v13.0.0-rolling" } ]
[1,2] { _ }            ->  unchanged
```

New `TestCompiler_ArrayLiteralWithBlock` covers resource-element arrays, a multi-element array, and pins the scalar cases so the fix cannot regress them. It was confirmed to fail with the original panic before the change.

`./mqlc/...`, `./mqlc/parser`, and `./llx/...` are green.

`make test/go/plain` reports three failures — `cli/config`, `cli/selfupdate`, and `providers` (build, missing generated mocks). All three reproduce on an unmodified checkout of `main`, so they are pre-existing and unrelated to this change.

### DIFF
--- a/mqlc/array_block_test.go
+++ b/mqlc/array_block_test.go
@@ -1,0 +1,55 @@
+// Copyright Mondoo, Inc. 2024, 2026
+// SPDX-License-Identifier: BUSL-1.1
+
+package mqlc_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.mondoo.com/mql/v13/mqlc"
+)
+
+// TestCompiler_ArrayLiteralWithBlock covers an array literal whose elements are
+// resources, with a block attached.
+//
+// The elements compile to refs, so the stack is already non-empty by the time the
+// block is handled. compileOperand only pushed the value onto the stack when the
+// stack looked empty, so for this shape the array primitive was never pushed and
+// the operand's own ref stayed 0 -- leaving the block bound to a chunk that is
+// not on the stack. Computing checksums then hit
+//
+//	panic: cannot compute checksum for chunk, it doesn't seem to reference a
+//	       function on the stack        (llx/chunk.go, Function.checksumV2)
+//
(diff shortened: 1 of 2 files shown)


arlimus (approved):
Nice catch, kudos for fixing!

PR #9447, opened 2026-07-26
✨ macos: surface system_profiler obtained_from as the package origin

tas50:
## Description

macOS was the one package backend that reported an **empty origin for every package**. `system_profiler SPApplicationsDataType` already classifies every bundle by how it was obtained, but `sysProfilerItem` never parsed the field, so it was dropped at the parser.

```go
type sysProfilerItem struct {
	Name    string `plist:"_name"`
	Version string `plist:"version"`
	Path    string `plist:"path"`
	// (obtained_from was never read)
}
```

Values Gatekeeper assigns: `apple`, `mac_app_store`, `ios_app_store`, `identified_developer`, `unknown`.

### Why `origin` rather than a new field

`Origin` is where the other backends already put exactly this:

| backend | `Origin` holds |
|---|---|
| flatpak | the remote it was installed from (`flathub`) |
| freebsd | the ports origin path (`security/sudo`) |
| netbsd | `PKGPATH` |
| **macOS** | **how the bundle was obtained** (new) |

Since macOS left `Origin` empty until now, this is **purely additive** — no existing value changes, no schema change, and it's queryable immediately as `packages.list { origin }`. Server-side, `Origin` is only passed through (`policy/conversion.go`), so nothing downstream changes behavior.

If reviewers would rather have a dedicated `obtainedFrom` field than overload `origin`, that's an easy pivot — it just costs an `.lr` field plus codegen. I went with `origin` because the flatpak precedent already establishes "where this package came from" as its meaning.

### Why this is worth having

The motivating consumer is **remediation**, not inventory. An App Store app can't be upgraded by `brew upgrade` — the generated script dead-ends at `brew.tmpl.sh`:

> `Package <name> is not installed by brew. We currently cannot automatically upgrade it.`

Today nothing downstream can tell an App Store install from a direct download, so there's no way to route those to "update via the App Store" instead of emitting a script that no-ops. This change supplies the missing signal.

It also makes iOS apps on Apple Silicon identifiable. They're reported alongside native Mac apps and were previously indistinguishable — worth knowing, since an iOS app sharing a `CFBundleName` with a desktop product would otherwise be silently treated as that product.

## How to Test

```bash
go test ./providers/os/resources/packages/ -run TestMacOsXPackageParser -count=1 -v
```

The fixture gained a `mac_app_store` entry (WireGuard) and an `ios_app_store` entry (Victory), and the test now asserts provenance on those plus the pre-existing `apple` and `identified_developer` entries.

**Verified against real data**, not just the fixture — parsing a live capture of 503 applications through `ParseMacOSPackages` reproduces the raw `system_profiler` distribution exactly:

```
parsed 503 packages
  origin="apple"                300
  origin="unknown"              137
  origin="identified_developer"  52
  origin="mac_app_store"         13
  origin="ios_app_store"          1
```

On a Mac, this should now return non-empty origins:

```bash
cnspec run local -c 'packages.where(format == "macos") { name version origin }'
```

## Notes

`providers/os/resources/os.lr` gained documentation for the per-backend meaning of `origin`. Regenerated with `./lr go` + `./lr versions`; `os.lr.versions` was unchanged (comment-only edits don't shift versions) and `os.resources.json` is a gitignored build artifact, so no generated files are in this diff.

### DIFF
--- a/providers/os/resources/os.lr
+++ b/providers/os/resources/os.lr
@@ -1000,6 +1000,20 @@ package @defaults("name version") {
   // (see the `package` accessor on tool resources): the installing source
   // could not be identified. A known origin means the package is
   // manager-tracked and listed in `packages`.
+  //
+  // The exact meaning is per-backend, so compare it against values you know
+  // the backend emits rather than across platforms:
+  //
+  //   flatpak       the remote it was installed from, e.g. "flathub"
+  //   freebsd       the ports origin path, e.g. "security/sudo"
+  //   netbsd        PKGPATH
+  //   macOS         how the bundle was obtained, as classified by Gatekeeper:
+  //                 "apple", "mac_app_store", "ios_app_store",
+  //                 "identified_developer", or "unknown"
+  //
+  // On macOS this distinguishes App Store installs from everything else, which
+  // matters when choosing an update path: an App Store app cannot be upgraded
+  // with `brew upgrade`.
   origin() string
 
   // Latest version the package manager reports as available
--- a/providers/os/resources/packages/macos_packages.go
+++ b/providers/os/resources/packages/macos_packages.go
@@ -26,6 +26,13 @@ type sysProfilerItem struct {
 	Name    string `plist:"_name"`
 	Version string `plist:"version"`
 	Path    string `plist:"path"`
+	// Where the bundle came from, as classified by Gatekeeper:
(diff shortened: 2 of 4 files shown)


czunker (approved):
We should later move this to a separate field.

Same for flatpak. Perhaps something like `package_source`.
Origin has a different meaning in the deb context.

PR #9452, opened 2026-07-27
🐛 os: detect python packages installed in virtualenvs

### LINKED ISSUE #9450: python.packages misses virtualenv site-packages (app venvs invisible in SBOM)
## Summary

`python.packages` only searches a fixed set of system paths, so Python packages installed into a **virtualenv** are never detected. This is the dominant layout for containerized Python apps, so the application's own dependencies — usually the interesting ones — are missing from the SBOM entirely.

## Reproduction

Using any container image that installs its app into a venv (very common for Python service images):

```
$ mql run container <image> -c "python.packages.length"
python.packages.length: 0
```

The packages are there — pointing the resource at the venv explicitly finds them:

```
$ mql run container <image> -c 'python("/app/.venv/lib/python3.13/site-packages").packages.where(name == "litellm").map(name + " " + version + " " + purl)'
python.packages.where.map: [
  0: "litellm 1.93.0 pkg:pypi/litellm@1.93.0"
]
```

So the parser is fine; only the search paths are wrong.

## Cause

`providers/os/resources/python.go`:

```go
var defaultPythonPaths = []string{
	// Linux
	"/usr/local/lib/python*",
	"/usr/local/lib64/python*",
	"/usr/lib/python*",
	"/usr/lib64/python*",
	...
}
```

No venv locations. `fsutil.WalkGlob` uses `afero.Glob`, which supports `*` but **not** `**`, so each directory level has to be spelled out explicitly — a single recursive pattern won't work.

## Impact

Any image whose Python dependencies live in a venv reports zero Python packages. Since `sbom/generator/generator.go` collects `python.packages` into the SBOM, those packages are absent from the SBOM and cannot be matched against PyPI advisories.

## Suggested fix

Add a bounded set of globs covering the common container/venv conventions, rather than a full filesystem walk (which would be costly on large images). Roughly:

```
/app/.venv/lib/python*
/app/venv/lib/python*
/opt/*/lib/python*
/opt/*/.venv/lib/python*
/srv/*/lib/python*
/usr/src/app/.venv/lib/python*
/code/.venv/lib/python*
/workspace/.venv/lib/python*
/home/*/.venv/lib/python*
/venv/lib/python*
/.venv/lib/python*
```

A more principled alternative is to locate `pyvenv.cfg` marker files at a bounded depth and resolve each venv's `lib/python*/site-packages` from there — more robust against non-standard directory names, at the cost of a larger change.

Same gap likely applies to the Windows venv layout (`<venv>\Lib\site-packages`, no `lib/pythonX.Y` level), which is worth handling in the same change.

czunker:
## What

`python.packages` only searched a fixed set of system paths, so packages installed into a **virtualenv** were never detected. That is the dominant layout for containerized Python apps, so the application's own dependencies — usually the interesting ones — were missing from the SBOM entirely.

## Verification against a real image

An image that installs its app into `/app/.venv`:

| | before | after |
|---|---|---|
| `python.packages.length` | **0** | **174** |
| SBOM entries | 48 (apk only) | 222 (174 pypi + 48 apk) |
| the app's own package | *absent* | `pkg:pypi/litellm@1.93.0` |

No change on a system-python image (`python:3.12-slim`: 1 → 1) or an image with no Python (`alpine`: 0 → 0).

## Approach

A bounded glob list rather than a recursive filesystem walk, which would be costly on large images. `afero.Glob` supports `*` but not `**`, so every directory level is spelled out.

Two things worth a reviewer's eye:

**`*` matches dot-directories here.** Unlike shell globbing, `filepath.Match` treats `.` as an ordinary character, so `/app/*/lib/python*` matches `/app/.venv/lib/python3.13`. The whole approach depends on this, so `TestDefaultPythonPaths_GlobMatchesDotDirectories` pins it. Also means one pattern covers `.venv`, `venv`, `env`, and any other venv directory name.

**One bad match used to abort everything.** `collectPythonPackagesInPaths` returned the error, so a single glob landing on a file rather than a directory (`ReadDir` → `ENOTDIR`, which is not `os.IsNotExist`) killed the entire search and `python.packages` failed wholesale. Broadening the globs makes that materially likelier, so it now skips and continues. `TestCollectPythonPackagesInPaths_SkipsUnreadableMatches` covers it — deliberately on a real filesystem via `BasePathFs`, because `MemMapFs` does not reproduce the `ENOTDIR` distinction and the test silently passes without the fix against an in-memory FS.

Windows venvs put `site-packages` directly under `Lib` with no `pythonX.Y` level, so those patterns are shaped accordingly.

Closes #9450

### DIFF
--- a/providers/os/resources/python.go
+++ b/providers/os/resources/python.go
@@ -28,22 +28,54 @@ import (
 	"go.mondoo.com/mql/v13/types"
 )
 
+// defaultPythonPaths lists the directories that *contain* a site-packages or
+// dist-packages directory; collectPythonPackagesInPaths appends those names.
+//
+// These are glob patterns evaluated by afero.Glob, which supports "*" but not
+// "**", so every directory level has to be spelled out. Note that, unlike shell
+// globbing, "*" here does match dot-directories, so "/app/*/lib/python*" picks
+// up "/app/.venv/lib/python3.13".
 var defaultPythonPaths = []string{
 	// Linux
 	"/usr/local/lib/python*",
 	"/usr/local/lib64/python*",
 	"/usr/lib/python*",
 	"/usr/lib64/python*",
+	// Virtualenvs. Applications -- container images especially -- commonly
+	// install their dependencies into a venv instead of the system paths above,
+	// so without these the application's own packages are missing entirely.
+	// The venv directory name varies (.venv, venv, env, ...), hence the "*".
+	"/venv/lib/python*",
+	"/.venv/lib/python*",
+	"/app/*/lib/python*",
+	"/code/*/lib/python*",
+	"/workspace/*/lib/python*",
+	"/srv/*/lib/python*",
+	"/opt/*/lib/python*",
+	"/usr/src/*/*/lib/python*",
+	// per-user venvs, including the virtualenvwrapper/poetry layout that nests
+	// one directory deeper
+	"/root/*/lib/python*",
+	"/root/.virtualenvs/*/lib/python*",
+	"/home/*/.venv/lib/python*",
+	"/home/*/.virtualenvs/*/lib/python*",
 	// Windows
 	"C:\\Python*\\Lib",
 	"C:\\Program Files\\Python*\\Lib",
 	// per-user installs across all profiles
 	"C:\\Users\\*\\AppData\\Local\\Programs\\Python\\Python*\\Lib",
+	// Windows venvs put site-packages directly under Lib, with no pythonX.Y level
+	"C:\\venv\\Lib",
+	"C:\\.venv\\Lib",
+	"C:\\Users\\*\\.venv\\Lib",
+	"C:\\Users\\*\\*\\.venv\\Lib",
 	// macOS
 	"/opt/homebrew/lib/python*",
 	"/System/Library/Frameworks/Python.framework/Versions/*/lib/python*",
 	// we use 3.x to exclude the macOS 'Current' symlink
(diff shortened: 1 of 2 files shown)


czunker:
Thanks — both suggestions applied in fcbe172.

**Non-dotted venv names under user homes**: good catch, that was an inconsistency I introduced. The container roots matched any venv directory name via `*`, but `/home/*/.venv/...` was pinned to the dotted spelling. Switched to `/home/*/*/lib/python*` (and the same for `/root` and `/Users`), so `venv` and `env` now work everywhere `.venv` does. Added test cases for all three.

**Nesting depth**: added one more level for the container roots — `/app/*/*/lib/python*` and friends — so a project directory carrying its own venv (`/app/myservice/.venv`) is covered. Those roots are shallow and purpose-built, so the extra `ReadDir` per entry is cheap.

I deliberately did **not** extend user homes to that depth. Without `**` each extra level costs a `ReadDir` per directory, and a shared machine can have many users with many project directories each — `/home/*/*/*/lib/python*` is the one pattern here with a genuinely unbounded fan-out. So `~/projects/app/.venv` remains a gap.

Rather than leave that implicit, it is now documented in the code comment and pinned by `TestDefaultPythonPaths_DeepUserProjectVenvNotCovered`, which asserts the case is *not* found and explains why. If someone later decides the coverage is worth the scan cost, that test fails and forces the trade-off to be made deliberately. The `path` argument on the python resource covers such a layout in the meantime.

Re-verified after the change: the container image still reports 174 packages with no duplicates, and a local macOS system goes from 34 to 61 packages with no measurable slowdown.

PR #9468, opened 2026-07-28
🐛 azure: let the inventory name the sign-in method

preslavgerchev:
## Problem

An Azure connection with no client secret or certificate falls back to the default sign-in chain, which probes every method in turn. The managed identity probe alone retries for up to 15 seconds before giving up, and **every discovered asset gets its own connection** — so that cost is paid per asset. A scan of a subscription full of assets spends minutes probing credentials it was never going to get, and fills the log with one of these per asset:

```
ERROR → no Azure credentials were provided; trying to sign in with your local Azure CLI session, Azure environment variables, or a managed identity
ERROR x failed to get managed identity token (max retries reached) num_attempts=3
```

Timestamps from a live runner show ~19s burned per asset on a chain that had already definitively failed on the previous one.

## What this does

Connections can name the method(s) they authenticate with, via the `auth-method` option (or `--auth-method`), as a comma-separated list of `cli`, `env`, `managed-identity`, `workload-identity`:

```yaml
spec:
  assets:
  - connections:
    - type: azure
      options:
        client-id: <client-id>
        tenant-id: <tenant-id>
        auth-method: workload-identity
```

That builds a one-link chain — no `az` exec probe, no 3×5s managed identity retry.

- Unset (as well as the explicit `default` / `auto`) keeps the full chain, so nothing changes for callers that don't opt in.
- Values are normalized, so `Workload_Identity` is accepted.
- An unrecognized value is an **error**, not a silent fall-back to the full chain: a typo that quietly reinstated the probing would be invisible in exactly the deployment that asked to avoid it.
- An explicit `auth-method` beats a federated token file, since that file is often a leftover env var in the pod spec. With no `auth-method` set, the existing token-file short-circuit is untouched.

## Workload identity was unreachable anyway

`GetDefaultChainedToken` built its `WorkloadIdentityCredentialOptions` with a `TenantID` but no `ClientID`, and `GetTokenFromCredential` discarded the tenant and client it was handed (it passed an empty `DefaultAzureCredentialOptions{}`). So `NewWorkloadIdentityCredential` fell back to `AZURE_CLIENT_ID` and, absent that, failed to *construct* — at which point `BuildChainedToken` dropped it from the chain without a word. Both values now reach it, along with the federated token file.

## Supporting fixes

- `BuildChainedToken` swallowed every constructor error, so naming one method and getting nothing back surfaced only the SDK's bare `at least one credential required`. When the chain ends up empty it now reports why the named source is unusable.
- The sign-in guidance in `enrichTokenError` names the methods actually tried, and drops the `az login` advice when the CLI was never in the chain.

## API change

`azidentity.DefaultAzureCredentialOptions` has nowhere to put a `ClientID`, so `GetDefaultChainedToken` now takes an `azauth.ChainedTokenOptions` (which embeds it), and `GetTokenFromCredential` takes it as a trailing argument. All in-repo callers are updated.

Two call sites **outside** this repo will need a companion PR when they bump their pinned mql version:
- `server/apps/cnspec-runner/extendedscanner/extended_azure.go` — needs a trailing arg. Worth wiring to `auth-method` off `pCfg.Options` while there: it gates on `scan-vms: "true"` and calls `GetTokenFromCredential` with the same nil credential, so it is a second, independent source of the same log spam in the runner pod.
- `server/cnspec/cli/reporter/azure_service_bus_handler.go` — `&azidentity.DefaultAzureCredentialOptions{}` becomes `&azauth.ChainedTokenOptions{}`.

## Not addressed here

The credential is still rebuilt per asset with no process-level cache, so a scan does N token acquisitions instead of one. Naming the method collapses the per-asset cost from ~15s to near-zero, which makes this much less urgent, but it is still there.

## Test plan

- [x] `go test ./providers-sdk/v1/util/azauth/...` — new coverage for method parsing/normalization/rejection, chain restriction, `ClientID` plumb-through, and the guidance wording
- [x] `go test ./providers/azure/...` — new coverage for the `auth-method` option, its precedence over a token file, and invalid values
- [x] `go build ./...` clean for azure, ms365, k8s, os
- [x] `gofmt` / `go vet` clean
- [x] Provider CLI JSON regenerates with the new `--auth-method` flag

### DIFF
--- a/providers-sdk/v1/util/azauth/token.go
+++ b/providers-sdk/v1/util/azauth/token.go
@@ -48,13 +48,18 @@ var DefaultCredentialMethods = []CredentialMethod{
 	CredentialMethodWorkloadIdentity,
 }
 
-// credentialMethodDescriptions renders each method the way we talk about it in
-// user-facing sign-in guidance.
-var credentialMethodDescriptions = map[CredentialMethod]string{
-	CredentialMethodCLI:              "your local Azure CLI session",
-	CredentialMethodEnv:              "Azure environment variables",
-	CredentialMethodManagedIdentity:  "a managed identity",
-	CredentialMethodWorkloadIdentity: "workload identity federation",
+// credentialMethodNames renders a method selection as a plain list; an empty
+// selection renders the full default chain. These are the names auth-method
+// accepts, so a message built from them doubles as a usage hint.
+func credentialMethodNames(methods []CredentialMethod) string {
+	if len(methods) == 0 {
+		methods = DefaultCredentialMethods
+	}
+	names := make([]string, 0, len(methods))
+	for _, m := range methods {
+		names = append(names, string(m))
+	}
+	return strings.Join(names, ", ")
 }
 
 // ParseCredentialMethods reads a comma-separated method list, e.g.
@@ -77,23 +82,15 @@ func ParseCredentialMethods(s string) ([]CredentialMethod, error) {
 		}
 
 		method := CredentialMethod(name)
-		if _, ok := credentialMethodDescriptions[method]; !ok {
+		if !slices.Contains(DefaultCredentialMethods, method) {
 			return nil, fmt.Errorf("unknown Azure credential method %q, expected one of %s",
-				strings.TrimSpace(part), strings.Join(credentialMethodNames(), ", "))
+				strings.TrimSpace(part), credentialMethodNames(nil))
 		}
 		if !slices.Contains(methods, method) {
 			methods = append(methods, method)
 		}
 	}
 	return methods, nil
-}
-
-func credentialMethodNames() []string {
-	names := make([]string, 0, len(DefaultCredentialMethods))
-	for _, m := range DefaultCredentialMethods {
-		names = append(names, string(m))
-	}
-	return names
 }
 
 // AllowsMethod reports whether m is permitted by the selection. An empty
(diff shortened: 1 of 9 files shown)


preslavgerchev:
Addressed in 352b7ec.

**`azureinstancesnapshot/provider.go` — federated token file dropped** (🟡, raised twice): real bug, fixed. I'd added `Methods` there without `FederatedTokenFile`, so `--federated-token-file` plus `--auth-method workload-identity` silently lost the path.

One correction to the suggested patch: the option key is `azure-federated-token-file` (`connection.OptionFederatedTokenFile`), not `federated-token-file` — the latter is the CLI flag name, which `ParseCLI` maps to the prefixed option. The first review had it right, the follow-up didn't.

Also skipped the `os.Getenv` fallback from the suggestion: `azidentity` already falls back to `AZURE_FEDERATED_TOKEN_FILE` when `TokenFilePath` is empty, so only the option needs forwarding. The env read in `connection.go` exists for a different reason — it decides whether to take the workload-identity short-circuit before any credential is built.

**Test doesn't prove the plumbing** (🔵): correct, and it applied to my `connection.go` test too. `TestSelectAzureCredential_AuthMethodWorkloadIdentity` supplied the file via `AZURE_FEDERATED_TOKEN_FILE`, which `azidentity` reads directly, so it passed whether or not the option was forwarded. Added `..._TokenFileFromOption`, which unsets the env var so the option is the only source. Verified it discriminates by temporarily removing the forwarding: the new test fails, the env-var one still passes.

**`github-advanced-security` hardcoded-credentials alerts** (token.go:31, :33): false positives. The flagged lines are `CredentialMethodCLI CredentialMethod = "cli"` and `CredentialMethodEnv CredentialMethod = "env"` — enum values naming a sign-in method, not secrets. The rule appears to key on an identifier containing "Credential" assigned a short string literal; note it didn't flag `"managed-identity"` or `"workload-identity"` two lines below, which are the same kind of constant. Leaving the code as is rather than renaming to dodge the regex.

PR #9496, opened 2026-07-29
🐛 azure: attribute the sign-in log line and stop probing past workload identity

preslavgerchev:
## Problem

A workload-identity Azure scan logs a sign-in **per asset** that names no method, and pays for it:

```
scan for asset runner-ubk-Etac-Dev-208314 (Azure Compute VM) completed
+0.4s   → no Azure credentials were provided, trying the configured sign-in methods methods=workload-identity
+7.5s   → no Azure credentials were provided, trying the configured sign-in methods methods="cli, env, managed-identity, workload-identity"
+15.0s  x failed to get managed identity token (max retries reached) num_attempts=3
scan for asset agent-uiit-yeah-Dev (Azure Compute VM) completed
```

Two connections per asset, only one carrying `auth-method`. The one without it walks the chain in default order and burns 15s on the managed identity probe before reaching a credential that was always going to answer — once per asset, for the length of the scan.

Working out *which* connection that is was not possible from the logs. The line named neither its caller nor the connection it belonged to, and worse, the connections configured correctly were the only ones it could ever describe: `selectAzureCredential` had a shortcut where a token file with no method selection returned a bare workload identity credential and **logged nothing at all**. So every line came from a connection configured some other way, with no way to tell which one.

## Changes

**Managed identity goes last in the default chain.** It is the only method that cannot tell us it will not work. The others self-select — `NewEnvironmentCredential` and `NewWorkloadIdentityCredential` fail to construct without their configuration and `BuildChainedToken` drops them; the CLI credential fails in milliseconds when `az` is not on PATH. `NewManagedIdentityCredential` always constructs and only discovers there is no instance metadata endpoint by asking it: 5s a try, three tries. Behind the others, nothing reaches it unless nothing else worked. This is also the order azidentity's own `DefaultAzureCredential` uses.

That fixes the 15s stall for any deployment with a working credential earlier in the chain, whether or not `auth-method` ever arrives.

**Attribution.** `ChainedTokenOptions.Source` names the caller, and the log line now carries `source`, `tenant-id`, `client-id`, and whether a federated token file was found:

```
no Azure credentials were provided, trying the configured sign-in methods
  methods="cli, env, workload-identity, managed-identity"
  source=azure-connection tenant-id=… client-id=… federated-token-file=false
```

Sources are set at all five call sites — `azure-connection`, `azure-snapshot-connection`, `ms365-connection`, `k8s-aks-kubelogin`, `acr-auth-helper`. Anything still logging `source=unknown` is a caller left to wire up (cnspec-runner's extended Azure scanner is the remaining one, in the server repo).

**No more silent path.** The shortcut in `selectAzureCredential` is gone; every sign-in goes through `GetTokenFromCredential` and says who it was. This also means the fallback keeps its chain instead of collapsing to a workload-identity-only credential, so a stale token file degrades rather than hard-fails.

**One home for the tenant and client.** `GetTokenFromCredential` took `tenantId` and `clientId` positionally *and* a `ChainedTokenOptions` carrying `ClientID` and (through the embedded `DefaultAzureCredentialOptions`) `TenantID`. The positional pair overrode the options when non-empty, so a caller that set only the options had them honored on the chain path and ignored on the certificate and secret paths. They now come from the options alone.

An explicit `auth-method` selection is still honored exactly as given, in the order given.

## Breaking change for callers outside this repo

`GetTokenFromCredential(cred, tenantId, clientId, opts)` → `GetTokenFromCredential(cred, opts)`. One caller outside mql: `apps/cnspec-runner/extendedscanner/extended_azure.go` in the server repo, which needs the tenant and client moved into the options struct (and can set `Source: "extended-scanner"` while it is there).

## What this does not fix

The second connection per asset still has no `auth-method`, and this PR does not find it — it makes it name itself. Whatever it is, it no longer costs 15s to get past a managed identity endpoint that isn't there.

## Tests

- default chain keeps every method but pins managed identity last
- `hasFederatedTokenFile`: option and `AZURE_FEDERATED_TOKEN_FILE`, nil-safe
- `credentialSource`: passthrough, `unknown` when unset
- existing `providers/azure/connection` tests pass unchanged (vault credential still wins over a token file; a named method still narrows the chain rather than reordering it)

Help text for `--auth-method` in both providers now lists the methods in the order they are tried.

### DIFF
--- a/providers/azure/connection/azureinstancesnapshot/provider.go
+++ b/providers/azure/connection/azureinstancesnapshot/provider.go
@@ -91,14 +91,13 @@ func NewAzureSnapshotConnection(id uint32, conf *inventory.Config, asset *invent
 	}
 	// an empty token file leaves azidentity to fall back to
 	// AZURE_FEDERATED_TOKEN_FILE, so only the option needs forwarding here
-	chainOpts := &azauth.ChainedTokenOptions{
+	token, err := azauth.GetTokenFromCredential(cred, &azauth.ChainedTokenOptions{
+		TenantID:           conf.Options["tenant-id"],
 		ClientID:           conf.Options["client-id"],
 		FederatedTokenFile: conf.Options["azure-federated-token-file"],
 		Methods:            methods,
 		Source:             "azure-snapshot-connection",
-	}
-	chainOpts.TenantID = conf.Options["tenant-id"]
-	token, err := azauth.GetTokenFromCredential(cred, chainOpts)
+	})
 	if err != nil {
 		return nil, err
 	}
--- a/providers/azure/connection/azureinstancesnapshot/provider.go
+++ b/providers/azure/connection/azureinstancesnapshot/provider.go
@@ -91,12 +91,14 @@ func NewAzureSnapshotConnection(id uint32, conf *inventory.Config, asset *invent
 	}
 	// an empty token file leaves azidentity to fall back to
 	// AZURE_FEDERATED_TOKEN_FILE, so only the option needs forwarding here
-	token, err := azauth.GetTokenFromCredential(cred, conf.Options["tenant-id"], conf.Options["client-id"],
-		&azauth.ChainedTokenOptions{
-			FederatedTokenFile: conf.Options["azure-federated-token-file"],
-			Methods:            methods,
-			Source:             "azure-snapshot-connection",
-		})
+	chainOpts := &azauth.ChainedTokenOptions{
+		ClientID:           conf.Options["client-id"],
+		FederatedTokenFile: conf.Options["azure-federated-token-file"],
+		Methods:            methods,
+		Source:             "azure-snapshot-connection",
+	}
+	chainOpts.TenantID = conf.Options["tenant-id"]
+	token, err := azauth.GetTokenFromCredential(cred, chainOpts)
 	if err != nil {
 		return nil, err
 	}
--- a/providers/azure/connection/azureinstancesnapshot/provider.go
+++ b/providers/azure/connection/azureinstancesnapshot/provider.go
@@ -95,6 +95,7 @@ func NewAzureSnapshotConnection(id uint32, conf *inventory.Config, asset *invent
 		&azauth.ChainedTokenOptions{
 			FederatedTokenFile: conf.Options["azure-federated-token-file"],
 			Methods:            methods,
+			Source:             "azure-snapshot-connection",
 		})
 	if err != nil {
 		return nil, err
--- a/providers/azure/connection/connection.go
+++ b/providers/azure/connection/connection.go
@@ -74,14 +74,17 @@ func selectAzureCredential(conf *inventory.Config) (azcore.TokenCredential, erro
 		federatedTokenFile = os.Getenv("AZURE_FEDERATED_TOKEN_FILE")
 	}
 
-	// A token file with no method selection is unambiguously workload identity
-	// federation; azauth puts that method first rather than probing its way
-	// down to it. This used to shortcut straight to a bare workload identity
-	// credential here, which skipped the sign-in log line entirely -- so a
-	// connection that authenticated this way was invisible, and the only
-	// connections that showed up in the logs were the ones configured
-	// differently. Every path now goes through GetTokenFromCredential and says
-	// who it was.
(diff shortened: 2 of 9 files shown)


providers/azure/connection/connection.go
```diff
@@ -74,17 +74,25 @@ func selectAzureCredential(conf *inventory.Config) (azcore.TokenCredential, erro
...
+	// end of it (see azauth.DefaultCredentialMethods), so there is nothing left
+	// to shortcut past.
+	chainOpts := &azauth.ChainedTokenOptions{
+		ClientID:           clientId,
 		FederatedTokenFile: federatedTokenFile,
 		Methods:            methods,
-	})
+		Source:             "azure-connection",
+	}
+	chainOpts.TenantID = tenantId
```
preslavgerchev: do this in chain opts directly, why outside

providers/azure/connection/azureinstancesnapshot/provider.go
```diff
@@ -91,11 +91,14 @@ func NewAzureSnapshotConnection(id uint32, conf *inventory.Config, asset *invent
...
-			FederatedTokenFile: conf.Options["azure-federated-token-file"],
-			Methods:            methods,
-		})
+	chainOpts := &azauth.ChainedTokenOptions{
+		ClientID:           conf.Options["client-id"],
+		FederatedTokenFile: conf.Options["azure-federated-token-file"],
+		Methods:            methods,
+		Source:             "azure-snapshot-connection",
+	}
+	chainOpts.TenantID = conf.Options["tenant-id"]
```
preslavgerchev: do this in chain opts directly, why outside

providers/ms365/connection/connection.go
```diff
@@ -89,8 +89,13 @@ func NewMs365Connection(id uint32, asset *inventory.Asset, conf *inventory.Confi
...
 		return nil, err
 	}
-	token, err := azauth.GetTokenFromCredential(cred, tenantId, clientId,
-		&azauth.ChainedTokenOptions{Methods: methods})
+	chainOpts := &azauth.ChainedTokenOptions{
+		ClientID: clientId,
+		Methods:  methods,
+		Source:   "ms365-connection",
+	}
+	chainOpts.TenantID = tenantId
```
preslavgerchev: do this in chain opts directly, why outside

providers/azure/connection/connection.go
preslavgerchev: Fixed in d5e98ac28 — it goes in the literal now.

It was outside because `TenantID` lived in the embedded `azidentity.DefaultAzureCredentialOptions`, so it sat a level down from the `ClientID` right beside it and could not be written alongside the rest without importing azidentity at every call site.

The embed was the problem: we only ever read three of its fields (`ClientOptions`, `DisableInstanceDiscovery`, `TenantID`), so those are declared directly on `ChainedTokenOptions` and the embed is gone. One literal configures the whole sign-in now.

providers/azure/connection/azureinstancesnapshot/provider.go
preslavgerchev: Fixed in d5e98ac28 — it goes in the literal now.

It was outside because `TenantID` lived in the embedded `azidentity.DefaultAzureCredentialOptions`, so it sat a level down from the `ClientID` right beside it and could not be written alongside the rest without importing azidentity at every call site.

The embed was the problem: we only ever read three of its fields (`ClientOptions`, `DisableInstanceDiscovery`, `TenantID`), so those are declared directly on `ChainedTokenOptions` and the embed is gone. One literal configures the whole sign-in now.

providers/ms365/connection/connection.go
preslavgerchev: Fixed in d5e98ac28 — it goes in the literal now.

It was outside because `TenantID` lived in the embedded `azidentity.DefaultAzureCredentialOptions`, so it sat a level down from the `ClientID` right beside it and could not be written alongside the rest without importing azidentity at every call site.

The embed was the problem: we only ever read three of its fields (`ClientOptions`, `DisableInstanceDiscovery`, `TenantID`), so those are declared directly on `ChainedTokenOptions` and the embed is gone. One literal configures the whole sign-in now.

PR #9500, opened 2026-07-29
🐛 os: drop the zero epoch from rpm versions read from the rpmdb

czunker:
## Problem

The same rpm package reports two different versions depending on how the asset was reached:

| Collection path | Used for | `version` |
|---|---|---|
| `rpm -qa --queryformat` | live hosts (local / ssh) | `3.9.25-7.el9_8.2` |
| rpmdb parse | container images, filesystem/device scans | `0:3.9.25-7.el9_8.2` |

The purl differs with it, so downstream the same package looks like two different packages purely as a function of scan type. Consumers that compare with proper rpm version semantics are unaffected (`0:x` and `x` are equal there), but anything matching a version by prefix finds nothing at all for rpm images.

## Cause

`ParseRpmPackages` has normalized a zero epoch away since 2022 — zero is rpm's default and carries no information, so it belongs in neither the version nor the purl. #4255 later added the epoch to the rpmdb path so real epochs would show up there too, but applied the prefix unconditionally:

```go
epoch := strconv.Itoa(pkg.EpochNum())   // EpochNum() returns 0 when there is no epoch
version = epoch + ":" + version         // ...so every package gets "0:"
```

`ParseApkDbPackages` already gets this right, and guards it the same way.

## Change

- `normalizeRpmEpoch` collapses rpm's three spellings of "no epoch" (`0`, `(none)`, empty) in one place, used by both paths.
- The rpmdb loop body moves into `newRpmPackageFromDB`, which makes the static path unit-testable — it had no coverage at all, which is how this survived. (`TestPhoton4ImageParser` and `TestSuSEParser` build `rpmdb.PackageInfo` fixtures but then serialize them into the command format and assert on `ParseRpmPackages`.)

Only the zero-epoch case changes. Packages with a real epoch were already consistent between the two paths.

## Verification

`registry.access.redhat.com/ubi9/ubi:latest`, scanned both ways with a locally built provider:

```
docker image     python3  3.9.25-7.el9_8.2  pkg:rpm/redhat/python3@3.9.25-7.el9_8.2?arch=aarch64&distro=rhel-9.8
docker container python3  3.9.25-7.el9_8.2  pkg:rpm/redhat/python3@3.9.25-7.el9_8.2?arch=aarch64&distro=rhel-9.8
```

Identical; before the change the image row read `0:3.9.25-7.el9_8.2`. Non-zero epochs on `ubi8` (`gmp 1:6.1.2-11.el8`, `dmidecode 1:3.5-1.el8`) are byte-for-byte unchanged.

`TestRpmEpochParityAcrossCollectionPaths` pins the invariant by running one fixture through both paths and asserting version, purl, epoch and CPEs agree. Reverting the fix fails it on the two zero-epoch cases and leaves the non-zero case passing.

## Note for consumers

Assets scanned from images will report new purls for their rpm packages once this ships, since the `0:` disappears. Any stored data keyed by purl will see those as new packages on the next scan.

### DIFF
--- a/providers/os/resources/packages/rpm_packages.go
+++ b/providers/os/resources/packages/rpm_packages.go
@@ -44,14 +44,11 @@ func ParseRpmPackages(pf *inventory.Platform, input io.Reader) []Package {
 		m := RPM_REGEX.FindStringSubmatch(line)
 		if m != nil {
 			name := m[1]
-			epoch := m[2]
+			epoch := normalizeRpmEpoch(m[2])
 			version := m[3]
 
-			// trim epoch if it is 0 or "(none)"
-			if epoch == "0" || strings.TrimSpace(epoch) == "(none)" {
-				epoch = ""
-			} else {
-				// only append the epoch if we have a non-zero value
+			// only prefix the epoch if we have a non-zero value
+			if epoch != "" {
 				version = epoch + ":" + version
 			}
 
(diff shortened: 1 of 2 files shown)


czunker:
Scope note, since the epoch question naturally extends past this PR.

Per the [rpm purl type definition](https://github.com/package-url/purl-spec/blob/main/types/rpm-definition.json), the version component is `version-release` — "the combined version and release of an RPM" — and the epoch is a separate optional qualifier. Its own example:

```
pkg:rpm/fedora/centerim@4.22.10-1.el6?arch=i686&epoch=1&distro=fedora-25
```

So for a package with no epoch, `3.9.25-7.el9_8.2` is the correct version component, which is what this PR produces. That part is unambiguous.

For a package **with** an epoch we currently emit it twice:

```
pkg:rpm/redhat/gmp@1:6.1.2-11.el8?arch=aarch64&distro=rhel-8.10&epoch=1
                   ^^                                          ^^^^^^^
```

Strictly, the version component should be `6.1.2-11.el8`. Both collection paths do this identically, so it is consistent — but consistent is not the same as spec-correct, and I don't want the PR read as claiming otherwise.

## Why `Package.Version` must keep the epoch anyway

I checked what actually consumes this before assuming the prefix was harmless to drop, and it is not.

Distro advisories publish their fixed versions in native rpm EVR form, epoch included — e.g. RHSA-2024:1102 fixes `gmp` at `1:6.1.2-11.el8_6.1`. Vulnerability matching compares the installed version against that string with rpm version semantics, where **the epoch dominates**: it is compared first, and a difference there decides the result before version or release is looked at.

Stripping the epoch from the installed version gives:

```
installed 1:6.1.2-11.el8_6.1   vs fixed 1:6.1.2-11.el8_6.1  ->  not vulnerable
installed   6.1.2-11.el8_6.1   vs fixed 1:6.1.2-11.el8_6.1  ->  VULNERABLE
installed 1:6.1.2-99.el8_6.1   vs fixed 1:6.1.2-11.el8_6.1  ->  not vulnerable
installed   6.1.2-99.el8_6.1   vs fixed 1:6.1.2-11.el8_6.1  ->  VULNERABLE
```

The last row is the problem: `-99` is far newer than the fix and still reports vulnerable, because epoch `0` loses to epoch `1` before the release is compared. Dropping the epoch would make every epoch-bearing rpm package permanently vulnerable to every advisory against it, with no upgrade able to clear it.

So the `1:` is load-bearing. Removing it from `Package.Version` would not be a spec cleanup, it would be a false-positive generator.

## What that means for the spec question

Making the **purl** spec-correct is still possible, because package identity in advisory data is matched on a versionless purl (`pkg:rpm/redhat/gmp`) with the version compared separately — so the purl's version component is not what drives that matching. But it requires decoupling two things that are one string today: `Package.Version` keeps the epoch for comparison, while the purl version component drops it. That is a design change, it re-keys anything stored against the versioned purl, and it buys little. Not worth chasing on its own.

The same check confirms the zero-epoch case was always harmless for vulnerability matching — `0:3.9.25-7.el9_8.2` and `3.9.25-7.el9_8.2` compare equal, since an absent epoch parses as `0`. That is exactly why this bug only ever surfaced in consumers that match a version by prefix rather than by rpm semantics.

This PR stays scoped to the path divergence, where there's no such ambiguity: zero epoch carries no information, and the two paths simply disagreed.

PR #9530, opened 2026-07-30
🐛 azure: stop five fields from crossing the plugin boundary unset

preslavgerchev:
## What

A day of production scan logs carried **73 paired errors**, each one a

```
provider returned no data and no error for a field; the field was never set on the resource (provider bug)
llx: encountered a primitive with no type information, coercing to null
```

The second line is anonymous, but the first carries structured `provider`/`resource`/`field`/`id` attribution. Grouping on it resolves all 73 to exactly five resource/field pairs, every one in the azure provider:

| count | resource | field | logged `id` |
|---|---|---|---|
| 24 | `azure.subscription.webService.appsite.outboundVnetRouting` | `allTrafficEnabled` | empty |
| 21 | `azure.subscription.webService.appsiteconfig` | `minTlsCipherSuite` | real |
| 14 | `azure.subscription.aksService.cluster.autoUpgradeProfile` | `nodeOSUpgradeChannel` | empty |
| 7 | `azure.subscription.aksService.cluster.advancedNetworking` | `transitEncryptionType` | empty |
| 7 | `azure.subscription.aksService.cluster.autoUpgradeProfile` | `upgradeChannel` | empty |

The empty-vs-real `id` split separates two different bugs.

## Cause 1 — conditionally-assigned args (21 errors, real ids)

Both `appsiteconfig` creators assigned `minTlsVersion`, `ftpsState`, `minTlsCipherSuite` and `scmMinTlsVersion` only when the SDK pointer was non-nil, and skipped nine more fields entirely when Azure returned no `properties` block:

```go
if entry.Properties.MinTLSCipherSuite != nil {
    args["minTlsCipherSuite"] = llx.StringData(string(*entry.Properties.MinTLSCipherSuite))
}
```

A field omitted from the args map is never *set* on the resource — not null, unset — so `TValue.ToDataRes` encodes it as an empty `DataRes`, and the client logs the pair above once per field, per app, per scan while reporting a typeless null.

Both creators now share `siteConfigArgs`, which assigns every declared field on every path (using the existing `stringEnumPtr` helper, which already carries a doc comment about this exact trap).

## Cause 2 — resources named after the path that reaches them (52 errors, empty ids)

`azure.subscription.aksService.cluster` plus its field `autoUpgradeProfile` spells a name that is itself a resource. `mqlc.compileResource` resolves the longest matching resource name first, so the full dotted path compiles to a **bare resource creation**, not a field read on a cluster:

```
azure.subscription.aksService.cluster.autoUpgradeProfile.upgradeChannel
  c1 FUNCTION id="azure.subscription.aksService.cluster.autoUpgradeProfile" bind=0   <- no args
  c2 FUNCTION id="upgradeChannel"
```

The runtime builds that resource with no id and no fields set — which is why every one of these logged with `id=` empty — and each field read on the husk reports anonymously while the query quietly evaluates against nulls. This is the anti-pattern already documented in `CLAUDE.md` ("a resource named `<parent>.<field>` queried by its dotted path becomes an empty husk").

The three affected sub-resources now fail in `init` with the resource name and a query that works. Only the bare path reaches this — parents build them through `CreateResource`, which does not run `init`.

## Verification

Live against a Mondoo tenant, all three bare paths now return their own error instead of the anonymous pair:

```
$ mql run azure --subscription ... -c 'azure.subscription.aksService.cluster.autoUpgradeProfile.upgradeChannel'
azure.subscription.aksService.cluster.autoUpgradeProfile cannot be queried on its own,
it is only available through the resource that owns it: try
azure.subscription.aksService.clusters { autoUpgradeProfile { upgradeChannel nodeOSUpgradeChannel } }
```

and the correct nested forms still run clean.

`siteConfigArgs` is covered by a unit test that reads `azure.lr`, derives the set of non-computed fields, and asserts the creator sets all of them — with populated properties and with nil properties. The tenant has no App Service or AKS instances, so the mapping itself was not exercised against live data; it is a pure function over the SDK model and the test covers both branches.

## Notes

- No `.lr` schema changes, so no `.lr.versions` entries and no version bump.
- `azure.permissions.json` picks up the already-shipped `13.33.6` version stamp; the permission set is unchanged.
- The compiler's greedy longest-name resolution is the underlying footgun and applies to every provider with a `<parent>.<field>`-named sub-resource. Fixing it centrally (e.g. not extending the path into a `private` resource) is a broader change than this bugfix warrants, but worth considering separately.

### DIFF
--- a/providers/azure/resources/aks.go
+++ b/providers/azure/resources/aks.go
@@ -114,8 +114,24 @@ func (a *mqlAzureSubscriptionAksServiceClusterAutoUpgradeProfile) id() (string, 
 	return a.Id.Data, nil
 }
 
+// See notReachableDirectly: the auto-upgrade profile is derived from a
+// cluster's ARM properties and has no lookup of its own.
+func initAzureSubscriptionAksServiceClusterAutoUpgradeProfile(_ *plugin.Runtime, _ map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error) {
+	return nil, nil, notReachableDirectly(
+		"azure.subscription.aksService.cluster.autoUpgradeProfile",
+		"azure.subscription.aksService.clusters { autoUpgradeProfile { upgradeChannel nodeOSUpgradeChannel } }")
+}
+
 func (a *mqlAzureSubscriptionAksServiceClusterAdvancedNetworking) id() (string, error) {
 	return a.Id.Data, nil
+}
+
+// See notReachableDirectly: the advanced networking profile is derived from a
+// cluster's network profile and has no lookup of its own.
+func initAzureSubscriptionAksServiceClusterAdvancedNetworking(_ *plugin.Runtime, _ map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error) {
+	return nil, nil, notReachableDirectly(
+		"azure.subscription.aksService.cluster.advancedNetworking",
+		"azure.subscription.aksService.clusters { advancedNetworking { enabled transitEncryptionType } }")
 }
 
 // aksSecurityFlags are the cluster security-profile toggles flattened out of
--- a/providers/azure/resources/azure.lr.go
+++ b/providers/azure/resources/azure.lr.go
@@ -1149,7 +1149,7 @@ func init() {
 			Create: createAzureSubscriptionWebServiceAppsite,
 		},
 		"azure.subscription.webService.appsite.outboundVnetRouting": {
-			// to override args, implement: initAzureSubscriptionWebServiceAppsiteOutboundVnetRouting(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error)
+			Init:   initAzureSubscriptionWebServiceAppsiteOutboundVnetRouting,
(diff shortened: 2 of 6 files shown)


preslavgerchev:
Closing in favour of mondoohq/cnspec#3156, which fixes the four enterprise-policy checks that were producing 52 of these 73 errors by writing the sub-resource's dotted path instead of going through the cluster/appsite accessor.

Two things this PR was doing that the policy fix does not, recorded so they aren't lost:

1. `siteConfigArgs` — `azure.subscription.webService.appsiteconfig` leaves `minTlsCipherSuite`, `minTlsVersion`, `ftpsState`, `scmMinTlsVersion` unset when the SDK pointer is nil, and nine more fields unset when Azure returns no properties block. Accounts for the remaining 21 errors. Worth noting this is **log noise, not wrong data** — unset and null both arrive client-side as `Type: Nil, Value: nil`.

2. Init guards on the three `<parent>.<field>`-named sub-resources, so a bare dotted path fails with a message naming a query that works instead of silently evaluating against nulls. The footgun stays open without them: any resource whose name matches the path used to reach it will husk the same way, and the compiler can't currently tell those apart from resources like `aws.ec2.instance` that are legitimately queried bare (the "has a Go init" fact never reaches the schema).

PR #9783, opened 2026-08-11
🌟 llx/checksum: canonical scan-content checksum algorithm

preslavgerchev:
Canonical home of the scan-content checksum algorithm from the unchanged-scan short-circuit ADR (mondoohq/server#18038): structural 64-bit content checksums for llx values (stdlib FNV-1a — no third-party hash dependency), shared by cnspec (writing checksums into the scandb while scanning, mondoohq/cnspec#3225) and the server (recomputing and verifying uploads, mondoohq/server#18380).

## Design

- **Structural canonicalization, never marshaled proto bytes.** Proto serialization (even `Deterministic`) is not guaranteed stable across library versions, so nested encoded messages (Dict payloads, AssetValue) are decoded and re-walked.
- **Every variable-length write is length-prefixed** — concatenation ambiguity (`"ab"+"c"` vs `"a"+"bc"`) cannot collide.
- **Domain literals separate hash spaces** (`row:data`, `row:resource`, …).
- **Arrays fold as multisets** (sorted element sub-checksums): provider enumeration order is not content — the algorithm's defining property.
- `AlgoVersion = "1"` — every scandb and every manifest announce it in their metadata, so a future algorithm change (including a stronger or wider hash) reads as "no usable baseline" instead of silently mismatching. There is no fold above the rows anywhere.

## Consumers

- cnspec's scandb writer computes row checksums with this package into per-row `checksum` columns (mondoohq/cnspec#3225 — currently pins a pseudo-version of this branch).
- The server (mondoohq/server#18380) **imports this package directly** (pseudo-version pin of this branch; the earlier bit-identical `policy/scancontent` port is gone) — the golden vectors here (`bc7e2a9cf30d5d03` / `211d667e209d5ed6`) are the single source pinning the bits, and the server's cross-repo parity test proves cnspec-written columns equal its own recompute bit-for-bit.

https://claude.ai/code/session_018HHX819u6q9PngxJfpCaST


llx/digest/digest.go
```diff
@@ -0,0 +1,337 @@
...
+
+// AlgoVersion identifies the canonicalization algorithm. It is a fold input:
+// digests produced by different versions never compare equal. Bump on any
+// change to the canonicalization rules below.
+//
+// v2: arrays are hashed as multisets (per-element sub-digests, sorted) —
+// provider enumeration order is not content. v1 hashed arrays in order and
+// produced false "changed" verdicts when e.g. macOS app enumeration
+// returned the same package set in a different order.
+const AlgoVersion = "2"
```
preslavgerchev: this must be 1, new feature

PR #9856, opened 2026-08-14
⚡ os: stop recompiling the CPE epoch regexp on every package

tas50:
## What

`NewPackage2Cpe` compiled `^\d+:(.*)$` **in its own body**. Compiling a regexp builds a syntax tree, a program, and a one-pass table — about 4.6 KB and 60 allocations — and it was thrown away immediately.

dpkg calls this function **twice per package**, rpm **three times**. On a 2,000-package host that is 4,000–6,000 compilations of the same six-character pattern, per scan.

This hoists it to a package var, and swaps the `map[string]*string` that existed only to label an error message for a fixed array plus a parallel name table.

## Numbers

```
BenchmarkNewPackage2Cpe   before   4021 ns/op   4955 B/op   82 allocs/op
BenchmarkNewPackage2Cpe   after    1151 ns/op    416 B/op   21 allocs/op
```

End to end, on a synthetic 2,000-package `dpkg/status` parse (real `ParseDpkgPackages`, real CPE and purl generation) — this is where the function is hottest:

| | time | allocated | allocations |
|---|---|---|---|
| before | 47.5 ms | 34.24 MB | 505,974 |
| after | **35.9 ms** | **15.84 MB** | **261,884** |

**−54% bytes and −48% allocations on package parsing**, from one file. This is GC pressure and peak RSS on package-heavy assets, not resident baseline.

## Behavior is unchanged

Return values are identical, including the empty **non-nil** slice returned when there is no product or version to build a CPE from — callers all use `len(...) > 0` or `append(..., x...)`, but preserving it costs nothing (`[]string{}` does not allocate) so I did.

The one visible difference: iterating a map gave the `WFNize` error a **random** field name when more than one field was malformed. The array makes that deterministic.

## Testing

The table test is expanded from 3 cases to 10, pinning epoch stripping (zero, multi-digit, and the trailing-colon edge), lowercasing, the release/update field, and the three no-CPE paths.

Two things worth calling out:

- **The expanded table passes unmodified against the old implementation.** That is what makes it evidence the change is behavior-preserving rather than a description of the new code.
- **The added allocation test fails on the old implementation** (`"82" is not less than or equal to "30"`) and passes on the new one, so it guards the hoist instead of just documenting it. A future edit that moves the regexp back into the function body fails CI.

One case documents a quirk I hit while writing it: `wfn.WFNize("v1:2.3")` truncates to `v1`, because a colon separates WFN components. That is long-standing nvdtools behavior on a path this PR does not touch (the epoch pattern is anchored to leading digits, so it never matches), and the test comment says so — otherwise the next reader will blame the epoch handling.

```
go test ./providers/os/resources/cpe/ ./providers/os/resources/packages/ ./providers/os/resources/languages/...
```

all pass.

## Context

Found while auditing os provider memory. `strings.makeGenericReplacer` under `packageurl-go` is the next-largest item in the same parse (another −44%); that one needs a local purl renderer because `go.mod:75` pins v0.1.5 deliberately. Separate PR.

### DIFF
--- a/providers/os/resources/cpe/pkg2cpe.go
+++ b/providers/os/resources/cpe/pkg2cpe.go
@@ -12,7 +12,13 @@ import (
 )
 
 // epochRegex matches a version that carries an epoch, e.g. `1:2.3.4`.
+// Compiled once: this function runs two to three times per package, so
+// compiling in the body cost a few kilobytes of garbage per package on every
+// scan.
 var epochRegex = regexp.MustCompile(`^\d+:(.*)$`)
+
+// Field names for the WFNize error, positionally matched to the loop below.
+var cpeFieldNames = [...]string{"vendor", "name", "version", "release", "arch"}
 
 func NewPackage2Cpe(vendor, name, version, release, arch string) ([]string, error) {
 	cpes := []string{}
@@ -29,15 +35,9 @@ func NewPackage2Cpe(vendor, name, version, release, arch string) ([]string, erro
 	}
 
 	var err error
-	for n, addr := range map[string]*string{
-		"vendor":  &vendor,
-		"name":    &name,
-		"version": &version,
-		"release": &release,
-		"arch":    &arch,
-	} {
+	for i, addr := range [...]*string{&vendor, &name, &version, &release, &arch} {
 		if *addr, err = wfn.WFNize(*addr); err != nil {
-			return cpes, fmt.Errorf("couldn't wfnize %s %q: %v", n, *addr, err)
+			return cpes, fmt.Errorf("couldn't wfnize %s %q: %v", cpeFieldNames[i], *addr, err)
 		}
 	}
 
--- a/providers/os/resources/cpe/pkg2cpe.go
+++ b/providers/os/resources/cpe/pkg2cpe.go
@@ -34,11 +34,15 @@ func NewPackage2Cpe(vendor, name, version, release, arch string) ([]string, erro
 		version = matches[1]
 	}
 
-	var err error
 	for i, addr := range [...]*string{&vendor, &name, &version, &release, &arch} {
-		if *addr, err = wfn.WFNize(*addr); err != nil {
+		// WFNize returns an empty string alongside its error, so the result is
+		// only assigned once it succeeded. Assigning first would leave the
+		// error message reporting "" instead of the value that failed.
+		wfnized, err := wfn.WFNize(*addr)
+		if err != nil {
 			return cpes, fmt.Errorf("couldn't wfnize %s %q: %v", cpeFieldNames[i], *addr, err)
 		}
+		*addr = wfnized
 	}
 
 	// A CPE needs both a product name and a version. When either is missing we
--- a/providers/os/resources/cpe/pkg2cpe_test.go
+++ b/providers/os/resources/cpe/pkg2cpe_test.go
@@ -12,56 +12,133 @@ import (
 
 func TestPkg2Gen(t *testing.T) {
 	tests := []struct {
-		vendor   string
 		name     string
+		vendor   string
+		pkg      string
 		version  string
+		release  string
 		arch     string
 		expected []string
 	}{
 		{
-			"tar",
-			"tar",
-			"1.34+dfsg-1",
-			"",
-			[]string{
+			name:    "escapes a plus in the version",
+			vendor:  "tar",
+			pkg:     "tar",
+			version: "1.34+dfsg-1",
+			expected: []string{
 				"cpe:2.3:a:tar:tar:1.34\\+dfsg-1:*:*:*:*:*:*:*",
 			},
 		},
 		{
-			"@coreui/vue",
-			"@coreui/vue",
-			"2.1.2",
-			"",
-			[]string{
+			name:    "escapes a scoped npm name",
+			vendor:  "@coreui/vue",
+			pkg:     "@coreui/vue",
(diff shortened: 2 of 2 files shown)


providers/os/resources/cpe/pkg2cpe.go
```diff
@@ -29,15 +35,9 @@ func NewPackage2Cpe(vendor, name, version, release, arch string) ([]string, erro
...
 	var err error
-	for n, addr := range map[string]*string{
-		"vendor":  &vendor,
-		"name":    &name,
-		"version": &version,
-		"release": &release,
-		"arch":    &arch,
-	} {
+	for i, addr := range [...]*string{&vendor, &name, &version, &release, &arch} {
 		if *addr, err = wfn.WFNize(*addr); err != nil {
```
chris-rock: This line drops the value it is about to report. `WFNize` returns `""` alongside its error, and the assignment lands before the check, so the message reads `couldn't wfnize vendor ""` rather than naming the input that failed.

Pre-existing, but the loop is being rewritten here anyway:

```go
wfnized, err := wfn.WFNize(*addr)
if err != nil {
    return cpes, fmt.Errorf("couldn't wfnize %s %q: %v", cpeFieldNames[i], *addr, err)
}
*addr = wfnized
```

Reachable in practice: `NewPackage2Cpe(`v\*endor`, ...)` takes this branch.

providers/os/resources/cpe/pkg2cpe_test.go
```diff
@@ -12,56 +12,133 @@ import (
...
 
+// NewPackage2Cpe runs two to three times per package, so the regexp and the
+// field-name table are package level rather than rebuilt per call. This pins
+// the allocation count so a future edit cannot quietly move them back into the
+// function body.
+func TestNewPackage2CpeAllocations(t *testing.T) {
+	allocs := testing.AllocsPerRun(100, func() {
+		_, _ = NewPackage2Cpe("red hat, inc.", "libfoo-dev", "2:1.2.3-2.el9", "", "x86_64")
+	})
+	assert.LessOrEqual(t, allocs, float64(30), "NewPackage2Cpe allocations regressed")
```
chris-rock: `main` already allocates 26 here, since the hoist landed in #9944 — so this passes with the map back in place and would not catch a revert of the change this PR is making. `float64(25)` still leaves headroom over the current 21 and actually guards the array.

chris-rock (commented):
Rebased onto `main` and force-pushed — this is mergeable again.

**Why it conflicted, and my apologies:** #9944 landed the same epoch-regexp hoist yesterday. I wrote it while reviewing an apk parser PR, profiled my way to `NewPackage2Cpe`, and did not check for an existing open PR first. That was mine to check. Your PR predates it.

## What survived the rebase, and what it is worth now

The hoist itself is already in `main`, so the rebased diff is the `map[string]*string` → fixed array plus the expanded tests. That part is still worth landing on its own terms — measured on this machine, current `main` vs this branch, same benchmark:

| | main (hoist only) | + this PR | change |
| --- | --- | --- | --- |
| ns/op | 1100 | 942 | -14.4% |
| B/op | 496 | 416 | -16.1% |
| allocs/op | 26 | 21 | -19.2% |

Five allocations per call, at two to three calls per package, on package-heavy assets.

**Your "passes unmodified against the old implementation" claim checks out.** I ran the expanded table against `main`'s implementation (hoist landed, map still in place) and all ten cases pass, so the table is evidence rather than a description of the new code.

**The determinism claim is real, and stronger than the PR body says.** With three malformed fields, 200 runs against `main`:

```
field names reported: map[vendor:149 name:35 version:16]
```

The array pins that to `vendor`.

## Three things worth adding

**1. The error path is reachable, and untested.** `wfn.WFNize` errors when a field contains an escaped wildcard — `v\*endor` returns `unquoted '*' inside the FSB fragment`. That is the one behavior this PR changes, and nothing pins it:

```go
_, err := NewPackage2Cpe(`v\*endor`, `n\*ame`, "1.2.3", "", "x86_64")
// couldn't wfnize vendor "": unquoted '*' inside the FSB fragment: "v*endor"
```

A case asserting the message names `vendor` would guard the positional pairing between `cpeFieldNames` and the loop, which is the thing an edit could silently break.

**2. That message loses the value it is reporting.** Look at what it prints: `couldn't wfnize vendor ""` — empty. `*addr, err = wfn.WFNize(*addr)` assigns before the error is checked, and `WFNize` returns `""` on failure, so the offending input is gone by the time `%q` formats it. Pre-existing, but this PR is rewriting that exact line, so:

```go
for i, addr := range [...]*string{&vendor, &name, &version, &release, &arch} {
    wfnized, err := wfn.WFNize(*addr)
    if err != nil {
        return cpes, fmt.Errorf("couldn't wfnize %s %q: %v", cpeFieldNames[i], *addr, err)
    }
    *addr = wfnized
}
```

**3. The allocation guard no longer bites where it should.** `main` is already at 26 allocations, so the `<= 30` threshold passes with the map still in place — reintroducing it would not trip the test. `<= 25` keeps the headroom you wanted while still catching that.

## One note on the body

The numbers in the description (`4021 → 1151 ns`, the 2,000-package dpkg parse) are against a pre-#9944 baseline that no longer exists. Whoever merges this should read them as the combined effect of both PRs, not of this one.

`go test ./resources/cpe/...` passes on the rebased branch: ten table cases, the allocation guard, and `TestPlatformCPE`.

PR #9864, opened 2026-08-14
⚡ os: size the tar extract buffer from the entry header

MaxRink:
## What allocates

The container image scanner extracts a file with `ExtractFileFromTarStream`. Every `fs.Open` on a tar or image connection goes through it.

The function copied the tar entry into a `bytes.Buffer` through a `bufio.Writer`. The buffer got no size hint, so it grew by doubling. A file of N bytes therefore allocated about 2N bytes, plus two 4 KiB bufio buffers.

The tar header states the entry size. This change reads the entry into a slice of that size, then returns a `bytes.Reader` over it.

The declared size comes from the header, and a container image can come from an untrusted registry. The size is therefore trusted only up to `maxTarEntryPrealloc`, which is 64 MiB. A larger entry grows a `bytes.Buffer` instead, so a malformed header alone cannot drive the allocation.

## Measured numbers

The benchmark builds a tar with small entries around one target entry, which models a flattened image layer. The old implementation is benchmarked in the same package against the same fixture.

```
go test ./providers/os/fsutil/ -run=X -bench=ExtractFileFromTarStream -benchmem -benchtime=10x -count=5
```

Median of 5 runs:

| entry size | B/op before | B/op after | change |
|---|---|---|---|
| 4 KiB | 1,605,973 | 1,594,916 | -0.7% |
| 1 MiB | 2,507,418 | 1,452,743 | -42.1% |
| 16 MiB | 33,964,915 | 17,178,860 | -49.4% |
| 64 MiB | 134,391,466 | 67,273,786 | -49.9% |
| 128 MiB | 268,607,544 | 201,493,172 | -25.0% |

The 128 MiB entry is above the cap, so it uses the incremental path. It still improves, because the buffer starts at the cap instead of at zero.

Live heap for one 128 MiB extract, measured with `runtime.ReadMemStats` around the call on the version without the cap: bytes allocated during the extract fell from 256.1 MB to 128.1 MB, and the live heap after the extract fell from 449.5 MB to 385.5 MB.

## Time cost

There is no time cost. Median ns/op:

| entry size | before | after |
|---|---|---|
| 4 KiB | 7,005,773 | 6,968,916 |
| 1 MiB | 2,708,559 | 2,197,377 |
| 16 MiB | 9,888,202 | 6,023,807 |
| 64 MiB | 21,718,433 | 15,883,143 |
| 128 MiB | 46,301,599 | 28,719,134 |

It is faster at every size, because it removes one full copy through the `bufio.Writer` and the repeated buffer growth. No gate is needed, because the output bytes are identical.

## Correctness

* `go test ./providers/os/fsutil/... ./providers/os/connection/tar/... -count=1` passes.
* `TestExtractFileFromTarStreamLyingHeader` writes an entry that holds 4 bytes and rewrites the header to declare 1 TiB. The call returns an error, and the allocation stays at the cap.
* `TestExtractFileFromTarStreamEmptyEntry` checks that a zero length match returns an empty reader and a nil error, the same as a path that is absent.
* `TestExtractFileFromTarStreamLargeEntry` covers an entry above the cap.
* A missing path still returns an empty reader and a nil error.
* Repeated entries with the same name still concatenate, as before. The loop reads the whole stream for that reason, and the code now says so.
* `ExtractFileFromTarStream` now returns `io.Reader` instead of `*bufio.Reader`. The only caller is `tar.FS.open`, and `tar.File` only calls `Read`.

### DIFF
--- a/providers/os/connection/tar/file.go
+++ b/providers/os/connection/tar/file.go
@@ -5,8 +5,8 @@ package tar
 
 import (
 	"archive/tar"
-	"bufio"
 	"errors"
+	"io"
 	"os"
 	"path/filepath"
 	"slices"
@@ -18,7 +18,7 @@ type File struct {
 	path   string
 	header *tar.Header
 	Fs     *FS
-	reader *bufio.Reader
+	reader io.Reader
 }
 
 func (f *File) Name() string {
--- a/providers/os/connection/tar/fs.go
+++ b/providers/os/connection/tar/fs.go
@@ -5,8 +5,8 @@ package tar
 
 import (
 	"archive/tar"
-	"bufio"
 	"errors"
+	"io"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -147,7 +147,7 @@ func (fs *FS) resolveSymlink(header *tar.Header) string {
 	return path
 }
 
-func (fs *FS) open(header *tar.Header) (*bufio.Reader, error) {
+func (fs *FS) open(header *tar.Header) (io.Reader, error) {
 	log.Debug().Str("file", header.Name).Msg("tar> load file content")
 
 	// open tar file
--- a/providers/os/fsutil/stream.go
+++ b/providers/os/fsutil/stream.go
@@ -49,10 +49,13 @@ func StreamFileAsTar(
 	}
 }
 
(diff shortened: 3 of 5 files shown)


imilchev (approved):
LGTM! Thank you!

PR #9872, opened 2026-08-14
⚡ llx: compile a regex operand once

MaxRink:
## What allocates

`opStringCmpRegex` and `opRegexCmpString` implement the MQL regex operators. Both called `regexp.MustCompile` on every evaluation.

The array forms apply the operator to every element through `cmpArrayOne`, and the pattern is the same for every element. A comparison over 5,000 items therefore compiled the same pattern 5,000 times. One compile allocates about 12 KB for the syntax tree, the instruction list and the one pass program.

Both operators also matched with `r.Match([]byte(s))`, which copies the string into a new byte slice.

This change caches the compiled pattern and matches with `MatchString`, which needs no copy.

The cache is bounded. A pattern normally comes from a policy literal, so the number of distinct patterns is small. A pattern can also come from scanned data, so the cache stops storing at 128 entries and never grows past that.

This affects both scanners, because policies for nodes and for container images both use regex checks.

## Measured numbers

`BenchmarkOpStringCmpRegex` applies one pattern to 5,000 strings, which models an array comparison.

```
go test ./llx/ -run=X -bench=BenchmarkOpStringCmpRegex -benchmem -benchtime=5x -count=3
```

| | before | after | change |
|---|---|---|---|
| B/op | 63,776,838 | 483 | -99.99% |
| allocs/op | 585,034 | 0 | -100% |
| ns/op | 75,349,517 | 1,315,464 | -98.3% |

## Time cost

There is no time cost. It is 57 times faster on the benchmark above. No gate is needed, because the match results are identical.

## Correctness

* `TestCompiledRegexMatchesMustCompile` compares the cached regex against a fresh `regexp.MustCompile` over 7 patterns and 8 inputs.
* `TestOpStringCmpRegexUnchanged` checks both operators against expected results, including a case insensitive pattern and non matching input.
* `TestCompiledRegexPanicsOnInvalidPattern` checks that an invalid pattern still panics, exactly like the `regexp.MustCompile` call it replaces.
* `TestCompiledRegexIsBounded` requests three times the cache limit of distinct patterns and checks that the cache stays at or below the limit, and that every pattern still matches.
* `TestCompiledRegexConcurrent` runs 32 goroutines against the cache. `go test ./llx/... -count=1 -race` passes.
* `go test ./llx/... ./mqlc/... ./mqlx/... -count=1` passes.
* `sync.Map` is safe for concurrent use, and a compiled `regexp.Regexp` is safe for concurrent matching.

Note on the wider suite: `go test ./...` fails to build `providers` and `providers/os/connection/device/linux` on a clean checkout of main, because the generated mocks are absent. That failure is present without this change.

### DIFF
--- a/llx/regex_cache.go
+++ b/llx/regex_cache.go
@@ -31,8 +31,9 @@ func compiledRegex(pattern string) *regexp.Regexp {
 
 	r := regexp.MustCompile(pattern)
 
-	// Stop storing once the cache is full. Entries are never evicted, so the cache holds at
-	// most regexCacheMax patterns for the life of the process.
+	// Stop storing once the cache is full. The check is intentionally lock-free. Concurrent
+	// callers can pass it before LoadOrStore updates the counter, so the cache holds
+	// approximately regexCacheMax patterns for the life of the process.
 	if regexCacheLen.Load() < regexCacheMax {
 		if _, loaded := regexCache.LoadOrStore(pattern, r); !loaded {
 			regexCacheLen.Add(1)
--- a/llx/regex_cache.go
+++ b/llx/regex_cache.go
@@ -6,7 +6,6 @@ package llx
 import (
 	"regexp"
 	"sync"
-	"sync/atomic"
 )
 
 // regexCacheMax bounds the number of cached patterns. A pattern usually comes from a policy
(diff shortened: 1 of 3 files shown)


llx/regex_cache.go
```diff
@@ -0,0 +1,43 @@
...
+// It panics on an invalid pattern, exactly like the regexp.MustCompile call it replaces.
+func compiledRegex(pattern string) *regexp.Regexp {
+	if v, ok := regexCache.Load(pattern); ok {
+		return v.(*regexp.Regexp)
+	}
+
+	r := regexp.MustCompile(pattern)
+
+	// Stop storing once the cache is full. The check is intentionally lock-free. Concurrent
+	// callers can pass it before LoadOrStore updates the counter, so the cache holds
```
imilchev, line 35: i think this may be a little too simplified. We should do LRU or FIFO cache instead. This is now defined as a global cache and we do have cnspec running as a long-running service. With the current implementation that would mean that the cache is used at the beginning but over time as policies change/grow, the cache will always be full and never evicted (make it useless)

PR #9873, opened 2026-08-14
⚡ os: split dpkg control fields without a regexp

MaxRink:
## What allocates

`ParseDpkgPackages` reads `/var/lib/dpkg/status`. That file holds about 37000 lines and 1631 packages on a normal Ubuntu 24.04 host. The parser ran the regexp `^(.+):\s(.+)$` on every line. Each `FindStringSubmatch` call allocated a submatch slice plus backtracking state. A heap profile of the parse put `regexp.FindStringSubmatch` at 19.9% of allocated bytes, and `bufio.Scanner.Text` at 11.2%, once the CPE regexp cost from #9856 is taken out.

`scanner.Text` allocated one string per line. The parser keeps at most seven fields per package, so most of those line strings were garbage.

## What changed

`dpkgControlField` replaces the regexp with a backwards byte scan. The greedy first group binds the regexp to the last colon that is followed by a whitespace character, so the scan walks backwards and returns at that point. The colon and the `\s` bytes are ASCII, so they never appear inside a multi-byte UTF-8 sequence, and a byte scan finds the same split point as a rune scan.

The loop now reads `scanner.Bytes`. Every field the package keeps is copied with an explicit string conversion, so nothing aliases the scanner buffer.

`DPKG_REGEX` has no remaining user and is removed.

## Numbers

In-repo benchmark, 1600 synthetic packages in the shape of a real status file:

```
go test ./providers/os/resources/packages/ -run xxx -bench 'BenchmarkParseDpkgPackages$' -benchmem -count=6
```

| | before | after | change |
| --- | --- | --- | --- |
| B/op | 28719436 | 25046591 | -12.8% |
| allocs/op | 414549 | 352089 | -15.1% |
| ns/op | 60668386 | 32527603 | -46.4% |

Same benchmark against a real 1.65 MB `/var/lib/dpkg/status` with 1631 packages:

| | before | after | change |
| --- | --- | --- | --- |
| B/op | 34141261 | 27707335 | -18.8% |
| allocs/op | 436917 | 363187 | -16.9% |
| ns/op | 96956175 | 35202926 | -63.7% |

Stacked on top of #9856, which removes the CPE regexp compile that dominates the rest of this path, the same real-file parse goes from 18825318 to 12892407 B/op, a further -31.5%.

## Correctness

A differential test parsed the real 1.65 MB status file with the old and the new implementation. All 1631 packages compared equal with `reflect.DeepEqual`.

A differential fuzz run compared `dpkgControlField` against the old regexp over 400000 random inputs. The alphabet covered letters, colons, every byte in the `\s` class, multi-byte runes and invalid UTF-8. Inputs carry no newline, because the caller passes lines from a `bufio.Scanner`. All inputs agreed on both the match decision and the two submatches.

`TestDpkgControlField` pins the semantics, including the greedy last-colon case and a version string whose epoch colon has no following whitespace.

`go test ./providers/os/resources/packages/` passes.

### DIFF
--- a/providers/os/resources/packages/dpkg_packages.go
+++ b/providers/os/resources/packages/dpkg_packages.go
@@ -5,6 +5,7 @@ package packages
 
 import (
 	"bufio"
+	"bytes"
 	"fmt"
 	"io"
 	"os"
@@ -24,10 +25,36 @@ const (
 )
 
 var (
-	DPKG_REGEX = regexp.MustCompile(`^(.+):\s(.+)$`)
 	// e.g. source with version: samba (2:4.17.12+dfsg-0+deb12u1)
 	DPKG_ORIGIN_REGEX = regexp.MustCompile(`^\s*([^\(]*)(?:\((.*)\))?\s*$`)
 )
+
+// isDpkgFieldSpace reports whether c belongs to the regexp `\s` class. Go
+// defines that class as [\t\n\f\r ].
+func isDpkgFieldSpace(c byte) bool {
+	return c == ' ' || c == '\t' || c == '\n' || c == '\f' || c == '\r'
+}
+
+// dpkgControlField splits a dpkg control line into its field name and value.
+// It returns the same key and value as the regexp `^(.+):\s(.+)$`. The greedy
+// first group binds that regexp to the last colon followed by a whitespace
+// character. At least one byte must precede the colon, and at least one byte
+// must follow the whitespace.
+//
+// The colon and the whitespace bytes are ASCII, so they never appear inside a
+// multi-byte UTF-8 sequence. A byte scan therefore finds the same split point
+// as a rune scan. The caller passes one line from a bufio.Scanner, so the input
+// carries no newline.
+//
+// The returned slices alias line. The caller must copy what it keeps.
+func dpkgControlField(line []byte) (key []byte, value []byte, ok bool) {
+	for i := len(line) - 3; i >= 1; i-- {
+		if line[i] == ':' && isDpkgFieldSpace(line[i+1]) {
+			return line[:i], line[i+2:], true
+		}
+	}
+	return nil, nil, false
+}
 
 // ParseDpkgPackages parses the dpkg database content located in /var/lib/dpkg/status
 func ParseDpkgPackages(pf *inventory.Platform, input io.Reader) ([]Package, error) {
@@ -55,9 +82,10 @@ func ParseDpkgPackages(pf *inventory.Platform, input io.Reader) ([]Package, erro
 	scanner := bufio.NewScanner(input)
 	pkg := Package{Format: DpkgPkgFormat}
 	state := STATE_RESET
-	var key string
 	for scanner.Scan() {
-		line := scanner.Text()
+		// Bytes avoids one string allocation per line. Every field we keep is
+		// copied into the package below, so nothing aliases the scanner buffer.
+		line := scanner.Bytes()
 
 		// reset package definition once we reach a newline
 		if len(line) == 0 {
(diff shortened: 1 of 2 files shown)


chris-rock (commented):
Same shape as #9879 and it holds up the same way. I verified it rather than reading it, and then found something underneath it that I think is the more valuable half of this PR.

## The port is faithful

- **Differential against the regexp it replaces**, 400,000 random lines over letters, colons, every byte of the `\s` class, multi-byte runes and invalid UTF-8: `dpkgControlField` agrees with `^(.+):\s(.+)$` on the match decision and both submatches, every time.
- **Whole-parser differential on real data**: the pre-PR implementation restored side by side, run over `/var/lib/dpkg/status` pulled out of `debian:12`, `debian:13`, `ubuntu:22.04`, `ubuntu:24.04` and an Ubuntu image with build-essential, python3-dev, git, curl and vim installed. **618 packages, all `reflect.DeepEqual`-identical.**
- **End to end**: `os` provider built from this branch and from `main`, both run against three live containers with `mql run docker container <id> -c "packages { ... }"`. Byte-for-byte identical output — 88, 92 and 259 packages.
- **Numbers reproduce** on a real 257 KB status file (259 packages):

| | main | this PR | change |
| --- | --- | --- | --- |
| ns/op | 9,695,139 | 3,597,839 | -62.9% |
| B/op | 5,288,906 | 4,253,299 | -19.6% |
| allocs/op | 69,931 | 57,556 | -17.7% |

The backwards scan is right about the greedy binding, and the ASCII argument for scanning bytes instead of runes is right too.

## What the equivalence preserves

The regexp is wrong, and this PR is a faithful port of it. On any Debian or Ubuntu host today:

```
$ mql run docker container <debian> -c 'packages.where(name == "libc6") { name description }'
  name: "libc6"
  description: ""
```

The status file says:

```
Description: GNU C Library: Shared libraries
 Contains the standard libraries that are used by nearly all programs on
 the system. ...
```

The greedy first group binds to the **last** colon followed by whitespace, so the key becomes `Description: GNU C Library`, no case matches, `state` never becomes `STATE_DESC`, and the whole description — summary and continuation lines — is dropped. Your table pins exactly this:

```go
{"Description: the GNU shell: a thing", "Description: the GNU shell", "a thing", true},
```

That is the bug, written down as expected behavior.

The same greedy split also ends a description early when a *continuation* line contains `: ` — `gpg-wks-client` loses everything from `For more information see: https://wiki.gnupg.org/WKS` onward.

Measured across the five real status files:

| image | packages | empty descriptions | descriptions truncated |
| --- | --- | --- | --- |
| debian:12 | 88 | 2 | 5 |
| debian:13 | 78 | 2 | 5 |
| ubuntu:22.04 | 101 | 2 | 6 |
| ubuntu:24.04 | 92 | 2 | 6 |
| ubuntu + build tools | 259 | 5 | 20 |

**7.7% of packages on the largest image carry a wrong description**, including `libc6`, `libc6-dev`, `libc-bin` and `bsdutils`.

## The fix is smaller than the workaround

Debian's control format says a field name contains neither space nor colon, and a continuation line begins with a space or tab. Both rules are simpler than the regexp:

```go
func dpkgControlField(line []byte) (key []byte, value []byte, ok bool) {
	// a continuation line belongs to the field above it
	if len(line) == 0 || line[0] == ' ' || line[0] == '\t' {
		return nil, nil, false
	}
	// a field name carries neither space nor colon, so the first colon ends it
	i := bytes.IndexByte(line, ':')
	if i < 1 || i+2 > len(line)-1 || !isDpkgFieldSpace(line[i+1]) {
		return nil, nil, false
	}
	return line[:i], line[i+2:], true
}
```

I ran that against the same five files. **Empty descriptions go to zero on every image** (2→0, 5→0), 5 to 20 descriptions per image are corrected, and **no other field changes on any package** — name, version, arch, status, origin and purl are identical throughout. It also costs a forward `IndexByte` instead of a backwards loop, so it is not slower.

Exactly one case in your table changes: the `"Description: the GNU shell: a thing"` one. Every other case still passes, as do `TestDpkgParser` and `TestDpkgParserStatusD`.

This is a behavior change to shipped data, so it belongs in its own commit rather than inside a change whose whole argument is that output is identical. Happy to open it as a follow-up on top of this one — say the word.

## Two smaller things, both shared with the apk parser

Neither is introduced here; #9946 just fixed both on the apk side.

1. **`scanner.Err()` is never checked, and this function already returns an `error`.** A line over `bufio.Scanner`'s 64 KB default ends the scan, and `return pkgs, nil` reports a short package list as a clean read. The longest line I measured on a real host is 3.6 KB, so this is latent rather than live — but it is a two-line fix here, `scanner.Buffer(nil, max)` plus returning the error.
2. **`add(pkg)` after the loop logs on every healthy host.** A status file ends with a blank line, so the trailing call always gets an empty package and emits `ignored deb packages since information is missing` once per parse. It also puts a zerolog call inside your new benchmark's measured loop, which is what makes `go test -bench` output unreadable. Guarding it with `if pkg.Name != ""` fixes both.

PR #9879, opened 2026-08-14
⚡ os: split apk database fields without a regexp

MaxRink:
## What allocates

`ParseApkDbPackages` reads `/lib/apk/db/installed`. It ran the regexp `^([A-Za-z]):(.*)$` on every line.

Most lines of that database are not package metadata. They list the directories and files each package owns, on `F:`, `R:` and `Z:` lines. The parser keeps none of them. Every one of those lines still paid a `FindStringSubmatch` call, a submatch slice, and a `scanner.Text` string.

The parser also kept the submatch strings directly. A submatch aliases the line string, so each field it stored held the whole line alive.

## What changed

`apkField` replaces the regexp with two byte checks. The key of an apk field is exactly one ASCII letter followed by a colon, so no pattern matching is needed.

The loop now reads `scanner.Bytes`. A line the parser discards costs no allocation. Every field it keeps is copied with an explicit string conversion, which also drops the aliasing of the full line.

`APK_REGEX` has no remaining user and is removed.

## Numbers

```
go test ./providers/os/resources/packages/ -run xxx -bench 'ParseApkDbPackages' -benchmem -count=6
```

The benchmark builds a database of 60 packages, each with 60 files, which is the shape of a small Alpine image.

| | before | after | change |
| --- | --- | --- | --- |
| B/op | 1708985 | 629924 | -63.1% |
| allocs/op | 33286 | 9222 | -72.3% |
| ns/op | 6973171 | 1712500 | -75.4% |

The win is larger than on the dpkg parser because the file lines dominate the apk database, and the parser throws all of them away.

## Correctness

A differential test parsed the same database with the old and the new implementation. All 60 packages compared equal with `reflect.DeepEqual`.

`TestApkField` pins the semantics, including an empty value, a file line, a two letter key, a digit key, a missing key and a leading space.

`go test ./providers/os/resources/packages/` passes.

### DIFF
--- a/providers/os/resources/packages/apk_packages.go
+++ b/providers/os/resources/packages/apk_packages.go
@@ -24,7 +24,21 @@ const (
 	ApkDbInstalledUsr = "/usr/lib/apk/db/installed"
 )
 
-var APK_REGEX = regexp.MustCompile(`^([A-Za-z]):(.*)$`)
+// apkField splits a line of the apk database into its one letter field key and
+// its value. It returns the same key and value as the regexp
+// `^([A-Za-z]):(.*)$`, which the parser used before.
+//
+// The returned value aliases line, so the caller must copy what it keeps.
+func apkField(line []byte) (key byte, value []byte, ok bool) {
+	if len(line) < 2 || line[1] != ':' {
+		return 0, nil, false
+	}
+	c := line[0]
+	if (c < 'A' || c > 'Z') && (c < 'a' || c > 'z') {
+		return 0, nil, false
+	}
+	return c, line[2:], true
+}
 
 // ParseApkDbPackages parses the database of the apk package manager located in
 // `/lib/apk/db/installed`
@@ -68,9 +82,11 @@ func ParseApkDbPackages(pf *inventory.Platform, input io.Reader) []Package {
 
 	scanner := bufio.NewScanner(input)
 	pkg := Package{}
-	var key string
 	for scanner.Scan() {
-		line := scanner.Text()
+		// Bytes avoids one string allocation per line. Most lines of the apk
+		// database list files, and the parser keeps none of them. Every field
+		// it does keep is copied below.
+		line := scanner.Bytes()
 
 		// reset package definition once we reach a newline
 		if len(line) == 0 {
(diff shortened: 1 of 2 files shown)


providers/os/resources/packages/apk_packages.go
```diff
@@ -68,9 +82,11 @@ func ParseApkDbPackages(pf *inventory.Platform, input io.Reader) []Package {
 
 	scanner := bufio.NewScanner(input)
```
chris-rock, line 83: Not for this PR, but this scanner still has the default 64KB line cap and nothing checks `scanner.Err()`, so a long `D:` line ends the scan and the packages behind it vanish with no error. Since you are already in this loop, I opened #9945 stacked on top of this branch with `scanner.Buffer(nil, apkMaxLine)` plus an `Err()` check.

Worth noting for whoever reviews that: passing a **nil** initial buffer is what makes it free (757,093 vs 757,075 B/op). The more obvious `make([]byte, 0, 64*1024)` costs +61KB per parse for nothing.

providers/os/resources/packages/apk_packages.go
```diff
@@ -68,9 +82,11 @@ func ParseApkDbPackages(pf *inventory.Platform, input io.Reader) []Package {
...
 	pkg := Package{}
-	var key string
 	for scanner.Scan() {
-		line := scanner.Text()
+		// Bytes avoids one string allocation per line. Most lines of the apk
+		// database list files, and the parser keeps none of them. Every field
+		// it does keep is copied below.
+		line := scanner.Bytes()
```
chris-rock, line 89: Confirming the aliasing question this raises, since it is the one thing that could go wrong with `Bytes()`: all seven retained fields go through an explicit `string(value)` in the switch below, so nothing escapes the scanner's buffer. It is strictly better than before — the old code stored regexp submatches, and each of those pinned the entire line.

chris-rock (commented):
Nice change, and the reasoning in the description is exactly right: the file lines dominate an apk database and the parser throws all of them away, so not allocating for them is where the win is.

I verified it independently rather than reading it, and it holds up.

**Equivalence with the regexp it replaces.** I ran `apkField` against `^([A-Za-z]):(.*)$` over 500,000 random lines built from apk keys, colons, spaces, tabs, CR, NUL and invalid UTF-8. They agree on match, key and value in every case. The only input I could construct where they differ is a line containing an embedded `\n`, which `bufio.ScanLines` never produces.

**Whole-parser equivalence on real data.** I restored the old implementation side by side with the new one and compared the resulting `[]Package` with `reflect.DeepEqual` on `/lib/apk/db/installed` pulled out of eleven containers: `alpine:3.16/3.19/3.21/latest`, `wolfi-base`, `nginx:alpine`, `postgres:17-alpine`, `python:3.13-alpine`, `node:22-alpine`, `golang:alpine`, and an image with 73 packages installed. **328 packages, all identical** — name, version, epoch, arch, origin, description, license, purl, CPEs.

**End to end.** Built the `os` provider from this branch and from `main`, ran both against live containers with `mql run docker container <id> -c "packages { ... }"`. Byte-for-byte identical JSON.

**Numbers on a real 435KB database** (rather than the synthetic one), which are better than the ones in the description, because a real database has an even higher ratio of file lines to package lines:

| | main | this PR | change |
| --- | --- | --- | --- |
| ns/op | 6,383,440 | 866,022 | -86.4% |
| B/op | 3,171,716 | 757,075 | -76.1% |
| allocs/op | 63,231 | 11,206 | -82.3% |

One framing note for the description: **wall-clock per scan does not move.** Three timed runs each against a live container: 1.45s on `main`, 1.47s here. The milliseconds saved disappear next to process start, provider handshake and container transport. The value is allocation pressure where it compounds — SBOM runs across many images, or a scanner holding many assets at once — not scan latency. Worth saying plainly so nobody reads "-75%" as "scans get faster".

Two follow-ups came out of reviewing this, both opened as separate PRs since this is an external branch we cannot push to:

- **#9945**, stacked directly on this one: the scanner still has its 64KB line cap and `scanner.Err()` is never checked, so a long dependency line silently truncates the package list. Also guards the trailing `add(pkg)` that logs `ignored apk package since information is missing` once per parse of every healthy asset — which is what makes this PR's new benchmark output unreadable.
- **#9944**, independent: profiling what is left after this lands showed 95% of the remaining allocations are in the per-package `add()`, not in parsing. `cpe.NewPackage2Cpe` compiles a regexp per package, for every ecosystem. Hoisting it takes this parser a further -44% B/op and -24% ns/op.

Neither is a request to change anything here.

PR #10235, opened 2026-08-20
🐛 providers: complete a schema-only install when the provider must run (v13)

preslavgerchev:
## Problem

A provider installed with `--schema-only` (#9062, on this branch as #10142) has its config and resource schema on disk but no binary. When anything actually starts it:

- `TryProviderUpdate` sees the version is current and does nothing, and
- the coordinator then execs a binary that does not exist, failing with an obscure `fork/exec ... no such file or directory`.

## Fix

- `TryProviderUpdate` checks for the binary **before** any freshness check and completes a schema-only installation by installing the full package at the installed version (dependencies included). A current schema is no use without its binary.
- With auto-update disabled, `unsafeStartProvider` now reports `provider 'x' is installed schema-only and has no binary to run; install it fully or enable auto-update` instead of the fork/exec error.

## Why v13

This backs mondoohq/cnspec#3536, which makes scans pull only provider **schemas** for bundle requirements (fixing scans dying on providers that don't ship binaries for the scanning platform, e.g. db2 on darwin/arm64) and installs binaries per asset. With this fix, any runtime path that starts a schema-only provider self-heals; cnspec pins the v13 line. A forward-port to main follows separately.

## Verified

- `go build ./providers/` passes; `go test ./providers/` on this branch fails to build **before and after** this change (pre-existing missing `NewMockProviderPlugin` mock on v13), so the identical change was validated on main's checkout where the full providers suite passes.
- End-to-end with a cnspec build: deleting a provider's binary (leaving json + resources.json) and scanning prints `provider 'os' is installed schema-only, downloading its binary`, installs it at the pinned version, and the scan completes.

### DIFF
--- a/providers/coordinator.go
+++ b/providers/coordinator.go
@@ -314,6 +314,11 @@ func (c *coordinator) unsafeStartProvider(id string, update UpdateProvidersConfi
 	// fork/exec error, so catch schema-only installations here. With
 	// auto-update enabled, TryProviderUpdate above already completed such an
 	// installation, so this only triggers when auto-update is off.
+	// The disk is probed instead of trusting provider.HasBinary: that field
+	// is only populated by readProviderDir, so a Provider handed to the
+	// coordinator any other way (e.g. SetProviders) reads false even when
+	// the binary exists, and probing also reflects installs that happened
+	// after the provider list was cached.
 	if !config.ProbeFile(provider.binPath()) {
 		return nil, errors.New("provider '" + provider.Name + "' is installed schema-only and has no binary to run; install it fully or enable auto-update")
 	}
--- a/providers/coordinator.go
+++ b/providers/coordinator.go
@@ -15,6 +15,7 @@ import (
 	"github.com/muesli/termenv"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
+	"go.mondoo.com/mql/v13/cli/config"
 	"go.mondoo.com/mql/v13/logger"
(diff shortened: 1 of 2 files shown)


providers/coordinator.go
```diff
@@ -309,6 +310,14 @@ func (c *coordinator) unsafeStartProvider(id string, update UpdateProvidersConfi
 		}
 	}
 
+	// Without a binary the exec below can only fail with an obscure
+	// fork/exec error, so catch schema-only installations here. With
+	// auto-update enabled, TryProviderUpdate above already completed such an
+	// installation, so this only triggers when auto-update is off.
+	if !config.ProbeFile(provider.binPath()) {
```
preslavgerchev, line 322: isnt there a provider.HasBinary field?

preslavgerchev: There is, but it's a cached snapshot: `HasBinary` is only populated by `readProviderDir` (ListActive/ListAll and InstallIO results). A `Provider` that reaches the coordinator any other way — e.g. an embedder calling `SetProviders` with hand-built structs — has it `false` by zero value even when the binary exists, and the check would then refuse to start a perfectly good provider. Probing the disk is authoritative at the moment we're about to exec (it also reflects a completion that happened after the provider list was cached, e.g. by a concurrent scan), and it's a single stat() on the rare provider-start path. Same reasoning for the check in `TryProviderUpdate`. Pushed a comment into the code so the next reader doesn't have to ask.

PR #10267, opened 2026-08-20
🛑 mql run exits 0 when no asset could be connected

tas50:
## Problem

`mql run` exits **0** when it could not connect to a single asset.

Asset runtime creation (`discovery/asset_explorer.go:129`) and asset connection (`apps/mql/cmd/plugin.go`, `connectAll`) both log the failure and carry on, so that one unreachable asset does not abort a scan of many. That is the right behaviour for a partial failure. When *every* asset fails, though, `explorer.Connected()` is empty, the query loop has nothing to iterate, and `RunQuery` returns `nil` — so the command prints a fatal connection error and reports success.

Anything scripting `mql run` therefore cannot distinguish "every asset passed" from "no asset was ever scanned". `--exit-1-on-failure` does not help: it governs query *result* failures, and no query ever ran.

Reproducible on `main` with any unreachable target:

```console
$ mql run ssh root@192.0.2.1 -c 'asset.name'
! could not find keys in ssh agent
x unable to create runtime for asset error="rpc error: code = Unknown desc = no authentication method defined" asset=
$ echo $?
0
```

This surfaced while verifying #8441. The negative control scan failed with `googleapi: Error 400 ... Source zonal disk and target zonal disk must be in the same zone` and still exited 0, which is exactly the signal a CI job would need and would not get.

## Fix

Collapse the no-usable-asset case into an error. The explorer already records both failure points in `Errors()`, and duplicate assets return before being recorded there, so one check covers everything without new plumbing:

```go
connectAll(explorer, explorer.Discovered())
allAssets := explorer.Connected()
if err := noUsableAssetsError(len(allAssets), explorer.Errors()); err != nil {
    return err
}
```

`noUsableAssetsError` reports the first underlying cause rather than a generic message, since that cause is what a user needs.

**Partial failures are deliberately unchanged.** If any asset connected, its queries ran, and `--exit-1-on-failure` continues to govern result failures. Only the all-or-nothing case changes.

## Verification

Against a build of this branch:

| command | before | after |
|---|---|---|
| `mql run ssh root@192.0.2.1 -c 'asset.name'` | exit **0** | exit **1**, `could not connect to asset <unnamed asset>: ... no authentication method defined` |
| `mql run local -c 'asset.platform'` | exit 0 | exit 0, unchanged |

`TestNoUsableAssetsError` covers the single-failure, all-failed, unnamed-asset, partial-success, all-connected, and nothing-discovered cases.

## Compatibility

This is a behaviour change: a script that today sees exit 0 from a wholly failed run will now see exit 1. That is the point of the fix, but it is worth calling out for anyone whose automation has been passing on runs that never scanned anything.

## Tests this changed, and why

Three subtests in `TestProvidersEnvVarsLoading` depended on the old behaviour, and they are worth reading as evidence rather than as collateral damage.

**`command WITHOUT path`** ran `mql run fs` with neither `--path` nor `MONDOO_PATH`. That cannot connect (`missing filesystem mount path, use 'path' option`), and the test asserted **exit 0 with an empty package list**. That is exactly the failure mode this PR is about: a caller who forgets `--path` gets a result that reads as "this filesystem has no packages installed." It now asserts the misconfiguration is reported.

**The two `ssh localhost` subtests** assert on flag and config-binding messages in stderr. Whether `ssh://localhost` connects depends on the machine running the test — on CI it is `connection refused` — so their exit code was incidental rather than their subject. The exit-code assertion is dropped and the stderr assertions that carry their meaning are unchanged. Swapping `0` for `1` there would have hardcoded "no sshd on the runner", which is more fragile than not asserting it.

No other test in the repo asserted on this behaviour.

### DIFF
--- a/apps/mql/cmd/plugin.go
+++ b/apps/mql/cmd/plugin.go
@@ -144,6 +144,9 @@ func (c *mqlPlugin) RunQuery(conf *run.RunQueryConfig, runtime *providers.Runtim
 	// Recursively connect all discovered assets depth-first
 	connectAll(explorer, explorer.Discovered())
 	allAssets := explorer.Connected()
+	if err := noUsableAssetsError(len(allAssets), explorer.Errors()); err != nil {
+		return err
+	}
 
 	if conf.Format == "json" {
 		_ = out.WriteString("[")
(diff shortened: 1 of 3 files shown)


arlimus (approved):
Makes sense, kudos!

PR #10280, opened 2026-08-21
🐛 release-providers: route bump PR to v{major} for older majors

philipbalinov:
## Summary

`release-providers.yml` fires on `release: [published]` and opens a "bump provider versions" PR — but `base:` was hardcoded to `main`. In the v14-on-main / v13-support model, a `v13.35.x` release tag would try to open its bump PR against `main` (v14 code) with a massive v13-vs-v14 diff.

Complements Ivan's [#10060](https://github.com/mondoohq/mql/pull/10060) which stopped `providers.yaml` from firing on push to `main`: that fix prevents provider publishes from main, this fix keeps the version-bump PR chain routed to the right branch for v13 patches.

## Fix

The base decision now lives in `.github/scripts/determine-provider-release-base.sh`. It reads main's `go.mod` (via `gh api`) to learn the module major, then:
- `release_major == main_module_major` → `base=main`
- `release_major != main_module_major` → `base=v{release_major}` (e.g. `v13`)
- workflow_dispatch (no release tag) → `base=main`

Sparse-checkouts `.github/scripts` from `main` first (so the routing script is available before we know which branch we're targeting), then the full checkout uses the resolved base as `ref:` so `peter-evans/create-pull-request` opens a clean PR containing just the version bumps on top of the target branch's HEAD.

## Testability

- **`bats` unit tests** in `.github/scripts/tests/determine-provider-release-base.bats` — 11 cases covering matching/mismatched majors, v-prefixed and bare versions, pre-release suffixes, both v13-on-main and v14-on-main worlds, and the "no module line" and "missing tag" error paths. Runs in CI via the new `test-scripts.yml` workflow on any PR that touches `.github/scripts/**`.

### Run bats locally

```
bats -r .github/scripts/tests
```

## Motivation
Prereq for cutting v13.x provider patches while main tracks v14.

### DIFF
--- a/.github/scripts/determine-provider-release-base.sh
+++ b/.github/scripts/determine-provider-release-base.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+# Copyright Mondoo, Inc. 2024, 2026
+# SPDX-License-Identifier: BUSL-1.1
+#
+# Determine base branch for the provider version-bump PR opened after
+# an mql release.
+#
+# When main's own module major matches the incoming release's major,
+# the PR opens against main. Otherwise it opens against the v{major}
+# support branch (e.g. a v13.35.1 release while main is on v14 -> v13).
+#
+# Inputs:
+#   $1 (required)  incoming release tag (e.g. v13.35.1 or 14.0.0-rc.1)
+#   $MAIN_GOMOD    content of main's go.mod. If unset, fetched via
+#                  `gh api` using $GH_REPO (owner/repo) and $GH_TOKEN.
+#
+# Output: prints the base branch to stdout ("main" or "v{major}").
+#         Exits non-zero if main's module major cannot be determined.
+set -euo pipefail
+
+TAG="${1:?release tag required as first argument}"
+
+V="${TAG#v}"
+RELEASE_MAJOR="${V%%.*}"
+
+if [ -z "${MAIN_GOMOD+set}" ]; then
+  # MAIN_GOMOD not passed in at all — fetch it. An explicit empty value
+  # is respected (and will fail the "Could not extract" check below), so
+  # tests can exercise failure paths without touching the network.
+  : "${GH_REPO:?GH_REPO or MAIN_GOMOD required}"
+  MAIN_GOMOD=$(gh api "/repos/${GH_REPO}/contents/go.mod?ref=main" --jq '.content' | base64 -d)
+fi
+
+# module go.mondoo.com/mql/v13  -> 13
+MAIN_MAJOR=$(printf '%s' "$MAIN_GOMOD" \
+  | grep -oE '^module go\.mondoo\.com/mql/v[0-9]+' \
+  | head -1 \
+  | grep -oE '[0-9]+$' || true)
+
+if [ -z "$MAIN_MAJOR" ]; then
+  echo "Could not extract main's module major from go.mod" >&2
+  exit 1
+fi
+
+if [ "$RELEASE_MAJOR" = "$MAIN_MAJOR" ]; then
+  echo main
+else
+  echo "v${RELEASE_MAJOR}"
+fi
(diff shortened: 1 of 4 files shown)


vjeffrey (commented):
Nice to see the routing logic pulled into a testable script with real `bats` coverage. Two blockers before this can land, though — and the first one isn't in the code you wrote.

### 1. The script can never succeed against today's `main`

The major is extracted with:

```bash
grep -oE '^module go\.mondoo\.com/mql/v[0-9]+'
```

but `main`'s `go.mod` is now:

```
module go.mondoo.com/mql
```

That changed yesterday in 55e66c9e — **"🧹 v14 -> non-versioned go-mod (#10234)"** (2026-08-20). The major suffix was deliberately dropped for v14. Tag `v13.35.2` still carries `module go.mondoo.com/mql/v13`, so the assumption held right up until it didn't.

Result: `MAIN_MAJOR` comes back empty, the script `exit 1`s, and **every `release: published` run fails at "Determine base branch"** — no bump PR at all, for any major. Worth noting the `"missing module line fails loudly"` bats case is effectively encoding main's current real `go.mod` shape as the error path.

### 2. Merging to `main` doesn't affect the case this is for

`release: published` runs the workflow file from the tag's commit, not from the default branch. From the actual run history:

```
release  head_branch=v13.35.2  head_sha=243f3cb0  success
release  head_branch=v13.35.1  head_sha=243f3cb0  success
```

`243f3cb0` is the tag's commit, and it's an ancestor of the `v13` branch (`main...v13.35.2` reports diverged, 15 ahead / 165 behind). The `v13` branch's own `release-providers.yml:125` still reads `base: main`. So a `v13.35.3` release executes the *old* workflow and opens the v13-vs-v14 PR against main exactly as it does today. And on `main`, releases are v14, so `base=main` and nothing changes there either.

You half-spotted this — the sparse-checkout comment says *"the v13 branch's tag may predate this script's existence"* — but the same reasoning applies to the workflow steps themselves, and those can't be sparse-checked-out from main. **This needs backporting to `v13` to have any effect.**

### Suggested fix that resolves both

Stop asking `go.mod` and ask whether the support branch exists. `v13` exists today; `v14` doesn't:

```bash
MAJOR="${TAG#v}"; MAJOR="${MAJOR%%.*}"
if gh api "repos/${GH_REPO}/branches/v${MAJOR}" >/dev/null 2>&1; then
  echo "v${MAJOR}"
else
  echo main
fi
```

Correct right now (v13 tags → `v13`, v14 tags → `main`), self-maintaining across the v14 cut with no edit needed, and independent of a `go.mod` convention that just moved. The existing bats suite maps over cleanly — swap the `MAIN_GOMOD` fixture for a stubbed `gh` on `PATH`.

### Smaller notes

- `RELEASE_MAJOR="${V%%.*}"` doesn't validate semver. A tag like `latest` yields base `vlatest`, which surfaces as a confusing `actions/checkout` failure several steps later rather than a clear error. Worth a guard now that this script decides where release automation writes.
- `concurrency: group: release-providers` is global across branches, so a v13 and a v14 release close together serialize rather than run in parallel. `cancel-in-progress: false` means nothing is lost — fine, just worth knowing once two majors are live.

### Things I checked that are fine

- The `MAIN_GOMOD` unset-vs-empty handling via `${MAIN_GOMOD+set}` is correct, and the comment accurately describes it.
- `BRANCH_NAME` is timestamped, so main-targeted and v13-targeted bump PRs can't collide on the head branch.
- The second `actions/checkout` does call `disableSparseCheckout()` when the `sparse-checkout` input is absent ([`git-source-provider.ts:257`](https://github.com/actions/checkout/blob/v7.0.1/src/git-source-provider.ts#L257)), so the full tree materializes correctly after the sparse first pass. No issue there.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

vjeffrey (approved):

PR #10283, opened 2026-08-21
🐛 os: drop macOS system_profiler entries that are not application bundles

### LINKED ISSUE #10282: macOS packages: non-application entries emitted with empty versions and unusable PURLs
## Summary

On macOS, the `packages` resource emits entries whose `name` is not an application
name at all, but a bundle-identifier fragment or a framework/SDK directory name.
Every one of them has an **empty version**, which produces a PURL with no version
component.

## Observed packages

All from arm64 macOS hosts. These are verbatim `packages` output:

```json
{ "name": "com",                       "version": "", "installed": true, "arch": "arm64",
  "purl": "pkg:macos/macos/com?arch=arm64" }
{ "name": "org.eclipse",               "version": "", "installed": true, "arch": "arm64",
  "purl": "pkg:macos/macos/org.eclipse?arch=arm64" }
{ "name": "Microsoft.NETCore.App",     "version": "", "installed": true, "arch": "arm64",
  "purl": "pkg:macos/macos/Microsoft.NETCore.App?arch=arm64" }
{ "name": "Microsoft.AspNetCore.App",  "version": "", "installed": true, "arch": "arm64",
  "purl": "pkg:macos/macos/Microsoft.AspNetCore.App?arch=arm64" }
```

Also observed with the same shape (empty version, non-app name):
`k8s`, `esxi`, `gcp`, `gcp.project.cloudRunService`, `cloudflare`, `testapp`, `TestApp`.

Note `com` and `org.eclipse` — those look like the leading components of reverse-DNS
bundle identifiers (`com.*`, `org.eclipse.*`). `Microsoft.NETCore.App` and
`Microsoft.AspNetCore.App` are the shared-framework directory names under
`/usr/local/share/dotnet/shared/`.

A second, distinct group has legitimate Apple bundle names but is still versionless:
`ShortcutsActions`, `PassViewer`, `liquiddetectiond`. Those are probably the
"bare `.app` daemon with no `Info.plist`" case the code already comments on, and may
be working as intended — worth separating from the group above when triaging.

## Where it comes from

`providers/os/resources/packages/macos_packages.go` — `ParseMacOSPackages` assigns
`system_profiler`'s `_name` directly with no validation:

```go
pkgs[i].Name = entry.Name
...
pkgs[i].PUrl = purl.NewPackageURL(
    platform, purl.TypeMacos, entry.Name, version, purl.WithQualifiers(purlQualifiers),
).String()
```

`bundleVersionFromInfoPlist` already falls back to `CFBundleVersion` when
`system_profiler` reports no version, and returns `""` when there is no
`Info.plist`. So a versionless entry means neither source had a version.

`entry.Path` is captured into `pkgs[i].Files` but is never used to validate the
entry — that's the most likely place to hang a fix, since the path should reveal
whether these are real app bundles.

## Repro

1. On an arm64 macOS host — ideally a developer machine with the .NET SDK and an
   Eclipse install, since those two reliably produce entries above:

   ```bash
   system_profiler SPApplicationsDataType -xml \
     | plutil -convert json - -o - \
     | jq '._SPCommandLineArguments? // .[0]._items[] | {_name, version, path}' \
     | jq -s '[.[] | select(.version == null or .version == "")]'
   ```

   Confirm whether `system_profiler` itself reports these `_name` values, and what
   `path` each one carries. That single answer splits the bug:

   - **`system_profiler` reports them** → it is walking non-application directories,
     and mql needs to filter (e.g. require the path to end in `.app`, or require a
     readable `Info.plist`).
   - **`system_profiler` does not report them** → the plist decode in
     `ParseMacOSPackages` is mis-associating `_name`, and the bug is ours.

2. Then compare against the resource output:

   ```bash
   cnspec run local -c 'packages.where(version == "") { name version purl }'
   ```

## Impact

- **Unusable PURLs.** `pkg:macos/macos/com?arch=arm64` has no version, so nothing
  downstream can ever match it to advisory data or compare it against a range.
- **Inventory noise.** These appear in software inventory as if they were installed
  applications. `com` and `gcp` are not products, and a user cannot act on them.
- **Ambiguity.** A one-word name like `gcp` or `com` could collide with a real
  product name in any name-based lookup.

## Suggested direction

Skip entries that have no usable version *and* whose path is not a real `.app`
bundle, rather than emitting a versionless PURL. If they are worth keeping for
inventory completeness, they at least need to be distinguishable from real
applications. Logging the skipped entries at debug level with their path would make
the next report of this much cheaper to diagnose.

tas50:
Fixes #10282

## What was happening

`system_profiler SPApplicationsDataType` is not a list of installed applications. It enumerates **every path carrying a bundle-like extension**, and it names each item after the basename minus its final dot component. That is why reverse-DNS directories collapse into fragments — `group.is.workflow.my.app` is reported as `group.is.workflow.my`, and by the same mechanism a `com.*` or `org.eclipse.*` directory is reported as `com` or `org.eclipse`.

`ParseMacOSPackages` emitted every item verbatim, so those directories became inventory entries with an empty version and an unusable purl like `pkg:macos/macos/com?arch=arm64`.

The issue framed this as a binary triage: either `system_profiler` reports these names (→ filter on a `.app` suffix), or our plist decode mis-associates `_name` (→ our bug). Running the repro on an arm64 host gives a third answer: `system_profiler` **does** report them, but a `.app` suffix filter would not help — **110 of the 111 junk paths already end in `.app`**. They are Firefox origin-storage directories, app-group script containers, and `Library/HTTPStorages` entries.

Verbatim, from this host:

```
https+++bsky            /Users/…/Firefox/Profiles/…/storage/default/https+++bsky.app
group.is.workflow.my    /Users/…/Library/Application Scripts/group.is.workflow.my.app
com.apple.ctcategories  /Users/…/Library/HTTPStorages/com.apple.ctcategories.service
liquiddetectiond        /System/Library/CoreServices/liquiddetectiond.app
```

## The fix

The discriminator is `Contents/Info.plist`, not the path suffix — an application bundle always has one. When `system_profiler` reports no version **and** no `Info.plist` can be read, the entry is skipped and logged at debug level with its path, as the issue suggested.

`bundleVersionFromInfoPlist` already opened that file for the version fallback, so this adds no extra I/O — it now returns whether the bundle was found alongside the version.

**The check is gated on an already-empty version, deliberately.** `/Applications/Victory.app` is a wrapped iOS app: a `WrappedBundle` symlink into `Wrapper/`, with no top-level `Contents/Info.plist`. It reports a version, and dropping it would be a regression. Versioned entries are never checked.

## Measured on a real arm64 host

| | before | after |
|---|---|---|
| items from `system_profiler` | 497 | 497 |
| packages reported | 497 | **386** |
| versionless packages | 116 | **4** |

All 111 dropped entries lack a `Contents/Info.plist`. The 4 survivors are genuine bundles (`CFBundlePackageType => APPL`) that simply ship no version keys — `PassViewer`, `ShortcutsActions`, `Proton Mail Uninstaller`, `qFlipper`. That is the second group the issue asked to separate out during triage; they are real installed applications, so they are kept rather than dropped.

## Test plan

`macos_packages_test.go` now covers all four quadrants of version × `Info.plist`:

- versioned + `Info.plist` → kept (`Preview`, `Contacts`, `Firefox`)
- versionless + `Info.plist` **with** a version key → version recovered (`Microsoft Teams (PWA)`, from `CFBundleVersion`)
- versionless + `Info.plist` **without** version keys → kept, empty version (`qFlipper`)
- versioned + **no** `Info.plist` → kept (`Victory`, the wrapped-iOS-app regression guard)
- versionless + **no** `Info.plist` → dropped (`liquiddetectiond`, `https+++bsky`, `group.is.workflow.my`)

The testdata paths are the real shapes observed on the host, not invented ones.

```
go test ./resources/packages/...   # ok
go vet ./resources/packages/       # clean
```

Also driven end-to-end through a real local connection against this host's live `system_profiler` output to produce the table above.

### DIFF
--- a/providers/os/resources/packages/macos_packages.go
+++ b/providers/os/resources/packages/macos_packages.go
@@ -64,35 +64,60 @@ func ParseMacOSPackages(conn shared.Connection, platform *inventory.Platform, in
 		return nil, errors.New("format not supported")
 	}
 
-	pkgs := make([]Package, len(data[0].Items))
-	for i, entry := range data[0].Items {
-		// We need a special handling for Firefox to determine ESR installations
-		purlQualifiers := getPurlQualifiers(conn, entry)
-
+	pkgs := make([]Package, 0, len(data[0].Items))
+	for _, entry := range data[0].Items {
 		// system_profiler only surfaces CFBundleShortVersionString as the
 		// version. Some bundles (e.g. PWAs) ship a version only in
 		// CFBundleVersion, so fall back to the bundle's Info.plist when
 		// system_profiler reports no version.
 		version := entry.Version
 		if version == "" {
-			version = bundleVersionFromInfoPlist(conn, entry.Path)
+			// system_profiler lists every path carrying a bundle-like
+			// extension, not only application bundles: Firefox origin storage
+			// (https+++example.app), app-group script containers
+			// (group.is.workflow.my.app) and daemon directories all show up. It
+			// names them after the basename minus its final dot component, so a
+			// reverse-DNS directory collapses into a fragment like "com" or
+			// "org.eclipse". An application bundle always carries a
+			// Contents/Info.plist, so its absence means this is not an installed
+			// application: drop the entry instead of reporting it with a
+			// versionless purl that can never match advisory data.
+			//
+			// Entries that do report a version are never checked, because some
+			// real applications have no Contents/Info.plist. Wrapped iOS apps
+			// keep theirs under Wrapper/ and would otherwise be lost.
+			bundleVersion, isBundle := bundleVersionFromInfoPlist(conn, entry.Path)
+			if !isBundle {
+				log.Debug().
+					Str("name", entry.Name).
+					Str("path", entry.Path).
+					Msg("skipping entry that is not an application bundle")
+				continue
+			}
+			version = bundleVersion
 		}
 
-		pkgs[i].Name = entry.Name
-		pkgs[i].Version = version
-		pkgs[i].Format = MacosPkgFormat
-		pkgs[i].FilesAvailable = PkgFilesIncluded
-		pkgs[i].Arch = platform.Arch
-		pkgs[i].PUrl = purl.NewPackageURL(
-			platform, purl.TypeMacos, entry.Name, version, purl.WithQualifiers(purlQualifiers),
-		).String()
+		// We need a special handling for Firefox to determine ESR installations
+		purlQualifiers := getPurlQualifiers(conn, entry)
+
+		pkg := Package{
+			Name:           entry.Name,
+			Version:        version,
+			Format:         MacosPkgFormat,
+			FilesAvailable: PkgFilesIncluded,
+			Arch:           platform.Arch,
+			PUrl: purl.NewPackageURL(
+				platform, purl.TypeMacos, entry.Name, version, purl.WithQualifiers(purlQualifiers),
+			).String(),
+		}
 		if entry.Path != "" {
-			pkgs[i].Files = []FileRecord{
+			pkg.Files = []FileRecord{
 				{
 					Path: entry.Path,
 				},
 			}
 		}
+		pkgs = append(pkgs, pkg)
 	}
 
 	return pkgs, nil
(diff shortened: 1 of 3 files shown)


czunker (approved):
LGTM

Thanks @tas50

PR #10347, opened 2026-08-24
⚡ deduplicate concurrent resource resolution in the provider SDK

imilchev:
Closes mondoohq/server#18630 (the `NewResource` half; see [Not in scope](#not-in-scope) for the rest).

## The defect

Two concurrent resolutions of the same resource both entered its `init`, both spent the API call, and the loser's freshly built resource was discarded on the way out. The cache cannot prevent this: the cache key is the *resolved* resource's `MqlID()`, which is not knowable until `init` has already done the work.

The argument set **is** knowable up front, and two calls with identical arguments must by definition produce the same resource. `ResolveResource` collapses them onto one flight.

Two related fixes came with it:

**Child connections now inherit flights along with the cache.** `service.go` gives a child runtime its parent's `Resources` map. Scoping flights per-runtime would have left sibling assets sharing a cache while duplicating the work that fills it — the two scopes have to agree.

**The generated code had a check-then-act race.** It did `Get` → miss → `Set`. Two callers that both miss the `Get` both `Set`, so **two objects escape for one cache key**, and every field memoized on the loser is invisible to the rest of the graph. `syncx.Map.GetOrSet` (atomic `LoadOrStore`) makes the cache converge.

## Measurement

A/B against a **neutered** build — the same binary with the deduplication bypassed — so the probe is proven to reach the changed code. 16 assets, `cnspec scan aws --parallelism 8`, three runs each:

| | no dedup | with dedup |
|---|---|---|
| `DescribeVpcs` (20 distinct VPCs) | 97 | **20** — exactly 1.00× |
| `DescribeVolumes` | 10 | **2** |
| targeted init calls, 3 runs | 108 / 108 / 107 | **22 / 22 / 22** |
| init executions | 399–440 of 399–440 asked | 450–501 of 670–781 asked |
| wall time (mean of 3) | 10.0s | 10.6s |
| check results | — | **identical on all 16 assets** |

~80% fewer targeted calls, and the deduplicating build does **more** traversal work (643 VPC resolutions vs 407) while making a fifth of the calls. The only difference in the JSON reports is the per-run random incognito asset MRN.

**Wall time does not move**, and the table says so deliberately. These are single-item `Describe` calls running alongside the region-fanout list calls that dominate the critical path. The win is API quota and rate-limit headroom, not latency.

**This only pays off above `--parallelism 1`.** At parallelism 1 both builds are identical — 108 calls each, max observed init concurrency 1 — because `llx` executes a single asset strictly serially. It is worth doing now because parallelism defaults to 8 for aws.

## Design notes

- **`sync.Once` was the obvious idea and is wrong twice over.** It would seal `NotReady`, which must stay retryable (`providers/os/resources/command.go:27` depends on it), and there is nowhere to put it: `TValue[T]` is assigned by value at 1,048 hand-written sites, so a lock inside it is a copylocks violation *and* a proactive assignment would replace the lock while another goroutine holds it.
- **Key canonicalisation refuses rather than guesses.** Every part is length-prefixed so no value can forge a separator, the type travels with the value so `"1"` and `1` cannot collapse, and any argument that is not a plain scalar declines a key outright. A missed deduplication is free; a collision returns the wrong resource.
- **Errors are shared within a flight and never cached.** `singleflight` forgets the key when the flight completes, so the next caller retries. Covered by a test.
- **Nested resolution.** Distinct flights never block each other, which is what every typed accessor relies on. Same-key re-entry from inside an `init` would deadlock — but that is infinite recursion today, so it is not a new broken program. Tested.

## Diff shape

+932 / −3,737 across 89 files. The generated `NewResource` / `CreateResource` bodies collapse to one SDK call each, which is where the deletions come from and why the logic is now unit-testable in `providers-sdk/v1/plugin` instead of duplicated into 82 providers.

## Tests

New `resource_factory_test.go`. Both halves are mutation-verified — removing the single-flight fails the concurrency test, reverting `GetOrSet` fails the convergence test.

The convergence test needed real work to be load-bearing: a plain stress loop caught the regression 1 run in 10, and a channel gate made it *worse* (Go wakes goroutines serially, so the first one finishes its `Get`/`Set` before the others are scheduled). A spin barrier sized to `GOMAXPROCS` over 100 rounds gets it to **10/10**, while the correct implementation passes 20 consecutive runs in 0.5s.

## Verification

- all 82 providers build
- 243 packages pass; `-race` clean on `providers-sdk/v1/plugin`
- `golangci-lint` v2: 0 issues on the changed packages
- codegen reproducible — regenerating every `.lr` produces no further diff
- one pre-existing failure in `llx` (`TestRawDataJson_Umlauts`, a UTF-8 replacement-char assertion); `llx` is untouched by this PR

## Not in scope

The issue's `GetOrCompute` half is left alone. It is unmeasured by the issue's own admission, and the pervasive provider idiom is *one compute populating several fields* — a per-field lock does not deduplicate that at all, which is why `mqlCommand` hand-rolls its own. Worth measuring before paying for it.

Same for a single-flight on the client at `providers/runtime.go:582`, which would also save the gRPC round-trip — but it would hand one `*llx.RawData` to all waiters where each gets its own today, so it wants a measurement rather than an assumption.

### DIFF
--- a/providers-sdk/v1/mqlr/lrcore/go.go
+++ b/providers-sdk/v1/mqlr/lrcore/go.go
@@ -184,38 +184,7 @@ func NewResource(runtime *plugin.Runtime, name string, args map[string]*llx.RawD
 		return nil, errors.New("cannot find resource " + name + " in this provider")
 	}
 
-	if f.Init != nil {
-		cargs, res, err := f.Init(runtime, args)
-		if err != nil {
-			return res, err
-		}
-
-		if res != nil {
-			mqlId := res.MqlID()
-			id := name + "\x00" + mqlId
-			if x, ok := runtime.Resources.Get(id); ok {
-				return x, nil
-			}
-			runtime.Resources.Set(id, res)
-			return res, nil
-		}
-
-		args = cargs
-	}
-
-	res, err := f.Create(runtime, args)
-	if err != nil {
-		return nil, err
-	}
-
-	mqlId := res.MqlID()
-	id := name + "\x00" + mqlId
-	if x, ok := runtime.Resources.Get(id); ok {
-		return x, nil
-	}
-
-	runtime.Resources.Set(id, res)
-	return res, nil
+	return plugin.ResolveResource(runtime, name, args, f)
 }
 
 // CreateResource is used by the runtime of this plugin to create resources.
@@ -227,19 +196,7 @@ func CreateResource(runtime *plugin.Runtime, name string, args map[string]*llx.R
 		return nil, errors.New("cannot find resource " + name + " in this provider")
 	}
 
-	res, err := f.Create(runtime, args)
-	if err != nil {
-		return nil, err
-	}
-
-	mqlId := res.MqlID()
-	id := name + "\x00" + mqlId
-	if x, ok := runtime.Resources.Get(id); ok {
-		return x, nil
-	}
-
-	runtime.Resources.Set(id, res)
-	return res, nil
+	return plugin.InstantiateResource(runtime, name, args, f)
 }
 `
 }
--- a/providers-sdk/v1/plugin/resource_factory.go
+++ b/providers-sdk/v1/plugin/resource_factory.go
@@ -1,0 +1,165 @@
+// Copyright Mondoo, Inc. 2024, 2026
+// SPDX-License-Identifier: BUSL-1.1
+
+package plugin
+
+import (
+	"sort"
+	"strconv"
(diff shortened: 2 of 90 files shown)


arlimus (changes requested):
Working on getting this into v14 - amazing stuff!!
The newly introduced cross-asset resolution makes this a bit blocking because of the flight group sharing, which causes problems there. Working on resolving it

PR #10467, opened 2026-08-26
🐛 network: parse the url query and fragment

tas50:
## What

Three fields on the `url` resource reported nothing.

**`query` was never populated.** `initUrl` fills in scheme, user, password, host, port, path, `rawQuery` and `rawFragment` — but not `query`, and there is no computed accessor for it either. Every read crossed the plugin boundary unset:

```
$ mql run host … -c 'url("https://u:p@example.com:8443/a/b?x=1&y=2#frag").query'
x provider returned no data and no error for a field; the field was never set
  on the resource (provider bug)  field=query resource=url
x llx: encountered a primitive with no type information, coercing to null
  query: null
```

This hit 25 of 25 hosts in a sweep — it is not conditional on anything.

**`rawFragment` was empty for ordinary fragments.** It reads Go's `u.RawFragment`, which the standard library only fills in when a fragment carries escaping that differs from the canonical encoding. `#section` leaves it empty, so the fragment was reported as absent.

**`string` dropped the fragment entirely** — escaped or not. `URL.String` renders the fragment from `Fragment` and consults `RawFragment` only when the two agree, so setting `RawFragment` alone rendered nothing.

## Changes

- `query` is parsed from the URL. A query may repeat a key and the field is a `map`, so the first occurrence wins — what `url.Values.Get` returns and what most servers read. Noted in the code.
- `rawFragment` falls back to the decoded fragment when `RawFragment` is empty, so it reports the fragment either way and still preserves unusual escaping as sent.
- `string` supplies both `Fragment` and `RawFragment`, so the reassembled URL round-trips.

Before / after on the same input:

```
url("https://u:p@example.com:8443/a/b?x=1&y=two#frag")
  query        null  ->  {x: "1", y: "two"}
  rawFragment  ""    ->  "frag"
  string       "https://u:p@example.com:8443/a/b?x=1&y=two"
            -> "https://u:p@example.com:8443/a/b?x=1&y=two#frag"
```

## Test plan

`url_internal_test.go`:

- **query** — splits parameters, is an empty map rather than absent when there is no query, keeps the first of a repeated key, decodes percent-encoding, and leaves `rawQuery` as sent.
- **fragment** — an ordinary `#section`, one with unusual escaping (`#frag%2Fx`) kept as sent, and none at all.
- **round trip** — six URLs reassembled through `string()` and compared against the input, covering fragment, query, both, userinfo with a port, and neither.

`go test ./...` in `providers/network/` passes. Verified live: the provider-bug warning is gone and all three fields report correctly.

## Note on scope

`rawFragment` and `string` were filed as a lower-severity finding than `query`, but all three live in the same function and the fragment ones are two lines. Shipping the `query` fix while leaving the fragment broken in the same resource seemed worse than fixing both.

## Schema

No `.lr` change — these fields are already declared; they were simply never filled in.

### DIFF
--- a/providers/network/resources/url.go
+++ b/providers/network/resources/url.go
@@ -11,6 +11,7 @@ import (
 
 	"go.mondoo.com/mql/llx"
 	"go.mondoo.com/mql/providers-sdk/v1/plugin"
+	"go.mondoo.com/mql/types"
 )
 
 func initUrl(runtime *plugin.Runtime, args map[string]*llx.RawData) (map[string]*llx.RawData, plugin.Resource, error) {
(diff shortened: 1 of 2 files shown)


arlimus (approved):
Thank you for fixing it!

PR #7818, opened 2026-05-22
🧹 rpm: switch queryformat delimiter from __ to ASCII RS

czunker:
## Summary

Replaces the \`__\` field separator in the rpm queryformat / RPM_REGEX with ASCII Record Separator (\`0x1E\`, expressed as \`\\036\` in the rpm queryformat).

Stacked on **#7814** (\`package.installDate\`); please review/merge that PR first. GitHub will auto-retarget this PR to \`main\` once #7814 lands.

## Why

The bot flagged this on #7813 (twice): a summary or license field containing a literal \`__\` would split into the wrong column. It was already latent before this PR series — every existing field (vendor, summary, modularity) is theoretically vulnerable — but the addition of \`%{LICENSE}\` made it slightly more likely since SPDX expressions occasionally contain underscores.

ASCII RS is purpose-built for this: rpm rejects non-printable characters in any header tag (NAME, VENDOR, SUMMARY, LICENSE, MODULARITYLABEL), so RS cannot legitimately appear in any field. The parser becomes unambiguous and no longer relies on \`__\` being absent from free-text strings.

## Mechanical changes

- **\`RPM_REGEX\`** switches from \`__\` literals to \`\x1e\` with \`[^\x1e]\` negated classes for the variable-width segments — eliminates the regex backtracking ambiguity that \`(.*?)\` between \`__\` separators implied.
- **\`RpmPkgManager.queryFormat()\`** emits \`\\036\` between fields. Raw Go string keeps the source readable.
- **\`providers/os/resources/reboot/rhel.go\`** reuses \`ParseRpmPackages\` on the kernel package; updated to match.
- **Test fixtures (.toml)** use \`\\u001E\` Unicode escape in stdout values (TOML basic strings decode to the actual RS byte that Go regex matches) and \`\\\\036\` in command keys (TOML decodes to literal \`\\036\`, matching the queryformat string that rpm receives).
- **\`TestPhoton4ImageParser\`** synthetic format string switches from \`__\` to \`\\x1e\`.

## Behaviour

No change for any well-formed rpm output. Removes the class of "package summary contains \`__\`" failures.

## Test plan

- [x] \`go test ./providers/os/resources/packages/...\` — pass
- [x] \`go test ./providers/os/resources/reboot/...\` — pass
- [x] \`golangci-lint run\` — 0 issues
- [x] \`gofmt\` — clean
- [ ] CI green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10606, opened 2026-08-31
🐛 os: rpm 7 reports no packages, MODULARITYLABEL needs a lower version bound

tas50:
## Impact

On RHEL, CentOS and Oracle Linux **7**, `packages` returns an **empty list with
no error** while the host has a fully populated rpm database. Every
package-based vulnerability, patch and compliance check evaluates against
nothing, and nothing anywhere reports a problem. RHEL 7 ELS runs to 2028, so
this is live.

```
# docker.io/library/centos:7, arm64, os provider 13.40.9
$ rpm -qa | wc -l
151
$ mql run local -c 'packages.length'
packages.length: 0          # no error, no warning
```

## Why it is invisible

mql asks rpm for `%{MODULARITYLABEL}`, a tag that only exists in rpm **4.14+**.
CentOS 7 ships rpm **4.11.3**:

```
$ rpm --version
RPM version 4.11.3
$ rpm -qa --queryformat '...__%{MODULARITYLABEL}\n'
error: incorrect format: unknown tag
error: incorrect format: unknown tag
...
$ echo $?
0                                    # <-- rpm EXITS 0
$ rpm -qa --queryformat '%{NAME} %{VERSION}\n' | wc -l
151                                  # <-- fine without the tag
```

rpm writes an error per package to **stderr**, writes **nothing** to stdout,
and **exits 0**. An exit-status check cannot catch this; the only signal is
empty stdout against a populated database, which parses cleanly to an empty
package list.

## Root cause

`modularitySupportedByPlatform` in `providers/os/resources/packages/rpm_packages.go`
had an **upper** bound (`< 10`, because RHEL 10 dropped modularity) but **no
lower** bound, so version 7 fell into the supported branch. Modularity arrived
in RHEL **8**; the supported window is `[8, 10)`.

`amazonlinux` is not in the switch, which is why Amazon Linux 2 (also rpm 4.11)
reports its packages correctly. That was the control that confirmed the
diagnosis.

## Verification

Built a linux/arm64 os provider and ran it in containers.

| image | rpm | `rpm -qa` | `packages.length` before | after |
|---|---|---|---|---|
| `docker.io/library/centos:7` | 4.11.3 | 151 | **0** | **151** |
| `registry.access.redhat.com/ubi8/ubi` | 4.14.3 | 185 | 185 | 185 |
| `registry.access.redhat.com/ubi9/ubi` | 4.16.1.3 | 188 | 188 | 188 |

The risk of this change is suppressing modularity where it *is* supported, so
two controls:

- The debug trace shows the tag is still requested on 8 and 9 and dropped only
  on 7:
  ```
  ubi9  : rpm -qa --queryformat '...__%{INSTALLTIME}__%{MODULARITYLABEL}\n'
  ubi8  : rpm -qa --queryformat '...__%{INSTALLTIME}__%{MODULARITYLABEL}\n'
  cos:7 : rpm -qa --queryformat '...__%{INSTALLTIME}\n'
  ```
- Installing a modular package on ubi8 (`dnf install nodejs`) shows the label
  still parses into the purl:
  ```
  rpm : nodejs nodejs:10:8030020210225164533:229f0a1c
  mql : pkg:rpm/redhat/nodejs@1:10.24.0-1.module+el8.3.0+10166+b07ac28e
        ?arch=aarch64&distro=rhel-8.10&epoch=1
        &rpmmod=nodejs:10:8030020210225164533:229f0a1c
  ```

## Tests

`TestModularitySupportedByPlatform` now walks all five distro names in the
switch across versions 7, 7.9, 8, 8.10, 9, 9.4, 10, 10.1 plus an empty and a
malformed version, and adds unlisted names (`amazonlinux`, `fedora`, `sles`,
`opensuse`, `photon`) at every version.

Against the **unpatched** function the new table fails on exactly 10 subtests,
the five distros x {`7`, `7.9`}:

```
--- FAIL: TestModularitySupportedByPlatform/oraclelinux/"7"
--- FAIL: TestModularitySupportedByPlatform/oraclelinux/"7.9"
--- FAIL: TestModularitySupportedByPlatform/almalinux/"7"
...
    Error:   Not equal: expected: false / actual: true
```

## Follow-up (not in this PR)

`ParseRpmPackages` on empty input yields an empty list, so any future
queryformat that rpm cannot render fails the same silent way (as #7818 did). A
guard that treats "empty rpm output against a non-empty database" as an error
would catch the whole class, but that is a behaviour change worth its own PR.

### DIFF
--- a/providers/os/resources/packages/rpm_packages.go
+++ b/providers/os/resources/packages/rpm_packages.go
@@ -602,16 +602,21 @@ func modularitySupportedByPlatform(platform *inventory.Platform) bool {
 
 	switch platform.Name {
 	case "oraclelinux", "almalinux", "redhat", "centos", "rocky":
+		// The window is [8, 10): modularity was introduced with RHEL 8 and
+		// removed again in RHEL 10. Do not widen it. On RHEL 7 and older rpm is
+		// 4.11, which has no %{MODULARITYLABEL} tag: it writes an error per
+		// package to stderr, writes nothing to stdout, and still exits 0, so
+		// asking for the tag there returns an empty package list with no error.
 		vParser := semver.Parser{}
-		cmp, err := vParser.Compare(platform.Version, "10")
+		lower, err := vParser.Compare(platform.Version, "8")
 		if err != nil {
 			return false
 		}
-		if cmp < 0 {
-			supported = true
-		} else {
-			supported = false
+		upper, err := vParser.Compare(platform.Version, "10")
+		if err != nil {
+			return false
 		}
+		supported = lower >= 0 && upper < 0
 	}
 
 	return supported
--- a/providers/os/resources/packages/rpm_packages_test.go
+++ b/providers/os/resources/packages/rpm_packages_test.go
@@ -549,68 +549,55 @@ func TestRedHatModularParser(t *testing.T) {
 	assert.Equal(t, p.PUrl, fPkg.PUrl)
 }
 
+// TestModularitySupportedByPlatform pins the version window in which the
+// %{MODULARITYLABEL} queryformat tag may be requested: modularity arrived in
+// RHEL 8 and was removed again in RHEL 10, so only 8.x and 9.x qualify.
+//
+// The version-7 rows guard a regression that silently emptied the package
+// inventory. The guard used to carry only the upper bound, so RHEL, CentOS and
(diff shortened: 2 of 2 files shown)


czunker (approved):
LGTM

PR #10665, opened 2026-09-03
🐛 os: ollama integration models are empty when the config omits the tag

tas50:
## What

`ollama.config.integration.models` returned `[]` on a host that had the model pulled.

Ollama treats an untagged model name as implicitly `:latest` — `ollama run qwen3.5` and `ollama run qwen3.5:latest` are the same model — but the model store reports the fully qualified name. The join in `ollama.go` was an exact string match, so a config recording the bare name matched nothing.

Observed with a config naming `qwen3.5` against a store holding `qwen3.5:latest`.

## Fix

`matchOllamaModel` resolves in three steps:

1. exact name
2. the implicit `:latest` form, when the configured name carries no tag
3. the reverse — a config pinned to `:latest` against a store entry reported bare

A bare name still does **not** match a differently tagged entry: `qwen3.5` must not silently resolve to `qwen3.5:1b`. That case is pinned by a test.

## Test

`TestMatchOllamaModel` is a table test over a pure helper, covering the bare↔tagged cases in both directions plus the negatives (differently-tagged entry, unknown model, empty name).

Confirmed it can fail — removing the `:latest` fallback gives:

```
--- FAIL: TestMatchOllamaModel/bare_name_matches_the_:latest_entry
    expected: true
    expected: string("tagged")
```

Found while verifying the os provider against Linux containers for the v13→v14 diff. This field is part of that diff.

### DIFF
--- a/providers/os/resources/ollama.go
+++ b/providers/os/resources/ollama.go
@@ -564,11 +564,36 @@ func (i *mqlOllamaConfigIntegration) models() ([]any, error) {
 		if !ok {
 			continue
 		}
-		if m, ok := byName[name]; ok {
+		if m, ok := matchOllamaModel(byName, name); ok {
 			out = append(out, m)
 		}
 	}
 	return out, nil
+}
+
+// matchOllamaModel resolves a configured model name against the model store.
+//
+// Ollama treats an untagged name as implicitly :latest - `ollama run qwen3.5`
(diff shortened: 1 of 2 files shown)


arlimus (approved):
oh nice correction, kudos @tas50 !!

PR #10711, opened 2026-09-07
🐛 providers: move the Windows plugin port range off 10000 and make it configurable

AdamVB:
## Problem

Providers are hashicorp/go-plugin subprocesses. On Linux and macOS the host reaches them over a Unix socket, but go-plugin hard-codes loopback TCP on Windows and lets the *client* pick the port range. Our `plugin.ClientConfig` never set one, so go-plugin's default of 10000-25000 applied, tried in ascending order. The first provider a scan starts (usually `os`) therefore always took port 10000, which is a popular application port (Webmin, plenty of Java stacks), and any product that needs it fails its own port check while a scan is running.

There was no workaround either: go-plugin passes the range to the subprocess as `PLUGIN_MIN_PORT`/`PLUGIN_MAX_PORT` *appended after* the host environment, so setting those on the host was silently overridden.

## Fix

- **New default range 50000-65535**, inside the IANA dynamic range where no service is supposed to claim a fixed port. go-plugin skips busy and excluded ports (Hyper-V/WSL reservations), and each running provider takes one port.
- **`PLUGIN_MIN_PORT` / `PLUGIN_MAX_PORT` from the host environment are honored**, so go-plugin's own knobs work the way their names suggest. Both must be set together.
- **New `provider_port_range` config key** (`"min-max"`, e.g. `50000-50100`) in `mondoo.yml`, also readable as `MONDOO_PROVIDER_PORT_RANGE`. It lives in the shared `cli/config` layer, so cnspec picks it up with its next mql bump without a code change.
- **Precedence:** config key > `PLUGIN_*` env pair > default. A present-but-malformed value fails provider start with a clear error rather than silently keeping the range the operator was trying to leave.

Linux and macOS keep using Unix sockets and are unaffected: the fields are set but the Unix listener ignores them.

## Verification

- Unit tests for the resolution order and every malformed shape (`providers/plugin_ports_test.go`), and for the config key including the env binding through the real `InitViperConfig` (`cli/config/config_test.go`).
- End to end on macOS: spawned the `os` provider under each setting and read `PLUGIN_MIN_PORT`/`PLUGIN_MAX_PORT` from the child's environment. Default, config key, env pair, config-beats-env, and both error paths all behaved as designed. The child's environment is exactly what go-plugin's Windows listener reads, but Windows itself was not exercised here.
- `golangci-lint run ./providers/ ./cli/config/`: 0 issues.

## Follow-ups

- The docs site needs an entry for `provider_port_range` / `MONDOO_PROVIDER_PORT_RANGE`; no in-repo doc lists `mondoo.yml` keys.
- Longer term the right fix is Unix sockets on Windows (AF_UNIX exists since Windows 10 1803 / Server 2019), which needs an upstream go-plugin change.

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -212,6 +212,20 @@ func GetUpdatesURL() string {
 	return viper.GetString("updates_url")
 }
 
+// KeyProviderPortRange is the config key for the loopback TCP port range that
+// provider subprocesses may listen on, written as "min-max" (for example
+// "50000-50100"). It comes from mondoo.yml or, through viper's env binding,
+// from MONDOO_PROVIDER_PORT_RANGE. Only Windows uses TCP for the provider
+// transport; other platforms talk to providers over Unix sockets and ignore it.
+const KeyProviderPortRange = "provider_port_range"
+
+// GetProviderPortRange returns the provider_port_range setting, or an empty
+// string when unset. Parsing and the default live with the providers package,
+// which is the one that hands the range to the plugin runtime.
+func GetProviderPortRange() string {
+	return viper.GetString(KeyProviderPortRange)
+}
+
 // GetFeatures returns the features from viper config.
 // This can be called after InitViperConfig() to get features before cobra initialization.
 func GetFeatures() mql.Features {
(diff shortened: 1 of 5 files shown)


imilchev (approved):
LGTM

PR #10756, opened 2026-09-11
backport 🐛 os/packages: recover the real .NET release from the packed MSI version (#10710)

tas50:
Backport of #10710 to `v13`.

## What it fixes

A .NET runtime install can report a version nobody recognises — `64.120.56788` for what the product, its own DisplayName, and `dotnet --list-runtimes` all call **8.0.30**.

Microsoft's installers write the MSI `ProductVersion` into Add/Remove-Programs, and Windows Installer constrains that to major ≤ 255 / minor ≤ 255 / build ≤ 65535. .NET packs the release to fit — first component = major × 8 — so the value shares no scale with anything:

| .NET release | reported | | .NET release | reported |
|---|---|---|---|---|
| 2.1.30 | `16.120.30411` | | 6.0.36 | `48.144.23141` |
| 3.1.32 | `24.192.31915` | | 8.0.30 | `64.120.56788` |
| 5.0.17 | `40.68.31213` | | 10.0.10 | `80.40.55332` |

Installing the **standalone MSI** rather than the bundle — what a managed Intune/SCCM rollout does — registers only that entry, so the packed value is the only .NET runtime version cnspec can report for the host, and the DisplayName is the only place the real release appears at all.

The fix recovers the release from the DisplayName, in `createPackage` before the PURL is built, so `Version` and `PUrl` cannot disagree, gated on `windows/app`. It is applied to both the bundle and the MSI entry, which describe a single install: collapsing them onto the same release makes their PURLs identical, so one runtime stops producing two findings carrying the same CVEs.

## What changed relative to the original

Nothing in the logic. `windows_packages.go` cherry-picked cleanly, and the two new test files are byte-identical to the merged versions apart from their imports, which were rewritten to the v13 module path:

```
go.mondoo.com/mql/providers-sdk/v1/inventory
  -> go.mondoo.com/mql/v13/providers-sdk/v1/inventory
```

`git diff origin/main HEAD` over the three files shows only those import lines.

Authorship on the commit is preserved as the original author.

## Verification on this branch

```
go build ./providers/os/...                      ok
go vet ./providers/os/resources/packages/        ok
go test ./providers/os/resources/packages/       ok
```

The backported tests all pass, including the three per-host-shape tests that are the point of the change:

```
--- PASS: TestDotNetMsiOnlyHost
--- PASS: TestDotNetBundleHost
--- PASS: TestDotNetSideBySideReleases
--- PASS: TestNormalizeDotNetPackedVersion            (24 cases)
--- PASS: TestCreatePackageNormalizesDotNetVersion
```

No provider version bump here — the release flow owns that.

### DIFF
--- a/providers/os/resources/packages/windows_dotnet_hostshapes_test.go
+++ b/providers/os/resources/packages/windows_dotnet_hostshapes_test.go
@@ -1,0 +1,218 @@
+// Copyright Mondoo, Inc. 2024, 2026
+// SPDX-License-Identifier: BUSL-1.1
+
+package packages
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.mondoo.com/mql/v13/providers-sdk/v1/inventory"
+)
+
+// The .NET runtime installers register a DIFFERENT SET of Add/Remove-Programs
+// entries depending on how the runtime was deployed, and the per-entry unit
+// tests in windows_dotnet_version_test.go cannot see that. These tests run the
+// real ParseWindowsAppPackages over whole host shapes.
+//
+// The two JSON blobs below were dumped off a clean Windows 11 ARM64 VM with the
+// same registry query cnspec itself runs (installedAppsScript), once after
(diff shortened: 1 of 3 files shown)


czunker (approved):
Thanks @tas50

PR #10849, opened 2026-09-14
✨ together: model the cluster management plane and its nodes

tas50:
This was written stacked on #10841, which touched the same file. #10841 merged while this branch was being finished, so the branch is now rebased straight onto `main` and this PR is a single commit.

A Together AI GPU cluster shipped with its compute described and its administrative surface invisible. Whether the Kubernetes dashboard is up, whether a jumphost is deployed, whether SSH access is signed by the cluster's own CA, what load balancer fronts it, whether ingress is on: none of it was readable. Nor was any machine in the cluster, including GPU workers that answer on a public IPv4.

**Everything here is already in the `Beta.Clusters.List` response.** Confirmed by reading `together.Cluster` in together-go v0.12.0: `ClusterConfig`, `OidcConfig`, `GPUWorkerNodes` and `ControlPlaneNodes` are all fields of the record the provider already decodes in `clusters()`. Zero new API calls, no new permission.

## What is added

`together.cluster`:

| field | source | why it matters |
|---|---|---|
| `kubernetesDashboardEnabled` | `cluster_config.kubernetes_dashboard_enabled` | a web console with its service account's permissions, on a cluster holding model weights |
| `jumphostEnabled` | `cluster_config.jumphost_enabled` | a reachable host from which the cluster network can be entered |
| `sshCaEnabled` | `cluster_config.ssh_ca_enabled` | false means SSH access is not brokered by OIDC-signed certificates |
| `loadBalancer` | `cluster_config.load_balancer` | NONE, TRAEFIK, NGINX or ISTIO (the complete set from the SDK's `Any of` comment) |
| `ingressEnabled` | `cluster_config.ingress.enabled` | inbound traffic accepted |
| `oidcGroupClaim`, `oidcGroupPrefix`, `oidcUsernameClaim`, `oidcUsernamePrefix` | `oidc_config` | a group claim that is unset, or names a claim the IdP does not assert, leaves group role bindings in place and silently unmatched |
| `nodes` | `gpu_worker_nodes` + `control_plane_nodes` | below |

Note `loadBalancer` is read from `cluster_config.load_balancer`, **not** from the legacy top-level `install_traefik` that sits next to it in the same struct. The fixture sets `install_traefik: false` on a cluster whose load balancer is `TRAEFIK`, so reading the wrong one fails the test.

**Null, not false, when the API reported nothing.** `cluster_config` and `oidc_config` are optional in the response. Rather than letting the SDK's zero value through, each control is carried only when `respjson.Field.Valid()` says the API reported it. A cluster whose configuration block is absent reports null, and `kubernetesDashboardEnabled == false` then *fails* on it instead of passing a check nobody measured. A reported `false` still reads as `false` (asserted separately). The two shipped fields `oidcIssuer`/`oidcClientId` keep their existing empty-string behavior; changing them would change shipped semantics for no gain in this PR.

## The node resource shape, and why

**One resource, `together.cluster.node`, with a `role` discriminator (`GPU_WORKER` / `CONTROL_PLANE`), not two collections.**

The question worth asking sweeps both kinds:

```
together.clusters.all(nodes.none(isPublic))
```

A control plane node fronting the cluster API is at least as interesting as a worker, and with two collections every exposure check has to be written twice. The half that gets forgotten is the control plane. The two records also overlap on nearly everything they carry: `node_id`, `host_name`, `status`, `public_ipv4`, network(s), `memory_gib`, `num_cpu_cores`.

The four fields only a GPU worker reports (`numGpus`, `instanceId`, `autoRemediationEnabled`, `markedForDeletion`) are **null** on a control plane node, not zero, so nothing claims a measurement the record does not carry. `ClusterControlPlaneNode` reports a single `network` where the worker reports `networks`; both reach the list field.

**`isPublic`** is computed from the public IPv4 the response carries, which is what `isPublic` means on the ~20 other resources that declare it (CLAUDE.md §5).

**`__id` = `<clusterId>/<role>/<nodeId>`.** Per CLAUDE.md §4, the key has to carry every dimension the thing repeats along: a node repeats across clusters, across the two collections inside one cluster, and by its own identifier. Two nodes sharing a key would be reported as one, silently carrying the first one's values. Hostname is the fallback if a record omits `node_id` (it is `api:"required"`, so this is defensive), and the position within the cluster is the last resort so that two anonymous records cannot collapse onto each other. The synthetic key is not exposed; `id` is the API's own `node_id`.

## Kubeconfig cannot leak

⚠️ `Cluster.KubeConfig` (`kube_config`) is a **full kubeconfig with cluster credentials, returned inline in this exact list response**.

- Nothing in this change reads `c.KubeConfig`. No `dict` field, no passthrough, no field that serializes the cluster object or a node record wholesale. Every field is assigned individually from a named SDK field.
- `TestKubeConfigNeverReachesTheSchema` proves it mechanically: it first asserts the fixture really does carry a kubeconfig and that the SDK decodes it into `Cluster.KubeConfig` (so the response genuinely hands the provider the secret), then enumerates **every declared field** of `together.cluster` and of each `together.cluster.node` through the generated `getDataFields` accessors, flattens each result including array members and map values, and fails if the marker appears anywhere. A passthrough or dict field added over the cluster object later fails this test without anyone having to remember the rule.
- Proven non-vacuous by mutation: assigning `llx.StringData(c.KubeConfig)` to `cudaVersion` fails the test.

Also checked and deliberately **not** modeled: `FineTuningNewParams.WandbAPIKey` and `BetaModelRemoteUploadNewParams.Token` are request-only params, never returned. The existing `together.deployment.environmentVariableNames` (names only, no values) remains the precedent for anything env-carrying; nothing added here carries environment content.

## Tests

`providers/together/resources/cluster_test.go` decodes a cluster fixture through the real client against an `httptest` server. Addresses come from the documentation range 203.0.113.0/24 and every identifier is invented.

| test | what it catches | mutation proven |
|---|---|---|
| `...ExposureControlsComeFromTheClusterConfig` | mistyped or crossed mapping; reading `install_traefik` instead of `cluster_config.load_balancer` | — |
| `...OidcClaimsComeFromTheOidcConfig` | crossed claim/prefix pair (the two prefixes differ in the fixture) | — |
| `...ExposureControlsAreNullWhenTheApiReportsNothing` | absent config reported as a confident `false` | dropping the `Valid()` check ⇒ `kubernetesDashboardEnabled should be null, got false` |
| `...NodesCoverBothCollections` | a collection dropped; control plane `network` not reaching `networks` | — |
| `...NodeIsPublicFollowsThePublicAddress` | predicate not moving with the address, in both directions | `return true, nil` ⇒ fails |
| `...NodeFieldsTheRecordDoesNotCarryAreNull` | worker-only fields reading 0/false on a control plane node, and absent optional flags reading false | same `Valid()` mutation ⇒ `worker markedForDeletion should be null, got false` |
| `...ClusterNodeIDSeparatesClusterRoleAndNode` | a key dimension missing | dropping `role` from the key ⇒ duplicate key |
| `...EveryNodeGetsItsOwnCacheEntry` | two nodes collapsing into one cache entry | covered by the same mutation |
| `...KubeConfigNeverReachesTheSchema` | any field carrying the credential | `cudaVersion` ⇒ kubeconfig fails the sweep |

The existing `together_test.go` and `whoami_test.go` still pass.

## Verification

Inside `providers/together`:

- `gofmt -l .` clean, `go build ./...`, `go test ./...` pass, `go vet ./...` clean, `go mod tidy` no diff
- `./mqlr generate providers/together/resources/together.lr --dist providers/together/resources` run; `.lr.versions` entries land at `14.0.0`, which is the provider's `Version` (`14.0.0-rc.1`) per the precedent set in #10833. `Version` itself is untouched. No breaking-change warnings: every entry in the diff is additive.
- `make providers/build/together` succeeds; the provider carries no `permissions.json`.
- Schema resolves offline against the built plugin (`PROVIDERS_PATH=… mql run together --info -c …`), which is what rules out the two codegen naming traps:

```
- together.cluster
  - nodes  type=[]together.cluster.node
- together.cluster.node
  - isPublic  type=bool
  - publicIpv4  type=string
  - numGpus  type=int
```

`nodes` resolves as a list of the node resource rather than collapsing to a single resource, and the node fields resolve rather than coming back as an empty husk. `together.clusters.all(nodes.none(isPublic))` and `together.clusters.none(kubernetesDashboardEnabled)` both compile.

## What this rests on

**No live Together account was available.** What is proven: the schema compiles against the built provider, the decode paths, the null handling, the key construction, and the absence of the kubeconfig from every declared field. What is **not** proven against a real API response: that a live control plane node really omits the GPU worker fields as the SDK types say, that `load_balancer` only ever takes the four documented values, and that `cluster_config` is present on a live cluster at all (the null path is the honest answer if it is not). These need an account with a provisioned GPU cluster.


arlimus (approved):
Kudos for expanding it!

PR #10850, opened 2026-09-14
🌟 mistral: model connectors and document libraries, and stop discarding four decoded fields

tas50:
## What

Two new Mistral surfaces plus four fields the provider already decoded and threw away.

### `mistral.connectors()` → `mistral.connector`

`GET /v1/connectors`. A connector is a remote server the models in a workspace may call, so it is both an egress path and a supply-chain surface. Fields: `id`, `name`, `description`, `server`, `protocol`, `visibility`, `ownerType`, `supportedAuthMethods`, `privateToolExecution`, `createdAt`, `modifiedAt`.

```
mistral.connectors.where(visibility == "shared_org") { name server }
```

**No authentication material is modeled.** The connector payload does carry credential-shaped fields: `connection_credentials`, and on each supported auth method both `headers` (names, `is_secret`) and `global_headers` — whose spec text says the value "is plaintext in memory so create round-trips and encryption-at-rest keep the real value; secrets are redacted only on JSON serialization". None of that is decoded. `supportedAuthMethods` is the list of `method_type` names and nothing else. A test feeds a payload containing all of it and asserts the re-encoded struct contains none of it.

### `mistral.libraries()` → `mistral.library` + `accesses()`

`GET /v1/libraries` for the corpus metadata, `GET /v1/libraries/{id}/share` for who can read it. `accesses()` is a lazy accessor, one call per library, so it costs nothing unless queried. Document contents are not modeled.

```
mistral.libraries.where(nbDocuments > 0) { name accesses { shareWithType shareWithUuid role } }
```

### Tier A fields, all previously decoded and discarded

| field | why |
|---|---|
| `mistral.fineTuningJob.integrations` | a Weights and Biases integration streams training telemetry to a third party; unreviewed third-party egress on a training job was invisible |
| `mistral.model.replacementModel` | we shipped `deprecation` with no migration target, which made the deprecated-model query unactionable |
| `mistral.batchJob.errors` | `failedRequests` was a number with no reason; this separates a rate-limit failure from a content-filter refusal |
| `mistral.batchJob.metadata` | the only attribution a batch job carries |

`replacementModel` is a **typed accessor returning `mistral.model`**, not the raw `deprecation_replacement_model` id string, per the typed-reference gate in CLAUDE.md §2 step 1.5. It resolves through the already-fetched workspace model list, so a query over every deprecated model costs one list call rather than one call per model; a replacement the list does not carry falls back to a single `GET /v1/models/{id}`, and a replacement that cannot be read reports null rather than failing the model it hangs off.

## Endpoint shapes, and where they come from

Every shape below was read from Mistral's published OpenAPI document at `https://docs.mistral.ai/openapi.yaml`, not from prose docs:

- `GET /v1/connectors` → `PaginatedConnectors` = `{items: [Connector], pagination: {next_cursor, page_size}}`. **Keyset** pagination (`cursor`, `page_size`, default 100, max 1000), not the offset envelope the older list endpoints use. Enums are closed in the spec: `ConnectorProtocol` = mcp/http/turbine, `ResourceVisibility` = shared_global/shared_org/shared_workspace/private, `ConsumerType` = user/org/workspace/system, `OutboundAuthenticationType` = oauth2/bearer/none/github_app/slack_app.
- `GET /v1/libraries` → `ListLibrariesResponse` = `{data: [Library], next_page_token}`. The offset `page` parameter is marked deprecated in the spec ("offset paging re-scans earlier pages and is being phased out"), so the walk follows `next_page_token`.
- `GET /v1/libraries/{library_id}/share` → `ListSharingResponse` = `{data: [Sharing]}`, unpaginated. `share_with_uuid` is nullable. `EntityType` = User/Workspace/Org and `ShareEnum` = Viewer/Editor both carry `x-speakeasy-unknown-values: allow`, and the `Sharing` response types `role`/`share_with_type` as bare strings rather than those enums, so the `.lr` says "such as" rather than presenting a closed set.

Both new list walks stop on a repeated cursor/token as well as on a null one, so a server that echoes the cursor back cannot spin the walk forever.

Every nullable string on the new resources reaches MQL through `llx.StringDataPtr`, so an absent value reads `null` rather than `""`: `server`, `description`, `ownerId`, `shareWithUuid`, and the integration's `name`/`runName`/`url`. "This connector has no server URL" and "this connector's server URL is the empty string" are different facts, and a policy should be able to tell them apart.

`mistral.library.access` has no id of its own. Its `__id` is composed as `<libraryId>/<shareWithType>/<shareWithUuid>` and passed through the magic `"__id"` argument, with no public `id` field. The entity kind is in the key deliberately: an organization-wide share reports **no** `share_with_uuid`, so keying on the id alone would collide every id-less share on a library into one entry. Same reasoning for `mistral.fineTuningJob.integration`, keyed `<jobId>/<type>/<project>/<runName>`.

## Verification

**There is no Mistral API key available in this environment, so nothing here has been run against a live Mistral account.** Every claim about response shapes rests on the vendor's published OpenAPI document, not on observed responses. Specifically unverified against a real workspace: whether `Library.owner_type` and `Sharing.role` carry the capitalized `User`/`Workspace`/`Org` and `Viewer`/`Editor` spellings in practice, and the `integrations` payload on a fine-tuning job (see the note below — that endpoint is no longer in the public spec at all, so its shape here rests on the in-tree decoder written when the provider shipped plus the deprecated-fine-tuning docs).

What *was* run: the provider was built and driven end to end against a local stand-in API serving the documented payloads, which exercises the schema wiring the Go tests cannot reach — field resolution, `__id` composition, the lazy accessor, and null handling:

| query | result |
|---|---|
| `mistral.connectors { name server protocol visibility ownerType supportedAuthMethods privateToolExecution createdAt }` | both cursor pages collected; `supportedAuthMethods: ["bearer"]`; the second connector's absent `server` and absent `created_at` both read `null` |
| `mistral.libraries { name description ownerId }` | the library with no description and no owner id reads `null` for both, not `""` |
| `mistral.fineTuningJobs { ... integrations { name runName url } }` | the integration with no run yet reads `null` for all three, not `""` |
| `mistral.connectors` (JSON) piped through `grep -ci 'header-value-must-not-be-decoded\|credential\|global_headers'` | **0** |
| `mistral.libraries { ... accesses { shareWithType shareWithUuid role } }` | both share entries reported distinctly; the Org share's `shareWithUuid` is `null`, not `""` |
| `mistral.models { id replacementModel { id maxContextLength } }` | in-list hit resolved; no-successor model `null`; a replacement absent from the listing resolved through the single-model fallback. The stand-in API logged exactly two requests for the whole query, `GET /v1/models` and one `GET /v1/models/{id}` for the replacement the listing did not carry, which is the cost claim above |
| `mistral.fineTuningJobs { id integrations { type project name runName url } }` | two integrations on one job reported separately, so the composed `__id` holds; the job with none reads `[]` |
| `mistral.batchJobs { id failedRequests metadata errors { message count } }` | both errors and both metadata keys land |
| `mistral.fineTuningJobs.all(integrations.length == 0)` | compiles and fails on the job that has integrations |

## Tests

Decode tests for every new record, absent-optional cases, and both pagination walks. Every test was checked by mutation:

| mutation | test that failed |
|---|---|
| `json:"visibility"` → `json:"visibilty"` | `TestListConnectors_DecodesEveryModeledField` (`expected "shared_org", actual ""`) |
| repeated-cursor guard removed | `TestListConnectors_RepeatedCursorTerminates` and `TestListLibraries_RepeatedTokenTerminates` (server returns "walk did not terminate") |
| `shareWithType` dropped from the access `__id` | `TestLibraryAccessID_SeparatesSharesWithNoEntityID` |
| `project` dropped from the integration `__id` | `TestIntegrationID_SeparatesIntegrationsOnOneJob` (3 ids collapse to 2) |

The stuck-cursor servers return a 500 after a few requests rather than looping, so a missing guard fails the suite in milliseconds instead of hanging it.

## Out of scope, deliberately

- **The Mistral Admin API** (`/v1/admin/api-keys` with `last_used`/`expiration_date`, audit logs, users, workspaces, spend limits). It is genuinely the most audit-valuable surface Mistral offers, but it is Enterprise-only, its keys come from a different console, it authenticates with an `x-api-key` header rather than `Authorization: Bearer`, and the docs state a standard API key never grants admin access regardless of the caller's role. It needs a second connection mode, not fields on `mistral`. Modeled here it would read empty for every non-Enterprise user, which is worse than not shipping it.
- **`GET /v1/users/me`** and `/v1/users/me/{organizations,workspaces}`. These require a dashboard session (`DashboardUserContextAuth`), not an API key. They look like the answer to "who owns this workspace" and would read empty.

That absence is also why `mistral.library.ownerId` and `mistral.library.access.shareWithUuid` stay raw strings rather than typed accessors: there is no user, workspace, or org resource in this provider for them to point at, and the endpoints that would back one need auth this provider cannot obtain.

## Flagging, not acting on

`mistral.fineTuningJob` rests entirely on `GET /v1/fine_tuning/jobs`, which Mistral files under **"Deprecated Fine-tuning"** — and which is **no longer present in the public OpenAPI document at all** (only `/v1/fine_tuning/models/{model_id}` and its archive routes remain). Adding `integrations` to it is still worth doing, since a job that streams telemetry off-workspace is a real finding today, but the resource's future needs a decision: keep it against a deprecated endpoint, or retire it alongside the vendor.

## Housekeeping

- `.lr.versions` entries at `14.0.0`, matching `Version: "14.0.0-rc.1"` in `providers/mistral/config/config.go`; codegen assigned these, and `Version` is untouched.
- `gofmt` clean, `go build ./...` and `go test ./...` green inside `providers/mistral/`, `go mod tidy` shows no diff, `golangci-lint` reports 0 issues, and `mqlr generate` is idempotent (a re-run leaves no diff).


arlimus (approved):
Lots more context, kudos!

PR #10851, opened 2026-09-14
⭐ weaviate: model group-based RBAC, and when an API key was last used

tas50:
**Stacked on #10847** — both change `rbac.go`'s role builder and user lister, and `group.roles` depends on that PR's fix (a group's roles come back in full, and must not be erased by a name-only reference arriving later). Base this on #10847, or merge that first; the diff against it is the second commit only.

## 1. Group-based RBAC

On an OIDC deployment, roles are granted to **groups** as well as users. `weaviate.role.assignedUsers` was the only assignment edge modeled, so "who holds root?" answered wrong in the worst direction: often **empty**, while an entire directory group held it. An empty list satisfies every assertion made about it.

Both directions, per CLAUDE.md Step 1.5 — the reverse edge is the one that carries the audit:

| | source |
|---|---|
| `weaviate.role.assignedGroups() []weaviate.group` | `Roles().GroupAssignmentGetter().WithRole(name)` |
| `weaviate.instance.groups() []weaviate.group` | `Groups().OIDC().GetKnownGroups()` — a real LIST |
| `weaviate.group.roles() []weaviate.role` | `Groups().OIDC().RolesGetter().WithIncludeFullRoles(true)` |

`weaviate.group` carries `groupId`, `groupType` and `roles()`. Typed references both ways rather than a list of names, so `weaviate.instance.groups.where(groupId == "platform-admins").first.roles.first.permissions` walks straight through.

`WithIncludeFullRoles(true)` brings each role's permissions back with it, so the permissions of a role held **only** by a group resolve without another request — and thanks to #10847 those full roles are not subsequently erased by a name-only reference to the same role.

**Identity.** A group is keyed by `serverID/group/<groupType>/<groupId>`. One name can belong to two kinds of group, and a key built from the name alone makes the second resolve to the first through the resource cache. `groupTypeOIDC` is the only kind Weaviate assigns roles to today (`models.GroupTypeOidc` is the sole enum value), and the SDK only exposes a roles getter under `OIDC()`, so `group.roles` returns empty for any other kind rather than guessing at an endpoint.

## 2. `weaviate.user.lastUsedAt` and `createdAt`

`users.UserInfo.LastUsedAt` was already on the struct; the lister was not passing `.WithLastUsedTime()`, which sets `?includeLastUsedTime=true`. Without it the server sends `"lastUsedAt":"0001-01-01T00:00:00.000Z"` for **every** user, so every key looked as though it had never been used. One flag.

This is what answers "which API keys are still valid but have not been used in 90 days".

**Null, not year 1.** Both timestamps arrive as the zero time when the server has none — a never-used key, and an environment-defined user that has no creation time. Reported by value that would date them to January of the year 1 and make any staleness check true for a key issued this morning, so `nonZeroTime` maps the zero time to null.

Confirmed on a live 1.38.9, which is also where the `createdAt` behaviour below comes from:

```json
{"active":true,"createdAt":"2026-09-14T20:28:08.905Z","dbUserType":"db_user",
 "lastUsedAt":"0001-01-01T00:00:00.000Z","roles":["escalator"],"userId":"analyst"}
{"active":true,"createdAt":"0001-01-01T00:00:00.000Z","dbUserType":"db_env_user",
 "lastUsedAt":"0001-01-01T00:00:00.000Z","roles":["root"],"userId":"root-user"}
```

`apiKeyFirstLetters` is deliberately **not** modeled — it trips secret scanners and buys nothing. Nothing from `client.Data()`, `client.GraphQL()` or `client.Experimental().Search()` is touched.

## Tests

Unit tests, each proved to fail by mutating the implementation:

| test | mutation | observed failure |
|---|---|---|
| `TestNonZeroTime` | drop the `t.IsZero()` guard | `nonZeroTime(zero) = 0001-01-01 00:00:00 +0000 UTC, want nil` |
| `TestGroupResourceIDCarriesGroupType` | drop `groupType` from the id | `two kinds of group named platform-admins share the id ".../group/platform-admins"` |
| `TestGroupsOfDifferentTypesAreDistinctResources` | same | `two kinds of group named platform-admins resolved to one resource` |

Plus `TestIsNotFound`.

**One gap I want to be explicit about, because I found it by mutating rather than by assuming:** deleting `.WithLastUsedTime()` still compiles and still passes every unit test. Nothing offline can catch it — the flag only changes a query string. `TestIntegrationUserTimestamps` is therefore written to catch exactly that signature and nothing else: since a 1.38.9 server sends `createdAt` **whether or not** the request asks for timestamps, `createdAt` proves nothing about the flag, and my first draft of this test asserted on it and would have passed with the flag removed. It now asserts that at least one user reports a non-null `lastUsedAt` — the test authenticates as one of the listed users, so "every user is null" is the signature of the missing flag.

## Verification

**What rests on the compiler and tests:** everything above except the live run below. Unit tests, `go build`, `go vet`, `golangci-lint` (0 issues), `mqlr generate` reporting 14 additive and 0 breaking schema changes, and `go test ./...` inside `providers/weaviate/`.

**What was run against a real server:** I brought up Weaviate 1.38.9 in Docker with RBAC and dynamic DB users while working on #10847 and verified the permission and user shapes there, including the JSON above. The daemon on that machine then went away, before I could re-run it against **this** branch.

**What still needs a real server, and is not verified:**

- `weaviate.instance.groups`, `weaviate.role.assignedGroups` and `weaviate.group.roles` have not been run against a live server at all. They need an **OIDC-backed** deployment with group assignments, which a local API-key Weaviate cannot provide.
- The **404 handling** is the specific thing I would check first. I treat a 404 from the group endpoints as "no group holds a role here", on the reasoning that a server without OIDC has no group endpoint and failing every `weaviate.instance.groups` query on the majority of deployments would be worse. I have **not** observed that 404 — it is inferred from `except.NewUnexpectedStatusCodeErrorFromRESTResponse` being the only non-200 path in the SDK. If the server answers 200 with an empty list instead, the code is still correct and the 404 branch is simply dead.
- `lastUsedAt` reporting a **real** instant. I confirmed the zero-time case live; I have not seen a used key come back with a real time, which is the half `TestIntegrationUserTimestamps` is built to assert.

Reproducing the fixture takes about a minute:

```
docker run -d -p 8085:8080 \
  -e AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED=false \
  -e AUTHENTICATION_APIKEY_ENABLED=true \
  -e AUTHENTICATION_APIKEY_ALLOWED_KEYS=root-user-key \
  -e AUTHENTICATION_APIKEY_USERS=root-user \
  -e AUTHENTICATION_DB_USERS_ENABLED=true \
  -e AUTHORIZATION_RBAC_ENABLED=true \
  -e AUTHORIZATION_RBAC_ROOT_USERS=root-user \
  cr.weaviate.io/semitechnologies/weaviate:1.38.9

WEAVIATE_TEST_HOST=127.0.0.1:8085 WEAVIATE_TEST_KEY=root-user-key \
  WEAVIATE_TEST_SEEDED=1 go test ./resources/ -run TestIntegration -v
```

## Privileged access

Groups, roles and users all need RBAC-admin permissions. The new resources follow the provider's existing rule: `isForbidden(err)` (401/403) maps to an empty list rather than an error, and 404 joins it for the group endpoints.

That makes **empty ambiguous between "no group holds this role" and "this credential cannot see group assignments"**, and a policy asserting "no group holds root" passes in both cases. `weaviate.instance.rbacEnabled` separates the RBAC-is-off case but not the under-privileged one. This is the house rule rather than a choice made here, but it is worth stating plainly on a PR whose whole subject is who holds which role.

## Not changed

`weaviate.role.assignedUsers` stays `[]string`. It is shipped, and changing a shipped field's type is not on the table; `assignedGroups` returning typed resources alongside it is an inconsistency I would rather have than a break.


arlimus (approved):
kudos!

PR #10899, opened 2026-09-16
🧹 os: drop the unused bootspec key constant

tas50:
Removes a dead constant in `kernel_nixos.go` that golangci-lint fails on.

```
providers/os/resources/kernel_nixos.go:30:2: const nixosBootspecV1Key is unused (unused)
```

`nixosBootspecV1Key` names the bootspec namespace, but the only place that string is needed is the struct tag on `nixosBootJSON`, and a struct tag cannot read a constant — so it was dead the moment it was written. The RFC 125 pointer it carried moves onto the struct, where a reader looking at the tag will be.

The lint was red on #10893 and it merged that way. `only-new-issues` means it stays quiet until someone touches this file, and then it fires on lines they did not write.

Companion to #10898, which does the same on v13, where #10895 carried the constant across. Both branches end up identical here.

## Verification

`golangci-lint run --config=.github/.golangci.yaml ./providers/os/resources/` → `0 issues` (the repo root config does not catch this one; CI's does). `go build ./providers/os/...` and the `TestNixos*` kernel tests pass.


tas50:
Closing: #10896 landed the identical fix on main first, with better wording on the struct comment. Rebasing this branch onto main dropped its commit as empty, so there is nothing left in it. The v13 side is covered by #10898.

PR #10913, opened 2026-09-17
✨ os: add jetbrains for IDE plugin inventory

tas50:
Adds `jetbrains`, an inventory of the plugins installed into JetBrains IDEs across every user account and every installed product, following the shape `vscode` already uses for editor extensions.

## What it reports

**The plugin id, not just the display name.** The two differ:

```xml
<id>org.intellij.plugins.markdown</id>
<name>Markdown</name>
```

The id is what the IDE and the JetBrains Marketplace key on, and what survives a rename or a translated display name. Without it a plugin cannot be stably identified or joined to marketplace data.

**Plugins bundled with the IDE as well as the ones a user installed**, separated by `bundled`. That distinction decides whether the resource says anything at all: on the macOS host this was developed against, **ten JetBrains products are installed and not one has a `plugins` directory**. Every plugin present is bundled, 244 of them, so reading only the user directories would report zero plugins for a machine full of IDEs.

**Two derived facts.** `enabled` reads the product's `disabled_plugins.txt`, since a disabled plugin is installed but not executing and only one of those is a risk. `outsideBuildRange` reports a plugin that declares support only for IDE builds older than the one it is installed into, which needs the build number the IDE states in `product-info.json`.

## Schema

```
jetbrains {
  plugins() []jetbrains.plugin
  paths() []string
}

private jetbrains.plugin @defaults("product name version enabled") {
  identifier  string   // org.jetbrains.plugins.github
  name        string   // GitHub
  version     string
  vendor      string
  category    string
  description string
  product     string   // GoLand 2026.2
  path        string
  bundled     bool
  enabled     bool
  sinceBuild  string
  untilBuild  string
  outsideBuildRange bool
  uid         int
}
```

Metadata comes from `META-INF/plugin.xml` inside the plugin, read through the connection's filesystem so it works against an image or a remote host exactly as it does locally. Three layouts are handled: descriptor inside a jar under `lib/`, unpacked on disk, and a plugin that is a single jar.

## Two things the live run corrected

Both are regression tests now, and neither was reachable by reading the code.

**`Contents/plugins` is a standard macOS bundle directory, not a JetBrains marker.** Treating its presence as the signal matched **23 unrelated applications** — 1Password, Bitwarden, Zoom, Tailscale, WireGuard, Goodnotes and more — against 3 real IDEs. An application is a JetBrains product only when it also ships `product-info.json`, which conveniently is the same file that supplies the build number `outsideBuildRange` needs.

**An IDE under an absolute root is installed once for the machine.** Scanning `/Applications` and `/opt` per user reported every bundled plugin once per account: on a Linux tree with `root` and one login, the same plugin came back twice. On a shared host that multiplies a few hundred plugins by the number of logins.

A third assumption survived only because it was measured: the jar named after the plugin directory holds the descriptor for most plugins, but **18 of the 139 bundled with Rider keep it elsewhere** (`markdown/lib/intellij.markdown.jar`, `DatabaseTools/lib/database-plugin.jar`). The name-matched jar is tried first and the rest are scanned after; reading only the first would have silently missed 13%.

## Verification

**macOS, ten installed JetBrains products:**

```
jetbrains.plugins.length   244
jetbrains.paths.length       3      Rider, DataSpell, MPS
```

244 is exact rather than approximately right: the three plugin directories hold **247** entries, of which three are bookkeeping files (`plugin-classpath.txt` twice, `build.properties` once). Per product, 139 / 55 / 50 against 140 / 56 / 51 entries on disk.

A record, showing the id/name distinction:

```
identifier: "com.intellij.database"
name:       "Database Tools and SQL"
product:    "DataSpell 2026.1.3"
version:    "261.26222.84"
vendor:     "JetBrains"
category:   "Database"
bundled:    true
enabled:    true
```

**A Linux tree via a filesystem scan**, for the paths a Mac cannot exercise — the Linux layout where plugins sit directly in the product directory rather than under `plugins/`, a user-installed plugin, a disabled one, and one pinned to an old IDE branch:

| plugin | uid | bundled | enabled | outsideBuildRange |
|---|---|---|---|---|
| `org.jetbrains.plugins.go` | 0 | **true** | true | false |
| `com.example.current` (until-build `262.*`) | 1000 | false | true | false |
| `com.example.stale` (until-build `241.*`) | 1000 | false | **false** | **true** |

That run also covers the wildcard form of `until-build`, and the descriptor living in a non-matching jar name.

## Tests

19 tests. The fixtures under `testdata/jetbrains/` are captured from the plugins bundled with the installed Rider 2026.2, trimmed to the metadata element with the values verbatim, and cover the shapes real descriptors take: `id` and `name` in either order, `vendor` carrying a `url` attribute alongside its text, and `category` absent.

Named breaks the tests catch: a descriptor with neither id nor name must error rather than produce a plugin with an empty identifier (which would collide with every other such plugin in the resource cache); an application with `Contents/plugins` but no product marker must not be a product; a system-wide IDE must be scanned once; `product-info.json` that is malformed or names no product must read as "not a product" rather than a product with empty fields.

One test pins something the security tooling raised on a neighbouring PR: a marketplace `plugin.xml` is third-party input, so the parser must not expand entities. Go's `encoding/xml` does not, and `TestJetBrainsPluginXMLDoesNotExpandEntities` keeps it that way.

## No purl field

`vscode.extension` carries one, so the absence is deliberate. I checked the official [purl type list](https://github.com/package-url/purl-spec): there is `vscode-extension` and `chrome-extension`, but **no JetBrains type**. Marketplace plugins do resolve from `plugins.jetbrains.com/maven` as `group:id:version`, but I could not pin the Marketplace's group convention from the documentation. A shipped field's format cannot be changed later, so it is left out rather than guessed at. Proposing a `jetbrains-plugin` type upstream is the honest fix and would be a good follow-up.

## Gates

`gofmt`, `typos`, `go build ./providers/os/...`, `go test ./providers/os/...` and `cd .github && golangci-lint run ../providers/os/...` all clean — the last two report only the pre-existing `snapshot.MockVolumeMounter` typecheck noise present on `main`.

The 18 new `.lr.versions` entries read `14.0.0`, matching the release the current `14.0.0-rc.1` is a candidate for, which is what `mqlr generate` produces.


arlimus (approved):
Wooot!! amazing

PR #10946, opened 2026-09-19
🐛 os: report the epoch of a deb package

tas50:
## Problem

dpkg keeps the epoch inside the `Version` field of `/var/lib/dpkg/status`, so a package versioned `1:2.38.1-5+deb12u3` reported `epoch: ""`. rpm reports the two separately and `ParseRpmPackages` fills both `Package.Epoch` and the epoch-prefixed `Package.Version`, so the epoch reached the purl's `epoch=` qualifier and the CPE's update component on an rpm host and nowhere on a deb host.

`ParseDpkgPackages` was already passing `pkg.Epoch` to `purl.WithEpoch` and to `cpe.NewPackage2Cpe` — the value was simply never assigned.

Found while sweeping package detection across 34 distro containers.

### Measured across the sweep

| family | images | pkgs with an epoch in `version` | `epoch` field set |
|---|---|---|---|
| rpm | 13 (fedora 41/42/43, alma 9/10, rocky 9, oracle 9, ubi 9/10, amzn 2/2023/2027, openeuler, mageia, altlinux, azurelinux, leap, tumbleweed) | 3–24 per image | **all of them** |
| dpkg | 8 (debian 11/12/13, ubuntu 20.04/22.04/24.04, kali) | 11–14 per image | **none** |

On `debian:12`, 11 of 88 packages carry an epoch — `bsdutils`, `diffutils`, `libattr1`, `libaudit-common`, `libaudit1`, `libcap2`, `libcrypt1`, `libgmp10`, `login`, `passwd`, `zlib1g`.

## Change

Split the epoch off the version with a small helper and keep `Version` exactly as dpkg wrote it — the same pairing the rpm reader produces.

Only a leading run of digits before the first colon counts. A Debian `upstream_version` may itself contain a colon, but policy allows that only when an epoch is present, so the first colon is the epoch separator exactly when what precedes it is numeric. `0` is reported as no epoch, matching `normalizeRpmEpoch`.

## Shipped values that change

For the ~13% of deb packages carrying an epoch, two values change — in both cases *to match what rpm already emits*:

```
purl   before  pkg:deb/debian/bsdutils@1:2.38.1-5%2Bdeb12u3?arch=arm64&distro=debian-12.15
       after   pkg:deb/debian/bsdutils@1:2.38.1-5%2Bdeb12u3?arch=arm64&distro=debian-12.15&epoch=1

cpe    before  cpe:2.3:a:bsdutils:bsdutils:2.38.1-5\+deb12u3:*:*:*:*:*:arm64:*
       after   cpe:2.3:a:bsdutils:bsdutils:2.38.1-5\+deb12u3:1:*:*:*:*:arm64:*
```

rpm today, for comparison: `pkg:rpm/fedora/xz-libs@1:5.8.1-4.fc42?arch=aarch64&distro=fedora-42&epoch=1`.

**Please sanity-check this against advisory matching** if anything downstream keys on the exact deb purl string. The epoch is already present inside the purl version either way; this adds the qualifier that rpm purls have always carried. `version` is untouched, and the package list is untouched.

`TestDpkgParser`'s golden expectation for `libaudit1` (the one epoch-bearing package in that fixture) is updated to the new values — that diff is the behavior change, stated explicitly rather than hidden.

## Scope

dpkg only. pacman has the same gap (4 packages on the Arch image), but its reader is being restructured in #10945; I'll follow up there once that lands rather than conflict with it.

## Tests

- `TestEpochFromVersion` — real epoch-bearing versions from debian:12, a multi-digit epoch, explicit `0:` and `00:`, an empty version, a leading `:`, and three non-numeric-before-colon cases that must not be read as an epoch.
- `TestDpkgParserEpoch` — an epoch-bearing and a plain stanza; asserts the field, that `version` keeps the epoch, and that the purl gains/omits the qualifier. Removing the `epochFromVersion` call makes every assertion but the version one fail.

## Verification

- `go test ./resources/packages/` — pass
- Live on `debian:12` with the patched provider: `packages.where(epoch != "").length` = **11**, `dpkg-query -W -f '${Version}\n' | grep -c ':'` = **11**, `packages.length` = 88 (unchanged)


tas50:
@chris-rock @czunker This is technically correct but let me know how this impacts vuln matching

czunker (approved):
Thanks @tas50

PR #10948, opened 2026-09-19
🐛 os: report the epoch of a pacman package

tas50:
Follow-up to #10946, promised in its description.

## Problem

pacman keeps the epoch inside the version, the way dpkg does. A package versioned `1:1.8.13-1` reported `epoch: ""`.

Both readers are affected — the `pacman -Q` one and the local-database one added in #10945 — so the field was empty however an Arch host was scanned.

### Measured, `menci/archlinuxarm:latest`

4 of 137 packages carry an epoch, and none reported it:

```
$ pacman -Q | grep ':'
iptables 1:1.8.13-1
lz4      1:1.10.0-2
nftables 1:1.1.7-3
zlib     1:1.3.2-3
```

| | before | after |
|---|---|---|
| `packages.length` | 137 | 137 |
| `packages.where(epoch != "").length` | **0** | **4** |

```
before: {"name":"zlib","version":"1:1.3.2-3","epoch":"",
         "purl":"pkg:alpm/arch/zlib@1:1.3.2-3?arch=aarch64&distro=archarm-rolling"}
after:  {"name":"zlib","version":"1:1.3.2-3","epoch":"1",
         "purl":"pkg:alpm/arch/zlib@1:1.3.2-3?arch=aarch64&distro=archarm-rolling&epoch=1"}
```

## Change

`version` stays exactly as pacman wrote it and `Epoch` carries the value on its own — the pairing the rpm and dpkg readers produce. Reuses `epochFromVersion` from #10946 rather than a second copy of the rule.

As for deb, the purl gains the `epoch=` qualifier that rpm purls have always carried. Same tradeoff as #10946, on 4 packages rather than ~13% of the inventory.

## Stacking

The first commit is #10946's (`epochFromVersion` lives there). **Merge #10946 first**; this then rebases to a single commit. The branch is cut from `main` *plus* that commit, so it already contains #10945 and will not conflict with the merged pacman reader.

## Tests

`TestPacmanEpoch` covers both readers with the real epoch-bearing packages from the container: `pacman -Q` output for `iptables`, and a desc record for `lz4`, each paired with an epoch-free package (`bash`) that must keep `epoch: ""` and gain no purl qualifier. Removing either `epochFromVersion` call makes its subtest fail.

## Verification

- `go test ./resources/packages/` — pass
- Live on `menci/archlinuxarm:latest`: `packages.where(epoch != "").length` = **4**, matching `pacman -Q | grep -c ':'` = **4**; `packages.length` still 137


czunker (approved):
LGTM
Thanks @tas50

